PR #132, opened 2021-06-25
Generate `never` as AST reference type for missing cross references

### LINKED ISSUE #130: Grammar without references generates invalid `ast.js`.
Langium version: 0.1.0
Package name: ?

## Steps To Reproduce

1. Make a grammar without references (`...=[Person|ID]`).
2. Run `langium:generate`.

## The current behavior

`src/generated/ast.ts` has a syntax error at `export type <language name>AstReference = ;`

## The expected behavior

I guess `never` would work as RHS value if there are no references.

msujew:
Closes #130 

Assuming the user grammar contains no cross references, the problem in the mentioned issue occurs. This change instead adds `never` as the type alias for `AstReference`, keeping it in line with the generated `getReferenceType` method that always throws without any cross references.

### DIFF
--- a/packages/langium-cli/src/generator/ast-generator.ts
+++ b/packages/langium-cli/src/generator/ast-generator.ts
@@ -71,7 +71,7 @@ function generateAstReflection(grammar: Grammar, interfaces: Interface[]): Gener
     );
     reflectionNode.children.push(
         'export type ', grammar.name, 'AstReference = ',
-        crossReferenceTypes.map(e => `'${e.type}:${e.feature}'`).join(' | '),
+        crossReferenceTypes.map(e => `'${e.type}:${e.feature}'`).join(' | ') || 'never',
         ';', NL, NL
     );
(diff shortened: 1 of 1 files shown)


dhuebner (approved):

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -71,7 +71,7 @@ function generateAstReflection(grammar: Grammar, interfaces: Interface[]): Gener
     );
     reflectionNode.children.push(
         'export type ', grammar.name, 'AstReference = ',
-        crossReferenceTypes.map(e => `'${e.type}:${e.feature}'`).join(' | '),
+        crossReferenceTypes.map(e => `'${e.type}:${e.feature}'`).join(' | ') || 'never',
```
spoenemann, line 74: Does that work? I think arrays are always truthy, even when empty, so the length should be checked explicitly.

spoenemann: Oh I see, that's tricky...

spoenemann (approved):

Issue #131, opened 2021-06-25
Add generator CLI to domainmodel and statemachine examples


spoenemann:
We should add a CLI (similar to #121) to each of the domainmodel and statemachine examples. These should generate code from an input file. The target language is open for discussion.

My proposals:
 - domainmodel &mdash; generate Java classes with private fields and corresponding getters and setters.
 - statemachine &mdash; generate C code with a `main` function that reads from stdin and prints state transitions according to the events entered by the user.

This issue depends on #122: we should finalize the generator API before adding more uses of it.

PR #133, opened 2021-06-25
Improve Generator API

### LINKED ISSUE #122: Improve the API of CompositeGeneratorNode
The generator node API should be improved to make appending and indenting generated code easier.

We currently use it heavily in `langium-cli`. The goal of this task is to make the generator code more compact and readable.

msujew:
Closes #122 

I've already rewritten the `ast-generator.ts` so that uses the new API. WDYT?

If we all agree on the API, I'll rewrite the other generators as well.


dhuebner (approved):
I like the new API! 
Do we have some generator or node processor tests already?

msujew:
@dhuebner There aren't any, yet. I'll write some on monday.

packages/langium/src/generator/node.ts
```diff
@@ -6,45 +6,38 @@
 
 import { EOL } from 'os';
 
-export type GeneratorNode = CompositeGeneratorNode | IndentNode | TextNode | NewLineNode | string;
+export type GeneratorNode = CompositeGeneratorNode | IndentNode | NewLineNode | string;
```
spoenemann, line 9: I'd propose to rename this file to `generator-node.ts`.

packages/langium/src/generator/node-processor.ts
```diff
@@ -70,17 +58,15 @@ class Context {
     }
 }
 
-export function process(node: GeneratorNode, defaultIndentation?: string | number): string {
+export function processNode(node: GeneratorNode, defaultIndentation?: string | number): string {
```
spoenemann, line 61: Maybe even more specific, like `processGeneratorNode`? We need to consider that currently we export everything into the index, so users import it from `'langium'`.

packages/langium/src/generator/node.ts
```diff
@@ -6,45 +6,38 @@
...
 
 export class CompositeGeneratorNode {
 
-    private _children: GeneratorNode[] = [];
+    readonly children: GeneratorNode[] = [];
```
spoenemann, line 13: What do you think of renaming this to `content` or `contents`?

spoenemann (commented):
I agree it looks really good! Just a few remarks.

packages/langium/src/generator/generator-node.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    ifNotEmpty  = false;
+
+    constructor(lineDelimiter?: string, ifNotEmpty = false) {
+        this.lineDelimiter = lineDelimiter ?? EOL;
+        this.ifNotEmpty = ifNotEmpty;
+    }
+}
+
+export const NL = new NewLineNode();
```
spoenemann, line 70: I'm wondering whether we should define this with `\n` instead of using the OS-specific EOL.

spoenemann (approved):
:+1:

spoenemann:
@Pluralia this PR is important for #131.

packages/langium/src/generator/generator-node.ts
msujew: I thought about that as well. Using `\n` led to git issues on Windows, since the file 'changed'. That why I changed it to the OS-EOL.

PR #142, opened 2021-06-30
Embed grammar json in typescript file

msujew:
To facilitate easy use of the `langium` package without the need to always copy the `grammar.json`, we generate a TypeScript file that is easily embedded by webpack.


spoenemann (approved):
How shall we handle cases like this, where we change the generated code so a file is no longer generated and used? Shall we include a "clean" step that deletes all files in the target directory? If yes, shall that be opt-in or opt-out via options?

msujew:
I would say a cleanup step would be opt-out via the `langium` config options. Especially since the `generated` directory is usually git ignored, right?

spoenemann:
Ok could you create an issue for that?

PR #156, opened 2021-07-02
Add normalisation check in the arithmetics example

Pluralia:


examples/arithmetics/src/language-server/arithmetics-validator.ts
```diff
@@ -9,15 +9,50 @@ export class ArithmeticsValidationRegistry extends ValidationRegistry {
...
+            Division: validator.checkDivByZero,
+            Definition: validator.checkNormalisable
         };
         this.register(checks, validator);
     }
 }
 export class ArithmeticsValidator {
     checkDivByZero(div: Division, accept: ValidationAcceptor): void {
-        if (isNumberLiteral(div.right) && Number((div.right as NumberLiteral).value) === 0) {
+        if (isNumberLiteral(div.right) && +div.right.value === 0) {
```
msujew, line 20: Is there a specified reason why you start the `div.right.value` statement with a `+`?

examples/arithmetics/src/language-server/arithmetics-validator.ts
```diff
@@ -9,15 +9,50 @@ export class ArithmeticsValidationRegistry extends ValidationRegistry {
...
         }
     }
+
+    checkNormalisable(def: Definition, accept: ValidationAcceptor): void {
+        const context = new Map<Expression, number>();
+
+        const makeOp = (expr: Addition | Subtraction | Multiplication | Division, op: (x: number, y: number) => number): void => {
+            const subExprs = [expr.left, expr.right];
+            subExprs.forEach(e => evalExpr(e));
+            const [left, right] = subExprs.map(e => isNumberLiteral(e) ? +e.value : context.get(e));
```
msujew, line 31: Is there a specified reason why you start the `e.value` statement with a +?

examples/arithmetics/src/language-server/arithmetics-validator.ts
```diff
@@ -9,15 +9,50 @@ export class ArithmeticsValidationRegistry extends ValidationRegistry {
...
+            if (left && right && op(left, right).toString().length <= 8) {
```
msujew, line 32: You should check for `undefined` explicitly here, since `0` is also falsy. Therefore, an expression like `0 + 0 - 0 * 0` will not be picked up by this validation.

Use `left !== undefined && right !== undefined` instead.

msujew (commented):
Thanks, that looks good already. Just a few small remarks.

examples/arithmetics/src/language-server/arithmetics-validator.ts
Pluralia: Type of `div.right.value` is `string`. If I want to compare it with `0`, I should covert it to `number`.

examples/arithmetics/src/language-server/arithmetics-validator.ts
Pluralia: To apply `op`, we have to have `left` and `right` of type `number`. So, I just convert `string` of `e.value` to `number`.

examples/arithmetics/src/language-server/arithmetics-validator.ts
msujew: I see. Could you add `returns string` to the `NUMBER` terminal rule? It will automatically convert it to a JS `number`.

msujew (approved):
LGTM

PR #158, opened 2021-07-02
Add `langium-config.json` support

### LINKED ISSUE #79: Support configuration in external file
We should support configuration via an external json file additionally to the embedded config in package.json. By doing that we can easily provide a json schema in langium-vscode so users get editor support for the config (is that possible for package.json extensions?)

We should have a default for the file name, but also allow passing a config file path as argument.

Candidates for default name:
 * langiumconfig.json
 * langium-config.json
 * .langiumconfig.json
 * .langium-config.json
 * .langiumrc.json

I like `langium-config.json` most. IMO `.langiumrc.json` is not really applicable because `rc` means ["run commands"](https://en.wikipedia.org/wiki/Run_commands).

msujew:
Closes #79

Also adds a few other things to the `langium-cli` package:

* Console colors for errors, warnings, success messages, etc.
* Catch any IO exceptions and log them appropriately.
* Grammar validation is now executed before trying to run the CLI. (logging any errors)
* JSON schema validation for config files.
* A watch mode.


packages/langium-cli/schema.json
```diff
@@ -0,0 +1,54 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "definitions": {
+        "config": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "grammar": {
+                    "type": "string"
```
spoenemann, line 9: Please add descriptions to all properties.

packages/langium-cli/src/generate.ts
```diff
@@ -4,74 +4,104 @@
...
-    const packageContent = fs.readFileSync(file).toString();
-    const pack = <Package>JSON.parse(packageContent);
-
+export function generate(config: LangiumConfig): void {
     const services = createLangiumGrammarServices();
+    const relPath = config[RelativePath];
 
-    const grammarFile = pack.langium.grammar ?? 'src/grammar.langium';
-    const grammarFileContent = fs.readFileSync(grammarFile).toString();
+    const grammarFile = config.grammar ?? 'src/grammar.langium';
```
spoenemann, line 26: I think we should make the `grammar` property mandatory and drop this default value.

packages/langium-cli/src/langium.ts
```diff
@@ -4,20 +4,36 @@
...
+    if (!configs.length) {
+        console.error('Could not find a langium configuration. Please add a langium-config.json to your project or a langium section to your package.json.'.red);
+        process.exit(1);
+    }
+    const validation = validate(configs, schema);
+    if (!validation.valid) {
+        console.error('Your langium configuration is invalid.'.red);
+        process.exit(1);
+    }
+    configs.forEach(callback);
```
spoenemann, line 38: I think we will need to change this when we support multiple languages, but it's ok to process each language independently for now.

packages/langium/package.json
```diff
@@ -24,7 +24,7 @@
     "watch": "tsc --watch",
     "lint": "eslint ./src --ext .js,.ts",
     "test": "jest --coverage",
-    "langium:generate": "node ../langium-cli/bin/langium generate",
+    "langium:generate": "langium generate",
```
spoenemann, line 27: Shouldn't we keep the local path to ensure that we use the local version?

packages/langium/src/documents/document-builder.ts
```diff
@@ -33,13 +32,12 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
         this.process(document);
+        const diagnostics = this.documentValidator.validateDocument(document);
+        document.diagnostics = diagnostics;
 
         if (this.connection) {
-            const diagnostics = this.documentValidator.validateDocument(document);
             // Send the computed diagnostics to VS Code.
             this.connection.sendDiagnostics({ uri: document.uri, diagnostics });
-        } else {
-            resolveAllReferences(parseResult.value);
```
spoenemann, line 42: Attaching the diagnostics to the document might unnecessarily increase memory usage. We could add an optional callback parameter `(diagnostics: Diagnostic[]) => void` to the `build` method instead.

packages/langium/package.json
msujew: We don't need it, since npm workspaces has no issues with cyclic dependencies. So the `langium` call will always go to the local `langium-cli` package.

spoenemann: Ok, but please change the dependency to a fixed version: `"langium-cli": "0.1.0"` (without the `^`)

packages/langium/src/documents/document-builder.ts
```diff
@@ -29,17 +28,19 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
         this.documentValidator = services.validation.DocumentValidator;
     }
 
-    build(document: LangiumDocument): void {
+    build(document: LangiumDocument, diagnostics?: Diagnostic[]): void {
```
spoenemann, line 31: I feel we should find a cleaner solution than passing a mutable array. What about returning an object like this?
```typescript
return {
    get diagnostics() {
         // If not done before, compute the diagnostics, otherwise return the previously computed diagnostics.
    }
};
```

spoenemann: We could add even more data to the result object:
```typescript
export type BuildResult = {
     readonly parseResult: ParseResult<AstNode>
     readonly diagnostics: Diagnostic[]
}
```
With this, we can omit the `undefined` check when accessing the parse result from the document.

msujew: Yeah, I like this way better.

packages/langium/src/documents/document-builder.ts
```diff
@@ -29,18 +33,28 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+        let diagnostics: Diagnostic[] | undefined;
+        const validator = this.documentValidator;
         if (this.connection) {
-            const diagnostics = this.documentValidator.validateDocument(document);
-            // Send the computed diagnostics to VS Code.
-            this.connection.sendDiagnostics({ uri: document.uri, diagnostics });
-        } else {
-            resolveAllReferences(parseResult.value);
+            diagnostics = validator.validateDocument(document);
+            if (this.connection) {
```
spoenemann, line 44: Redundant check for `this.connection`.

spoenemann (approved):
Great!

PR #159, opened 2021-07-04
Fix deep partial type for services

### LINKED ISSUE #119: PartialLangiumServices is too partial
The type `PartialLangiumServices` is meant for overriding the default and generated services of Langium. We use `DeepPartial` here because we want to allow overriding only selected services. A simple `Partial` is not sufficient because we have groups. The problem: `DeepPartial` is too deep and allows specifying incomplete services.

Example:
```
    references: {
        ScopeProvider: () => ({})
    }
```

Here I can provide an empty object `{}` instead of a proper scope provider because the type system accepts `DeepPartial<ScopeProvider>`.

Langium version: 0.1.0
Package name: langium

## Steps To Reproduce

Add the code snippet above to the user module of a language.

## The current behavior

The incomplete scope provider is accepted by the TypeScript compiler.

## The expected behavior

The type system should reject incomplete services.

msujew:
Closes #119 

You can test this by trying to bind a service to an empty implementation like `AstReflection: () => ({})`.

### DIFF
--- a/packages/langium/src/services.ts
+++ b/packages/langium/src/services.ts
@@ -72,7 +72,10 @@ export type LangiumDefaultServices = {
 
 export type LangiumServices = LangiumGeneratedServices & LangiumDefaultServices
 
-type DeepPartial<T> = {
+// We basically look into T to see whether its type definition contains any methods. (with T[keyof T])
+// If it does, it's one of our services and therefore should not be partialized.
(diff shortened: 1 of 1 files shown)


danieldietrich (approved):
LGTM

(btw - I experimented with better types for DI that allow the inject function to accept an arbitrary number of arguments. The arguments are also deeply merged. I will go on with that after my vacation.)

PR #160, opened 2021-07-06
Add a CLI for the domainmodel example

Pluralia:
The CLI generates Java classes with private fields and corresponding getters and setters.
Closes [issue/131](https://github.com/langium/langium/issues/131).


examples/domainmodel/example/generated/example/blog/Blog.java
```diff
@@ -0,0 +1,22 @@
+package example.blog;
+
+class Blog {
```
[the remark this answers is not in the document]
Pluralia: I think it's better to push them. It will allow to take a look at generated files without CLI execution, directly in the repository.
So, in general we should add such things in a gitignore, but I'd prefer to push it for `Langium` usage examples.

examples/cli-util/src/cli-util.ts
```diff
@@ -0,0 +1,50 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import * as fs from 'fs';
+import colors from 'colors';
+import { DefaultDocumentValidator, Grammar, LangiumDocumentConfiguration, LangiumServices } from 'langium';
+import { Package } from 'langium-cli/src/package';
+
+export function extractGrammar(fileName: string, configFileName: string, services: LangiumServices): Grammar {
```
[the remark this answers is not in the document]
Pluralia: I also would prefer to add `cli` utility methods in a `src/cli`: `examples` is not the best directory for code, which can be reused by users. However, we should discuss, which functions have to be made as utility: function `extractGrammar` "hardcodes" and combine too many things, what can be not so nice for flexibility of user projects.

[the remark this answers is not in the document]
spoenemann: Yes we can think about extracting common functionality at some point, but for now I'd prefer duplicating this code than adding an additional package.

examples/cli-util/src/cli-util.ts
```diff
@@ -0,0 +1,50 @@
...
+    const document = LangiumDocumentConfiguration.create(`file:${fileName}`, languageId, 0, fileContent);
+    services.documents.DocumentBuilder.build(document);
+    if (!document.parseResult) {
+        console.error(`Failed to parse the grammar file ${fileName}`);
+        process.exit(1);
+    }
+ 
+    const validationErrors = new DefaultDocumentValidator(services)
+        .validateDocument(document)
+        .filter(e => e.severity === 1);
```
spoenemann, line 38: Validation is also done by the document builder. We should allow passing a callback function to `DocumentBuilder.build` to process diagnostics.

examples/domainmodel/example/generated/example/blog/Blog.java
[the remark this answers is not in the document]
spoenemann: I see your point, but that also means we need to take care of keeping those generated files updated. I propose adding a README to the example package describing how to use the CLI.

examples/cli-util/src/cli-util.ts
msujew: This will be implemented by #158

spoenemann:
Please move `cli.ts` and `generator.ts` to a subdirectory `cli`. Currently we have code for multiple processes in one package, so we should use a clear directory structure to avoid confusion.

`cli.ts` could be renamed to `index.ts` so it can be invoked with `node lib/cli` (equivalent to `node lib/cli/index.js`).

examples/cli-util/src/cli-util.ts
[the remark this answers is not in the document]
Pluralia: Okay, I'll remove `cli-util` and add an issue about extracting common functionality.

examples/domainmodel/src/cli/cli-util.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import * as fs from 'fs';
+import colors from 'colors';
+import { DefaultDocumentValidator, Grammar, LangiumDocumentConfiguration, LangiumServices } from 'langium';
+
+export function extractGrammar(fileName: string, languageId: string, extensions: string[], services: LangiumServices): Grammar {
```
msujew, line 11: Why does this return a `Grammar` if its then casted into a `DomainModel` either way? I would convert it into a generic method, simply returning `T`.

examples/domainmodel/src/cli/generator.ts
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    private generateEntity(entity: Entity, fileNode: CompositeGeneratorNode): void {
+        const maybeExtends = entity.superType ? ` extends ${entity.superType.$refName}` : '';
+        fileNode.append(`class ${entity.name}${maybeExtends} {`, NL);
+        fileNode.indent(classBody => {
+            const featureData = entity.features.map(f => this.generateFeature(f, classBody));
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            featureData.forEach(([generateField, _1, _2]) => generateField());
```
msujew, line 56: You can omit `_1` and `_2` here.

examples/domainmodel/src/cli/generator.ts
```diff
@@ -0,0 +1,89 @@
...
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            featureData.forEach(([_0, generateSetter, generateGetter]) => { generateSetter(); generateGetter(); } );
```
msujew, line 58: You can omit `_0` here.

examples/domainmodel/src/cli/generator.ts
```diff
@@ -0,0 +1,89 @@
...
+        });
+        fileNode.append('}', NL);
+    }
+
+    private generateFeature(feature: Feature, classBody: IndentNode): [() => void, () => void, () => void] {
```
msujew, line 63: If every array return type is the same, we could just as well return `Array<() => void>`

examples/domainmodel/src/cli/generator.ts
```diff
@@ -0,0 +1,89 @@
...
+import _ from 'lodash';
+import { CompositeGeneratorNode, Grammar, IndentNode, NL, processGeneratorNode } from 'langium';
+import { AbstractElement, Domainmodel, Entity, Feature, isDomainmodel, isEntity, isPackageDeclaration, Type } from '../language-server/generated/ast';
+
+export class DomainModelGenerator {
+    private domainmodel: Domainmodel;
+    private destination: string;
+    private path: string;
+
+    constructor(grammar: Grammar, fileName: string, destination = '.') {
```
msujew, line 17: This should not accept a `Grammar` but either `unknown` or `DomainModel`

examples/domainmodel/src/cli/cli-util.ts
Pluralia: `Grammar` extends `AstNode` and has more functionality, so if I return `T`, I lose it. Actually, I don't need this functionality now, for domain model CLI, but it's income in extracting common API.

examples/domainmodel/src/cli/generator.ts
Pluralia: Yes, but I want to underline, that here we have exactly 3 functions. Should I anyway replace it to `Array`?

msujew: Actually, yeah, nevermind.

examples/domainmodel/src/cli/cli-util.ts
msujew: Right, we usually use a generic constraint for that in most of the Langium code which would look like this `extract<T extends AstNode>(...): T`. Though it is not necessary for this case.

examples/domainmodel/src/cli/cli-util.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (!new RegExp(`^.*${extensions?.join('|')}$`).test(fileName)) {
+        console.error(`Please, choose a file with one of these extensions: ${extensions}.`);
+        process.exit(1);
+    }
+
+    if (!fs.existsSync(fileName)) {
+        console.error(`File ${fileName} doesn't exist.`);
+        process.exit(1);
+    }
+    const fileContent = fs.readFileSync(fileName).toString();
```
msujew, line 21: Use `fs.readFileSync(fileName, 'utf-8')` instead of `fs.readFileSync(fileName).toString()`

examples/domainmodel/src/cli/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+program
+    .command('generate')
+    .argument('<file>', 'the .dmodel file')
+    .option('-d, --destination <dir>', 'destination directory of generating')
+    .description('generate Java classes from .dmodel file')
+    .action((fileName: string, opts: GenerateOptions) => {
+        const metaData = new DomainModelLanguageMetaData();
+        const domainmodel = extractAstNode<Domainmodel>(fileName, metaData.languageId, metaData.extensions, createDomainModelServices());
+        new DomainModelGenerator(domainmodel, fileName, opts.destination).generate();
```
msujew, line 28: I'd like to see a success message once the domain model generator finishes.

examples/domainmodel/src/cli/generator.ts
```diff
@@ -0,0 +1,83 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import * as fs from 'fs';
```
msujew, line 7: Use `import fs from 'fs';` instead of `import * as fs from 'fs';`

examples/domainmodel/src/cli/cli-util.ts
```diff
@@ -0,0 +1,48 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import * as fs from 'fs';
```
msujew, line 7: Use `import fs from 'fs';` instead of `import * as fs from 'fs';`

msujew (approved):
LGTM, though I have left some smaller nitpick comments.

examples/domainmodel/src/cli/cli-util.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    const document = LangiumDocumentConfiguration.create(`file:${fileName}`, languageId, 0, fileContent);
+    services.documents.DocumentBuilder.build(document);
+    if (!document.parseResult) {
+        console.error(`Failed to parse the grammar file ${fileName}`);
+        process.exit(1);
+    }
+
+    const validationErrors = new DefaultDocumentValidator(services)
+        .validateDocument(document)
```
spoenemann, line 31: Validation is already done by the DocumentBuilder. Please use the BuildResult object, wich contains the parseResult (so there's no need to check whether it's present or not) and the diagnostics.

examples/domainmodel/src/cli/generator.ts
```diff
@@ -0,0 +1,84 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import fs from 'fs';
+import _ from 'lodash';
```
spoenemann, line 8: Please add `lodash` as an explicit dependency to package.json.

examples/domainmodel/src/cli/generator.ts
```diff
@@ -0,0 +1,84 @@
...
+import { CompositeGeneratorNode, IndentNode, NL, processGeneratorNode } from 'langium';
+import { AbstractElement, Domainmodel, Entity, Feature, isEntity, isPackageDeclaration, Type } from '../language-server/generated/ast';
+
+export class DomainModelGenerator {
```
spoenemann, line 12: I'd propose to implement this as a plain function instead:
```typescript
export function generateJava(ctx: GeneratorContext) {
    // ...
}

export interface GeneratorContext {
    domainmodel: Domainmodel;
    destination: string;
    path: string;
}
```

Reason: Here the `class` construct is used to store the parameters for the generator. This is how you would do it in Java or other object-oriented languages, but the TypeScript / JavaScript world is different. I know we already use classes a lot in our code base, but that's for other reasons:
 * Combine one or more service functions in a service class
 * Enable fine-grained customization with template methods
So our usage of classes is constrained to _service implementations_. We compared this with possible alternatives that work with plain functions, but found the class-based code to be most readable in that case. But in cases where plain functions can be used, I'd prefer them over classes.

Since this is a fundamental decision, it would be good to discuss it more broadly. @langium/committers WDYT?

msujew: I agree with Miro here: We should try to stay away from classes as much as possible.

dhuebner: Sounds reasonable to me as well

jbicker: I say "Yes!" to that too.

Pluralia: Okay, I'll replace this to a plain function (in other CLIs too)

examples/domainmodel/src/cli/generator.ts
```diff
@@ -0,0 +1,79 @@
...
+    fileNode.indent(classBody => {
+        const featureData = entity.features.map(f => generateFeature(f, classBody));
+        featureData.forEach(([generateField, , ]) => generateField());
+        featureData.forEach(([, generateSetter, generateGetter]) => { generateSetter(); generateGetter(); } );
+    });
+    fileNode.append('}', NL);
+}
+
+function generateFeature(feature: Feature, classBody: IndentNode): [() => void, () => void, () => void] {
+    const name = feature.name;
```
spoenemann, line 55: There is a problem when the feature name is the same as a Java keyword. For example, the qualified-names example can't be compiled because there is a feature named `this`:
```
example/qualifiednames/baz/E3.java:4: error: as of release 8, 'this' is allowed as the parameter name for the receiver type only
    private E4 this;
               ^
```
The solution could be a consistent renaming in such cases, e.g. using `_` as prefix (`_this`).

.gitignore
```diff
@@ -6,4 +6,5 @@ lib/
 out/
 node_modules/
 *.vsix
-*.tsbuildinfo
\ No newline at end of file
+*.tsbuildinfo
+examples/*/example/generated
```
spoenemann, line 10: The output files now seem to be generated into subfolders named after the source file: `blog` and `qualifiednames`. So this ignore line has no effect.

Please create a new `.gitignore` file in `examples/domainmodel` listing the generated subfolders to be ignored.

examples/domainmodel/bin/cli.js
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
```
spoenemann, line 1: Please remove the file extension `.js`: command line tools usually have none.

spoenemann (commented):
Thanks! Here are my last three comments, then it's good to go!

examples/domainmodel/src/cli/generator.ts
Pluralia: I propose to leave generation how it is:
- If I use my DSL for generating Java classes, I know about Java keywords and can not use them as names
- It's necessary to consider all Java keywords
- If we support features renaming, we need to support it for classes too
- Maybe, renaming / adding prefix is not the most expected behaviour

However, names clashing is still a problem now. So, I renamed `this` --> `that`.

examples/domainmodel/src/cli/generator.ts
```diff
@@ -0,0 +1,79 @@
...
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import fs from 'fs';
+import _ from 'lodash';
+import { CompositeGeneratorNode, IndentNode, NL, processGeneratorNode } from 'langium';
+import { AbstractElement, Domainmodel, Entity, Feature, isEntity, isPackageDeclaration, Type } from '../language-server/generated/ast';
+
+export function generateJava(domainmodel: Domainmodel, fileName: string, destination: string = '.') {
+    const path = fileName.replace(/\..*$/, '').replaceAll(/[.-]/g, '');
```
spoenemann, line 13: The `replaceAll` string method is supported only by NodeJS 15, but we want to support NodeJS 12 as minimum version. Please remove all usages of that method and implement this differently.

See NodeJS compatibility here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility

spoenemann (approved):
Great work!

PR #161, opened 2021-07-06
Add support for hidden tokens in CST and improve `longer_alt` token behavior for terminals

### LINKED ISSUE #136: Parser generator does not take terminals into account for `longer_alt`
Usually, tokens that start with the same characters have to include the `longer_alt` property to determine what token may come next. This is already working for keyword based token, but not for terminals.

Langium version: 0.1.0
Package name: `langium-cli`

## Steps To Reproduce

1. Create a terminal starting with a `//` (like `SL_COMMENT`)
2. Use `/` as a keyword in a rule.
3. `//` will never be parsed correctly.

## The current behavior

The terminal starting with `//` will never be parsed because it is in conflict with the `/` keyword.

## The expected behavior

The terminal is parsed correctly.

msujew:
Closes #136

Right now, Chevrotain only supports a single terminal for the `longer_alt` property. For this reason, we use a single `COMMENT` terminal for both single line and multi line comments in the `arithmetics` example.


dhuebner (commented):
Two things I noticed during fast testing:
- no syntax highlighting :-(
- error is reported (typing/CA) inside ML comment here: `def foo:  2 /*  */ + 4;` (notice that there must be a warning "Expression could be normalized to constant 6" to reproduce)

Bildschirmfoto 2021-07-08 um 14 31 12

msujew:
1. Fixed the missing `#comments` syntax highlighting pattern. 
2. What would be the expected behavior for the second issue in your opinion?

dhuebner:
> * Fixed the missing `#comments` syntax highlighting pattern.

Works!

> * What would be the expected behavior for the second issue in your opinion?

Most important: no error (annoying popup):
 `[Error - 12:40:28 PM] Request textDocument/completion failed.
  Message: Request textDocument/completion failed with message: Cannot read property 'text' of undefined
  Code: -32603 
` 
I don't know why this error pops up only if the part of ML commit is inside an issuer marker  
I would expect the same behavior (default hippie CA) like in a cases without a validation marker: `/*  plain ML comment */`


In a long term we need to think about how a user can provide CA inside a ML comment (e.g. like in Javadoc), but for such a case one will probably define a new rule

msujew:
Alright, I see. CA for comments or strings (which is also behaving strange) is out of scope for this PR. How does Xtext approach this issue? Does it simply skip CA for terminals, which shouldn't display CA?

dhuebner:
I thought it‘s a hidden rule problem. Seem to be not, so I think we can merge it. In xtext you can provide CA for keywords and other Terminal rules like strings. It also possible to provide proposals inside hidden nodes, but not out of the box

dhuebner (approved):

PR #169, opened 2021-07-09
New parser infrastructure

msujew:
This new parser infrastructure allows three things:

1. We no longer have to generate a parser + tokens. Everything is build in memory.
2. We no longer have to generate a grammar access as it was only used by the generated parser.
3. Enables future playground support, since the parser does not need to be generated anymore.

Performance:

I was actually curious how this change would impact performance. Running the parser 1000 times over the langium grammar took 6300ms on avg with the old infrastructure, while the new one took only 4800ms. I would have expected it to perform worse, but it seems like it actually performs better.


msujew:
@spoenemann WDYT of introducing a service like `TokenBuilder` that allows users to create their own tokentypes based on terminals and keywords. It would allow easy customization of the tokenizer like we planned for using e.g. different lexer modes.

spoenemann:
Thanks for the proposal! There is only one potential drawback that comes to my mind: understanding and debugging the parser becomes much harder. It will likely end up setting conditional breakpoints. This is an issue that we should not underestimate, so the advantages need to be weighed against that.

I see enabling playground as a nice-to-have, but would not align architectural decisions to that feature.

msujew:
@spoenemann I guess that makes sense. We could still keep the generated parser and add some sort of debug mode to the langium-cli. That way debugging the grammar would still work fine.

FYI this change has a few more advantages than just enabling playground support. For example, the newly introduced `TokenBuilder` service allows users to easily control the creation of tokens from terminals and keywords. This allows to introduce different lexer modes, case insensitive keywords, etc, without much overhead. This wouldn't be possible with a purely generated approach.

dhuebner:
@msujew 
> We could still keep the generated parser and add some sort of debug mode to the langium-cli

You mean activating debugging parser generation with a setting/flag? 

Overall it sounds very good to mean. I especially like the playground "feature" :)

msujew:
> You mean activating debugging parser generation with a setting/flag?

Right. With that flag in place, parser generation happens as it currently is. But that would lead to a lot of overhead, as every parser feature has to be duplicated. We also run the risk of divergent behavior on the generated and the in-memory parser. E.g. the generated parser uses different tokens than the in-memory parser, once the user overrides the `TokenBuilder`.

msujew:
> For the framework itself it was a bit more important (e.g. trying to get towards partial parsing)

I think this is Miro's main concern. There are still a few unfinished parser features like parameterized rules. For this we could make good use of the debuggability of the parser.

> I think there are more robust ways to refer to grammar elements if necessary.

There are, and we are using them. For example in validations, when we want to show the validation on a certain cst node associated with a grammar element we can simply pass `{ node: astNode, property: 'name' } ` to the validators accept method. The `'name'` part is actually derived from the type of `astNode` and is correctly autocompleted and validated by TypeScript. Making it a lot safer for grammar refactoring.

> The strongly typed AST interfaces where a game changer. Would you get rid of these as well?

Definitely not. There are still a few files to be generated by the `langium-cli` like the AST interfaces, the generated DI module and the serialized grammar.

meysholdt:
if you can allow two editors in VS-Code side-by-side: one to edit the grammar and the older with an example that's being parsed+validated... that would be a killer show-case and inner-dev-loop. I would have loved to have this for Xtext :) 

If you also add visualizations / logs about what the parser is doing, this may be the better way to debug a parser compared to stepping through generated code. 

it's clear that in the editor some IDE features would not be available since you wan to be careful about loading code from a workspace into the IDE that's editing it.

msujew:
@meysholdt exactly, that's basically what I'm trying to achieve here. Interpreting the grammar would allow for some sort of playground support. This is currently planned as a component of our Langium website. I'm not sure how the solution with vscode would work, but an online playground has proven to be quite useful for different projects.

> it's clear that in the editor some IDE features would not be available

I don't see a reason for this, at least for the online playground. Currently all language features of Langium should be able to run in an 'interpreter' mode in a browser environment.

packages/langium/src/parser/token-builder.ts
```diff
@@ -0,0 +1,94 @@
...
+import { partialMatches } from '../utils/regex-util';
+import { stream } from '../utils/stream';
+
+export interface TokenBuilder {
+    buildTokens(grammar: Grammar): TokenType[];
+}
+
+export class DefaultTokenBuilder implements TokenBuilder {
+
+    buildTokens(grammar: Grammar): TokenType[] {
```
spoenemann, line 19: Could we keep a generated `TokenType[]` array to speed up the startup time of Langium?

packages/langium/src/parser/langium-parser.ts
```diff
@@ -4,365 +4,212 @@
...
 }
 
-export const DatatypeSymbol = Symbol('Datatype');
-
-type RuleResult = () => any;
+type Method = () => void;
 
-export class LangiumParser {
-    readonly grammarAccess: GrammarAccess;
+export class LangiumParser extends LangiumBaseParser {
```
spoenemann, line 25: I think we should use composition instead of inheritance here. The two classes have different responsibilities..

packages/langium/src/parser/token-builder.ts
msujew: Yes/no. This is a premature optimization. I'll keep this in mind if we notice any startup performance issues 👍

packages/langium/src/parser/langium-parser.ts
msujew: You're right. I changed it that way 👍

spoenemann (approved):
Great!

PR #170, opened 2021-07-10
Generate language meta data

### LINKED ISSUE #168: Provide Language id and file extension information
I need a service where I can get the language id and the file extension.
Language id can be fetched from the document, but I do not have one when indexing

msujew:
Closes #168 

Adds a dedicated `LanguageMetaData` service to each language.

```ts
export interface LanguageMetaData {
    languageId: string;
    extensions: string[];
}
```


examples/arithmetics/src/language-server/generated/meta-data.ts
```diff
@@ -0,0 +1,9 @@
+/******************************************************************************
+ * This file was generated by langium-cli 0.1.0.
+ * DO NOT EDIT MANUALLY!
+ ******************************************************************************/
+
+export class ArithmeticsLanguageMetaData {
+    languageId = 'arithmetics';
+    extensions = ['.calc'];
```
dhuebner, line 8: Do we need the leading `.` ?  I saw some js API returns extensions wit a dot `Path.extname`, but some do not

packages/langium/src/grammar/generated/meta-data.ts
```diff
@@ -0,0 +1,9 @@
+/******************************************************************************
+ * This file was generated by langium-cli 0.1.0.
+ * DO NOT EDIT MANUALLY!
+ ******************************************************************************/
+
+export class LangiumGrammarLanguageMetaData {
+    languageId = 'langium';
+    extensions = ['langium'];
```
dhuebner, line 8: Here is `.` missing :)

dhuebner:
Great! Exactly what I need. We need to decide about having a leading `.` or not. 
I tend to prefer the no-dot solution

Pluralia:
Leading `.` are redundant, however, it's nice to have it to support the association: "if a string starts with `.`, it's an extension". It highlights extensions over all other metadata.

msujew:
I prefer the `.` variation as well, since `path.extname` will return the ext with a `.`

examples/arithmetics/src/language-server/generated/meta-data.ts
msujew: The `extensions` now always start with a leading dot.

dhuebner (approved):

examples/arithmetics/src/language-server/generated/meta-data.ts
```diff
@@ -0,0 +1,9 @@
+/******************************************************************************
+ * This file was generated by langium-cli 0.1.0.
+ * DO NOT EDIT MANUALLY!
+ ******************************************************************************/
+
+export class ArithmeticsLanguageMetaData {
```
spoenemann, line 6: In this case I'd rather use a plain object than a class.

packages/langium/src/grammar/language-meta-data.ts
```diff
@@ -0,0 +1,10 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+export interface LanguageMetaData {
+    languageId: string;
+    extensions: string[];
```
spoenemann, line 9: Shall we name this property `fileExtensions` for clarity? "extensions" is quite generic.

examples/arithmetics/src/language-server/generated/meta-data.ts
msujew: I see, that makes sense. Should the object this be generated into a `meta-data.ts` file, or should it be build in the generated injection module? like this:

```ts
LanguageMetaData: () => ({
    languageId: 'id',
    extensions: ['.id']
})
```

spoenemann: Exporting a `const` for it could be useful so you can access it directly in your language implementation. But it could be in the `generated/module.ts` instead of a new file to keep the number of generated files low.

msujew: Alright, yeah I agree with that.

PR #172, opened 2021-07-12
WS Index. Cross file linking

dhuebner:
- scoping, linking, CA, go-to works
- find global references works
- `AstNode -> Path` & `Path -> AstNode` works


packages/langium/src/index/ast-node-locator.ts
```diff
@@ -0,0 +1,60 @@
...
+            if(currentValue.length === 0)
+                return previousValue;
+            const propertyIndx = currentValue.split('@');
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            const value = (previousValue as any)[propertyIndx[0]];
+            return (propertyIndx.length === 2) ? value[propertyIndx[1]] : value;
+        }, document.parseResult?.value);
+    }
+
+    protected pathSegment(node: AstNode, container: AstNode): string {
```
msujew, line 44: WDYT about supporting XPath here? The current syntax looks a lot like Xtext, I would argue for something more web-native.

dhuebner: As the default implementation?  Maybe, if it fits all our needs. In the signature probably not. I would like to keep it `string` as I already saw projects that extended at least the fragment URI part

msujew: Yes, I wouldn't change the return type, but maybe using the XPath schema would improve interaction with third-party libraries. It's just a thought though, I'm fine with the current implementation.

dhuebner:
Some progress on syncing index and document state across documents:
document-sync

msujew:
@spoenemann As vscode does evil stuff with its URIs that are passed to the language server, I resorted to using the `vscode-uri` package to deal with that. You probably know that one from Theia already.

Do we want to use the URI internally as well? Currently, everything is simply strings, which seems highly error-prone and unsafe to me, due to differences in schema and uri encoding.

Also, this is ready for review, I think I addressed everything that we mentioned earlier. At least during manual testing I couldn't notice any further issues.

packages/langium/src/utils/ast-util.ts
```diff
@@ -177,4 +179,17 @@ export function findLocalReferences(targetNode: AstNode, lookup: AstNode): Strea
     };
     streamAllContents(lookup).forEach(process);
     return stream(refs);
+}
+
+export function findAllReferences(targetNode: AstNode, astNodePath: string, referenceLookUp: IndexManager): Stream<AstNodeReferenceDescription> {
```
spoenemann, line 184: Utility functions that take an injectable service as argument are a smell; see comment above.

packages/langium/src/lsp/language-server.ts
```diff
@@ -36,25 +36,53 @@ export function startLanguageServer(services: LangiumServices): void {
...
+        if (newDocument.parseResult?.value) {
+            services.index.IndexManager.update(newDocument);
+        }
+        langiumDocs.all.filter(doc => isAffected(doc, change.document.uri)).forEach(
+            doc => {
+                documentBuilder.build(doc);
+                services.index.IndexManager.update(doc);
+            }
+        );
+        documentBuilder.build(newDocument);
```
spoenemann, line 80: Should this logic be part of `LangiumDocuments`?  Or of `IndexManager`?

packages/langium/src/references/linker.ts
```diff
@@ -5,26 +5,40 @@
...
+        this.services = services;
         this.scopeProvider = services.references.ScopeProvider;
-        this.reflection = services.AstReflection;
     }
 
     link(node: AstNode, referenceName: string, referenceId: string): AstNode | undefined {
+        const description = this.linkingCandiates(node, referenceName, referenceId);
+        if (description)
+            return this.loadAstNode(description);
+        return undefined;
```
spoenemann, line 30: We discussed that the linker should be responsible of creating Reference objects, so the logic in the `buildReference` method in LangiumParser should be moved here.

spoenemann (commented):
Thanks @msujew! Yes, probably it makes sense to work with a URI object instead of strings so we can ensure a consistent format. We started with string because that's what the LSP API offers.

An aspect that is currently missing is to ensure that the validator does not link cross-references to other documents while validating them. This could be done by using the `getCandidate` method of the linker (see remark below).

packages/langium/src/documents/document.ts
```diff
@@ -4,24 +4,114 @@
...
 import { AstNodeDescription } from '../references/scope';
+import { LangiumServices, LanguageMetaData, Stream, stream } from '..';
+import { URI } from 'vscode-uri';
 
-export interface LangiumDocument extends TextDocument {
+export interface LangiumDocument {
     parseResult?: ParseResult
     precomputedScopes?: PrecomputedScopes
+    outdated: boolean
+    textDocument: TextDocument
```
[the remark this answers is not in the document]
msujew: Well, the `TextDocument` provides us with a lot of utility. E.g. getting the full text and also offset/position conversion, which we often need outside of the LSP as well. I don't see a reason to not use it.

packages/langium/src/documents/document.ts
```diff
@@ -4,24 +4,114 @@
...
+export class DefaultTextDocumentFactory implements TextDocumentFactory {
+
+    protected readonly languageMetaData: LanguageMetaData;
+
+    constructor(services: LangiumServices) {
+        this.languageMetaData = services.LanguageMetaData;
+    }
+
+    create(uri: string): TextDocument {
+        const content = fs.readFileSync(URI.parse(uri).fsPath, 'utf-8');
```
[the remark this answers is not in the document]
msujew: Yes, I also thought about this. In the long run we would want some kind of `langium/node` namespace which encapsulates file system stuff and so on. That way we would keep everything web-compatible.

packages/langium/src/documents/document.ts
```diff
@@ -4,24 +4,114 @@
...
+        this.parser = services.parser.LangiumParser;
+    }
+
+    create(textDocument: TextDocument): LangiumDocument {
+        const doc: LangiumDocument = {
+            outdated: false,
+            textDocument
+        };
+        const parseResult = this.parser.parse(doc);
+        doc.parseResult = parseResult;
```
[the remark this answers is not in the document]
msujew: We should, yes. It doesn't make a lot of sense to create a document without parsing it in the first place 👍

packages/langium/src/documents/document.ts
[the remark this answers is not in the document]
spoenemann: > getting the full text and also offset/position conversion

The full text is available through the CST. We don't need that conversion as long as we stick to using offsets internally, but we could consider adding utility functions for this like in the Xtext node model.

> I don't see a reason to not use it.

We should look into memory consumption, and be careful with potential problems with duplicate instances of TextDocument. Do we track all these instances in `TextDocuments`?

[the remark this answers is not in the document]
msujew: > We should look into memory consumption, and be careful with potential problems with duplicate instances of TextDocument. Do we track all these instances in TextDocuments?

We don't we throw them away after the document is opened in the editor. Also, memory consumption shouldn't be an issue, as we don't copy the text, and just pass a reference to it to our newly created `TextDocument`. The only overhead is the text document object itself, which should be quite minimal.

packages/langium/src/index/ast-descriptions.ts
```diff
@@ -0,0 +1,112 @@
...
+            const name = this.services.references.NameProvider.getName(rooNode);
+            if (name) {
+                descr.push(this.createDescription(rooNode, name, document));
+            }
+            streamContents(rooNode).forEach(content => {
+                const name = this.services.references.NameProvider.getName(content.node);
+                if (name) {
+                    descr.push(this.createDescription(content.node, name, document));
+                }
+            });
```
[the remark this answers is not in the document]
dhuebner: @spoenemann 
> that is those directly contained in the root node

I thought `streamContents(rooNode)`does exactly that

packages/langium/src/lsp/language-server.ts
dhuebner: I was thinking about it as I implemented this parts, but I can't identify any existing service which can/should own this code. This lines are connecting/synching parts between Index and Documents storage; isAffected belongs to that probably too.

packages/langium/src/documents/document-builder.ts
```diff
@@ -55,6 +59,32 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+            // Send the computed diagnostics to VS Code.
+            this.connection.sendDiagnostics({ uri: document.textDocument.uri, diagnostics });
+        }
+        return diagnostics;
+    }
+
+    documentChanged(uri: string): void {
+        this.langiumDocuments.invalidateDocument(uri);
+        const newDocument = this.langiumDocuments.createOrGetDocument(uri);
+        if (newDocument.parseResult?.value) {
```
spoenemann, line 76: `parseResult` is mandatory, see above.

packages/langium/src/documents/document-builder.ts
```diff
@@ -63,3 +93,6 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
     }
 
 }
+function isAffected(document: LangiumDocument, changedUri: string): boolean {
+    return changedUri !== document.textDocument.uri;
+}
```
spoenemann, line 98: Is this a placeholder? Change affection should be overridable (a template method or separate service) and should consider references between documents.

packages/langium/src/documents/document-builder.ts
```diff
@@ -55,6 +59,32 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
         };
     }
 
+    validate(document: LangiumDocument): Diagnostic[] | undefined {
+        let diagnostics: Diagnostic[] | undefined;
+        const validator = this.documentValidator;
+        if (this.connection) {
```
spoenemann, line 65: This method is used only in `build`, which already checks whether `connection` is available. Do we need this method at all? We could do the same within `build` as before.

packages/langium/src/documents/document.ts
```diff
@@ -4,24 +4,115 @@
...
+        this.textDocuments = services.documents.TextDocuments;
+        this.textDocumentFactory = services.documents.TextDocumentFactory;
+        this.langiumDocumentFactory = services.documents.LangiumDocumentFactory;
+    }
+
+    get all(): Stream<LangiumDocument> {
+        return stream(this.documentMap.values());
+    }
+
+    createOrGetDocument(uri: string): LangiumDocument {
```
spoenemann, line 92: Maybe `getOrCreateDocument` to emphasize the precedence?

packages/langium/src/documents/document.ts
```diff
@@ -4,24 +4,115 @@
...
+        }
+        langiumDoc = this.langiumDocumentFactory.fromTextDocument(textDoc);
+        this.documentMap.set(uri, langiumDoc);
+        return langiumDoc;
+    }
+
+    invalidateDocument(uri: string): void {
+        const langiumDoc = this.documentMap.get(uri);
+        if (langiumDoc) {
+            langiumDoc.valid = false;
```
spoenemann, line 109: Sorry for the back-and-forth, but seeing this I wonder what the `valid` flag and the "invalidate" methods mean. Do we find a better wording? Maybe
```typescript
langiumDoc.dirty = true;
```
and name the methods `markDirty`, `markAllDirty`?

spoenemann: Also, I don't find any code that sets `valid` to `true` again after it has been invalidated. Is that missing in `DocumentBuilder`?

packages/langium/src/index/ast-descriptions.ts
```diff
@@ -0,0 +1,113 @@
...
+export interface AstNodeDescriptionProvider {
+    createDescription(node: AstNode, name: string, document: LangiumDocument): AstNodeDescription;
+    createDescriptions(document: LangiumDocument): AstNodeDescription[];
+}
+
+export interface ReferenceDescriptionProvider {
+    createDescriptions(document: LangiumDocument): ReferenceDescription[];
+}
+
+export class DefaultAstNodeDescriptionsProvider implements AstNodeDescriptionProvider {
```
spoenemann, line 34:
```suggestion
export class DefaultAstNodeDescriptionProvider implements AstNodeDescriptionProvider {
```

packages/langium/src/index/ast-descriptions.ts
```diff
@@ -0,0 +1,113 @@
...
+        const rooNode = document.parseResult?.value;
+        if (rooNode) {
+            const name = this.nameProvider.getName(rooNode);
+            if (name) {
+                descr.push(this.createDescription(rooNode, name, document));
+            }
+            streamContents(rooNode).forEach(content => {
+                const name = this.nameProvider.getName(content.node);
+                if (name) {
+                    descr.push(this.createDescription(content.node, name, document));
```
spoenemann, line 65: The index should not contain references to the actual AST nodes. Maybe we need two different variants of `createDescription`, one for local scope and one for global scope?

packages/langium/src/index/index-manager.ts
```diff
@@ -0,0 +1,125 @@
...
+import { AstNode } from '../syntax-tree';
+import { stream, Stream } from '../utils/stream';
+import { getDocument } from '../utils/ast-util';
+import { AstReflection } from '..';
+
+export interface IndexManager {
+    initializeRoot(rootUri: string): void;
+    initializeWorkspace(folders: WorkspaceFolder[] | null): void;
+    update(document: LangiumDocument): void;
+    /* Use streams? */
```
spoenemann, line 25: Yes, please remove comment.

packages/langium/src/index/index-manager.ts
```diff
@@ -0,0 +1,125 @@
...
+                if(refDescr.targetUri === targetDocUri && refDescr.targetPath === astNodePath) {
+                    result.push(refDescr);
+                }
+            });
+        });
+        return stream(result);
+    }
+
+    allElements(referenceType?: string): Stream<AstNodeDescription> {
+        return stream(Array.from(this.simpleIndex.values()).flat().filter(e => referenceType ? this.astReflection.isSubtype(e.type, referenceType) : true));
```
spoenemann, line 67: More efficient:
```typescript
return stream(this.simpleIndex.values()).flat().filter(...);
```

packages/langium/src/lsp/language-server.ts
```diff
@@ -36,25 +36,41 @@ export function startLanguageServer(services: LangiumServices): void {
...
+    const langiumDocs = services.documents.LangiumDocuments;
     const documents = services.documents.TextDocuments;
     const documentBuilder = services.documents.DocumentBuilder;
     documents.onDidChangeContent(change => {
-        documentBuilder.build(change.document);
+        documentBuilder.documentChanged(change.document.uri);
+    });
+    documents.onDidClose(() => {
+        if (documents.keys().length === 0)
+            langiumDocs.invalidateAllDocuments(); // clean up all cached document
```
spoenemann, line 72: Doesn't that just mark them dirty? Cleaning up would mean removing all LangiumDocuments from internal maps etc.

packages/langium/src/index/index-manager.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    constructor(services: LangiumServices) {
+        this.astReflection = services.AstReflection;
+        this.languageMetaData = services.LanguageMetaData;
+        this.langiumDocuments = () => services.documents.LangiumDocuments;
+        this.astNodeDescriptionProvider = () => services.index.AstNodeDescriptionProvider;
+        this.referenceDescriptionProvider = () => services.index.ReferenceDescriptionProvider;
+    }
+
+    documentDescriptions(): ReadonlyMap<string, ReferenceDescription[]> {
```
spoenemann, line 49: This is currently not used.

spoenemann (commented):
General remark: please add comments to interfaces, public properties and methods, at least where it's not self-explaining.

packages/langium/src/index/index-manager.ts
msujew: Is it really more efficient though? If we iterate more than once over the returned stream, we need to flatten and filter everything on every call, don't we? As everything is lazily evaluated.

packages/langium/src/documents/document.ts
msujew: Well we throw a document simply away after it has changed, so we never need to set its `valid` state to `true`.

packages/langium/src/lsp/language-server.ts
dhuebner: I will remove it

packages/langium/src/index/index-manager.ts
spoenemann: That's responsibility of the caller. If the caller knows they will iterate multiple times, they can convert it to an array first. If they iterate only once, it's more efficient to operate on the stream.

packages/langium/src/documents/document-builder.ts
dhuebner: Moved to service implementation. 
> and should consider references between documents.

Yes, I know. I would like to keep it that simple and create tests first. Afterwards I will provide a more complex and performant version.

packages/langium/src/documents/document-builder.ts
dhuebner: If I don't check this.connection I will need to do `!.`: `this.connection!.sendDiagnostics({ uri: document.textDocument.uri, diagnostics });`

> Do we need this method at all? 

I think yes. First, to easily override only this aspect of building a document. Second, I am still not sure that validation should be a part of document building. For not opened documents I wold like to do only parsing and linking for example. I think we can target that refactoring in the next PR's, what do you think?

packages/langium/src/documents/document-builder.ts
```diff
@@ -25,27 +28,28 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+        this.indexManager = services.index.IndexManager;
     }
 
     build(document: LangiumDocument): BuildResult {
-        const parseResult = this.parser.parse(document);
-        document.parseResult = parseResult;
+        if (!document.parseResult) {
+            const parseResult = this.parser.parse(document);
+            document.parseResult = parseResult;
+        }
```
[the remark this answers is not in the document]
dhuebner: Will fix places where I already have changes, but will not touch the rest. I would like to open a new PR/Issue for the parseResult clean-up to reduce the amount of unrelated changes in this PR.

packages/langium/src/index/index-manager.ts
dhuebner: Removed

packages/langium/src/index/index-manager.ts
dhuebner: >That's responsibility of the caller

> they can convert it to an array first

My impression is that we should probably keep the `allElements()` as it was by returning really all references as an array and add an additional API `elementsByType(referenceType?: string): Stream<>` or maybe even `elementsBy(filter: (node: AstNodeDescription) => boolean): Stream<AstNodeDescription>;`

Anyway, please do such kind of optimization in a separate Issue/PR.

packages/langium/src/index/ast-descriptions.ts
dhuebner: I'm not sure we should introduce a new interface for that, but it is definitely something we should discuss in a new issue. 
Currently we set `AstNodeDescription.node` to undefined before adding the AstNodeDescription into the index, which is fine so far.

packages/langium/src/documents/document-builder.ts
spoenemann: Yes, the default (but overridable) behavior should be to validate only open documents, which can be checked with the TextDocuments service. Could you create an issue so we don't forget?

packages/langium/src/index/index-manager.ts
spoenemann: Ok, please create an issue then.

packages/langium/src/documents/document-builder.ts
spoenemann: I'd propose to make the method protected and remove it from the interface.

packages/langium/src/documents/document.ts
spoenemann: Oh, so that's what `valid` means. That's definitely something that should be documented in the LangiumDocument interface. It was not obvious to me.

packages/langium/src/references/scope.ts
```diff
@@ -90,13 +97,17 @@ export class DefaultScopeProvider implements ScopeProvider {
...
-        // TODO use the global scope (index) as outermost scope
-        let result: Scope = EMPTY_SCOPE;
+        let result: Scope = this.getGlobalScope(referenceType);
         for (let i = scopes.length - 1; i >= 0; i--) {
             result = new SimpleScope(scopes[i], result);
         }
         return result;
     }
+
+    // TODO use the global scope (index) as outermost scope
```
spoenemann, line 107: Can we remove this comment?

spoenemann (approved):
I'm ok with merging this. I found two problems with the domainmodel example, though. Please create issues for these:

 - We should customize the AstNodeDescriptionProvider for that example so it exports all entities with their qualified names.
 - Find references does not work properly. Example: change line 5 in blog.dmodel to `many posts: E1`, then "Find All Reference" shows one hit from blog.dmodel and one from qualified-names.dmodel when started from the new reference, while it shows two hits from qualified-names.dmodel when started from the declaration.

msujew (approved):
I also observed some interesting issues regarding 'Find references', but they can be solved in another PR. LGTM!

PR #173, opened 2021-07-13
Generate meta data as a plain object

msujew:
Removes unnecessarily created files and simplifies the `LanguageMetaData` setup.


packages/langium-cli/src/generator/module-generator.ts
```diff
@@ -21,10 +21,16 @@ export function generateModule(grammar: langium.Grammar, config: LangiumConfig):
     node.append(
         'import { ', grammar.name, "AstReflection } from './ast';", NL,
         'import { ', grammar.name, "GrammarAccess } from './grammar-access';", NL,
-        'import { ', grammar.name, "LanguageMetaData } from './meta-data';", NL,
         "import { Parser } from './parser';", NL, NL
     );
 
+    node.append('const metaData = {', NL);
```
spoenemann, line 27: Shall we export this so you can access it directly in your language implementation?

packages/langium-cli/src/generator/module-generator.ts
```diff
@@ -21,10 +21,16 @@ export function generateModule(grammar: langium.Grammar, config: LangiumConfig):
...
+    node.indent(metaData => {
+        metaData.append(`languageId: '${config.languageId}',`, NL);
```
spoenemann, line 29: What if this property is undefined? Shall we make it mandatory in the config?

packages/langium-cli/src/generator/module-generator.ts
```diff
@@ -21,10 +21,16 @@ export function generateModule(grammar: langium.Grammar, config: LangiumConfig):
...
+        metaData.append(`fileExtensions: [${config.extensions && config.extensions.map(e => appendQuotesAndDot(e)).join(', ')}]`, NL);
```
spoenemann, line 30: Seeing this I think we could rename `extensions` to `fileExtensions` in the config as well, WDYT?

packages/langium-cli/src/generator/module-generator.ts
msujew: Yes #158 already makes it mandatory.

packages/langium-cli/src/generator/module-generator.ts
msujew: Yeah, that makes sense.

packages/generator-langium/langium-template/package.json
```diff
@@ -29,7 +29,7 @@
     "langium": {
         "languageId": "<%= language-id %>",
         "grammar": "src/language-server/<%= language-id %>.langium",
-        "extensions": <%= file-extension %>,
+        "fileExtensions": <%= file-extension %>,
```
spoenemann, line 32: This needs an entry in the `CHANGELOG.md`. Could you create one in `langium-cli`?

You can take https://github.com/eclipse/lsp4j/blob/master/CHANGELOG.md as an example format.

examples/arithmetics/src/language-server/generated/module.ts
```diff
@@ -6,14 +6,18 @@
 import { LangiumGeneratedServices, LangiumServices, Module } from 'langium';
 import { ArithmeticsAstReflection } from './ast';
 import { ArithmeticsGrammarAccess } from './grammar-access';
-import { ArithmeticsLanguageMetaData } from './meta-data';
 import { Parser } from './parser';
 
+const metaData = {
```
spoenemann, line 11: This needs to be re-generated.

spoenemann (approved):

PR #175, opened 2021-07-15
Add CLI to Yeoman template

Pluralia:
Closes the [issue/157](https://github.com/langium/langium/issues/157).
Adds [hello-world example](https://github.com/langium/langium/tree/pluralia/yeoman-template-cli/examples/hello-world), which is base for the yeoman-template CLI and can be removed after the yeoman-template update.


examples/hello-world/src/cli/index.ts
```diff
@@ -0,0 +1,26 @@
+import colors from 'colors';
+import { Command } from 'commander';
+import { Model } from '../language-server/generated/ast';
+import { HelloWorldLanguageMetaData } from '../language-server/generated/meta-data';
+import { createHelloWorldServices } from '../language-server/hello-world-module';
+import { extractAstNode } from './cli-util';
+import { HelloWorldGenerator } from './generator';
+const metaData = new HelloWorldLanguageMetaData();
```
msujew, line 8: Instead of instantiating a new meta data object here, we should use the one provided by the `LangiumServices`.

spoenemann:
@Pluralia could you apply the same changes as for my comments to #160?

Pluralia:
Should I remove the hello-world example after adding it into yeoman template?

msujew:
> Should I remove the hello-world example after adding it into yeoman template?

@Pluralia yes, please :)

spoenemann:
Looks good, please proceed adding the CLI code to the template. Two remarks:

 * Please leave out the `example` folder
 * Don't generate any file header comment (we don't want to force our copyright and license on the user's code)

spoenemann:
One detail in the console output: there's a double-slash
```
JavaScript code generated successfully: example/generated//example.js
```

PR #178, opened 2021-07-16
Add folding lsp feature

msujew:
Also adds some CST utility methods, like the ability to stream over cst nodes.

Currently has no tests. I will add them once the testing support is done and merged.

### DIFF
--- a/packages/langium/src/default-module.ts
+++ b/packages/langium/src/default-module.ts
@@ -16,6 +16,7 @@ import { DefaultCompletionProvider } from './lsp/completion/completion-provider'
 import { RuleInterpreter } from './lsp/completion/rule-interpreter';
 import { DefaultDocumentHighlighter } from './lsp/document-highlighter';
 import { DefaultDocumentSymbolProvider } from './lsp/document-symbol-provider';
+import { DefaultFoldingRangeProvider } from './lsp/folding-range-provider';
 import { DefaultGoToResolverProvider } from './lsp/goto';
 import { DefaultReferenceFinder } from './lsp/reference-finder';
 import { DefaultRenameHandler } from './lsp/rename-refactoring';
@@ -38,6 +39,9 @@ export type DefaultModuleContext = {
 export function createDefaultModule(context: DefaultModuleContext = {}): Module<LangiumServices, LangiumDefaultServices> {
     return {
         parser: {
+            GrammarConfig: () => ({
+                multilineCommentRules: ['ML_COMMENT']
+            }),
             LangiumParser: (injector) => createLangiumParser(injector),
             ValueConverter: () => new DefaultValueConverter(),
             TokenBuilder: () => new DefaultTokenBuilder()
@@ -56,7 +60,8 @@ export function createDefaultModule(context: DefaultModuleContext = {}): Module<
             },
             Connection: () => context.connection,
             DocumentSymbolProvider: (injector) => new DefaultDocumentSymbolProvider(injector),
-            ReferenceFinder:  (injector) => new DefaultReferenceFinder(injector),
+            FoldingRangeProvider: (injector) => new DefaultFoldingRangeProvider(injector),
+            ReferenceFinder: (injector) => new DefaultReferenceFinder(injector),
             GoToResolver: (injector) => new DefaultGoToResolverProvider(injector),
             DocumentHighlighter: (injector) => new DefaultDocumentHighlighter(injector),
             RenameHandler: (injector) => new DefaultRenameHandler(injector)
--- a/packages/langium/src/lsp/folding-range-provider.ts
+++ b/packages/langium/src/lsp/folding-range-provider.ts
@@ -1,0 +1,138 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { FoldingRange, FoldingRangeKind } from 'vscode-languageserver';
+import { LangiumDocument } from '../documents/document';
+import { LeafCstNodeImpl } from '../parser/cst-node-builder';
+import { LangiumServices } from '../services';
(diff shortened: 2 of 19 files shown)


packages/langium/src/services.ts
```diff
@@ -43,12 +44,14 @@ export type LangiumLspServices = {
...
     DocumentHighlighter: DocumentHighlighter
     DocumentSymbolProvider: DocumentSymbolProvider
+    FoldingRangeProvider: FoldingRangeProvider
     GoToResolver: GoToResolver
     ReferenceFinder: ReferenceFinder
 }
 
 export type LangiumDefaultServices = {
     parser: {
+        MultilineComment: string
```
spoenemann, line 54: I'd propose to put this into a object like this (naming can be discussed):
```typescript
export interface GrammarConfig {
    multilineCommentRules: string[]
}
```

packages/langium/src/lsp/folding-range-provider.ts
```diff
@@ -0,0 +1,112 @@
...
+                    }
+                } while (!result.done);
+            }
+
+            this.collectCommentFolding(document, root, acceptor);
+        }
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    protected isHandled(node: AstNode): boolean {
```
spoenemann, line 61: Please add comments explaining these template methods.

packages/langium/src/services.ts
msujew: Sure, makes sense.

packages/langium/src/test/langium-test.ts
```diff
@@ -55,6 +52,22 @@ export function expectSymbols(services: LangiumServices, cb: ExpectFunction): (i
     };
 }
 
+export function expectFoldings(services: LangiumServices, cb: ExpectFunction): (input: ExpectedBase) => void {
```
spoenemann, line 55: General remark: Wouldn't it be nicer to use a more expressive local name for the callback, like `expectEqual`?

packages/langium/src/lsp/folding-range-provider.ts
```diff
@@ -0,0 +1,120 @@
...
+        const { start, end } = node.range;
+        const startPosition = document.positionAt(start);
+        let endPosition = document.positionAt(end);
+        // Don't generate foldings for nodes that are less than 3 lines
+        if (endPosition.line - startPosition.line < 2) {
+            return undefined;
+        }
+        // As we don't want to hide the end token like 'if { ... --> } <--',
+        // we simply select the end of the previous line as the end position
+        endPosition = document.positionAt(document.offsetAt({ line: endPosition.line, character: 0 }) - 1);
```
spoenemann, line 116: This is only useful if the end token is actually a closing brace. For grammar rules, for example, it would be better to hide the whole definition. Can this be solved here generically, or do we need a customization for the grammar language?

msujew: I would argue that the default behavior should not include the last line of the object in the folding. This is basically how every programming language does it:

image

But yes, it makes sense for the grammar language. I'll implement a template method for this.

packages/langium/src/lsp/folding-range-provider.ts
```diff
@@ -0,0 +1,141 @@
...
+        }
+        return FoldingRange.create(startPosition.line, endPosition.line, startPosition.character, endPosition.character, kind);
+    }
+
+    /**
+     * Template method to determine whether the folding range for this cst node should include its last line.
+     * Returns true by default for ast nodes which end in braces and for comments.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    protected addLastFoldingLine(node: CstNode, kind?: string): boolean {
```
spoenemann, line 127: I'd name this `includeLastFoldingLine`.

packages/langium/src/lsp/folding-range-provider.ts
```diff
@@ -0,0 +1,141 @@
...
+        // we simply select the end of the previous line as the end position
+        if (!this.addLastFoldingLine(node, kind)) {
+            endPosition = document.positionAt(document.offsetAt({ line: endPosition.line, character: 0 }) - 1);
+        }
+        return FoldingRange.create(startPosition.line, endPosition.line, startPosition.character, endPosition.character, kind);
+    }
+
+    /**
+     * Template method to determine whether the folding range for this cst node should include its last line.
+     * Returns true by default for ast nodes which end in braces and for comments.
```
spoenemann, line 124: Looking at the code, it's the other way around.

packages/langium/src/default-module.ts
```diff
@@ -33,6 +34,9 @@ export type DefaultModuleContext = {
 export function createDefaultModule(context: DefaultModuleContext = {}): Module<LangiumServices, LangiumDefaultServices> {
     return {
         parser: {
+            GrammarConfig: () => ({
+                multilineCommentRules: ['ML_COMMENT']
+            }),
```
spoenemann, line 39: We should consider specifying comment rules in a way that the CLI can use that information as well, e.g. for #204. That could be through a grammar language extension (see #185) or through JSON configuration. WDYT?

msujew: I don't like either approach, but I propose a third. Identify whether the token is a comment by two properties:

1. It's a hidden token
2. It doesn't match a whitespace string

A similar approach is already used in the parser interpreter which sorts terminals depending on whether they're able to match a whitespace. Of course, we can always let the user override the computed grammar config. Do you think that's a path we can take?

spoenemann: Yes, I think it's totally fine to assume that all tokens that are hidden but contain non-whitespace characters are comments.

msujew: Alright, I would tackle this in a different issue. This PR should be ready for a last review 👍

spoenemann (approved):
Ok, could you create an issue for detection of comment terminal rules?

PR #179, opened 2021-07-16
Language specific unit tests

kaisalmen:
Allow to run language specific unit test with jest.
Adds debugging support to jest test in VSCode.
Builds upon langium-config support.

### DIFF
--- a/packages/langium/src/references/references.ts
+++ b/packages/langium/src/references/references.ts
@@ -6,7 +6,7 @@
 
 import { findAssignment } from '../grammar/grammar-util';
 import { LangiumServices } from '../services';
-import { CstNode } from '../syntax-tree';
+import { CstNode, Reference } from '../syntax-tree';
 import { isReference } from '../utils/ast-util';
 import { findRelevantNode } from '../utils/cst-util';
 import { NameProvider } from './naming';
@@ -36,23 +36,38 @@ export class DefaultReferences implements References {
             if (assignment && nodeElem) {
                 // eslint-disable-next-line @typescript-eslint/no-explicit-any
                 const reference = (nodeElem as any)[assignment.feature] as unknown;
+
                 if (isReference(reference)) {
-                    const ref = reference.ref;
-                    if (ref && ref.$cstNode) {
-                        const targetNode = this.nameProvider.getNameNode(ref);
-                        if (!targetNode) {
-                            return ref.$cstNode;
-                        }
-                        else {
-                            return targetNode;
+                    return this.processReference(reference);
+                }
+                else if (Array.isArray(reference)) {
+                    for (const ref of reference) {
+                        if (isReference(ref)) {
+                            const target = this.processReference(ref);
+                            if (target && target.text === sourceCstNode.text) return target;
                         }
                     }
-                } else {
+                }
+                else {
                     const nameNode = this.nameProvider.getNameNode(nodeElem);
                     if (nameNode === sourceCstNode) {
                         return sourceCstNode;
                     }
                 }
+            }
+        }
+        return undefined;
+    }
+
+    private processReference(reference: Reference): CstNode | undefined {
+        const ref = reference.ref;
+        if (ref && ref.$cstNode) {
+            const targetNode = this.nameProvider.getNameNode(ref);
+            if (!targetNode) {
+                return ref.$cstNode;
+            }
+            else {
+                return targetNode;
             }
         }
         return undefined;
--- a/packages/langium/src/test/langium-test.ts
+++ b/packages/langium/src/test/langium-test.ts
@@ -1,0 +1,158 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { CompletionItem, DocumentSymbol, Range } from 'vscode-languageserver';
+import { LangiumDocument, LangiumDocumentConfiguration } from '../documents/document';
+import { BuildResult } from '../documents/document-builder';
+import { LangiumServices } from '../services';
+import { getDocument } from '../utils/ast-util';
+
+export function parseHelper(services: LangiumServices): (input: string) => BuildResult {
+    const metaData = services.LanguageMetaData;
+    const documentBuilder = services.documents.DocumentBuilder;
+    return input => {
+        const randomNumber = Math.floor(Math.random() * 10000000) + 1000000;
+        const document = LangiumDocumentConfiguration.create(`file:/${randomNumber}${metaData.fileExtensions[0]}`, metaData.languageId, 0, input);
+        const buildResult = documentBuilder.build(document);
+        return buildResult;
+    };
+}
+
+export type ExpectFunction = (actual: unknown, expected: unknown) => void;
+
+interface ExpectedBase {
+    text: string
+    indexMarker?: string
+    rangeStartMarker?: string
+    rangeEndMarker?: string
+}
+
+export interface ExpectedSymbols extends ExpectedBase {
+    expectedSymbols: DocumentSymbol[]
+}
+
+export function expectSymbols(services: LangiumServices, cb: ExpectFunction): (input: ExpectedSymbols) => void {
+    return (input) => {
+        const document = parseDocument(services, input.text);
+        if (!document.parseResult) {
+            throw new Error('Could not parse document');
+        }
+        const symbolProvider = services.lsp.DocumentSymbolProvider;
+        const symbols = symbolProvider.getSymbols(document);
+        cb(symbols.length, input.expectedSymbols.length);
+        for (let i = 0; i < input.expectedSymbols.length; i++) {
+            const expected = input.expectedSymbols[i];
+            const item = symbols[i];
+            if (typeof expected === 'string') {
+                cb(item.name, expected);
+            } else {
+                cb(item, expected);
+            }
+        }
+    };
+}
+
+export interface ExpectedCompletion extends ExpectedBase {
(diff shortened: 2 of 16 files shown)


packages/langium/src/references/references.ts
```diff
@@ -36,18 +36,19 @@ export class DefaultReferences implements References {
...
-                    if (ref && ref.$cstNode) {
-                        const targetNode = this.nameProvider.getNameNode(ref);
-                        if (!targetNode) {
-                            return ref.$cstNode;
-                        }
-                        else {
-                            return targetNode;
+                    return this.processReference(reference);
+                }
+                else if (reference instanceof Array) {
```
msujew, line 43: Use `Array.isArray(reference)` instead of `reference instanceof Array`

packages/langium/src/utils/file-utils.ts
```diff
@@ -0,0 +1,18 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import fs from 'fs-extra';
+import path from 'path';
+
+export function buildRealpath(pathToRepoPackage: string, packageFile: string): string {
```
msujew, line 10: Do you use this in the actual `langium` package? If not, this probably belongs to the `test` directory.

packages/langium/test/languages/domain-model/lsp/goto-resolver.test.ts
```diff
@@ -0,0 +1,76 @@
...
+ ******************************************************************************/
+
+import { Diagnostic, Position } from 'vscode-languageserver';
+import { buildRealpath, createLangiumGrammarServices, LangiumDocument, LangiumDocumentConfiguration, LangiumGrammarServices, loadUriContent } from '../../../../src';
+
+let services: LangiumGrammarServices;
+let document: LangiumDocument;
+
+beforeAll(() => {
+    const realPathFile = buildRealpath('packages/langium', 'test/languages/domain-model/domain-model.langium');
```
msujew, line 14: Why not use `path.join(__dirname, '..', 'domain-model.langium')`?

packages/langium/test/languages/domain-model/lsp/goto-resolver.test.ts
```diff
@@ -0,0 +1,76 @@
...
+            },
+            // `    DataType | Entity;` - Must find definiton of Entity
+            position: Position.create(15, 16)
+        };
+        const locationLink = goToResolver.goToDefinition(document, textPos);
+        expect(locationLink.length).toBe(1);
+        expect(locationLink[0].targetSelectionRange.start.line).toBe(20);
+        expect(locationLink[0].targetSelectionRange.start.character).toBe(0);
+        expect(locationLink[0].targetSelectionRange.end.line).toBe(20);
+        expect(locationLink[0].targetSelectionRange.end.character).toBe(6);
```
msujew, line 40: I thought about a more modular approach to testing. For example, I'd like to see a testing service somewhere in the `langium` package (or a new `langium-test` package, although I'm against that), that can be used like this:

```ts
test('Find declaration', () => {
    langiumTest.expectGoToDefinition('<input-text>', position, {
        start: {
            line: 20,
            character: 0
        },
        end: {
            line: 20,
            character: 6
        }
    });
});
```

Where `expectGoToDefinition` is something like this `expectGoToDefinition(input: string, position: Position, expected: Range)`. Maybe we need to initialize it with some kind of `expect` method, so that it can be used with other test frameworks than jest.

msujew: For that we would also need an equivalent of the `ParseHelper` from Xtext in Langium. This should now be easily possible, since the language meta data is included in the `LangiumServices`

packages/langium/src/references/references.ts
kaisalmen: Will do

packages/langium/src/utils/file-utils.ts
kaisalmen: Wasn't sure. Do you see further need for this beyond the tests?

packages/langium/test/languages/domain-model/lsp/goto-resolver.test.ts
kaisalmen: Because I didn't know `__dirname` is available. This basically resolves the need for the file-utils. It will be removed. Already tested this locally.

packages/langium/src/utils/file-utils.ts
kaisalmen: It will go away. See https://github.com/langium/langium/pull/179#discussion_r673150611

msujew: Not right now, but I also can't envision a use case where this becomes necessary. I'm especially concerned with the hardcoded stuff in there, at least in the `buildRealpath` method. The other method is fine, imo.

packages/langium/test/languages/domain-model/lsp/goto-resolver.test.ts
msujew: While this would generally improve test readability and maintenance (as everything is tested in one place), I'm mostly concerned with downstream tests. e.g. A developer trying to test their own language/framework based on Langium. Just as with the rest of the framework, we want to provide the developer with a nice and handy API to test their code.

kaisalmen: Ok, so something like a collection of tests (select-what-you-need, extendable) that you execute in the context of a specific language. On every call (what you outlined above) you formulate the expectations. This could even be driven completely with configuration.

msujew:
FYI I took over this PR, due to vacation.

spoenemann (commented):
Thanks for this first step towards testing language services! I have three main remarks:

 - I'd try to avoid using classes for test helpers such as `DefaultLangiumTest` and `DefaultParseHelper`. We can get the same functionality with plain functions. Instead of `createTestEnvironment` we could have three functions that return functions for `expectSymbols`, `expectCompletion` and `expectGoToDefinition`.
 - For service tests that require a cursor position (offset), I liked the option in Xtext to pass an input string that marks the position with a special character series. The following example uses the marker `<|>`:
```typescript
const model = `
grammar g
Aaaa: name="A";
aaaa: name="a";
Bbbb: name="B";
C: a=aa<|>;`;
```
 - Can we avoid duplicating the `domainmodel` and `statemachine` examples as test languages in `langium/test`? Of course we can add test languages later, e.g. to reproduce a reported bug. But tests that use our example languages should be written in the examples themselves, not in the core package.

msujew:
> Instead of createTestEnvironment we could have three functions that return functions for expectSymbols, expectCompletion and expectGoToDefinition.

This would require us to always pass our `LangiumServices` as well as our assertion function into each call. That would make these functions quite unwieldy. Or how would you expect to call them?

msujew:
@spoenemann In addition to index markers `<|>`, I added range markers starting with `<|` and ending with `|>`.

packages/langium/src/test/langium-test.ts
```diff
@@ -0,0 +1,123 @@
...
+            if (typeof expected === 'string') {
+                cb(item.name, expected);
+            } else {
+                cb(item, expected);
+            }
+        }
+    };
+}
+
+export function expectCompletion(services: LangiumServices, cb: ExpectFunction): (input: string, index: number, ...expectedItems: Array<string | CompletionItem>) => void {
```
spoenemann, line 41: I'd propose to define the actual test parameters as an object, so it's easier to add optional parameters like changing the `<|>` marker string, and parameters of boolean and number type (like `index`) are more readable on the call site.

packages/langium/src/test/parse-helper.ts
```diff
@@ -0,0 +1,33 @@
...
+import { LangiumDocumentConfiguration } from '../documents/document';
+import { BuildResult, DocumentBuilder } from '../documents/document-builder';
+import { LanguageMetaData } from '../grammar/language-meta-data';
+import { LangiumServices } from '../services';
+
+export interface ParseHelper {
+    parse(input: string): BuildResult
+}
+
+export class DefaultParseHelper implements ParseHelper {
```
spoenemann, line 16: While we're transforming classes to functions in the context of tests, shall this be a function as well?

packages/langium/src/index.ts
```diff
@@ -29,11 +29,12 @@ export * from './parser/langium-parser';
 export * from './references/linker';
 export * from './references/naming';
 export * from './references/scope';
+export * from './test/langium-test';
+export * from './test/parse-helper';
```
spoenemann, line 33: Please create an `index.ts` in `test` and export the stuff there, so we avoid including the test support in production code. This means you'd import that from `'langium/lib/test'` instead of `'langium'`.

packages/langium/src/test/parse-helper.ts
```diff
@@ -0,0 +1,20 @@
...
+export function parseHelper(services: LangiumServices): (input: string) => BuildResult {
+    const metaData = services.LanguageMetaData;
+    const documentBuilder = services.documents.DocumentBuilder;
+    return input => {
+        const randomNumber = Math.floor(Math.random() * 10000000) + 1000000;
+        const document = LangiumDocumentConfiguration.create(`file:/${randomNumber}${metaData.fileExtensions[0]}`, metaData.languageId, 0, input);
+        const buildResult = documentBuilder.build(document);
+        return buildResult;
+    };
+}
```
spoenemann, line 20: We could move this to `langium-test.ts`, or do you see a reason to keep a separate file?

spoenemann (approved):
:+1:

PR #180, opened 2021-07-20
Delete target folder before generation

### LINKED ISSUE #167: Clean up previously generated files before regenerating them
Since we are still in a beta-phase and commonly change which files are being generated by the `langium-cli` package, we should clean up the target directory to remove any unwanted files.

msujew:
Closes #167

Simply deletes the `output` folder before generation, therefore removing any old files.

### DIFF
--- a/packages/langium-cli/src/generate.ts
+++ b/packages/langium-cli/src/generate.ts
@@ -56,7 +56,12 @@ export function generate(config: LangiumConfig): boolean {
 
     const output = path.join(relPath, config.out ?? 'src/generated');
     console.log(`${getTime()}Writing generated files to ${output.white.bold}`);
-    mkdirWithFail(output);
+    if (rmdirWithFail(output)) {
+        return false;
+    }
+    if (mkdirWithFail(output)) {
+        return false;
+    }
 
     const genAst = generateAst(grammar, config);
     writeWithFail(path.join(output, 'ast.ts'), genAst);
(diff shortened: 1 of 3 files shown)


spoenemann (commented):
Looks good, but I'm thinking whether bad things can happen when we delete a directory without warning.

Idea: check whether the directory contains any items that are *not* generated by the current CLI run. If there are such items, display them and ask the user whether to delete them.

msujew:
@spoenemann Sure, makes sense. I added a yes/no check beforehand if the generator notices any additional files in the generated directory. For this, I had to make the whole generator async in order to interact with the console input.

packages/langium-cli/package.json
```diff
@@ -34,7 +34,8 @@
     "fs-extra": "^9.1.0",
     "jsonschema": "^1.4.0",
     "langium": "0.1.0",
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "yesno": "^0.3.1"
```
spoenemann, line 38: NodeJs has the built-in `readline` package. Shall we use that instead of pulling another dependency?

See an example here:
https://github.com/eclipse/openvsx/blob/6674bf4719174b8ba6b6582ef533419caa1d3174/cli/src/util.ts#L158-L182

packages/langium-cli/src/generate.ts
```diff
@@ -58,45 +58,74 @@ export function generate(config: LangiumConfig): boolean {
 
     const output = path.join(relPath, config.out ?? 'src/generated');
     console.log(`${getTime()}Writing generated files to ${output.white.bold}`);
-    mkdirWithFail(output);
+
+    if (await rmdirWithFail(output, ['ast.ts', 'grammar.ts', 'grammar-access.ts', 'parser.ts', 'module.ts'])) {
+        return false;
+    }
+    if (await mkdirWithFail(output)) {
+        return false;
```
spoenemann, line 66: Wouldn't it be easier to reject the promise?

msujew: Well, kind of. We would have to deal with the `generate` function throwing inside of the other code, which would be quite annoying (especially in the watch mode). So I went with this direction.

spoenemann: I see. I think the code would be better understandable if you use a more specific return type:
```typescript
export type ResultType = 'success' | 'failure';
```
or something like that.

spoenemann (approved):
:+1:

PR #181, opened 2021-07-22
Fix lifting of common properties

### LINKED ISSUE #176: Common properties are lifted into the super type regardless of their own type
Langium version: 0.1.0
Package name: `langium-cli`

## Steps To Reproduce

1. Create a grammar like this:
```
Token: TokenA | TokenB;
TokenA: name=('A1' | 'A2');
TokenB: name=('B1' | 'B2' | 'B3');
```
2. Generate the AST and see that the `Token` type contains a single property `name: 'A'`.

## The current behavior

The `name` property is not put into consideration when lifting fields.

## The expected behavior

Properties are not lifted, if their types aren't compatible.

Pluralia:
Closes #176

Checks properties types before lifting.


msujew (approved):
Looks good to me 👍

PR #182, opened 2021-07-22
Add hover lsp service

msujew:
Also improves some parser behavior regarding cst nodes of datatype rules and fragments, as well as hidden nodes.

### DIFF
--- a/packages/langium/src/lsp/hover-provider.ts
+++ b/packages/langium/src/lsp/hover-provider.ts
@@ -1,0 +1,88 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { Hover, HoverParams } from 'vscode-languageserver';
+import { LangiumDocument } from '../documents/document';
+import { GrammarConfig } from '../grammar/grammar-config';
+import { References } from '../references/references';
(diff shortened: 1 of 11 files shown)


packages/langium/src/lsp/hover-provider.ts
```diff
@@ -0,0 +1,93 @@
...
+            const hiddenNodes: CstNode[] = [];
+            for (const node of cstNode.children) {
+                if (node.hidden) {
+                    hiddenNodes.push(node);
+                } else {
+                    break;
+                }
+            }
+            const lastNode = hiddenNodes[hiddenNodes.length - 1];
+            if (lastNode && lastNode instanceof LeafCstNodeImpl && lastNode.tokenType.name === 'ML_COMMENT') {
```
spoenemann, line 64: Please use the new GrammarConfig (see #178).

packages/langium/src/lsp/language-server.ts
```diff
@@ -168,6 +168,14 @@ export function addDocumentHighlightsHandler(connection: Connection, services: L
     });
 }
 
+export function addHoverHandler(connection: Connection, services: LangiumServices): void {
+    const hoverProvider = services.lsp.HoverProvider;
+    connection.onHover(params => {
+        const document = paramsDocument(params, services);
+        return document && hoverProvider.getHoverContent(document, params.position);
```
spoenemann, line 175: Please pass the whole `params` so we don't need to change our API in case LSP adds more properties.

PR #220, opened 2021-09-15
Clean up undefined checks for LangiumDocument.parseResult

### LINKED ISSUE #209: Clean up undefined checks for LangiumDocument.parseResult
We need to check for uses of this `parseResult` property where it's still assumed to be optional and clean up.

dhuebner:
Closes  #209


msujew:
I think we can remove the check in the [document builder](https://github.com/langium/langium/blob/6a3f31dbc7457ae46fc1d71b8d7ff8d177c76013/packages/langium/src/documents/document-builder.ts#L87-L89) as well. It seems a bit redundant to me.

dhuebner:
@msujew 
> I think we can remove the check in the

Sure, no longer need to check. Commit is updated

msujew (approved):
LGTM, thanks!

PR #221, opened 2021-09-15
Replace `string` with `URI` type

msujew:
Instead of dealing with raw strings everywhere, we should use the `URI` type to enforce some safety in our APIs.


packages/langium/src/documents/document-builder.ts
```diff
@@ -103,7 +104,7 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
         document.precomputedScopes = this.scopeComputation.computeScope(document);
     }
 
-    protected isAffected(document: LangiumDocument, changedUri: string): boolean {
-        return changedUri !== document.textDocument.uri;
+    protected isAffected(document: LangiumDocument, changedUri: URI): boolean {
+        return changedUri.toString() !== document.uri.toString();
```
spoenemann, line 108: Istn't there a way to compare the URI objects?

packages/langium/src/references/scope.ts
```diff
@@ -18,7 +19,7 @@ export interface AstNodeDescription {
     node?: AstNode
     type: string // AstNodeType?
     name: string // QualifiedName?
-    documentUri: string // DocumentUri?
+    documentUri: URI // DocumentUri?
```
spoenemann, line 22: The comment here was referring exactly to changing the type to something more concrete; it can be removed now.

packages/langium/src/documents/document-builder.ts
msujew: There is none. However, we could write our own utility functions for that. I'm also wondering, whether we should re-export the URI in langium, as that will be used a fair bit by downstream users.

spoenemann (approved):

packages/langium/src/documents/document-builder.ts
spoenemann: A utility function might be useful in case we need to normalize the URI somehow, but let's keep that in mind for later.

Re-exporting the URI makes sense because it's an important part of our API.

PR #222, opened 2021-09-16
Fixed unassigned call of a datatype rule

spoenemann:
This fixes the case when a datatype rule has an unassigned call of another datatype rule. In that case the result is a primitive data type, so we cannot transfer properties like we do for other unassigned rule calls.

### DIFF
--- a/packages/langium/src/utils/ast-util.ts
+++ b/packages/langium/src/utils/ast-util.ts
@@ -24,13 +24,23 @@ export function isReference(obj: unknown): obj is Reference {
 export function getContainerOfType<T extends AstNode>(node: AstNode | undefined, typePredicate: (n: AstNode) => n is T): T | undefined {
     let item = node;
     while (item) {
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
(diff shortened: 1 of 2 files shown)


packages/langium/src/utils/ast-util.ts
```diff
@@ -33,6 +32,17 @@ export function getContainerOfType<T extends AstNode>(node: AstNode | undefined,
     return undefined;
 }
 
+export function hasContainerOfType<T extends AstNode>(node: AstNode | undefined, typePredicate: (n: AstNode) => n is T): boolean {
```
msujew, line 35: I fell like it doesn't make a lot of sense to include the generic parameter in here:
```suggestion
export function hasContainerOfType(node: AstNode | undefined, typePredicate: (n: AstNode) => boolean): boolean {
```

spoenemann: But then the type predicate has the type `(n: AstNode) => n is AstNode`, which looks strange, and the compiler complains that line 41 (the _else_ case of `if (typePredicate(item))`) is not reachable.

PR #224, opened 2021-09-16
Specify minimum node version in engines field

### LINKED ISSUE #192: Missing engines declaration in package.json
All our packages should declare a minimum NodeJS version that is required to use them.

msujew:
Closes #192 

We use Node `12.10.0` as the minimum version now, as that is required by `yeoman@5.3.0`.

### DIFF
--- a/examples/arithmetics/package.json
+++ b/examples/arithmetics/package.json
@@ -58,7 +58,7 @@
         "vscode-languageserver": "^7.0.0"
     },
     "devDependencies": {
-        "@types/node": "^14.17.3",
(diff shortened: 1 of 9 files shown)


spoenemann (commented):
Thanks! The `engines` declaration should be in all published packages (except the vscode extension) because node / npm can use it to check whether the client can run the package.

spoenemann:
Additionally, we could specify the minimum npm version in the root package.json, see https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines

spoenemann (approved):
:+1:

PR #226, opened 2021-09-16
Validate unused rules and improve reference finder

### LINKED ISSUE #197: Validate reachability of rules
Unreachable rules should be marked with warnings. Due to #91 there is no need to declare "dummy rules" like in Xtext.

msujew:
Closes #197

### DIFF
--- a/packages/langium/src/grammar/langium-grammar-validator.ts
+++ b/packages/langium/src/grammar/langium-grammar-validator.ts
@@ -4,8 +4,12 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
+import { DiagnosticTag } from 'vscode-languageserver-types';
+import { References } from '../references/references';
+import { LangiumServices } from '../services';
+import { streamAllContents } from '../utils/ast-util';
 import { ValidationAcceptor, ValidationCheck, ValidationRegistry } from '../validation/validation-registry';
-import { AbstractRule, Grammar, isTerminalRule, Keyword, LangiumGrammarAstType, ParserRule, TerminalRule, UnorderedGroup } from './generated/ast';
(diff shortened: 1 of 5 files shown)


packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -103,6 +113,20 @@ export class LangiumGrammarValidator {
...
     }
 
+    checkGrammarForUnusedRules(grammar: Grammar, accept: ValidationAcceptor): void {
+        let entryRule = true;
+        for (const rule of grammar.rules) {
+            if (isParserRule(rule) && entryRule) {
+                entryRule = false;
+                continue;
+            }
+            const references = this.references.findReferences(rule);
```
spoenemann, line 123: This criterion is not sufficient. We want to mark rules as unreachable if they are referenced only by other unreachable rules.

Algorithmic approach: start a DFS from the entry rule and mark all visited rules. Then find all unmarked rules and give them a warning.

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -103,6 +113,20 @@ export class LangiumGrammarValidator {
...
+            if (references.length === 0) {
+                accept('hint', 'This rule is declared but never referenced', { node: rule, property: 'name', tags: [DiagnosticTag.Unnecessary] });
```
spoenemann, line 125: In Xtext it's only a hint because we need dummy rules to shape the generated AST metamodel. In the light of #91, we could mark unreachable rules with warnings. Or shall we wait until #91 is actually implemented?

msujew: I would wait for #91.

spoenemann (approved):

Issue #213, opened 2021-09-14
Update domainmodel example to use the index 


dhuebner:
We should customize the AstNodeDescriptionProvider for that example so it exports all entities with their qualified names.

PR #229, opened 2021-09-17
Update domainmodel example to use the index

dhuebner:
Update domainmodel example to use the index #213

Also added a test to cover a bug in node locator service

### DIFF
--- a/examples/domainmodel/example/blog.dmodel
+++ b/examples/domainmodel/example/blog.dmodel
@@ -1,7 +1,6 @@
-datatype String
-
 entity Blog {
     title: String
+    date: complex.Date
     many posts: Post
 }
 
--- a/examples/domainmodel/src/language-server/domain-model-index.ts
+++ b/examples/domainmodel/src/language-server/domain-model-index.ts
@@ -1,0 +1,36 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { AstNodeDescription, DefaultAstNodeDescriptionProvider, LangiumDocument, LangiumServices, streamAllContents } from 'langium';
+import { DomainModelNameProvider } from './domain-model-naming';
+import { isPackageDeclaration, isType, PackageDeclaration } from './generated/ast';
+
+export class DomainModelDescriptionProvider extends DefaultAstNodeDescriptionProvider {
+
+    constructor(services: LangiumServices) {
+        super(services);
+    }
+
+    /**
+     * Exports only types (`DataType or `Entity`) with their qualified names.
+     */
+    createDescriptions(document: LangiumDocument): AstNodeDescription[] {
+        const descr: AstNodeDescription[] = [];
+        streamAllContents(document.parseResult.value).forEach(content => {
+            const modelNode = content.node;
+            if (isType(modelNode)) {
+                let name = this.nameProvider.getName(modelNode);
+                if (name) {
+                    if (isPackageDeclaration(modelNode.$container)) {
+                        name = (this.nameProvider as DomainModelNameProvider).getQualifiedName(modelNode.$container as PackageDeclaration, name);
+                    }
+                    descr.push(this.createDescription(modelNode, name, document));
+                }
+            }
+        });
+        return descr;
+    }
+}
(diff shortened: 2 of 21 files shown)


packages/langium/src/index/ast-node-locator.ts
```diff
@@ -33,19 +33,19 @@ export class DefaultAstNodeLocator implements AstNodeLocator {
...
         let container: AstNode | undefined = node.$container;
-        const path: string[] = [''];
+        const path: string[] = [];
         while (container) {
             path.push(this.pathSegment(node, container));
             node = container;
             container = container.$container;
         }
-        return path.join(this.segmentSeparator);
+        return this.segmentSeparator + path.reverse().join(this.segmentSeparator);
```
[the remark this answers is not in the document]
dhuebner: Is there a performance difference between both?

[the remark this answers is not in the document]
msujew: Well creating the list and reversing it is double as costly as creating it in the right direction in the first place.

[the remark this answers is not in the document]
dhuebner: I thought shift will reindex the whole array each time it is called

msujew (approved):
LGTM!

examples/domainmodel/src/language-server/domain-model-index.ts
```diff
@@ -0,0 +1,36 @@
...
+ ******************************************************************************/
+
+import { AstNodeDescription, DefaultAstNodeDescriptionProvider, LangiumDocument, LangiumServices, streamAllContents } from 'langium';
+import { DomainModelNameProvider } from './domain-model-naming';
+import { isPackageDeclaration, isType, PackageDeclaration } from './generated/ast';
+
+export class DomainModelDescriptionProvider extends DefaultAstNodeDescriptionProvider {
+
+    constructor(services: LangiumServices) {
+        super(services);
```
spoenemann, line 14: This constructor can be omitted (supertype constructor is implicitly available).

examples/domainmodel/example/blog.dmodel
```diff
@@ -1,7 +1,6 @@
-datatype String
-
 entity Blog {
     title: String
+    date: complex.Date
```
spoenemann, line 3: This change broke the CLI: running `./bin/cli generate example/blog.dmodel` in the terminal no longer works because the datatype references can't be resolved.

How shall we handle this? Automatically include other `.dmodel` files in the same directory? Add a CLI option to specify a root path?

examples/domainmodel/src/language-server/domain-model-naming.ts
```diff
@@ -0,0 +1,32 @@
...
+export class DomainModelNameProvider extends DefaultNameProvider {
+
+    /**
+     * 
+     * @param qualifier if the qualifier is a `string`, simple string concatenation is done: `qualifier.name`.
+     *      if the qualifier is a `PackageDeclaration` fully qualified name is created: `package1.package2.name`.
+     * @param name simple name
+     * @returns qualified name separated by `.`
+     */
+    getQualifiedName(qualifier: PackageDeclaration | string, name: string): string {
```
spoenemann, line 23: The type casting from `NameProvider` to `DomainModelNameProvider` is a bad smell. Possible solutions:

 * Add support for qualified names in the framework ( :-1: )
 * Move this function to a separate service `QualifiedNameProvider` (specific to this example language)
 * ?

spoenemann (commented):
@dhuebner @msujew sorry for the after-the-fact review, but I found a few things that I think should be improved:

examples/domainmodel/src/language-server/domain-model-index.ts
dhuebner: I added an explicit super constructor call after I noticed that `services` field was `undefined`

examples/domainmodel/example/blog.dmodel
spoenemann: > > This change broke the CLI: running ./bin/cli generate example/blog.dmodel
>
> Is it possible to setup a CLI specific test case for that? 

Yes, we could add a test that runs the CLI, checks the exit code, and then checks whether the expected output file exists.

>> Add a CLI option to specify a root path?
>
> We should run the indexer in the source root folder/-s which is/are either specified by the user or using a good default

:+1:

spoenemann:
I'll create issues for my remarks.

PR #230, opened 2021-09-17
Fix eslint config for example projects

msujew:
Eslint ignored the global config because the `arithmetics` and `domainmodel` project had their own. This PR resolves that and fixes any linting issues that weren't previously reported.


spoenemann (approved):
:+1:

Issue #214, opened 2021-09-14
Review and fix find references/index using domainmodel example


dhuebner:
## Steps To Reproduce
Find references does not work properly. Example: change line 5 in blog.dmodel to many posts: E1, then "Find All Reference" shows one hit from blog.dmodel and one from qualified-names.dmodel when started from the new reference, while it shows two hits from qualified-names.dmodel when started from the declaration.

dhuebner:
@spoenemann 
The problem here is that I explicitly exclude `local` cross references from the index, see:
https://github.com/langium/langium/blob/6dd874b70d0debb08c52cad73cd8e142ca40873b/packages/langium/src/index/ast-descriptions.ts#L92-L94

These cross references are not interesting for a global index lookup (e.g. isAffected), so I'm not sure how to proceed here.
One option is to create all the reference descriptions for a document but mark them with `local = true` so we can skip them when doing a lookup for foreign cross references.

spoenemann:
But shouldn't the call to `findLocalReferences` ensure that the references that are local to the declaration are included as well?

PR #231, opened 2021-09-17
Review and fix find references/index using domainmodel example #214

dhuebner:
Adds local references to the indexed reference descriptions


examples/domainmodel/test/cross-refs.test.ts
```diff
@@ -0,0 +1,58 @@
...
+ entity Blog {
+    title: String
+    description: String
+    time: big.Int
+}
+`;
+describe('Cross references', () => {
+    const datatypeDoc: LangiumDocument = parseDocument(services, datatypeFile);
+    const referencingDoc: LangiumDocument = parseDocument(services, referencingFile);
+    (services.index.IndexManager as DefaultIndexManager).processDocuments([referencingDoc, datatypeDoc]);
```
spoenemann, line 36: Shall we expose that method in the interface? Type-casting services always looks like a bad smell to me.

examples/domainmodel/test/cross-refs.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const referencingFile = `
+ entity Blog {
+    title: String
+    description: String
+    time: big.Int
+}
+`;
+describe('Cross references', () => {
+    const datatypeDoc: LangiumDocument = parseDocument(services, datatypeFile);
```
spoenemann, line 34: Shall we add a type parameter `<T = AstNode>` to LangiumDocument and also to `parseDocument` so we can avoid typecasting further below? This line could then be written as
```typescript
const datatypeDoc = parseDocument<DomainModel>(services, datatypeFile);
```
and `datatypeDoc` would have the inferred type `LangiumDocument<DomainModel>`.

spoenemann (commented):
Something is still not working properly with the Find References service. I changed the type of the `posts` attribute of `Blog` to `E1` and started Find References on `E1`. The number of results (3) is correct, but the last result in the list is corrupt:
Screen Shot 2021-09-23 at 13 29 41

DefaultReferences still uses `findLocalReferences`. Is this necessary now that the local references are part of the index?

examples/domainmodel/test/cross-refs.test.ts
dhuebner: Would be cool!

examples/domainmodel/test/cross-refs.test.ts
dhuebner: This is a test class, I'm not yet sure wie will need this somewhere in the real projects, but yes why not.

spoenemann: We should add documentation to the IndexManager methods about when and how to use them.

dhuebner:
> Something is still not working properly with the Find References service. I changed the type of the `posts` attribute of `Blog` to `E1` and started Find References on `E1`. The number of results (3) is correct, but the last result in the list is corrupt:

Yes, I noticed that, still had no time to look into it. :(

dhuebner:
@spoenemann 
Reference range is fixed now:
Bildschirmfoto 2021-09-23 um 16 03 21

>  type parameter <T = AstNode> to LangiumDocument 

Should I open a feature request for?

packages/langium/src/index/ast-descriptions.ts
```diff
@@ -23,10 +23,12 @@ export interface ReferenceDescription {
...
     /** how to find target AstNode inside the document */
     targetPath: string
-    /** start offset of the reference text */
-    start: number
-    /** end offset of the reference text */
-    end: number
+    /** start position of the reference text */
+    start: {line: number, character: number}
+    /** end position of the reference text */
+    end:  {line: number, character: number}
```
spoenemann, line 29: Shouldn't we keep the offset-based positions like in Xtext? Why did you change this to line/character positions?

packages/langium/src/documents/document-builder.ts
```diff
@@ -85,14 +85,14 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
         const newDocument = this.langiumDocuments.getOrCreateDocument(uri);
+        this.build(newDocument);
         this.indexManager.update(newDocument);
         this.langiumDocuments.all.filter(doc => this.isAffected(doc, uri)).forEach(
             doc => {
                 this.build(doc);
                 this.indexManager.update(doc);
             }
         );
-        this.build(newDocument);
```
spoenemann, line 95: This conflicts with the changes in #244. @msujew we should test the domainmodel example thoroughly in your PR.

PR #232, opened 2021-09-21
Moved langium config from package.json to its own json file

### LINKED ISSUE #219: Change default project to generate langium-config.json
I'd propose to generate a `langium-config.json` file instead of embedding the configuration in the package.json (though we can keep supporting both variants in the CLI). This way we can offer JSON editing support to users (#218).

jbicker:
To test it with the next version note the [contribution guidelines](https://github.com/langium/langium/blob/3b4b511b60d61cb807c187c17b1a794f6cf3986a/CONTRIBUTING.md#altered-langium-for-own-language-projects).
Fixes #219

### DIFF
--- a/examples/arithmetics/langium-config.json
+++ b/examples/arithmetics/langium-config.json
@@ -1,0 +1,9 @@
+{
+    "languageId": "arithmetics",
+    "grammar": "src/language-server/arithmetics.langium",
+    "fileExtensions": [".calc"],
+    "out": "src/language-server/generated",
+    "textMate": {
+        "out": "syntaxes/arithmetics.tmLanguage.json"
+    }
+}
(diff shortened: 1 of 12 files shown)


spoenemann (commented):
Thanks! When the configuration is in a separate file, you don't need the additional `"langium"` property block, so you can move its content one level higher.

msujew:
FYI, I think the Langium config in the `langium` package itself can also be moved into its own file.

jbicker:
> FYI, I think the Langium config in the `langium` package itself can also be moved into its own file.

Oh yes. Totally forgot about that this is also a langium language package itself.

spoenemann (approved):
:+1:

PR #233, opened 2021-09-21
Fix generator watch mode

msujew:
With #172, we accidentally broke the support for watching langium grammars with the generator. The first grammar file that was read was cached and never invalidated. This change invalidates the grammar file without recreating the `services` instance.

### DIFF
--- a/packages/langium-cli/src/generate.ts
+++ b/packages/langium-cli/src/generate.ts
@@ -27,6 +27,7 @@ const services = createLangiumGrammarServices();
 export async function generate(config: LangiumConfig): Promise<GeneratorResult> {
     const relPath = config[RelativePath];
     const absGrammarPath = URI.file(path.resolve(relPath, config.grammar));
+    services.documents.LangiumDocuments.invalidateDocument(absGrammarPath);
     const document = services.documents.LangiumDocuments.getOrCreateDocument(absGrammarPath);
     const buildResult = services.documents.DocumentBuilder.build(document);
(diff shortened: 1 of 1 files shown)


spoenemann (approved):

packages/langium-cli/src/generate.ts
```diff
@@ -27,6 +27,7 @@ const services = createLangiumGrammarServices();
 export async function generate(config: LangiumConfig): Promise<GeneratorResult> {
     const relPath = config[RelativePath];
     const absGrammarPath = URI.file(path.resolve(relPath, config.grammar));
+    services.documents.LangiumDocuments.invalidateDocument(absGrammarPath);
```
dhuebner, line 30: It would be nice if we could reuse the same logic from the existing services.
Would `DocumentBuilder.documentChanged(uri: string)`  work here?
If not that we probably we need a better common API that does an update and returns a "fresh" Document.

msujew: Well the CLI in a watch mode is a special case, that we definitely will resolve inside of Langium. However, for now (0.2.0), I just want to fix the `langium-cli`.

For the future, we want to provide utilities for watching files in a CLI. This will then also include correctly calling `documentChanged` in a more generic way.

spoenemann: Could you describe that in a new issue?

msujew: @spoenemann Done #239

PR #240, opened 2021-09-22
Added json schema support to langium vscode

### LINKED ISSUE #218: Add JSON schema support for editing langium-config.json
VS Code offers a contribution point for [JSON schema](https://code.visualstudio.com/api/references/contribution-points#contributes.jsonValidation). We should use that so users have validation, completion and hover information when they edit langium-config.json files.

The schema is already available in `langium-cli`. We should copy it with an npm script to avoid duplicating its content.

jbicker:
and renamed the schema file

Fixes #218


.gitignore
```diff
@@ -7,4 +7,5 @@ out/
 node_modules/
 *.vsix
 *.tsbuildinfo
-examples/*/example/generated
\ No newline at end of file
+examples/*/example/generated
+packages/langium-vscode/langium-config-schema.json
```
spoenemann, line 11: In this case I'd prefer a local `.gitignore` file in `packages/langium-vscode`.

packages/langium-vscode/package.json
```diff
@@ -25,13 +25,21 @@
...
+      }
     ]
   },
   "activationEvents": [
     "onLanguage:langium"
   ],
   "main": "out/extension.js",
   "scripts": {
+    "prepare": "npm run copy-schema",
+    "copy-schema": "cp ../langium-cli/langium-config-schema.json ./langium-config-schema.json",
```
spoenemann, line 42: 1. Please move the schema into the existing `data` folder.
2. `cp` won't work on a Windows terminal. Shall we use an npm package for this (like we often use `rimraf` instead of `rm`)?

packages/langium-vscode/package.json
```diff
@@ -25,13 +25,21 @@
         "scopeName": "source.langium",
         "path": "./data/langium.tmLanguage.json"
       }
+    ],
+    "jsonValidation": [
+      {
+        "fileMatch": "langium-config.json",
+        "url": "./langium-config-schema.json"
```
spoenemann, line 32: This doesn't match anymore.

jbicker: Dammit

spoenemann (approved):
:+1:

PR #244, opened 2021-09-23
Async document builder

### LINKED ISSUE #238: Make all message processing (potentially) asynchronous
We should allow all LSP services to return either a value or a promise of it (see https://github.com/langium/langium/issues/186#issuecomment-913547234).

In addition, we should make the processing of `didChange` notifications asynchronous so we can cancel any previous processing when multiple changes are received in short time.

msujew:
Closes #238

This is resolved using the `setImmediate` JS function to delay the current execution to the next iteration of the event loop. Also requires the `async-mutex` package for us not to run into the document builder multiple times.

You can test this by creating a large domain model file (10000+ lines) and changing something that produces a validation warning really quickly. Using this approach, the validation will appear after the usual delay. Using the behavior on `main` this would result in blinking validation errors.


packages/langium/src/validation/document-validator.ts
```diff
@@ -69,25 +80,35 @@ export class DefaultDocumentValidator {
...
             }
         };
         runChecks(rootNode);
-        streamAllContents(rootNode).map(c => c.node).forEach(runChecks);
+        for (const node of streamAllContents(rootNode).map(c => c.node)) {
+            await delayNextTickAfter();
+            if (cancelToken.isCancellationRequested) {
+                break;
+            }
+            await runChecks(node);
```
spoenemann, line 110: In case a validation returns a promise, we could start more validations in parallel.

packages/langium/src/utils/promise-utils.ts
```diff
@@ -0,0 +1,34 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+export type MaybePromise<T> = T | Promise<T>
+
+/**
+ * Delays the execution of the current code to the next tick of the event loop.
```
spoenemann, line 10: Add a comment that this shouldn't be used in a loop to avoid creating too many promises?

spoenemann (changes requested):
That's a good start! I think we have to refactor the DocumentBuilder, though. I propose the following:

 * Create a dedicated method to be used by CLIs (maybe use `build` for that, but without parsing, see below).
 * `documentChanged` should work in separate phases:
   1. Invalidate, recreate and then `process` the changed document
   2. Collect all affected documents (this is probably the responsibility of the IndexManager)
   3. Update the index of the affected documents (including the changed one)
   4. Run validation on all affected documents
 * The DocumentBuilder should keep track of `dirty` documents, that is documents reported by `documentChanged` and affected ones.
 * When a `documentChanged` call is cancelled, the next invocation should consider the documents that are still dirty, meaning that the previous invocation did not finish updating or validating them.
 * We should consider generalizing the `valid` flag of LangiumDocument to something like `BuilderState`, which can be `'changed'`, `'parsed'`, `'processed'`, `'indexed'`, `'validated'`.
 * If we follow this approach, the DocumentBuilder could keep all documents in a single queue and process that in the phases as outlined above.

spoenemann:
> We should consider generalizing the valid flag of LangiumDocument to something like BuilderState, which can be 'changed', 'parsed', 'processed', 'indexed', 'validated'.

Or store these flags in separate boolean properties? Encoded in a bitmap? Maybe a simple string is enough.

packages/langium/src/utils/promise-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export function delayNextTick(): Promise<void> {
+    return new Promise(resolve => {
+        setImmediate(resolve);
+    });
+}
+
+let lastTick = -1;
```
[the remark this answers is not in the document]
msujew: I can see that leading to a lot of interruptions of the event loop, since they each have their own timer. Instead, what this does is that an interrupt will be triggered every 5 ms on a global level. I don't think we need higher fidelity, do we? Even if multiple callers call this method, it will only interrupt the event loop once, which is exactly what we want.

packages/langium/src/utils/promise-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Delays the execution of the current code to the next tick of the event loop if more milliseconds than specified have elapsed since the last call of this method.
+ * @param elapsed The amount of milliseconds that needs to elapse for delaying to the next tick.
+ */
+export async function delayNextTickAfter(elapsed = 5): Promise<void> {
+    const current = Date.now();
```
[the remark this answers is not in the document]
msujew: The `Date.now()` resolution is high enough for our use case. It is using a sub-millisecond resolution.

packages/langium/src/documents/document-builder.ts
```diff
@@ -54,55 +50,72 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+        const affectedDocuments = this.indexManager.getEffectedDocuments(document);
+        affectedDocuments.forEach(e => e.reset());
+        const relevantDocuments = Array.from(new Set([
+            document,
+            ...affectedDocuments,
+            // Also include all documents haven't completed the document lifecycle yet
+            ...this.langiumDocuments.all.filter(e => e.state < DocumentState.Validated)
+        ]));
+        await this.runCancelable(relevantDocuments, DocumentState.Processed, cancelToken, doc => this.process(doc));
+        await this.runCancelable(relevantDocuments, DocumentState.Indexed, cancelToken, doc => this.indexManager.update(doc, cancelToken));
```
spoenemann, line 97: I think we need to switch these two phases: first index, then process, because processing may involve resolving cross-references. Indexing has the constraint that no cross-references must be resolved in that phase (the same is true in Xtext).

The main reason for this is that I would propose to include precomputation of types for languages with typed expressions in the processing phase. Type computation involves a lot of following cross-references.

packages/langium/src/documents/document-builder.ts
```diff
@@ -54,55 +50,72 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+        await this.runCancelable(relevantDocuments, DocumentState.Validated, cancelToken, doc => this.validate(doc, cancelToken));
+    }
+
+    protected async runCancelable(documents: LangiumDocument[], state: DocumentState, cancelToken: CancellationToken, callback: (document: LangiumDocument) => MaybePromise<unknown>): Promise<void> {
```
spoenemann, line 101: Detail: I'd name the second argument `targetState`.

packages/langium/src/documents/document-builder.ts
```diff
@@ -54,55 +50,72 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+        if (cancelToken.isCancellationRequested) {
+            return;
+        }
```
spoenemann, line 104: I think checking the token is not necessary here since it's also done two lines below.

packages/langium/src/documents/document-builder.ts
```diff
@@ -54,55 +50,72 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+            await callback(document);
+            await delayNextTickAfter();
+        }
     }
 
     /**
      * Process the document by running precomputations. The default implementation precomputes the scope.
      */
     protected process(document: LangiumDocument): void {
         document.precomputedScopes = this.scopeComputation.computeScope(document);
```
spoenemann, line 118: I'd propose to pass the cancellation token into the scope computation.

packages/langium/src/documents/document.ts
```diff
@@ -10,15 +10,57 @@ import { TextDocuments } from 'vscode-languageserver/node';
...
+    constructor(textDocument: TextDocument, parseResult: ParseResult) {
+        this.parseResult = parseResult;
+        (this.parseResult.value as Mutable<AstNode>).$document = this;
+        this.state = DocumentState.Parsed;
+        this.textDocument = textDocument;
+        this.uri = URI.parse(textDocument.uri);
+    }
+
+    reset(): void {
+        this.state = DocumentState.Parsed;
```
spoenemann, line 57: When an affected document is reset, it does not need to be reindexed because its exported nodes should not depend on resolving cross-references (see above). So it should be sufficient to reset it to `Indexed`.

packages/langium/src/index/index-manager.ts
```diff
@@ -27,14 +26,21 @@ export interface IndexManager {
...
      */
-    update(document: LangiumDocument): void;
+    update(document: LangiumDocument, cancelToken?: CancellationToken): Promise<void>;
+
+    /**
+     * Returns all documents that could be affected by the specified document.
+     *
+     * @param document The document which may affect other documents.
+     */
+    getEffectedDocuments(document: LangiumDocument): Stream<LangiumDocument>;
```
spoenemann, line 43:
```suggestion
    getAffectedDocuments(document: LangiumDocument): Stream<LangiumDocument>;
```

packages/langium/src/references/linker.ts
```diff
@@ -51,15 +51,21 @@ export class DefaultLinker implements Linker {
 
     buildReference(node: AstNode, refNode: CstNode, text: string, crossRefId: string): Reference {
         const link = this.link.bind(this);
-        const reference: Reference & { _ref?: AstNode } = {
+        const reference: Reference & { _ref?: AstNode, _invalid: boolean, _hasListener: boolean } = {
             $refNode: refNode,
             $refName: text,
+            _invalid: false,
```
spoenemann, line 57: This seems to be unused.

packages/langium/src/utils/promise-util.ts
```diff
@@ -0,0 +1,35 @@
...
+let lastTick = -1;
+
+/**
+ * Delays the execution of the current code to the next tick of the event loop if more milliseconds than specified have elapsed since the last call of this method.
+ * @param elapsed The amount of milliseconds that needs to elapse for delaying to the next tick.
+ */
+export async function delayNextTickAfter(elapsed = 5): Promise<void> {
+    const current = Date.now();
+    if (lastTick < 0) {
+        lastTick = current;
```
spoenemann, line 28: Detail: In this case you can return without checking the delta.

packages/langium/src/documents/document.ts
```diff
@@ -10,15 +10,57 @@ import { TextDocuments } from 'vscode-languageserver/node';
...
     precomputedScopes?: PrecomputedScopes
-    valid: boolean
+    state: DocumentState
     textDocument: TextDocument
     uri: URI
+    reset(): void
+    onReset(listener: () => void): void
+}
+
+export class LangiumDocumentImpl implements LangiumDocument {
```
spoenemann, line 39: AFAICT the only reason to change to a class is the local handling of "reset listeners", but the only consumer is the linker with its created Reference objects. I'd propose to keep the simple interface and implement a more direct solution:
 * Add a `references: Reference[]` property to LangiumDocument.
 * The linker adds every new Reference to that array.
 * When the DocumentBuilder resets an affected document, it asks the Linker to unlink the contained references and then clears the array.
WDYT?

dhuebner:
> That's a good start! I think we have to refactor the DocumentBuilder, though. I propose the following:
> 
> * Create a dedicated method to be used by CLIs (maybe use `build` for that, but without parsing, see below).
> * `documentChanged` should work in separate phases:
>   
>   1. Invalidate, recreate and then `process` the changed document
>   2. Collect all affected documents (this is probably the responsibility of the IndexManager)
>   3. Update the index of the affected documents (including the changed one)
>   4. Run validation on all affected documents
> * The DocumentBuilder should keep track of `dirty` documents, that is documents reported by `documentChanged` and affected ones.
> * When a `documentChanged` call is cancelled, the next invocation should consider the documents that are still dirty, meaning that the previous invocation did not finish updating or validating them.
> * We should consider generalizing the `valid` flag of LangiumDocument to something like `BuilderState`, which can be `'changed'`, `'parsed'`, `'processed'`, `'indexed'`, `'validated'`.
> * If we follow this approach, the DocumentBuilder could keep all documents in a single queue and process that in the phases as outlined above.

It would be also nice to have a flag that mean ‚reentrant‘ processing, documents with this flag should  be queed for re-processing afterwards

packages/langium/src/documents/document-builder.ts
```diff
@@ -54,55 +50,72 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+    async documentChanged(uri: URI, cancelToken = CancellationToken.None): Promise<void> {
         this.langiumDocuments.invalidateDocument(uri);
         const newDocument = this.langiumDocuments.getOrCreateDocument(uri);
-        this.indexManager.update(newDocument);
-        this.langiumDocuments.all.filter(doc => this.isAffected(doc, uri)).forEach(
-            doc => {
-                this.build(doc);
-                this.indexManager.update(doc);
+        await delayNextTickAfter();
+        if (cancelToken.isCancellationRequested) {
```
spoenemann, line 81: Thought: what if we provide a special cancellation token where `isCancellationRequested` is of type `Promise<boolean>` and that internally calls `delayNextTickAfter` through a getter?
@msujew WDYT?

msujew: Sure, sounds good to me!

spoenemann: The advantage is that the caller of all these API methods can control themselves whether to interrupt the execution and after which time period.

The drawback is that the token interface is not compatible with that of `vscode-languageserver`. We could translate `vscode-languageserver` CancellationTokens into `langium` CancellationTokens with a utility function. Should we use a different name to avoid clashes?

msujew: For now, I thought about a `CancelToken` class that simply accepts a `CancellationToken` in its constructor. Or should we separate this stronger using an interface?

packages/langium/src/documents/document-builder.ts
msujew: Hm, we hit an issue with this. If the scope computation becomes async, loading grammars becomes async as well (since they need to precompute their scope locally). This in turn makes every service that requires the grammar (parser first and foremost) async as well. I don't think that's a path we can go down. Other ideas?

packages/langium/src/documents/document-builder.ts
spoenemann: I would prefer an interface, and it could be an extension of CancellationToken that adds a property. Example:
```typescript
export interface AsyncCancellationToken extends CancellationToken {
    readonly interruptAndCheck: Promise<boolean>;
}
```

Another thought, what do you think of this:
```typescript
export const OperationCanceled = Symbol('OperationCanceled');

export function interruptAndCheck(token: AsyncCancellationToken): Promise<void> {
    if (await token.interruptAndCheck) {
        throw OperationCanceled;
    }
}
```

This is similar to how it's done in Xtext:
https://github.com/eclipse/xtext-core/blob/83ae598a64313f511a2b2993dc23f0d17a092c9c/org.eclipse.xtext/src/org/eclipse/xtext/service/OperationCanceledManager.java#L85-L89

It's a way to avoid the frequent `if (...) { return; }` blocks.

msujew: Sure, sounds good as well 👍

spoenemann: It's probably not a good idea to choose the same name for the property and the utility function, though...

packages/langium/src/documents/document-builder.ts
spoenemann: I see three solutions:

 1. Leave the scope precomputation uninterruptible.
 2. Add the grammar to the `DefaultModuleContext` so you have to resolve it _before_ you create the services.
 3. Implement a synchronous scope precomputation as a utility function specifically for the grammar language.

I'm currently favoring solution 3, that would also allow us to hard-code it to consider only grammar rules so we would get even better performance for large grammars.

danieldietrich:
I have a proposal to change `CancelToken` a bit:

```diff
export class CancelToken {

    private source: CancellationToken;
-    private listeners: Array<() => unknown> = [];
+    private listeners: Array<() => unknown> | null = [];

    constructor(source: CancellationToken) {
        this.source = source;
-        this.source.onCancellationRequested(() => this.listeners.forEach(e => e()));
+        this.source.onCancellationRequested(() => {
+            this.listeners?.forEach(e => e())
+            this.listeners = null;
+        });
    }

    onCancel(listener: () => unknown): void {
+        if (this.listeners) {
+            this.listeners.push(listener);
+        } else {
+            listener();
+        }
    }

    get isCanceled(): Promise<boolean> {
        return delayNextTickAfter().then(() => this.source.isCancellationRequested);
    }
}
```

danieldietrich:
@msujew Sorry, I did not close this PR on purpose. Did not see it happen

spoenemann:
Another approach following up on my thought https://github.com/langium/langium/pull/244#discussion_r719157881:

```typescript
let lastTick = 0;
let globalInterruptionPeriod = 10;

/**
 * Reset the global interruption period and create a cancellation token source.
 */
export function startCancelableOperation(): AbstractCancellationTokenSource {
    lastTick = Date.now();
    return new CancellationTokenSource();
}

/**
 * Change the period duration for `interruptAndCheck` to the given number of milliseconds.
 * The default value is 10ms.
 */
export function setInterruptionPeriod(period: number) {
    globalInterruptionPeriod = period;
}

/**
 * This symbol may be thrown in an asynchronous context by any Langium service that receives
 * a `CancellationToken`. This means that the promise returned by such a service is rejected with
 * this symbol as rejection reason.
 */
export const OperationCanceled = Symbol('OperationCanceled');

/**
 * This function does two things:
 *  1. Check the elapsed time since the last call to this function or to `startCancelableOperation`. If the predefined
 *     period (configured with `setInterruptionPeriod`) is exceeded, execution is delayed with `delayNextTick`.
 *  2. If the predefined period is not met yet or execution is resumed after an interruption, the given cancellation
 *     token is checked, and if cancellation is requested, `OperationCanceled` is thrown.
 *
 * All services in Langium that receive a `CancellationToken` may potentially call this function, so the
 * `CancellationToken` must be caught (with an `async` try-catch block or a `catch` callback attached to
 * the promise) to avoid that event being exposed as an error.
 */
export async function interruptAndCheck(token: CancellationToken): Promise<void> {
    if (token === CancellationToken.None) {
        // Early exit in case cancellation was disabled by the caller
        return;
    }
    const current = Date.now();
    if (current - lastTick >= globalInterruptionPeriod) {
        lastTick = current;
        await delayNextTick();
    }
    if (token. isCancellationRequested) {
        throw OperationCanceled;
    }
}
```

This would combine `delayNextTickAfter` and the custom `CancelToken`, and it would enable cancellation checking with a single function call instead of an `if (cancelToken.isCancellationRequested) { return; }` block.

I'm sorry for the back-and-forth again. When it comes to fundamental decisions such as this, it takes a while until the thoughts are properly sorted in my mind 😅. But now I think solving the problem with utility functions is better than defining our own cancellation token.

For reference, here's the source of the vscode-languageserver `CancellationToken`:
https://github.com/microsoft/vscode-languageserver-node/blob/main/jsonrpc/src/common/cancellation.ts

packages/langium/src/documents/document-builder.ts
msujew: I implemented a special sync scope precomputation now like you proposed in 3. Works well 👍

packages/langium/src/documents/document-builder.ts
```diff
@@ -44,65 +40,78 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
-    build(document: LangiumDocument): BuildResult {
-        const parseResult = this.parser.parse(document);
-        document.parseResult = parseResult;
-        this.process(document);
-        let diagnostics: Diagnostic[] | undefined = this.connection ? this.validate(document) : undefined;
-        const validator = this.documentValidator;
+    async build(document: LangiumDocument, cancelToken = CancellationToken.None): Promise<BuildResult> {
+        await this.buildDocument(document, cancelToken);
         return {
             parseResult: document.parseResult,
```
spoenemann, line 58: The `parseResult` is not built in this function, but is already contained in the given document. We should remove it from the BuildResult.

packages/langium/src/documents/document-builder.ts
```diff
@@ -44,65 +40,78 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+        if (this.connection || forceDiagnostics) {
+            diagnostics = await validator.validateDocument(document, cancelToken);
+            await interruptAndCheck(cancelToken);
+            if (this.connection) {
+                // Send the computed diagnostics to VS Code.
+                this.connection.sendDiagnostics({ uri: document.textDocument.uri, diagnostics });
+            }
         }
+        await interruptAndCheck(cancelToken);
+        document.state = DocumentState.Validated;
```
spoenemann, line 75: Shouldn't this state change be moved to the `if` block above?

packages/langium/src/documents/document-builder.ts
```diff
@@ -4,23 +4,19 @@
...
 
 export interface DocumentBuilder {
-    build(document: LangiumDocument): BuildResult
-    /**
-     * Validates a document and creates diagnostic information.
-    * @param document to be validated
-    * @returns Diagnostic array or undefined if no validation issues found
-    */
-    validate(document: LangiumDocument): Diagnostic[] | undefined
+    build(document: LangiumDocument, cancelToken?: CancellationToken): Promise<BuildResult>
```
spoenemann, line 19: Please add some documentation to the public API methods, including a `@throws OperationCanceled` with some explanation.

packages/langium/src/documents/document-builder.ts
```diff
@@ -44,65 +40,78 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+        await this.buildDocument(newDocument, cancelToken);
+    }
+
+    protected async buildDocument(document: LangiumDocument, cancelToken: CancellationToken): Promise<void> {
+        const affectedDocuments = this.indexManager.getAffectedDocuments(document);
+        affectedDocuments.forEach(e => {
+            this.linker.unlink(e);
+            e.state = DocumentState.Indexed;
+        });
+        const relevantDocuments = Array.from(new Set([
```
spoenemann, line 92: What happens with the order of documents after throwing them into a Set? Could this lead to indeterministic execution order? Is there some kind of ordered set (tree set) to avoid this?

msujew: JavaScript's set use insertion order, they are basically arrays with a `has` method. Same for maps.

PR #247, opened 2021-09-28
Auto detect multiline comments

### LINKED ISSUE #204: Consider language-specific comment format in syntax highlighting
The generated syntax highlighting rules currently declare standard comments. We should adapt this to the actual comment format in the grammar, e.g. by looking at `ML_COMMENT` and `SL_COMMENT` rules.

### LINKED ISSUE #228: Automatically detect comment rules in grammar
For #204, we need some way to detect comment rules in grammars. I propose the following solution:

Every terminal that is hidden but does not match a white space string is a comment terminal by default.

msujew:
Closes #204
Closes #228

Automatically detects multiline comments by visiting the input regex. Courtesy of the `regex-to-ast` package that is included in Chevrotain.

This affects both the default `GrammarConfig` object that no longer assumes `ML_COMMENT` by default. Also changes the textMate generator to include all comments in a single comment regex.

### DIFF
--- a/examples/domainmodel/syntaxes/domain-model.tmLanguage.json
+++ b/examples/domainmodel/syntaxes/domain-model.tmLanguage.json
@@ -6,6 +6,9 @@
   ],
   "patterns": [
     {
+      "include": "#comments"
+    },
+    {
       "name": "keyword.control.domain-model",
       "match": "\\b(datatype|entity|extends|many|package)\\b"
     },
@@ -33,10 +36,10 @@
           }
         },
         {
-          "begin": "(^\\s+)?(?=//)",
+          "begin": "//",
           "beginCaptures": {
             "1": {
-              "name": "punctuation.whitespace.comment.leading.cs"
+              "name": "punctuation.whitespace.comment.leading.domain-model"
             }
           },
           "end": "(?=$)",
--- a/packages/langium/src/grammar/grammar-util.ts
+++ b/packages/langium/src/grammar/grammar-util.ts
@@ -44,6 +44,10 @@ export function isDataTypeRule(rule: ast.ParserRule): boolean {
     return false;
 }
 
+export function isCommentTerminal(terminalRule: ast.TerminalRule): boolean {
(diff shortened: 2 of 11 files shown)


examples/domainmodel/syntaxes/domain-model.tmLanguage.json
```diff
@@ -13,36 +17,5 @@
...
-              "name": "punctuation.whitespace.comment.leading.cs"
-            }
-          },
-          "end": "(?=$)",
-          "name": "comment.line.domain-model"
-        }
-      ]
-    }
-  }
+  ]
```
spoenemann, line 20: Multiline comment highlighting does not work with these changes.
Screen Shot 2021-09-28 at 15 09 44

packages/langium/src/grammar/grammar-util.ts
```diff
@@ -44,6 +44,10 @@ export function isDataTypeRule(rule: ast.ParserRule): boolean {
     return false;
 }
 
+export function isCommentTerminal(terminalRule: ast.TerminalRule): boolean {
+    return terminalRule.$container.hiddenTokens.some(e => e.ref === terminalRule) && !' '.match(terminalRule.regex);
```
spoenemann, line 48: Shall we use the regexp visitor to detect whitespace-only terminals similarly to how you did it to detect multiline comments?

spoenemann:
This will also fix #204, right?

examples/domainmodel/syntaxes/domain-model.tmLanguage.json
msujew: True, the textmate implementation only looks at a single line at a time. For multiline comments we need to use the repository. That requires some extensive changes in order to calculate how a comment starts and ends. I implemented some heuristic for that.

packages/langium/src/grammar/grammar-util.ts
msujew: I think this solution is the best, as it does not rely on our custom detection behavior with the visitor (probably missing a lot of cases).

spoenemann (approved):
:+1:

PR #248, opened 2021-09-28
Add the global maxLookahead configuration

Pluralia:
Closes [the issue/202](https://github.com/langium/langium/issues/202).
Allows to configure Chevrotain's `IParserConfig` as part of `LangiumConfig`; provides a new service `ParserConfig` similarly to the `LanguageMetaData` service.

### DIFF
--- a/examples/domainmodel/langium-config.json
+++ b/examples/domainmodel/langium-config.json
@@ -5,5 +5,12 @@
     "out": "src/language-server/generated",
     "textMate": {
         "out": "syntaxes/domain-model.tmLanguage.json"
+    },
+    "parserConfig": {
+      "chevrotainConfig": { 
+          "recoveryEnabled": true,
+          "nodeLocationTracking": "onlyOffset",
+          "maxLookahead": 3
+      }
     }
 }
--- a/examples/domainmodel/langium-config.json
+++ b/examples/domainmodel/langium-config.json
@@ -6,11 +6,9 @@
     "textMate": {
         "out": "syntaxes/domain-model.tmLanguage.json"
     },
-    "parserConfig": {
-      "chevrotainConfig": { 
-          "recoveryEnabled": true,
-          "nodeLocationTracking": "onlyOffset",
-          "maxLookahead": 3
-      }
+    "chevrotainParserConfig": {
+        "recoveryEnabled": true,
+        "nodeLocationTracking": "onlyOffset",
+        "maxLookahead": 3
     }
 }
--- a/packages/langium-cli/langium-config-schema.json
+++ b/packages/langium-cli/langium-config-schema.json
@@ -60,10 +60,6 @@
                             "description": "Enable computation of CST nodes location",
                             "type": "string"
                         },
-                        "errorMessageProvider": {
-                            "description": "A custom error message provider",
-                            "type": "object"
-                        },
                         "traceInitPerf": {
                             "description": "A flag to print performance tracing logs during parser initialization",
                             "type": ["boolean", "number"]
--- a/packages/langium-cli/langium-config-schema.json
+++ b/packages/langium-cli/langium-config-schema.json
@@ -40,49 +40,40 @@
                         "out"
                     ]
                 },
-                "parserConfig": {
-                    "description": "An object to describe the parser configuration",
+                "chevrotainParserConfig": {
+                    "description": "An object to describe the Chevrotain parser configuration",
                     "type": "object",
                     "properties": {
-                        "chevrotainConfig": {
-                            "description": "An object to describe the Chevrotain parser configuration",
-                            "type": "object",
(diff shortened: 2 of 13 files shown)


examples/domainmodel/langium-config.json
```diff
@@ -5,5 +5,12 @@
     "out": "src/language-server/generated",
     "textMate": {
         "out": "syntaxes/domain-model.tmLanguage.json"
+    },
+    "parserConfig": {
+      "chevrotainConfig": {
```
spoenemann, line 10: I'd propose to simplify this structure by eliminating one nesting level and name the group `chevrotainParserConfig`.

packages/langium-cli/langium-config-schema.json
```diff
@@ -40,6 +40,50 @@
...
+                                    "description": "A flag to support Dynamically defined Tokens",
+                                    "type": "boolean"
+                                },
+                                "nodeLocationTracking": {
+                                    "description": "Enable computation of CST nodes location",
+                                    "type": "string"
+                                },
+                                "errorMessageProvider": {
+                                    "description": "A custom error message provider",
+                                    "type": "object"
```
spoenemann, line 69: I guess this config option cannot be set via JSON, so we should rather remove it from here.

packages/langium/src/services.ts
```diff
@@ -26,6 +26,7 @@ export type LangiumGeneratedServices = {
     Grammar: Grammar
     AstReflection: AstReflection
     LanguageMetaData: LanguageMetaData
+    ParserConfig: ParserConfig
```
spoenemann, line 29: Please move this to the `parser` group:
```suggestion
    parser: {
        ParserConfig: ParserConfig
    }
```

packages/langium-cli/src/generator/module-generator.ts
```diff
@@ -13,29 +13,48 @@ export function generateModule(grammar: langium.Grammar, config: LangiumConfig):
...
+    }
+
     node.append('export const ', grammar.name, 'GeneratedModule: Module<LangiumServices, LangiumGeneratedServices> = {', NL);
     node.indent(moduleNode => {
         moduleNode.append(
             'Grammar: () => grammar(),', NL,
             'AstReflection: () => new ', grammar.name, 'AstReflection(),', NL,
-            'LanguageMetaData: () => languageMetaData', NL
+            'LanguageMetaData: () => languageMetaData,', NL,
+            `ParserConfig: () => ${config.parserConfig ? 'parserConfig' : 'defaultParserConfig'}`, NL
```
spoenemann, line 57: I'd propose a different approach: Mark the `ParserConfig` service as optional and merge the default options with the user defined options in the ChevrotainWrapper constructor:
```typescript
    constructor(tokens: TokenType[], config?: IParserConfig = {}) {
        const mergedConfig = {
            recoveryEnabled: true,
            nodeLocationTracking: 'onlyOffset',
            ...config
        };
        super(tokens, mergedConfig);
    }
```

spoenemann (commented):
Thanks!

examples/domainmodel/langium-config.json
Pluralia: It's made specially for simple adding of other parser configurations. I'll remove it.

packages/langium/src/parser/langium-parser.ts
```diff
@@ -31,6 +32,7 @@ export class LangiumParser {
     private readonly lexer: Lexer;
     private readonly nodeBuilder = new CstNodeBuilder();
     private readonly wrapper: ChevrotainWrapper;
+    private readonly config: IParserConfig | undefined;
```
spoenemann, line 35: We don't need to store that in a property if we only use it in the constructor.

packages/langium/src/parser/langium-parser.ts
```diff
@@ -302,8 +305,12 @@ class ChevrotainWrapper extends EmbeddedActionsParser {
 
     private analysed = false;
 
-    constructor(tokens: TokenType[]) {
-        super(tokens, { recoveryEnabled: true, nodeLocationTracking: 'onlyOffset' });
+    constructor(tokens: TokenType[], config?: IParserConfig) {
+        super(tokens, {
+            recoveryEnabled: true,
+            nodeLocationTracking: 'onlyOffset',
+            ...config
```
spoenemann, line 312: I'm wondering about this syntax being allowed although `config` is potentially undefined. I couldn't find any specification about that.

spoenemann (approved):
Just one remark and a question, then it's good to merge.

packages/langium/src/parser/langium-parser.ts
Pluralia: I've tried it a playground -- it's working nicely.

PR #250, opened 2021-09-29
Add langium logo to readme

msujew:
Adds the langium logo to the readme and updates some other bits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,27 @@
-# Langium
-[![npm](https://img.shields.io/npm/v/langium)](https://www.npmjs.com/package/langium)
-[![Build](https://github.com/langium/langium/actions/workflows/build.yml/badge.svg)](https://github.com/langium/langium/actions/workflows/build.yml)
-[![Gitpod Ready-to-Code](https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/langium/langium)
+<div id="langium-logo" align="center">
+    <a href="https://github.com/langium/langium">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,27 @@
-# Langium
-[![npm](https://img.shields.io/npm/v/langium)](https://www.npmjs.com/package/langium)
-[![Build](https://github.com/langium/langium/actions/workflows/build.yml/badge.svg)](https://github.com/langium/langium/actions/workflows/build.yml)
-[![Gitpod Ready-to-Code](https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/langium/langium)
+<div id="langium-logo" align="center">
```
spoenemann, line 1: From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div#attributes):

> The align attribute is obsolete; do not use it anymore. Instead, you should use CSS properties or techniques such as CSS Grid or CSS Flexbox to align and position `<div>` elements on the page.

Alternative: `style="text-align: center;"`

spoenemann (approved):
Thanks, just a detail:

README.md
msujew: Style properties are stripped from all html elements in Github Readme's. See [here](https://github.com/github/markup#github-markup):

> The HTML is sanitized, aggressively removing things that could harm you and your kin—such as script tags, **inline-styles**, and class or id attributes.

spoenemann: :-/ ok

PR #255, opened 2021-10-01
Made parseDocument generic

### LINKED ISSUE #245: Consider type parameter for LangiumDocument
It would be nice to have a type parameter so we don't need to cast the root ast node:
```ts
// case LangiumDocument<DomainModel>

const  document = parseDocument<DomainModel>(services, datatypeFile);
const DomainModel model = document.parseResult.value;
```

See https://github.com/langium/langium/pull/231#discussion_r714695475

jbicker:
Fixes #245


packages/langium/src/test/langium-test.ts
```diff
@@ -9,16 +9,17 @@ import { TextDocument } from 'vscode-languageserver-textdocument';
...
+export function parseHelper<T extends AstNode = AstNode>(services: LangiumServices): (input: string) => BuildResult<T> {
     const metaData = services.LanguageMetaData;
     const documentBuilder = services.documents.DocumentBuilder;
     return input => {
         const randomNumber = Math.floor(Math.random() * 10000000) + 1000000;
         const textDocument = TextDocument.create(`file:/${randomNumber}${metaData.fileExtensions[0]}`, metaData.languageId, 0, input);
-        const document: LangiumDocument = services.documents.LangiumDocumentFactory.fromTextDocument(textDocument);
-        const buildResult = documentBuilder.build(document);
+        const document = services.documents.LangiumDocumentFactory.fromTextDocument(textDocument) as LangiumDocument<T>;
+        const buildResult = documentBuilder.build<T>(document);
```
msujew, line 22: Perhaps we should make `fromTextDocument` generic as well with `<T extends AstNode = AstNode>` so we don't need to cast here?

packages/langium/src/test/langium-test.ts
```diff
@@ -122,9 +123,9 @@ export function expectGoToDefinition(services: LangiumServices, expectEqual: Exp
     };
 }
 
-export function parseDocument(services: LangiumServices, input: string): LangiumDocument {
-    const buildResult = parseHelper(services)(input);
-    const document = getDocument(buildResult.parseResult.value);
+export function parseDocument<T extends AstNode = AstNode>(services: LangiumServices, input: string): LangiumDocument<T> {
+    const buildResult = parseHelper<T>(services)(input);
+    const document = getDocument(buildResult.parseResult.value) as LangiumDocument<T>;
```
msujew, line 128: I'd propose we change `getDocument` to be generic as well.

packages/langium/src/documents/document-builder.ts
```diff
@@ -54,8 +55,8 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
         this.indexManager = services.index.IndexManager;
     }
 
-    build(document: LangiumDocument): BuildResult {
-        const parseResult = this.parser.parse(document);
+    build<T extends AstNode = AstNode>(document: LangiumDocument<T>): BuildResult<T> {
+        const parseResult = this.parser.parse(document) as ParseResult<T>;
```
msujew, line 59: I don't think this cast is necessary here.

```suggestion
        const parseResult = this.parser.parse<T>(document);
```

jbicker: Ah yes. Thanks. The parse method just got generic in the last change, wasn't at the moment I've did the upper change.

msujew (commented):
Alright, the code looks good to me 👍

One small remark: Inside of the different example CLI implementations, there are calls to these methods as well. Would you mind replacing the type casts with type arguments as well?

msujew:
@jbicker Still on it? There is a small merge conflict in the `langium-parser.ts` If you fix the other comment this PR is good to go!

jbicker:
squashed and force pushed the changes we directly talked about.

Issue #210, opened 2021-09-10
Let linker return AstNode or LinkingError


dhuebner:
As we talked wit @spoenemann about separating linking from the user validation we found out, that it would be a good idea to return a `LinkingError` instead of `undefined` when a cross reference can not be resolved.

```ts
export interface Linker {
    link(node: AstNode, referenceName: string, referenceId: string): AstNode | LinkingError;
    // ... 
}
```

The `_ref` implementation should also be changed to work with `LinkingError` instead of `undefined` (see `DefaultLinker.buildReference`)

Inside a `LinkingError ` we could store information about the problematic reference and an error message, that should be enough to produce a `Diagnostic` from.

sailingKieler:
Proposal: also rename `link(...)` to `getLinkedObject(...)`.

spoenemann:
> Proposal: also rename `link(...)` to `getLinkedObject(...)`.

...or rather `getLinkedNode(...)` because we usually refer to model elements as "AST nodes".

PR #256, opened 2021-10-04
Introduced 'LinkingError' as demanded in #210..…

sailingKieler:
updated Linker.link(), DefaultLinker.link(...) and DefaultLinker.buildReference(...) accordingly.


packages/langium/src/references/linker.ts
```diff
@@ -29,12 +36,27 @@ export class DefaultLinker implements Linker {
...
-            return this.loadAstNode(description);
+    getLinkedNode(node: AstNode, referenceString: string, referenceId: string): AstNode | LinkingError {
+        const description = this.getCandidate(node, referenceString, referenceId);
+        if (description === undefined) {
+            return {
+                cause: 'noDescription',
+                referenceId: referenceId,
+                referenceString: referenceString
+            };
+        } else {
```
dhuebner, line 47: `else` can be omitted as we are returning in `if`, right?

packages/langium/src/references/linker.ts
```diff
@@ -50,16 +72,15 @@ export class DefaultLinker implements Linker {
...
-        const reference: Reference & { _ref?: AstNode } = {
+        const getLinkedNode = this.getLinkedNode.bind(this);
+        const reference: Reference & { _ref?: AstNode | LinkingError } = {
             $refNode: refNode,
             $refName: text,
             get ref() {
-                if (reference._ref === undefined || !getDocument(reference._ref).valid) {
-                    // TODO handle linking errors
-                    reference._ref = link(node, text, crossRefId);
+                if (reference._ref === undefined || isAstNode(reference._ref) && !getDocument(reference._ref).valid) {
```
dhuebner, line 80: Should it be `or not isAstNode`?  `|| !isAstNode(reference._ref` ?

dhuebner: @msujew
I think one of the main ideas to have a LinkingError (not just `undefined`) is to prevent trying to resolve the reference when it is clear it can not be resolved.
Otherwise validation and other code that operates on a linked document will try to link again and again

PR #257, opened 2021-10-08
Reworked Stream API and implementation

### LINKED ISSUE #71: Improve Stream.concat behavior
As stated in [here](https://github.com/TypeFox/langium/pull/57#pullrequestreview-682794953), there is still some work left to do regarding the `stream.concat` implementation.

spoenemann:
Closes #71.

 * Added more useful functions to `Stream`, mostly inspired from the JavaScript `Array` functions.
 * Removed the `ArrayLikeStream` type as we're not using it anywhere and it's hard to maintain full compatibility between `Array` and `Stream`.
 * Reworked the internal implementation.
 * Added tests for all functions except some that are implicitly tested in other tests.

### DIFF
--- a/packages/langium/src/utils/stream.ts
+++ b/packages/langium/src/utils/stream.ts
@@ -27,9 +27,27 @@ export interface Stream<T> extends Iterable<T> {
     isEmpty(): boolean;
 
     /**
+     * Determines the number of elements in this stream.
+     */
+    count(): number;
+
+    /**
      * Collects all elements of this stream into an array.
      */
     toArray(): T[];
+
+    /**
+     * Collects all elements of this stream into a Set.
+     */
+    toSet(): Set<T>;
+
+    /**
+     * Collects all elements of this stream into a Map, applying the provided functions to determine keys and values.
+     *
+     * @param keyFn The function to derive map keys. If omitted, the stream elements are used as keys.
+     * @param valueFn The function to derive map values. If omitted, the stream elements are used as values.
+     */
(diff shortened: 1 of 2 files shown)


packages/langium/src/utils/stream.ts
```diff
@@ -4,34 +4,222 @@
...
+     *
+     * @param callbackfn This method calls the function once for each element in the stream, providing
+     *        the previous and current values of the reduction.
+     * @param initialValue If specified, `initialValue` is used as the initial value to start the
+     *        accumulation. The first call to the function provides this value as an argument instead
+     *        of a stream value.
+     */
+    reduce(callbackfn: (previousValue: T, currentValue: T) => T): T | undefined;
+    reduce(callbackfn: (previousValue: T, currentValue: T) => T, initialValue: T): T;
+    reduce<U>(callbackfn: (previousValue: U, currentValue: T) => U, initialValue: U): U;
```
msujew, line 123: Can we merge the definition of those methods by using `<U = T>`? The `U` type parameter doesn't have to used explicitly anyway, since TypeScript infers type parameters:

```ts
const a = s.stream(['A']); // a is `Stream<string>`
a.reduce((prev, cur) => prev + cur.length, 0); // TypeScript knows that `U` is `number`
```

spoenemann: I copied those declarations from `Array`, but yes we can simplify it.

msujew (approved):
LGTM! I'm ready to re-review if you plan to add more tests but I'm fine with merging this in its current state.

spoenemann:
I added more functionality and rewrote the tests. @msujew could you look again?

PR #271, opened 2021-10-27
Listen to LSP connection file changes

### LINKED ISSUE #243: Listen to file changes and update the index
Listen to file changes and update the index:
- delete
- rename
- create?

see connection.onDidChangeWatchedFiles

msujew:
Closes #243

Also fixes an issue related to the glob pattern which was used in the yeoman generator. The generated pattern was incorrect and changes were never picked up.

You can test this by opening the `domain-model` example workspace and **only** open one of the files in vscode. Deleting/Changing files in the workspace outside of vscode should invalidate references. Changing them back or restoring them should create valid files again.

### DIFF
--- a/packages/generator-langium/langium-template/src/extension.ts
+++ b/packages/generator-langium/langium-template/src/extension.ts
@@ -1,6 +1,5 @@
 import * as vscode from 'vscode';
 import * as path from 'path';
-import { workspace } from 'vscode';
 import {
     LanguageClient, LanguageClientOptions, ServerOptions, TransportKind
 } from 'vscode-languageclient/node';
@@ -34,12 +33,15 @@ function startLanguageClient(context: vscode.ExtensionContext): LanguageClient {
         debug: { module: serverModule, transport: TransportKind.ipc, options: debugOptions }
     };
 
+    const fileSystemWatcher = vscode.workspace.createFileSystemWatcher('**/*.<%= file-glob-extension %>');
+    context.subscriptions.push(fileSystemWatcher);
+
     // Options to control the language client
     const clientOptions: LanguageClientOptions = {
         documentSelector: [{ scheme: 'file', language: '<%= language-id %>' }],
         synchronize: {
             // Notify the server about file changes to files contained in the workspace
-            fileEvents: workspace.createFileSystemWatcher('**/*.<%= file-extension %>')
+            fileEvents: fileSystemWatcher
         }
     };
 
--- a/packages/langium/src/documents/document-builder.ts
+++ b/packages/langium/src/documents/document-builder.ts
@@ -27,16 +27,17 @@ export interface DocumentBuilder {
     build(document: LangiumDocument, cancelToken?: CancellationToken): Promise<BuildResult>
     /**
      * This method is called when a document change is detected.
-     * Implementation should updates the state of that `LangiumDocument` and make sure
+     * Implementation should update the state of associated `LangiumDocument` instances and make sure
      * that the index information of the affected documents are also updated.
      *
-     * @param uri of the document that was changed
+     * @param changed URIs of changed/created documents
+     * @param deleted URIs of deleted documents
      * @param cancelToken allows to cancel the current operation
      * @see IndexManager.update()
      * @see LangiumDocuments.invalidateDocument()
      * @throws `OperationCancelled` if cancellation is detected during execution
      */
-    documentChanged(uri: URI, cancelToken?: CancellationToken): Promise<void>;
+    update(changed: URI[], deleted: URI[], cancelToken?: CancellationToken): Promise<void>;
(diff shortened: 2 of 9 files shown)


dhuebner:
I checked the domain model example and can verify that it works. 

I have an additional question about FS watcher. Will the watcher created with `workspace.createFileSystemWatcher` be disposed automatically when client.stop() is called or is it not necessary at all?

packages/langium/src/documents/document-builder.ts
```diff
@@ -76,40 +77,54 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
         return diagnostics;
     }
 
-    async documentChanged(uri: URI, cancelToken = CancellationToken.None): Promise<void> {
-        this.langiumDocuments.invalidateDocument(uri);
-        const newDocument = this.langiumDocuments.getOrCreateDocument(uri);
+    async update(changed: URI[], deleted: URI[], cancelToken = CancellationToken.None): Promise<void> {
+        // Gather deleted documents first, as they won't be available later
+        // They are needed to correctly collect all affected documents
+        const deletedDocuments = deleted.map(e => this.langiumDocuments.getOrCreateDocument(e));
```
spoenemann, line 92: AFAICT we only need the URIs of the deleted documents, not their LangiumDocument instances.

packages/langium/src/index/index-manager.ts
```diff
@@ -23,26 +23,33 @@ export interface IndexManager {
...
-    update(document: LangiumDocument | LangiumDocument[], cancelToken?: CancellationToken): Promise<void>;
+    update(documents: LangiumDocument[], cancelToken?: CancellationToken): Promise<void>;
 
     /**
      * Returns all documents that could be affected by the specified document.
      *
      * @param document The document which may affect other documents.
      */
-    getAffectedDocuments(document: LangiumDocument): Stream<LangiumDocument>;
+    getAffectedDocuments(documents: LangiumDocument[]): Stream<LangiumDocument>;
```
spoenemann, line 52: Use the URIs as input so we don't have to resolve deleted Documents.

packages/langium/src/lsp/language-server.ts
```diff
@@ -82,23 +81,33 @@ export function startLanguageServer(services: LangiumServices): void {
...
-        } catch (err) {
-            if (err !== OperationCancelled) {
-                throw err;
-            }
-        }
+        changePromise = documentBuilder
+            .update(changed, deleted ?? [], changeTokenSource.token)
+            .catch(err => {
+                if (err !== OperationCancelled) {
+                    console.log('Error: ', err);
```
spoenemann, line 99: Log errors to `console.error`

packages/langium/src/lsp/language-server.ts
```diff
@@ -82,23 +81,33 @@ export function startLanguageServer(services: LangiumServices): void {
...
-        try {
-            changePromise = documentBuilder.documentChanged(URI.parse(change.document.uri), changeTokenSource.token);
-        } catch (err) {
-            if (err !== OperationCancelled) {
-                throw err;
-            }
-        }
+        changePromise = documentBuilder
+            .update(changed, deleted ?? [], changeTokenSource.token)
+            .catch(err => {
```
spoenemann, line 97: Why did you change to a callback instead of a `catch` block?

spoenemann (commented):
@dhuebner you are right. FileSystemWatcher extends Disposable, so we should add all created watchers to `context.subscriptions` so they're cleaned up when the extension is stopped.

packages/langium/src/lsp/language-server.ts
msujew: Because the `catch` block doesn't actually catch the exception there (since we don't await the promise). It only catches it if we are at the previous `await` instruction, but there's no guarantee that we actually hit that code.

packages/generator-langium/langium-template/src/extension.ts
```diff
@@ -34,12 +33,15 @@ function startLanguageClient(context: vscode.ExtensionContext): LanguageClient {
         debug: { module: serverModule, transport: TransportKind.ipc, options: debugOptions }
     };
 
+    const fileSystemWatcher = vscode.workspace.createFileSystemWatcher('**/*.<%= file-glob-extension %>');
+	context.subscriptions.push(fileSystemWatcher);
```
spoenemann, line 37: Nitpick: inconsistent whitespace

spoenemann (approved):
:+1:

packages/generator-langium/langium-template/src/extension.ts
msujew: Thanks, somehow a tab character found it's way into there 🤔

Issue #198, opened 2021-09-01
Validate cyclic rule definitions


spoenemann:
1. Cyclic rule definitions shall be marked with an error.
2. All tools working with the grammar shall be robust against cycles: throw an error, but don't run into endless loops / recursion.

spoenemann:
Point 2 can be omitted because we validate the grammar when running the CLI.

spoenemann:
After discussing with @Pluralia I realized that this doesn't make a lot of sense. For now we rely on the checks made by Chevrotain when validating the parser in the langium-cli. Eventually we might implement automatic transformation of left-recursive rules so this kind of checks becomes obsolete.

PR #272, opened 2021-10-27
Validate cyclic rule definitions

Pluralia:
Closes [issue/198](https://github.com/langium/langium/issues/198).
Proposes left recursion validation.

### DIFF
--- a/examples/domainmodel/src/language-server/domain-model.langium
+++ b/examples/domainmodel/src/language-server/domain-model.langium
@@ -5,7 +5,13 @@ Domainmodel:
     (elements+=AbstractElement)*;
 
 AbstractElement:
-    PackageDeclaration | Type;
+    PackageDeclaration | Type | R;
+
+R:
+    'a'? T 'a' | 'b';
+
+T:
+    '('? R ')';
 
 PackageDeclaration:
     'package' name=QualifiedName '{'
--- a/examples/domainmodel/src/language-server/domain-model.langium
+++ b/examples/domainmodel/src/language-server/domain-model.langium
@@ -5,22 +5,7 @@ Domainmodel:
     (elements+=AbstractElement)*;
 
 AbstractElement:
-    PackageDeclaration | Type | R;
-
-R:
-    'a'? T 'a' | 'b';
-
-T:
-    '('? R ')';
-
-TT:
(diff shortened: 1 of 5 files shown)


Pluralia:
I've considered the algorithm, proposed by @msujew, and took a decision postpone its implementation. The main reason is that currently there is no concrete application domain of this algorithm. How I understand we need to validate cyclic definitions to don't catch stack overflow in a process of grammar handling. However, the only one stack overflow which I've met for now was fixed by refactoring [here](https://github.com/langium/langium/pull/272/files#diff-15bec189480f5554c85cffd20f91da2123502e1eca86548030fa45cf97e0c8d5).

Actually, it doesn't mean we can drop validation of cyclic definitions or left recursion (it can be useful to prevent Chevrotain's errors), but we should determine exact cases / grammar rules that must be validated, because cycles not always have to be forbidden.

Regarding to the proposed algorithm of cyclic definition rules detection.

> We identify every mandatory rule call inside of a ParserRule instance by just walking down the AST, ignoring any optional (?/*) elements and collecting all mandatory rule calls inside

Can we omit optional element here?
`
EntryRule: (r+=R)* 'end';
R: R;
`
> Once we have all mandatory rule calls, we can just apply any old graph cycle detection algorithm on top of that.

Thanks to alternatives we have not simple graph, but "multidimensional" graph: each node with alternatives have some groups of children, that have to be considered "in parallel". For example, the `EntryRule` has children [X, Y], but X has 3 possible children groups: [R, Y], [S], [T]. It makes us consider 3 different graphs: where X has children from the 1st alternative, from the 2nd and the 3rd.
`
EntryRule: X Y;
X: R Y | S | T;
`
Rules R, Y, S and T also can have alternatives. So, the graph cycle detection algorithm has to be modified and becomes more complicated.

spoenemann:
Removing the milestone assignment as I think we need more time to think this through.

sailingKieler:
This PR is 3 month old now.
@Pluralia @spoenemann is it still valid?

I've no idea how much effort rebasing might be.
Currently, it doesn't seem that reviewing it makes sense.

spoenemann:
Let's close it for now. It was my fault: see #198

PR #275, opened 2021-10-30
Store LSP range on CST nodes

msujew:
Simplifies dealing with ranges when handling CST nodes by storing the position (line+character) info on the cst nodes. This alleviates the need to deal with range/offset transformation in different services.


spoenemann (approved):
:+1:

PR #276, opened 2021-11-01
Allow digits and uppercase letters in file extensions

dhuebner:
Made the validator for file extensions in the yeoman generator more lenient.


sailingKieler (approved):
👍

PR #278, opened 2021-11-01
Implement change affection computation #237

### LINKED ISSUE #237: Implement change affection computation
https://github.com/langium/langium/blob/83986e9e91f116a57e8263efb4941d2215eca45b/packages/langium/src/documents/document-builder.ts#L105-L107

This placeholder implementation should be replaced with an index lookup to avoid rebuilding the whole workspace on every change.

msujew:
Closes #237 

Computes which documents are affected by changes in different documents. By default this affects documents with linking errors or ones which reference the changed document.


dhuebner (approved):
LGTM

PR #288, opened 2021-11-05
Refactor terminal definitions

### LINKED ISSUE #185: Change the syntax for hidden terminals
I propose to change

```
grammar Foo
hidden(WS, SL_COMMENT, ML_COMMENT)

terminal WS: /\s+/;
terminal ML_COMMENT: /\/\*[\s\S]*?\*\//;
terminal SL_COMMENT: /\/\/[^\n\r]*/;
```

to

```
grammar Foo

hidden terminal WS: /\s+/;
```

The only potential drawback I can think of is that it becomes less clear which terminals shall be hidden in a scenario where you have multiple languages in one project. So let's keep this open until we have a clear path for handling multiple languages and reusability of grammar rules.

### LINKED ISSUE #194: Extend terminal syntax to enable composition
In Xtext you can use other terminal rules in a terminal definition. Shall we add this composition to our terminals as well?

The alternative is to use datatype rules, which already support composing terminals. The difference is that a terminal produces a single token, while a datatype rule can produce many tokens in the CST.

msujew:
Closes #185
Closes #194

* Allows hidden on terminal declarations
* Add validations and quick fixes for hidden tokens on grammar
* Enable composition of terminal rules

These changes would usually warrant two separate pull requests. However, dealing with multiple concurrent PRs that change the generated structure of the grammar does not sound like a good idea to me. So I simply did both changes at once in here.


packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -135,40 +132,43 @@ ParenthesizedElement returns AbstractElement:
...
+	{RegexToken} regex=RegexLiteral;
 
-Wildcard:
+Wildcard returns AbstractElement:
 	{Wildcard} '.';
 
-CharacterRange:
-	left=Keyword ('..' right=Keyword)?;
+CharacterRange returns AbstractElement:
+	{CharacterRange} left=Keyword ('..' right=Keyword)?;
```
spoenemann, line 173: Given that we can express the same with regular expressions, some of these constructs are redundant. We should still keep them to ease migration from Xtext. But what do you think about adding hints through validation to discourage some constructs in favor of the regexp equivalent?

spoenemann (commented):
Great! General remarks:
 * Add a validation that forbids using a hidden terminal in a parser rule or terminal rule.
 * Change the grammar in our generator-langium template.

packages/langium/src/grammar/langium-grammar.langium
msujew: Added a hint and a quick-fix for that 👍

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -96,33 +106,71 @@ export class LangiumGrammarValidator {
...
+    }
+
+    checkUsedHiddenTerminalRule(rule: ast.AbstractRule, accept: ValidationAcceptor): void {
+        if (ast.isTerminalRule(rule) && rule.hidden) {
+            return;
+        }
+        streamAllContents(rule).forEach(({ node }) => {
+            if (ast.isRuleCall(node) || ast.isTerminalRuleCall(node)) {
+                const ref = node.rule.ref;
+                if (ast.isTerminalRule(ref) && ref.hidden) {
```
spoenemann, line 140: Is it maybe better performance-wise to associate the validation with `RuleCall | TerminalRuleCall` and check the containing rule in such a case? This would avoid the additional `streamAllContents` call.

packages/langium/src/grammar/langium-grammar-code-actions.ts
```diff
@@ -58,4 +66,76 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
...
+        const rootCst = document.parseResult.value.$cstNode;
+        if (rootCst) {
+            const cstNode = findLeafNodeAtOffset(rootCst, offset);
+            const element = cstNode?.element;
+            const container = ast.isCharacterRange(element) ? element : getContainerOfType(element, ast.isCharacterRange);
+            if (container && container.right && container.$cstNode) {
+                const left = container.left.value;
+                const right = container.right.value;
+                return {
+                    title: 'Refactor into regex',
```
spoenemann, line 81: I would write the longer term "regular expression" in such user-facing strings, unless there's a real need to shorten it.

spoenemann (approved):
Two more remarks, otherwise good:

PR #290, opened 2021-11-05
Add `prepackage` script to regenerate all files

msujew:
Adds a new script that should to be run before publishing all npm packages. Regenerates all generated files in the workspace and compiles them.


spoenemann:
Currently we run `npm publish` separately in the individual packages because we need to update versions as part of the process. Eventually we will document this. I think in this case the top-level prepackage script will not be considered, right?

examples/domainmodel/src/language-server/generated/module.ts
```diff
@@ -14,7 +14,7 @@ export const languageMetaData: LanguageMetaData = {
 
 export const parserConfig: IParserConfig = {
     recoveryEnabled: true,
-    nodeLocationTracking: 'onlyOffset',
+    nodeLocationTracking: 'full',
```
dhuebner, line 17: Q: Is this somehow related to the build?

msujew:
Due to the comment by @spoenemann, I will close this for now.

PR #291, opened 2021-11-06
Optimize container lookups in parser

msujew:
Caches features and their assignments inside the parser. Using this way, the `getContainerOfType` assignment lookup is only needed once per feature, which greatly improves parsing performance; Up to 20% in my tests.


packages/langium/src/parser/langium-parser.ts
```diff
@@ -256,6 +262,17 @@ export class LangiumParser {
         return obj;
     }
 
+    private getAssignment(feature: AbstractElement): AssignmentElement {
+        if (!this.assignmentMap.has(feature)) {
+            const assignment = getContainerOfType(feature, isAssignment);
+            this.assignmentMap.set(feature, {
+                assignment: assignment,
+                crossRef: assignment && isCrossReference(assignment.terminal) || false
```
spoenemann, line 270: IMO this is more readable: `assignment ? isCrossReference(assignment.terminal) : false`

spoenemann (approved):
Thanks!

PR #293, opened 2021-11-09
Correctly escape keywords for textmate and remove symbols

### LINKED ISSUE #292: Remove symbol highlighting for generated textmate grammar
Usually programming languages for VSCode don't highlight symbols (brackets, parenthesis, etc.) in their textmate grammar. We should align to that decision for our generated textmate grammar.

msujew:
Closes #292 

Also fixes a bug that prevents correct highlighting when the input isn't regex escaped.


spoenemann (approved):
:+1:

Issue #68, opened 2021-06-15
Make entry rule explicit


spoenemann:
I don't like the fact that Xtext assumes the first rule to be the entry rule. I would prefer to make this explicit in Langium, e.g. with a keyword `main` or `root`. We could offer a code action that adds this keyword to the first rule to simplify migration from Xtext.

spoenemann:
Alternative keyword: `entry`

dhuebner:
As I could imagine to have more than one entry rule I like `entry` keyword much more, otherwise having more than one main entry rule will lead to the same ordering problem as without any keywords.
Maybe I'm mixing it with a parser "entry rule", we are talking hier about the `root` AST rule/type, than `entry` is not that good...

spoenemann:
> Maybe I'm mixing it with a parser "entry rule", we are talking hier about the root AST rule/type

Isn't that the same? Whatever entry rule is used by the parser will become the root node.

dhuebner:
With Chevrotain you can use any rule as an **entry** rule, but for the langium document we should have only one. That why `entry` keyword for a parser rule confuse me a bit. 

`root` keyword will reflect that the parser rule AST Type will became the root type.
`main` keyword says that a parser rule will be used as a main or starting entry rule for the Langium document.

PR #305, opened 2021-11-16
Make entry rule explicit

Pluralia:
Closes [#68](https://github.com/langium/langium/issues/68).

### DIFF
--- a/packages/langium/src/grammar/grammar-util.ts
+++ b/packages/langium/src/grammar/grammar-util.ts
@@ -321,7 +321,9 @@ export function getRuleType(rule: ast.AbstractRule | undefined): string {
 }
 
 export function getEntryRule(grammar: ast.Grammar): ast.ParserRule | undefined {
-    return grammar.rules.find(e => ast.isParserRule(e)) as ast.ParserRule;
+    const entryRules = grammar.rules.filter(e => ast.isParserRule(e) && e.entry) as ast.ParserRule[];
+    // returns undefined if more than 1 entry rules are found
+    return entryRules.length !== 1 ? undefined : entryRules[0];
 }
 
 export function loadGrammar(json: string): ast.Grammar {
--- a/packages/langium/src/grammar/grammar-util.ts
+++ b/packages/langium/src/grammar/grammar-util.ts
@@ -321,9 +321,7 @@ export function getRuleType(rule: ast.AbstractRule | undefined): string {
 }
 
 export function getEntryRule(grammar: ast.Grammar): ast.ParserRule | undefined {
-    const entryRules = grammar.rules.filter(e => ast.isParserRule(e) && e.entry) as ast.ParserRule[];
-    // returns undefined if more than 1 entry rules are found
-    return entryRules.length !== 1 ? undefined : entryRules[0];
+    return grammar.rules.find(e => ast.isParserRule(e) && e.entry) as ast.ParserRule;
 }
 
 export function loadGrammar(json: string): ast.Grammar {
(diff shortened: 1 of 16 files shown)


packages/langium/src/grammar/grammar-util.ts
```diff
@@ -321,7 +321,9 @@ export function getRuleType(rule: ast.AbstractRule | undefined): string {
 }
 
 export function getEntryRule(grammar: ast.Grammar): ast.ParserRule | undefined {
-    return grammar.rules.find(e => ast.isParserRule(e)) as ast.ParserRule;
+    const entryRules = grammar.rules.filter(e => ast.isParserRule(e) && e.entry) as ast.ParserRule[];
+    // returns undefined if more than 1 entry rules are found
+    return entryRules.length !== 1 ? undefined : entryRules[0];
```
spoenemann, line 326: Please just return the first rule marked as `entry`. We already have the validation to catch the case of more than one entry.

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -71,16 +71,14 @@ export class LangiumGrammarValidator {
...
-            if (isDataTypeRule(firstRule)) {
-                accept('error', 'The entry rule cannot be a data type rule.', { node: firstRule, property: 'name' });
-            } else if (firstRule.fragment) {
-                accept('error', 'The entry rule cannot be a fragment.', { node: firstRule, property: 'name' });
-            }
-        } else {
+    checkEntryGrammarRule(grammar: ast.Grammar, accept: ValidationAcceptor): void {
+        const entryRules = grammar.rules.filter(e => ast.isParserRule(e) && e.entry) as ast.ParserRule[];
+        if (entryRules.length === 0) {
             accept('error', 'This grammar is missing an entry parser rule.', { node: grammar, property: 'name' });
```
spoenemann, line 77: In case there is at least one parser rule, please assign this error to the first parser rule and add a code action to fix the problem by adding the `entry` keyword prefix. This will ease migration from Xtext.

spoenemann (commented):
Very good! Just two comments

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -71,16 +71,14 @@ export class LangiumGrammarValidator {
...
-    checkFirstGrammarRule(grammar: ast.Grammar, accept: ValidationAcceptor): void {
-        const firstRule = getEntryRule(grammar);
-        if (firstRule) {
-            if (isDataTypeRule(firstRule)) {
-                accept('error', 'The entry rule cannot be a data type rule.', { node: firstRule, property: 'name' });
-            } else if (firstRule.fragment) {
-                accept('error', 'The entry rule cannot be a fragment.', { node: firstRule, property: 'name' });
-            }
-        } else {
+    checkEntryGrammarRule(grammar: ast.Grammar, accept: ValidationAcceptor): void {
```
msujew, line 74: Suggestion: Please add another validation case here for fragment rules. A fragment rule can't be an entry rule.

Pluralia: The grammar was changed such way, that it's impossible to create a rule with `fragment` and `entry` words simultaneously, so, I think such validation for fragment rules is redundant.

msujew: I see, alright.

spoenemann (approved):
:+1:

Issue #200, opened 2021-09-03
Change cross-reference syntax


spoenemann:
The cross-reference syntax of Xtext is `[Type|TOKEN]`, where `Type` is the AST type of the reference and `TOKEN` is a terminal rule or datatype rule that shall be parsed here. This syntax is misleading because in the context of TypeScript `|` means an alternative.

Proposals: `[Type: TOKEN]`, `[TOKEN: Type]`

Of course omitting the token shall still be possible: `[Type]` (the token defaults to `ID`)

spoenemann:
Note: We should still allow the old syntax, but mark it with a validation error. This way we can offer a code action to fix it.

dhuebner:
> but mark it with a validation error

Let use at least warning

>  because in the context of TypeScript

I can't recall that we were worried about clashes with Java syntax as the Xtext grammar language was invented

spoenemann:
I understand why that decision was made: it's like the mathematical notation for sets `{a | a > 0}`. The same reasoning is behind the lambda syntax of Xtend. The point is that it's inconsistent with all major programming languages that we use. We should keep the good parts of the Xtext syntax, but change the bad parts. Easing migration is a separate topic, we could even offer a CLI to convert .xtext to .langium.

debovema:
A CLI to convert XText to Langium would be so nice 😍

spoenemann:
If we change the syntax as proposed, we could enable alternative types for cross-references: `[Type1 | Type2 | Type3]` (or `[Type1 | Type2 | Type3: TOKEN]`)

Pluralia:
The idea to have alternatives `[Type1 | Type2 | Type3]` is very nice! Should we open a new issue for that?

However, I have concerns about `[Type1 | Type2 | Type3: TOKEN]` syntax:

1. On the first view `TOKEN` belongs only to `Type3` -- it can confuse
2. What if we want reference on different tokens for different alternatives?

I'd propose to replace it with `[(Type1 | Type2 | Type3): TOKEN]`. Also, for the case, when `Type1` and `Type2` are referenced by `TOKEN_X` and `Type3` -- by `TOKEN_Y`: `[(Type1 | Type2): TOKEN_X | Type3: TOKEN_Y]``

spoenemann:
> Should we open a new issue for that?

I'd wait for real-world use cases first.

sailingKieler:
I just stumbled over this issue. In the spirit of
[...]

PR #306, opened 2021-11-17
Replace '|' with ':' in cross-references

Pluralia:
Closes [#200](https://github.com/langium/langium/issues/200).

### DIFF
--- a/packages/langium/src/grammar/langium-grammar-code-actions.ts
+++ b/packages/langium/src/grammar/langium-grammar-code-actions.ts
@@ -38,9 +38,9 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
                 return this.fixHiddenTerminals(diagnostic, document);
             case IssueCodes.UseRegexTokens:
                 return this.fixRegexTokens(diagnostic, document);
-            case IssueCodes.MakeRuleEntry:
+            case IssueCodes.EntryRuleTokenSyntax:
                 return this.addEntryKeyword(diagnostic, document);
-            case IssueCodes.CrossRefSyntaxFix:
+            case IssueCodes.CrossRefTokenSyntax:
                 return this.fixCrossRefSyntax(diagnostic, document);
             default:
                 return undefined;
@@ -119,7 +119,7 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
 
     private fixCrossRefSyntax(diagnostic: Diagnostic, document: LangiumDocument): CodeAction {
         return {
-            title: 'Replace \'|\' with \':\'',
+            title: "Replace '|' with ':'",
             kind: CodeActionKind.QuickFix,
             diagnostics: [diagnostic],
             isPreferred: true,
@@ -127,7 +127,7 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
                 changes: {
                     [document.textDocument.uri]: [{
                         range: diagnostic.range,
-                        newText: document.textDocument.getText(diagnostic.range).replace(/\|/, ':')
+                        newText: ':'
                     }]
                 }
             }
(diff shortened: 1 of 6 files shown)


packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -52,6 +53,7 @@ export namespace IssueCodes {
     export const RuleNameUppercase = 'rule-name-uppercase';
     export const HiddenGrammarTokens = 'hidden-grammar-tokens';
     export const UseRegexTokens = 'use-regex-tokens';
+    export const CrossRefSyntaxFix = 'cross-ref-syntax-fix';
```
spoenemann, line 56: The issue code should be named after the issue, not its quick fix. Here I'd name it `cross-ref-token-syntax`.

The same remark might apply to your other PR.

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -142,6 +144,12 @@ export class LangiumGrammarValidator {
         }
     }
 
+    checkCrossReferenceSyntax(crossRef: ast.CrossReference, accept: ValidationAcceptor): void {
+        if (crossRef.terminal && crossRef.$cstNode?.text.includes('|')) {
+            accept('error', '\'|\' is depreceted. Please, use \':\' instead.', { node: crossRef, code: IssueCodes.CrossRefSyntaxFix });
```
spoenemann, line 149:
```suggestion
            accept('error', "'|' is deprecated. Please use ':' instead.", { node: crossRef, code: IssueCodes.CrossRefSyntaxFix });
```

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -142,6 +144,12 @@ export class LangiumGrammarValidator {
         }
     }
 
+    checkCrossReferenceSyntax(crossRef: ast.CrossReference, accept: ValidationAcceptor): void {
+        if (crossRef.terminal && crossRef.$cstNode?.text.includes('|')) {
```
spoenemann, line 148: You shouldn't use `$cstNode.text` because that includes hidden tokens such as comments. The following cross-ref definition also triggers the validation error although it's syntactically valid:
```
[Foo:/*|*/ID]
```
The correct way to solve this would be to analyze the sub-tree of the CST and look for leaf tokens with the text `|`. You can look into `cst-util.ts` whether there is already a useful utility function for that, or add one if it helps.

An alternative would be to include information about the chosen syntax in the AST by changing the grammar grammar:
```
CrossReference returns AbstractElement:
	{CrossReference} '[' type=[ParserRule:ID] ((deprecatedSyntax?='|' | ':') ^terminal=CrossReferenceableTerminal )? ']';
```
Then you can read the `deprecatedSyntax` property to check which syntax was used. This is the simpler solution, and I think it's ok because we don't regard the grammar language AST format as public API.

packages/langium/src/grammar/langium-grammar-code-actions.ts
```diff
@@ -96,6 +98,23 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
...
+        return {
+            title: 'Replace \'|\' with \':\'',
+            kind: CodeActionKind.QuickFix,
+            diagnostics: [diagnostic],
+            isPreferred: true,
+            edit: {
+                changes: {
+                    [document.textDocument.uri]: [{
+                        range: diagnostic.range,
+                        newText: document.textDocument.getText(diagnostic.range).replace(/\|/, ':')
```
spoenemann, line 111: The diagnostic range might include comments. For example, when the code action is applied to the following cross-ref definition, only the comment is modified, but not the actual cross-ref token separator.
```
[Foo/*|*/|ID]
```
See my comment above for a solution. The CST subtree should give you sufficient information to find the exact position of the character that needs to be replaced.

spoenemann (approved):
Good work!

PR #308, opened 2021-11-23
Added new package langium-sprotty

spoenemann:
This is work in progress for integrating Langium with Sprotty. I will test it with the "states" example in [sprotty-vscode](https://github.com/eclipse/sprotty-vscode).

### DIFF
--- a/packages/langium-sprotty/src/diagram-server-manager.ts
+++ b/packages/langium-sprotty/src/diagram-server-manager.ts
@@ -29,7 +29,7 @@ export interface DiagramServerManager {
     removeClient(clientId: string): void
 }
 
-export class DefaultDiagramServerManager {
+export class DefaultDiagramServerManager implements DiagramServerManager {
 
     protected readonly connection?: Connection;
     protected readonly diagramGenerator: IDiagramGenerator;
--- a/packages/langium-sprotty/src/diagram-server-manager.ts
+++ b/packages/langium-sprotty/src/diagram-server-manager.ts
@@ -6,10 +6,11 @@
 
 import { CancellationToken, Connection } from 'vscode-languageserver';
 import { URI } from 'vscode-uri';
-import { ActionMessage, DiagramServer, IDiagramGenerator } from 'sprotty-protocol';
-import { DocumentState, equalURI, interruptAndCheck, LangiumDocument, stream } from 'langium';
-import { LangiumSprottyServices } from './sprotty-services';
+import { ActionMessage, DiagramOptions, DiagramServer, isRequestAction, RejectAction, RequestModelAction } from 'sprotty-protocol';
+import { DocumentState, equalURI, interruptAndCheck, LangiumDocument, ServiceRegistry, stream } from 'langium';
+import { LangiumSprottyServices, LangiumSprottySharedServices } from './sprotty-services';
 import { LangiumDiagramGeneratorArguments } from './diagram-generator';
+import { DiagramActionNotification } from './lsp';
 
 /**
  * A `DiagramServer` instance can handle exactly one client diagram. The host application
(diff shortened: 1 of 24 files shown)


spoenemann:
I will add an example to [sprotty-vscode](https://github.com/eclipse/sprotty-vscode) that demonstrates how this works. But I cannot add that example before this is merged and a snapshot version is published.

packages/langium/src/dependency-injection.ts
```diff
@@ -35,8 +35,8 @@ export type Module<I, T = I> = {
  * @param module3 (optional) third Module
  * @returns a new object of type I
```
rainerschuster, line 36: Add JavaDoc
```suggestion
 * @param module4 (optional) fourth Module
 * @returns a new object of type I
```

rainerschuster, line 36:
```suggestion
 * @param module4 (optional) fourth Module
 * @returns a new object of type I
```

rainerschuster (changes requested):
Fix JavaDoc for inject

packages/langium-sprotty/src/diagram-server-manager.ts
```diff
@@ -0,0 +1,117 @@
...
+     */
+    acceptAction(message: ActionMessage): Promise<void>
+
+    /**
+     * The client application notified closing of a specific diagram client.
+     */
+    removeClient(clientId: string): void
+}
+
+export class DefaultDiagramServerManager {
```
danieldietrich, line 32: (not only for documentation purposes) it would be better, if `DefaultDiagramServerManager` would implement `DiagramServerManager`

packages/langium-sprotty/src/lsp.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import {
+    Connection, NotificationType
+} from 'vscode-languageserver';
+import { ActionMessage } from 'sprotty-protocol';
+import { OperationCancelled } from 'langium';
+import { LangiumSprottyServices } from './sprotty-services';
+
+/* eslint-disable @typescript-eslint/no-namespace */
+export namespace DiagramActionNotification {
```
danieldietrich, line 15: Side-note: I ask myself why we do lint namespaces away, they seem to be useful for us. Or is there a better alternative? Maybe we should just allow namespaces (part of another PR).

packages/langium/src/dependency-injection.ts
```diff
@@ -35,8 +35,8 @@ export type Module<I, T = I> = {
...
  */
-export function inject<I1, I2, I3, I extends I1 & I2 & I3>(module1: Module<I, I1>, module2?: Module<I, I2>, module3?: Module<I, I3>): I {
-    const module = [module1, module2, module3].reduce(_merge, {}) as Module<I>;
+export function inject<I1, I2, I3, I4, I extends I1 & I2 & I3 & I4>(module1: Module<I, I1>, module2?: Module<I, I2>, module3?: Module<I, I3>, module4?: Module<I, I4>): I {
```
danieldietrich, line 38: _(ah - I spotted the fourth module 🙂, maybe I will be able to massage the types as part of my DI efforts, the way that we allow N modules)_

danieldietrich (changes requested):
Well done, Sir! Please find my review comments. Thx!

packages/langium/src/documents/document-builder.ts
```diff
@@ -126,6 +161,25 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
         }
+        await this.notifyBuildPhase(documents, targetState, cancelToken);
+    }
+
+    onBuildPhase(targetState: DocumentState, callback: DocumentBuildListener): void {
+        if (this.buildPhaseListeners.has(targetState)) {
+            this.buildPhaseListeners.get(targetState)!.push(callback);
+        } else {
+            this.buildPhaseListeners.set(targetState, [callback]);
+        }
```
msujew, line 172: We do this quite often throughout our codebase. Would it make sense to add a `MultiMap` to our utilities? Doesn't have to happen in this PR, but we should create an issue for this.

msujew (commented):
Aside from the issues the others have already mentioned, this looks good to me. I have a general question, but otherwise still looks good to go after addressing the other issues.

packages/langium/src/documents/document-builder.ts
danieldietrich: Maybe one 'utility' function would be sufficient?

```ts
// collections.ts
export function insert<K, V>(map: Map<K, V[]>, key: K, value: V): Map<K, V[]> {
    map.has(key) ? map.get(key).push(value) : map.set(key, [value]);
    return map;
}
```

msujew: Well, that leaves the issue of the `get` function to return `undefined` instead of an empty array, which would be preferable.

danieldietrich: Makes sense - a `MultiMap` type would be nice.

packages/langium/src/documents/document-builder.ts
```diff
@@ -116,6 +151,7 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
 
     protected async buildDocuments(documents: LangiumDocument[], cancelToken: CancellationToken): Promise<void> {
         await this.indexManager.update(documents.filter(e => e.state < DocumentState.Indexed), cancelToken);
+        await this.notifyBuildPhase(documents, DocumentState.Indexed, cancelToken);
```
msujew, line 154: Should we pass the filtered documents here, instead of all documents? Some of the documents have been indexed previously, but will still be passed to the callback. The same applies to the call of `notifyBuildPhase` of `runCancelable`.

msujew (approved):
The API looks good to me! I have a question about the semantics of `onBuildPhase` and which documents we pass in there.

packages/langium/src/utils/collections.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * Returns an array of all values associated with the given key. If no value exists,
+     * an empty array is returned.
+     *
+     * _Note:_ The returned array is assumed not to be modified. Use the `set` method to add a
+     * value and `delete` to remove a value from the multimap.
+     */
+    get(key: K): readonly V[] {
+        return this.map.get(key) ?? [];
```
danieldietrich, line 66: I had to look twice until I spotted the readonly :)

danieldietrich (approved):
This looks great, Miro!

packages/langium/src/utils/collections.ts
spoenemann: I used `ReadonlyArray<V>` first, but eslint complained about that :-)

spoenemann:
I adapted the code to the new multi-language support (#311). This required considerable changes. I will test it next week.

PR #311, opened 2021-11-25
Add multi-language support

### LINKED ISSUE #298: Support multiple languages in one language server
* Create a global DI container, move `LangiumDocuments` and `IndexManager` to it.
 * Create a global map of file extensions to `LangiumServices`.
 * Change `language-server.ts` so it uses the global map to look up the `LangiumServices` for each request.
 * Create a small, but useful multi-language example with cross-references between two languages in this repo.

### LINKED ISSUE #300: Add import statements to use grammar rules from other .langium files
Example:
```
grammar Foo

import './bar'
```

Alternative keywords: `use`, `include`

## Semantics

 * Grammar rules and terminal rules can be referenced from the same file and all imported files
 * Include all terminals from the same file and all imported files in the lexer
 * Consider all hidden terminals from the same file and all imported files

### LINKED ISSUE #301: Handle multi-language projects in generator
* It shall be possible to declare multiple languages, each with its own main `.langium` file declaring a grammar (#299), in `langium-config.json`.
 * We should use a common output folder for generated files and prefix the files with the language id where necessary.
 * The AST types shall be generated once for the whole project, so they should cover the union of all included languages.
 * In order to resolve imported files (#300), we should index the whole workspace folder before loading the main files.
 * The root folder for indexing shall be configurable. We should set it explicitly to `./src` in the `generator-langium` template to avoid indexing `node_modules`.

msujew:
Closes #298
Closes #300
Closes #301

Adds quite a lot of stuff:

* Rules referenced from other grammar files are now correctly validated; Only _imported_ grammars can be used for cross referencing. Quickfixes are provided.
* Services are now split into _shared_ and _non-shared_ services. Shared services instances are reused by different services implementations.
* Quite a lot of generator changes, as multiple grammars have to be generated into a single `ast`/`grammar`/`module` file.
* The langium-config.json schema has changed quite a bit.

The changes can be tested with the `statemachine` example package. I plan to add a dedicated simple grammar with dependencies for regression testing, but this should be enough for reviewing the code.


packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -11,20 +11,13 @@ hidden terminal SL_COMMENT: /\/\/[^\n\r]*/;
...
-AbstractMetamodelDeclaration :
-	GeneratedMetamodel | ReferencedMetamodel;
-
-GeneratedMetamodel :
-	'generate' name=ID ePackage=string ('as' alias=ID)?;
-
-ReferencedMetamodel :
-	'import' ePackage=string ('as' alias=ID)?;
-
+PackageImport:
```
spoenemann, line 19: "Package" seems like the wrong term to me. Maybe `GrammarImport`?

packages/langium/src/lsp/language-server.ts
```diff
@@ -11,11 +11,14 @@ import {
 import { TextDocument } from 'vscode-languageserver-textdocument';
 import { URI } from 'vscode-uri';
 import { LangiumDocument } from '../documents/document';
-import { LangiumServices } from '../services';
+import { LangiumServices, LangiumSharedServices } from '../services';
 import { OperationCancelled, startCancelableOperation } from '../utils/promise-util';
+import { ServiceRegistry } from '../service-registry';
 
 export function startLanguageServer(services: LangiumServices): void {
```
spoenemann, line 18: We should pass the shared services here.

packages/langium/src/lsp/language-server.ts
```diff
@@ -111,97 +114,88 @@ export function addDocumentsHandler(connection: Connection, documents: TextDocum
...
-    serviceCall: (document: LangiumDocument, params: P, cancelToken: CancellationToken) => HandlerResult<R, E>,
-    services: LangiumServices
+    serviceCall: (services: LangiumServices, document: LangiumDocument, params: P, cancelToken: CancellationToken) => HandlerResult<R, E>,
+    registry: ServiceRegistry
 ): RequestHandler<P, R | null, E> {
     return async (params: P, cancelToken: CancellationToken) => {
-        const document = paramsDocument(params, services);
+        const uri = URI.parse(params.textDocument.uri);
+        const concreteServices = registry.getService(uri);
+        const document = paramsDocument(params, concreteServices);
```
spoenemann, line 193: Now we call `URI.parse` twice. We should inline the `paramsDocument` function.

packages/langium/src/service-registry.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    add(ext: string, service: LangiumServices): void {
+        this.map[ext] = service;
+    }
+    getService(uri: URI): LangiumServices {
+        const ext = Utils.extname(uri);
+        if (ext in this.map) {
+            return this.map[ext];
+        } else {
+            throw new Error();
```
spoenemann, line 43: Is this the best place to generate an Error (if yes, please provide a useful message), or shall we rather return `undefined`?

packages/langium/src/service-registry.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { URI, Utils } from 'vscode-uri';
+import { LangiumServices } from './services';
+
+export interface ServiceRegistry {
+    getService(uri: URI): LangiumServices
+    readonly all: LangiumServices[]
+}
+
+export class SingleServiceRegistry implements ServiceRegistry {
```
spoenemann, line 15: We could provide a function that creates an `ExtensionServiceRegistry` (which could then be named `DefaultServiceRegistry`) for a single language.

packages/langium/src/services.ts
```diff
@@ -48,23 +47,36 @@ export type LangiumLspServices = {
...
         DocumentBuilder: DocumentBuilder
         LangiumDocuments: LangiumDocuments
         LangiumDocumentFactory: LangiumDocumentFactory
         TextDocuments: TextDocuments<TextDocument>
         TextDocumentFactory: TextDocumentFactory
+        IndexManager: IndexManager
+    }
+}
+
+export type LangiumSharedServices = LangiumDefaultSharedServices & LangiumGeneratedSharedServices
```
spoenemann, line 69: Please add documentation: what are the shared services, how do they relate to the `LangiumServices`?

packages/langium/src/default-module.ts
```diff
@@ -30,35 +30,58 @@ import { DefaultNameProvider } from './references/naming';
...
 import { DefaultScopeComputation, DefaultScopeProvider } from './references/scope';
 import { DefaultJsonSerializer } from './serializer/json-serializer';
-import { LangiumDefaultServices, LangiumServices } from './services';
+import { DeepPartial, LangiumDefaultServices, LangiumDefaultSharedServices, LangiumGeneratedServices, LangiumGeneratedSharedServices, LangiumServices, LangiumSharedServices } from './services';
 import { DefaultDocumentValidator } from './validation/document-validator';
 import { ValidationRegistry } from './validation/validation-registry';
+import { ExtensionServiceRegistry, ServiceRegistry, SingleServiceRegistry } from './service-registry';
 
-export type DefaultModuleContext = {
+export interface DefaultModuleContext {
```
spoenemann, line 38: Please rename to `SharedModuleContext`.

packages/langium/src/default-module.ts
```diff
@@ -30,35 +30,58 @@ import { DefaultNameProvider } from './references/naming';
...
     connection?: Connection
+    serviceRegistry?: ServiceRegistry
 }
 
-export function createDefaultModule(context: DefaultModuleContext = {}): Module<LangiumServices, LangiumDefaultServices> {
+export function injectSingleService<T extends LangiumServices>(
```
spoenemann, line 43: Can we provide a setup function that works well for an arbitrary number of languages? Otherwise the path from the single language (as we generate with `generator-langium`) to multiple languages is not clear for users.

packages/langium/src/default-module.ts
```diff
@@ -30,35 +30,58 @@ import { DefaultNameProvider } from './references/naming';
...
+    sharedModule: Module<LangiumSharedServices, LangiumDefaultSharedServices>,
```
spoenemann, line 44: The shared module could be created in this function?

packages/langium/src/default-module.ts
```diff
@@ -30,35 +30,58 @@ import { DefaultNameProvider } from './references/naming';
...
+}
+
+export function createSharedModule(context: DefaultModuleContext = {}): Module<LangiumSharedServices, LangiumDefaultSharedServices> {
     return {
-        parser: {
-            GrammarConfig: (injector) => createGrammarConfig(injector),
-            LangiumParser: (injector) => createLangiumParser(injector),
-            ValueConverter: () => new DefaultValueConverter(),
-            TokenBuilder: () => new DefaultTokenBuilder()
+        ServiceRegistry: () => context.serviceRegistry ?? new ExtensionServiceRegistry(),
```
spoenemann, line 57: What's the use case for creating the registry externally?

packages/langium/src/default-module.ts
```diff
@@ -30,35 +30,58 @@ import { DefaultNameProvider } from './references/naming';
...
             LangiumDocumentFactory: (injector) => new DefaultLangiumDocumentFactory(injector),
             DocumentBuilder: (injector) => new DefaultDocumentBuilder(injector),
             TextDocuments: () => new TextDocuments(TextDocument),
             TextDocumentFactory: (injector) => new DefaultTextDocumentFactory(injector),
+            IndexManager: (injector) => new DefaultIndexManager(injector)
+        }
+    };
+}
+
+export function createDefaultModule(shared: LangiumSharedServices): Module<LangiumServices, LangiumDefaultServices> {
```
spoenemann, line 72: Idea for simplifying the DI concept: Instead of generating separate modules for `LangiumGeneratedServices`, provide them as a simple object to be passed as additional parameter to `createDefaultModule`. Then they are included in the module returned here. Same approach for `LangiumGeneratedSharedServices` and `createSharedModule`.

It's only a small improvement, but would reduce the number of "services" and "module" types.

spoenemann (commented):
I think this is a good direction!

Once the PR reaches a more mature state, please change the `langium-generator` template accordingly and document the migration steps &ndash; not for all API changes, but at least for the API that is used by our template.

packages/langium/src/default-module.ts
msujew: Simple example might be something like special build files, which are not only identified by their file extension. See `package.json` which gets special handling by vscode (code completion, validation, etc). Or even files without extensions (`BUILD` files).

packages/langium/src/default-module.ts
msujew: What do you mean by that?

packages/langium/src/default-module.ts
msujew: Well, we have some sort of circular dependency. An instanceof `LangiumServices` needs a shared services instance, but the registry needs the `LangiumServices` instance. By providing the registry externally, we can provide the created `LangiumServices` instance back to the service registry.

spoenemann (approved):
:+1:

Issue #315, opened 2021-12-02
Add case-insensitive configuration for keywords parsing


Pluralia:
It would be great to add special syntax / parser's flag to define case-insensitive keywords. Currently it's possible to do only by a pretty exhausting pattern: use a regular expression of type `/[kK][eE][yY][wW][oO][rR][dD]/` for each keyword. The flag would allow to use usual `'keyword'` and say to parse it case-insensitive.

dhuebner:
This is a nice first step  the direction to case-insensitive languages!
Most case-insensitive languages have also qualifier case-insensitive. That why I would like to suggest to have that flag in a higher level e.g. `LanguageMetaData` or `GrammarConfig`

PR #316, opened 2021-12-03
Add case-insensitive configuration for keywords parsing

Pluralia:
Closes [#315](https://github.com/langium/langium/issues/315).
Adds a flag `caseInsensitive` in `LanguageMetaData`.

### DIFF
--- a/examples/arithmetics/example/example.calc
+++ b/examples/arithmetics/example/example.calc
@@ -1,14 +1,14 @@
-module example1
+MODULE example1
 
-def y: 1 + 3 - 99828932 / 2 + 2 - 1;
+Def y: 1 + 3 - 99828932 / 2 + 2 - 1;
 
-def x: 12 / 3 - 1;
+DEF x: 12 / 3 - 1;
 
 x * 2 - 4;
 
-def t: 4;
+deF t: 4;
 
-def func(t, x):
+DEF func(t, x):
     t * t * t + x;
 
 func(t, x);
--- a/examples/arithmetics/example/example.calc
+++ b/examples/arithmetics/example/example.calc
@@ -1,5 +1,9 @@
 MODULE example1
 
+@ImPlement
+\straNge\
+xX^(#*&^)$#*%*!^)}{:>:<q
+
 Def y: 1 + 3 - 99828932 / 2 + 2 - 1;
 
 DEF x: 12 / 3 - 1;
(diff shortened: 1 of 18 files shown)


packages/langium/src/parser/token-builder.ts
```diff
@@ -8,11 +8,11 @@ import { Lexer, TokenPattern, TokenType } from 'chevrotain';
 import { terminalRegex } from '..';
 import { Grammar, isKeyword, isTerminalRule, Keyword, TerminalRule } from '../grammar/generated/ast';
 import { streamAllContents } from '../utils/ast-util';
-import { partialMatches } from '../utils/regex-util';
+import { getCaseInsensitivePattern, partialMatches } from '../utils/regex-util';
 import { stream } from '../utils/stream';
 
 export interface TokenBuilder {
-    buildTokens(grammar: Grammar): TokenType[];
+    buildTokens(grammar: Grammar, caseInsensitive?: boolean): TokenType[];
```
spoenemann, line 15: I'd propose to wrap this in an object:
```suggestion
    buildTokens(grammar: Grammar, options?: { caseInsensitive?: boolean }): TokenType[];
```
This is better extensible and avoids a Boolean function parameter, which I regard as bad API in general.

packages/langium/src/utils/regex-util.ts
```diff
@@ -114,6 +114,14 @@ export function escapeRegExp(value: string): string {
     return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
 
+export function getCaseInsensitivePattern(keyword: string): string {
+    const regexLetters: string[] = [];
+    for (const letter of keyword) {
```
spoenemann, line 119: Please use the `map` function rather than a `for` loop with `push`.

examples/arithmetics/example/example.calc
```diff
@@ -1,14 +1,18 @@
-module example1
+MODULE example1
```
spoenemann, line 1: Please don't forget to revert the example before merging.

spoenemann (approved):
Good work! A few detail remarks below, feel free to merge after addressing them.

examples/arithmetics/example/example.calc
Pluralia: I'd propose to leave it to demonstrate the ability to make keywords case-insensitive.

PR #317, opened 2021-12-09
Add correct flat type

msujew:
Changes the return type for `Stream.flat` from `Stream<T>` to `FlatStream<T, D>`.

Contains a lot of casting, since the flat operation isn't really typesafe. For example a call like `stream.flat<5>()` will flatten the stream only by a single layer, while the actual type is flattened by 5 layers. Consumers of this method should never use the generic parameter and only call this using `stream.flat(5)`. (`Array.flat` has the same problem)

### DIFF
--- a/packages/langium/src/utils/stream.ts
+++ b/packages/langium/src/utils/stream.ts
@@ -19,7 +19,7 @@ export interface Stream<T> extends Iterable<T> {
     /**
      * Returns an iterator for this stream. This is the same as calling the `Symbol.iterator` function property.
      */
-    iterator(): Iterator<T, undefined>;
+    iterator(): IterableIterator<T>;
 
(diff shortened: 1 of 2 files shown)


packages/langium/src/utils/stream.ts
```diff
@@ -229,6 +229,13 @@ export interface Stream<T> extends Iterable<T> {
 
 }
 
+export type FlatStream<T, Depth extends number> = {
+    'done': Stream<T>,
+    'recur': T extends Array<infer InnerArr>
```
spoenemann: Use `Iterable` to support nested streams as well.

spoenemann (approved):
I'll add a commit to generalize `Array` to `Iterable`.

PR #330, opened 2021-12-23
Set supertype on ruletypes even without fields or actions

jbicker:
Fixes https://github.com/langium/langium/issues/152

### DIFF
--- a/examples/domainmodel/src/language-server/generated/ast.ts
+++ b/examples/domainmodel/src/language-server/generated/ast.ts
@@ -8,9 +8,7 @@
 import { AstNode, AstReflection, Reference, isAstNode } from 'langium';
 
 export interface AbstractElement extends AstNode {
-    readonly $container: Domainmodel | AbstractElement;
-    elements: Array<AbstractElement>
-    name: QualifiedName
+    readonly $container: Domainmodel | PackageDeclaration;
 }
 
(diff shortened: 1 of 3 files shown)


Pluralia (approved):
Great! Could you also commit the updated ast of the domainmodel example?

msujew (changes requested):
I think we need another solution for this, since this exhibits a lot of unexpected behavior. For example, a rule such as `X: Y;` (where Y is a parser rule) shouldn't generate a type `X`, which is the reason for the whole field and action check.

jbicker:
We discussed that with @spoenemann already. As far as I remember in case of AST generation such types should be generated. And as a designer of the Grammar and/or developer of a Code-Generator based on the AST I would expect to have such type inference.

jbicker:
> since this exhibits a lot of unexpected behavior.

@msujew: What exactly wouldn't be expected?

Pluralia:
How I understand, @msujew supposes different behavior when the rule `X` is alternatives and when it's a just "call", so 'X: Y' doesn't generate a type `X`, but `X: Y | Z` generates. Is it true?

jbicker:
Ah, ok. Thanks for the hint, @Pluralia .
But even then as a designer of the grammar I would expect such type hierarchy. I mean, it is the decision or responsibility of the one who designs the grammar, or?

Issue #294, opened 2021-11-09
Name clashing of terminal and non-terminal rules


Pluralia:
Langium version: 0.2.1

## Steps To Reproduce

Generate an AST by a grammar that contains the same word as a keyword and a rule name. For example, here the word `Function` gives name clashing.

```
Function:
	GlobalFunction | LocalizedText | Validator | ParameterList;

FunctionTypeReference:
	'Function' ParameterList ('to' returnType=TypeReference)?;
```

## The current behavior

Rises an error: `The grammar has both a Terminal(Token) and a Non-Terminal(Rule) named: <Function>.`

## The expected behavior

No errors. As a possible solution, support `^` like it's made for the keywords and property names clashing, so we have a keyword `'Function'` and a `^Function` rule.

Pluralia:
Due to [#372](https://github.com/langium/langium/pull/372#discussion_r786616304) another solution has to be presented.

spoenemann:
I propose to rename keyword tokens instead of grammar rules. I think that will be easier to understand for users.

Example: for a `'Function'` keyword, name the token `KW:Function`.

PR #337, opened 2022-01-04
Fix name clashing of terminal and non-terminal rules

Pluralia:
Closes [#294](https://github.com/langium/langium/issues/294).


msujew (commented):
Do you mind adding an automated test for this? I.e. create a grammar with a terminal/keyword and parser rule with the same name. Creating a parser for this grammar should not result in an error.

msujew:
@Pluralia For the tests, you can reuse this code, which I wrote for another PR:

https://github.com/langium/langium/blob/81b0aff271609707f411aed2e466aea5a3189598/packages/langium/test/parser/langium-parser-builder.test.ts#L76-L95

packages/langium/test/parser/langium-parser-builder.test.ts
```diff
@@ -73,6 +73,37 @@ describe('Predicated grammar rules', () => {
...
+        parserFromGrammar(parseResult.value);
+    });
+
+    test('Should have no lexer errors', () => {
+        expect(parseResult.lexerErrors).toEqual([]);
+    });
+
+    test('Should have no parser errors', () => {
+        expect(parseResult.parserErrors).toEqual([]);
+    });
```
msujew, line 103: These tests don't actually do anything. Instead, the actual test should consist in the `parserFromGrammar` method not to throw:

```ts
test('Should not throw on parser validation', () => {
  expect(() => {
    parserFromGrammar(parseResult.value);
  }).not.toThrow();
});
```

Pluralia: These tests fall, if I execute them on non-modified version and call `parserFromGrammar` [here](https://github.com/langium/langium/blob/0d5c796b2a9c4afa46121176b3344f1faf60a9a6/packages/langium/test/parser/langium-parser-builder.test.ts#L94), and I thought that they are working. Now I've got that any tests fall after `parserFromGrammar` executing, when it throws exeptions, so, my tests really do nothing. I rewrote them.

packages/langium/src/parser/langium-parser-builder.ts
```diff
@@ -282,4 +284,4 @@ function wrap(ctx: RuleContext, method: Method, cardinality: Cardinality): Metho
     } else {
         throw new Error();
     }
-}
+}
```
msujew, line 287: Nitpick: You removed the newline here

msujew (approved):
Looks good to me 👍

PR #341, opened 2022-01-06
Various cross reference related changes

### LINKED ISSUE #303: Enable keywords as cross-reference tokens
It should be possible to use a keyword as a cross-reference token. For example:
```
SelfReference:
	ref=[Symbol|'Self'];
```

msujew:
Closes #303 

* Allow keywords as cross reference tokens
* Fix cross reference behavior for omitted terminals
* Fix infinite recursion for data type rules
* Remove unused code

### DIFF
--- a/packages/langium/src/parser/langium-parser-builder.ts
+++ b/packages/langium/src/parser/langium-parser-builder.ts
@@ -6,7 +6,7 @@
 
 import { TokenType } from 'chevrotain';
 import { AbstractElement, Action, Alternatives, CrossReference, Grammar, Group, isAction, isAlternatives, isAssignment, isCrossReference, isGroup, isKeyword, isParserRule, isRuleCall, isTerminalRule, isUnorderedGroup, Keyword, ParserRule, RuleCall, UnorderedGroup } from '../grammar/generated/ast';
-import { Cardinality, getTypeName, isArrayOperator, isDataTypeRule } from '../grammar/grammar-util';
+import { Cardinality, findNameAssignment, getTypeName, isArrayOperator, isDataTypeRule } from '../grammar/grammar-util';
 import { LangiumServices } from '../services';
 import { hasContainerOfType, streamAllContents } from '../utils/ast-util';
 import { stream } from '../utils/stream';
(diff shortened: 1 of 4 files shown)


packages/langium/src/parser/langium-parser-builder.ts
```diff
@@ -167,12 +167,17 @@ function buildAction(ctx: RuleContext, action: Action): Method {
...
 }
 
-function buildCrossReference(ctx: RuleContext, crossRef: CrossReference): Method {
-    const terminal = crossRef.terminal;
+function buildCrossReference(ctx: RuleContext, crossRef: CrossReference, terminal = crossRef.terminal): Method {
     if (!terminal) {
-        const idx = ctx.consume++;
-        const idToken = getToken(ctx, 'ID');
-        return () => ctx.parser.consume(idx, idToken, crossRef);
+        if (!crossRef.type.ref) {
```
spoenemann, line 172: Currently the type of a cross-reference refers to a `ParserRule`, but that will change once we implement #91. Furthermore, the approach of finding an assignment to a `name` property does not work with custom `NameProvider` implementations.

What do you think of this approach: When a cross-reference token is not given, look for a terminal rule named `ID`. If that does not exist, create a standard ID rule in memory.

msujew: I would argue trying to reference a `Type` (#91) without specifying a terminal should result in a validation error. Relying on the existence of some `ID` terminal leads to a lot of implicit behavior which is difficult to debug.

Regarding the `NameProvider`: I'd agree, but again, users should specify the terminal, if they want to prevent the heuristic for finding the name terminal to run. In particular, overriding the actual `name` node probably will happen way less than someone using a `QualifiedName` or something similar as a name (instead of an `ID`). Therefore, I'd say to keep the heuristic, which works for the common case.

spoenemann: Xtext uses the `ID` convention, and I never found it to be problematic: https://www.eclipse.org/Xtext/documentation/301_grammarlanguage.html#cross-references

If we're concerned about too much implicit behavior, we can do it like Xtext and reject a grammar with implicit cross-reference token if no `ID` terminal is defined. Or make the token always mandatory (see also discussion at #200).

spoenemann (approved):

packages/langium/src/parser/langium-parser-builder.ts
spoenemann: After rethinking, I like your approach more than relying on an ID rule.

PR #343, opened 2022-01-06
Fix in cross-references

JoanGi:
This PR comes from [336](https://github.com/langium/langium/discussions/336), and the main goal is to evaluate a minor fix.

In resume: Cross-reference are not detected if this references are out of the scopes. In  [domain model example index ](https://github.com/langium/langium/blob/main/examples/domainmodel/src/language-server/domain-model-index.ts) fix it by using streamAllContent() instead of streamContent().  The initial proposal is to change [DefaultAstNodeDescriptionProvider ](https://github.com/langium/langium/blob/main/packages/langium/src/workspace/ast-descriptions.ts) to use streamAllContent(). It fix the issue, and cross-file cross-reference still works. 

Probably there will be cases where we want that the cross-reference cannot be detected if it is not in the scope. In this case, maybe developing a way to express the scope of the cross-reference(global/local) in the grammar could be a better solution.

Captura de pantalla 2022-01-06 222249
Captura de pantalla 2022-01-06 222134

### DIFF
--- a/packages/langium/src/workspace/ast-descriptions.ts
+++ b/packages/langium/src/workspace/ast-descriptions.ts
@@ -10,7 +10,7 @@ import { getReferenceId, Linker } from '../references/linker';
 import { NameProvider } from '../references/naming';
 import { LangiumServices } from '../services';
 import { AstNode, AstNodeDescription, ReferenceInfo } from '../syntax-tree';
-import { getDocument, isLinkingError, streamAllContents, streamContents, streamReferences } from '../utils/ast-util';
+import { getDocument, isLinkingError, streamAllContents, streamReferences } from '../utils/ast-util';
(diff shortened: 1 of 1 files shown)


dhuebner:
@JoanGi 
If I recall it correctly we explicitly decided to reduce indexing to the root node + the first containment level which is basically what  the `createDescriptions()` currently does.

I would rather propose to change the `AstNodeDescriptionProvider` behavior in your current language and keep the lightweight default  implementation as it is.

JoanGi:
Ok!,  have sense to keep lightweight the default implementation. Thank's for the efforts and the quick answer (that's really good!)!! :)

dhuebner:
Thanks for contributing @JoanGi !
By the way, the every first implementation of `createDescriptions()` was exactly like you suggested here ;-)

PR #345, opened 2022-01-07
Support semantic token highlighting

### LINKED ISSUE #342: Semantic token highlighting
See LSP feature [here](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_semanticTokens).

msujew:
Closes #342

#### How to test

Run the `Grammar Extension` and assert that the assignments, actions, and return types of grammars are correctly colored.

### DIFF
--- a/packages/langium/src/grammar/grammar-util.ts
+++ b/packages/langium/src/grammar/grammar-util.ts
@@ -164,6 +164,30 @@ export function findNodeForFeature(node: CstNode | undefined, feature: string | 
     return nodes[index];
 }
 
+export function findKeywordNode(node: CstNode | undefined, keyword: string): CstNode | undefined {
+    return findKeywordNodeInternal(node, keyword, node?.element);
+}
+
+export function findKeywordNodeInternal(node: CstNode | undefined, keyword: string, element: AstNode | undefined): CstNode | undefined {
+    if (!node || !element) {
+        return undefined;
+    }
+    if (ast.isKeyword(node.feature) && node.feature.value === keyword) {
+        return node;
+    }
+    if (node instanceof CompositeCstNodeImpl) {
+        for (const child of node.children) {
+            if (child.element === element) {
+                const childNode = findKeywordNodeInternal(child, keyword, element);
+                if (childNode) {
+                    return childNode;
+                }
+            }
+        }
+    }
+    return undefined;
+}
+
 /**
  * This `internal` declared method exists, as we want to find the first child with the specified feature.
  * When the own feature is named the same by accident, we will instead return the input value.
--- a/packages/langium/src/grammar/grammar-util.ts
+++ b/packages/langium/src/grammar/grammar-util.ts
@@ -12,6 +12,7 @@ import { LangiumServices } from '../services';
 import { AstNode, AstNodeDescription, CstNode } from '../syntax-tree';
 import { getContainerOfType, getDocument, Mutable, streamAllContents } from '../utils/ast-util';
 import { MultiMap } from '../utils/collections';
+import { streamCst } from '../utils/cst-util';
 import { escapeRegExp } from '../utils/regex-util';
 import { documentFromText, LangiumDocuments, PrecomputedScopes } from '../workspace/documents';
 import { createLangiumGrammarServices } from './langium-grammar-module';
@@ -165,6 +166,9 @@ export function findNodeForFeature(node: CstNode | undefined, feature: string | 
 }
 
 export function findKeywordNode(node: CstNode | undefined, keyword: string): CstNode | undefined {
+    if (node && ast.isKeyword(node.feature) && node.feature.value === keyword) {
+        return node;
+    }
     return findKeywordNodeInternal(node, keyword, node?.element);
 }
 
@@ -172,19 +176,21 @@ export function findKeywordNodeInternal(node: CstNode | undefined, keyword: stri
     if (!node || !element) {
         return undefined;
     }
-    if (ast.isKeyword(node.feature) && node.feature.value === keyword) {
-        return node;
-    }
-    if (node instanceof CompositeCstNodeImpl) {
-        for (const child of node.children) {
-            if (child.element === element) {
-                const childNode = findKeywordNodeInternal(child, keyword, element);
-                if (childNode) {
+    const treeIterator = streamCst(node).iterator();
+    let result: IteratorResult<CstNode>;
+    do {
+        result = treeIterator.next();
+        if (!result.done) {
+            const childNode = result.value;
+            if (childNode.element === element) {
+                if (ast.isKeyword(childNode.feature) && childNode.feature.value === keyword) {
                     return childNode;
                 }
+            } else {
+                treeIterator.prune();
             }
         }
-    }
+    } while (!result.done);
     return undefined;
 }
 
--- a/packages/langium/src/grammar/langium-grammar-semantic-token-provider.ts
+++ b/packages/langium/src/grammar/langium-grammar-semantic-token-provider.ts
@@ -5,28 +5,47 @@
  ******************************************************************************/
 
 import { SemanticTokenTypes } from 'vscode-languageserver';
-import { AbstractSemanticTokenProvider } from '../lsp/semantic-token-provider';
+import { AbstractSemanticTokenProvider, SemanticTokenAcceptor } from '../lsp/semantic-token-provider';
 import { AstNode } from '../syntax-tree';
 import { isAction, isAssignment, isParameter, isParameterReference, isParserRule, isTerminalRule } from './generated/ast';
 import { isDataTypeRule } from './grammar-util';
 
 export class LangiumGrammarSemanticTokenProvider extends AbstractSemanticTokenProvider {
 
-    protected highlightElement(node: AstNode): boolean {
+    protected highlightElement(node: AstNode, acceptor: SemanticTokenAcceptor): void {
         if (isAssignment(node)) {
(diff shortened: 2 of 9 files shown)


packages/langium/src/grammar/langium-grammar-semantic-token-provider.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    protected highlightElement(node: AstNode): boolean {
+        if (isAssignment(node)) {
+            this.highlightFeature(node, 'feature', SemanticTokenTypes.property);
+        } else if (isAction(node)) {
+            if (node.feature) {
+                this.highlightFeature(node, 'feature', SemanticTokenTypes.property);
+            }
+        } else if ((isParserRule(node) && isDataTypeRule(node) || isTerminalRule(node)) && node.type) {
+            this.highlightFeature(node, 'type', SemanticTokenTypes.type);
```
spoenemann, line 23: This seems not to work for me. It's supposed to highlight primitive return types, right?

packages/langium/src/lsp/language-server.ts
```diff
@@ -181,6 +186,59 @@ export function addRenameHandler(connection: Connection, services: LangiumShared
...
+            if (services.lsp.SemanticTokenProvider) {
+                return services.lsp.SemanticTokenProvider.semanticHighlightRange(document, params, cancelToken);
+            }
+            return new ResponseError<void>(0, '');
+        },
+        services
+    ));
+}
+
+export function createServerRequestHandler<P extends { textDocument: TextDocumentIdentifier }, R, PR, E = void>(
```
spoenemann, line 219: Shall we rename the other function to `createRequestHandler`?

packages/langium/src/lsp/semantic-token-provider.ts
```diff
@@ -0,0 +1,188 @@
...
+    [SemanticTokenModifiers.async]: 1 << 1,
+    [SemanticTokenModifiers.declaration]: 1 << 2,
+    [SemanticTokenModifiers.defaultLibrary]: 1 << 3,
+    [SemanticTokenModifiers.definition]: 1 << 4,
+    [SemanticTokenModifiers.deprecated]: 1 << 5,
+    [SemanticTokenModifiers.documentation]: 1 << 6,
+    [SemanticTokenModifiers.modification]: 1 << 7,
+    [SemanticTokenModifiers.readonly]: 1 << 8,
+    [SemanticTokenModifiers.static]: 1 << 9
+};
```
spoenemann, line 51: Strange that this is not part of `vscode-languageserver`.

packages/langium/src/lsp/semantic-token-provider.ts
```diff
@@ -0,0 +1,188 @@
...
+    full: {
+        delta: true
+    },
+    range: true
+};
+
+export interface SemanticTokenProvider {
+    semanticHighlight(document: LangiumDocument, params: SemanticTokensParams, cancelToken: CancellationToken): SemanticTokens
+    semanticHighlightRange(document: LangiumDocument, params: SemanticTokensRangeParams, cancelToken: CancellationToken): SemanticTokens
+    semanticHighlightDelta(document: LangiumDocument, params: SemanticTokensDeltaParams, cancelToken: CancellationToken): SemanticTokens | SemanticTokensDelta
```
spoenemann, line 67: We made the cancellation token optional for other services. Shall we do it here as well for consistency?

packages/langium/src/lsp/semantic-token-provider.ts
```diff
@@ -0,0 +1,188 @@
...
+            return -1;
+        } else if (startLine > this.range.end.line) {
+            return 1;
+        } else {
+            return 0;
+        }
+    }
+
+    /**
+	 * @return true to skip the children of this element, false otherwise
```
spoenemann, line 157: A boolean return type is hard to understand in this case when looking at a subclass implementation. I'd propose to either return something more specific (`void | 'prune'`) or pass the iterator as a parameter so `prune` can be called explicitly.

packages/langium/src/lsp/semantic-token-provider.ts
```diff
@@ -0,0 +1,188 @@
...
+
+export interface SemanticTokenProvider {
+    semanticHighlight(document: LangiumDocument, params: SemanticTokensParams, cancelToken: CancellationToken): SemanticTokens
+    semanticHighlightRange(document: LangiumDocument, params: SemanticTokensRangeParams, cancelToken: CancellationToken): SemanticTokens
+    semanticHighlightDelta(document: LangiumDocument, params: SemanticTokensDeltaParams, cancelToken: CancellationToken): SemanticTokens | SemanticTokensDelta
+}
+
+export type SemanticTokenAcceptor = (line: number, char: number, length: number, type: string, modifiers?: string[]) => void;
+
+export abstract class AbstractSemanticTokenProvider implements SemanticTokenProvider {
```
spoenemann, line 72: Please document which method shall be overridden in a subclass, or make that method abstract.

spoenemann (commented):
Thanks, this is a great start!

packages/langium/src/grammar/langium-grammar-semantic-token-provider.ts
msujew: It works on my machine 🤔

image

packages/langium/src/lsp/semantic-token-provider.ts
msujew: Usually a language server should only add those modifiers and types to the token legend, which are actually in use. Additionally, you can actually add custom modifiers and types, which are special to your IDE. I imagine there's some optimization going on, which is why they probably don't want to export some kind of "catch-all" token legend. In my first iteration the semantic token provider actually had a method to return all used types and modifiers, but that proved to be a lot of boilerplate code and really complex for multiple token providers.

packages/langium/src/grammar/langium-grammar-semantic-token-provider.ts
```diff
@@ -0,0 +1,32 @@
...
+import { AbstractSemanticTokenProvider } from '../lsp/semantic-token-provider';
+import { AstNode } from '../syntax-tree';
+import { isAction, isAssignment, isParameter, isParameterReference, isParserRule, isTerminalRule } from './generated/ast';
+import { isDataTypeRule } from './grammar-util';
+
+export class LangiumGrammarSemanticTokenProvider extends AbstractSemanticTokenProvider {
+
+    protected highlightElement(node: AstNode): boolean {
+        if (isAssignment(node)) {
+            this.highlightFeature(node, 'feature', SemanticTokenTypes.property);
```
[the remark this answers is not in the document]
msujew: Since the acceptor isn't actually used inside of the `highlightElement` method, I have another idea: Make `SemanticTokenAcceptor` into an interface:

```
interface SemanticTokenAcceptor {
  highlightToken(line: number, char: number, length: number, type: string, modifiers?: string[]): void
  highlightFeature(node: N, feature: Properties<N>, type: string, ...modifiers: string[]): void
  highlightKeyword(node: AstNode, keyword: string, type: string, ...modifiers: string[]): void
  highlightNode(node: CstNode, type: string, ...modifiers: string[]): void
}
```

That way, we don't have to rely so much on the protected methods of the `AbstractSemanticTokenProvider` and can make better use of the acceptor. Alternatively, the acceptor stays as a single function, but with a lot of overloading and without the `highlightKeyword` method. What do you think?

packages/langium/src/grammar/grammar-util.ts
```diff
@@ -164,6 +164,30 @@ export function findNodeForFeature(node: CstNode | undefined, feature: string |
     return nodes[index];
 }
 
+export function findKeywordNode(node: CstNode | undefined, keyword: string): CstNode | undefined {
+    return findKeywordNodeInternal(node, keyword, node?.element);
+}
+
+export function findKeywordNodeInternal(node: CstNode | undefined, keyword: string, element: AstNode | undefined): CstNode | undefined {
```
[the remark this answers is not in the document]
msujew: Right, we should use that 👍

spoenemann (approved):
:+1:

packages/langium/src/grammar/langium-grammar-semantic-token-provider.ts
spoenemann: Now it works here as well 🤷‍♂️

spoenemann:
Not related to this change, but could you fix this as well (should be easy)? When I write `Foo returns string: '';` one of the validation rules throws an error.

packages/langium/src/lsp/language-server.ts
```diff
@@ -39,7 +40,10 @@ export function startLanguageServer(services: LangiumSharedServices): void {
                 hoverProvider: {},
                 renameProvider: {
                     prepareProvider: true
-                }
+                },
+                semanticTokensProvider: languages.some(e => e.lsp.SemanticTokenProvider !== undefined)
+                    ? DefaultSemanticTokenOptions
```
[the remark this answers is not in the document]
msujew: A good idea, but I don't really see a use case for that right now. This would be a bit annoying to override, since it would either belong in the shared services (which seems kind of weird), or we simply take the first options that are appear in all services, which also seems wrong.

I would wait until someone requests this feature.

msujew:
> Not related to this change, but could you fix this as well (should be easy)? When I write Foo returns string: ''; one of the validation rules throws an error.

This was in error in the old `isDataType` rule function and was accidentally fixed by #341 :)

PR #346, opened 2022-01-08
Support rule parameter and arguments

msujew:
Accidentally fixes https://github.com/langium/langium/issues/313 (My tests wouldn't run without fixing it)

Makes use of Chevrotain's semantic predicate feature to build `GATES` from the guard conditions of alternatives. Also adds validations to check for the correct amount of arguments and unused rule parameters.

### DIFF
--- a/packages/langium/src/grammar/generated/ast.ts
+++ b/packages/langium/src/grammar/generated/ast.ts
@@ -9,7 +9,7 @@ import { AstNode, AstReflection, Reference } from '../../syntax-tree';
 import { isAstNode } from '../../utils/ast-util';
 
 export interface AbstractElement extends AstNode {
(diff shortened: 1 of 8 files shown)


spoenemann (commented):
Cool, so we can support rule parameters directly in the parser builder! IIRC Xtext "flattens" the grammar when it generates the ANTLR 3 grammar, so it duplicates the rules for each instance of a parameterized rule.

I think this would be a good time to add a fourth example showing both the multi-language and parameterized rules features. It would be good to see these in a live example. The example languages could be two aspects of a system with a commonly used part of the syntax.

Rough idea: one language for a static view and another language for a dynamic view of a system, both sharing a common expression language.

dhuebner:
Nice! I'm really looking forward to trying it out

spoenemann (approved):
As mentioned above, I'd like to experiment with parameterized rules in a live example. But that can be done separately, so I extracted it to a new issue.

~The sorting of container types would be nice to add~ (extracted to a new issue), otherwise I'm ok with the code changes.

PR #347, opened 2022-01-10
Validate duplicate imported grammar rules

### LINKED ISSUE #302: Validate duplicate rule name globally
When validating duplicate grammar rule names, we should consider imported names (#300).

msujew:
Closes #302

#### How to test

Add these lines to the `arithmetics` example grammar:

```
import '../../../domainmodel/src/language-server/domain-model';
import '../../../statemachine/src/language-server/statemachine';
```

Any terminals/rule names which already exist inside of the `domain-model` grammar will be highlighted with an error. This also works for transitive dependencies. Furthermore, both imports will be highlighted, since they export the same rules.

### DIFF
--- a/packages/langium/src/grammar/langium-grammar-validator.ts
+++ b/packages/langium/src/grammar/langium-grammar-validator.ts
@@ -41,6 +41,7 @@ export class LangiumGrammarValidationRegistry extends ValidationRegistry {
                 validator.checkEntryGrammarRule,
                 validator.checkUniqueRuleName,
                 validator.checkUniqueImportedRules,
+                validator.checkDuplicateImportedGrammar,
                 validator.checkGrammarHiddenTokens,
                 validator.checkGrammarForUnusedRules,
                 validator.checkGrammarImports
@@ -107,7 +108,7 @@ export class LangiumGrammarValidator {
     checkUniqueRuleName(grammar: ast.Grammar, accept: ValidationAcceptor): void {
         const ruleMap = new MultiMap<string, ast.AbstractRule>();
         for (const rule of grammar.rules) {
-            ruleMap.add(rule.name.toLowerCase(), rule);
+            ruleMap.add(rule.name, rule);
         }
         for (const name of ruleMap.keys()) {
             const rules = ruleMap.get(name);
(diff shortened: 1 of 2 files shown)


packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -112,6 +117,69 @@ export class LangiumGrammarValidator {
...
+        }
+    }
+
+    private getDuplicateExportedRules(outer: ast.Grammar[], inner: ast.Grammar[]): Set<string> {
+        const exclusiveOuter = outer.filter(g => !inner.includes(g));
+        const outerRules = exclusiveOuter.flatMap(e => e.rules);
+        const innerRules = inner.flatMap(e => e.rules);
+        const duplicates = new Set<string>();
+        for (const outerRule of outerRules) {
+            const outerName = outerRule.name.toLowerCase();
```
spoenemann, line 174: Why a case-insensitive comparison here?

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -112,6 +117,69 @@ export class LangiumGrammarValidator {
...
+            for (const innerRule of innerRules) {
+                const innerName = innerRule.name.toLowerCase();
+                if (outerName === innerName) {
+                    duplicates.add(innerRule.name);
+                }
+            }
+        }
+        return duplicates;
```
spoenemann, line 182: This check can potentially be computation-heavy for large projects. It's ok for now, but let's keep the performance aspect in mind.

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -112,6 +117,69 @@ export class LangiumGrammarValidator {
...
+        const imports = new Map<ast.GrammarImport, ast.Grammar[]>();
+        for (const imp of grammar.imports) {
+            const importedGrammars = resolveTransitiveImports(this.documents, imp);
+            imports.set(imp, importedGrammars);
+        }
+        const allDuplicates = new MultiMap<ast.GrammarImport, string>();
+        for (const outerImport of grammar.imports) {
+            const outerGrammars = imports.get(outerImport)!;
+            for (const innerImport of grammar.imports) {
+                if (outerImport === innerImport) {
```
spoenemann, line 150: This does not cover the case where a grammar is imported twice. But that could be done in a separate validation.

packages/langium/src/grammar/langium-grammar-validator.ts
msujew: There was a case-insensitive comparison in another place. But both don't actually make a lot of sense, so I removed them both 👍

spoenemann (approved):
Thanks!

PR #354, opened 2022-01-13
Added $containerProperty and $containerIndex

### LINKED ISSUE #329: Node path information in the AST
In https://github.com/langium/langium/pull/324 I added a `$path` property to AST nodes, mainly for enabling programmatic construction of ASTs. An example where this is necessary is when you provide a built-in library in memory. The AstNodeLocator currently depends on the node model to compute paths, which does not work for programmatic ASTs.

An alternative would be to add `$containerProperty` and maybe also `$containerIndex` to AST nodes instead of `$path`. This might also be useful for user code when you want to inspect where a node is contained. And in addition this approach would make  the `AstNodeContent` type redundant, so we could simplify the `streamContents` and `streamAllContents` utilities.

What do you think?

spoenemann:
Closes #329.

 * Added `$containerProperty` and `$containerIndex` to AstNode, removed `$path`.
 * Changed `DefaultAstNodeLocator.getAstNodePath` to use the new properties to create paths, making that service independent of the CST (important for programmatically created ASTs).
 * Extracted function `linkContentToContainer` from LangiumParser so it can be reused when creating ASTs programmatically.
 * `streamContents` and `streamAllContents` provide streams of AstNode instead of the previous AstNodeContent, which is no longer necessary because containment information is now included in the AstNode itself.
 * Moved `findLeafNodeAtOffset` from `ast-util` to `cst-util`.


msujew (approved):
Looks great! Please rebase your changes :)

PR #369, opened 2022-01-17
Re-export external APIs

### LINKED ISSUE #361: Re-export external APIs
Langium's generated code uses external APIs like `URI` of module `vscode-uri`, e.g. in `src/cli/cli-util.ts`, while no explicit dependency on `vscode-uri` is added to client projects' `package.json` files.

In the development scenario outlined in https://github.com/langium/langium/blame/main/CONTRIBUTING.md with the dependency to langium stated like `"langium": "file:../../../git/langium/packages/langium"` langium's dependencies are located in `"file:../../../git/langium/node_modules"`. `tsc` apparently cannot identify those dependencies correctly and flags the import
```
import { URI } from 'vscode-uri';
```
in `src/cli/cli-util.ts` as error. This can be avoided by re-exporting the required symbols and importing them in generated sources via langium.

msujew:
Closes #361. 

Additionally exports the `langium/lib/test` package in such a way that it can be consumed using `langium/test`.

### DIFF
--- a/examples/arithmetics/src/cli/cli-util.ts
+++ b/examples/arithmetics/src/cli/cli-util.ts
@@ -5,9 +5,9 @@
  ******************************************************************************/
 
 import fs from 'fs';
+import path from 'path';
 import colors from 'colors';
 import { AstNode, LangiumDocument, LangiumServices } from 'langium';
-import path from 'path';
 import { URI } from 'vscode-uri';
 
 export async function extractDocument<T extends AstNode>(fileName: string, extensions: string[], services: LangiumServices): Promise<LangiumDocument<T>> {
--- a/examples/domainmodel/src/cli/cli-util.ts
+++ b/examples/domainmodel/src/cli/cli-util.ts
@@ -5,9 +5,9 @@
  ******************************************************************************/
 
 import fs from 'fs';
+import path from 'path';
 import colors from 'colors';
 import { AstNode, LangiumDocument, LangiumServices } from 'langium';
-import path from 'path';
 import { URI } from 'vscode-uri';
 import { WorkspaceFolder } from 'vscode-languageserver';
 
--- a/examples/domainmodel/src/cli/generator.ts
+++ b/examples/domainmodel/src/cli/generator.ts
@@ -5,14 +5,14 @@
  ******************************************************************************/
 
 import fs from 'fs';
+import path from 'path';
+import colors from 'colors';
 import _ from 'lodash';
 import { CompositeGeneratorNode, IndentNode, NL, processGeneratorNode } from 'langium';
 import { AbstractElement, Domainmodel, Entity, Feature, isEntity, isPackageDeclaration, Type } from '../language-server/generated/ast';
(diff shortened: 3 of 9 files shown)


spoenemann (commented):
I haven't seen this kind of reexporting in other libraries, but I'd be ok with it. I'm not fully aware of the consequences, though. @danieldietrich WDYT?

_Important:_ Please add new paths to be included in the packaged library to the `"files"` entry of package.json.

msujew:
@spoenemann Theia does it as well: https://github.com/eclipse-theia/theia/tree/master/packages/core/shared

sailingKieler:
Tested it on Gitpod with the dev setup scenario described in https://github.com/langium/langium/blob/5fc873db5aa5ae7897f6f3aa73ffa697ee14771b/CONTRIBUTING.md#L51-L60

Works well 🙂

sailingKieler (approved):

danieldietrich:
Hi @msujew @spoenemann @sailingKieler for apps that have modules that depend on each other I have used peerDependencies in the past. It is a way to ensure that all app modules use the same version of a third party lib like React. However, peerDependencies are a smell because the actual dependency present at runtime may have breaking changes (e.g. semantically, not necessarily on the source/API level).

I do not have much experience in bundling and exposing external dependencies as part of the own library. Therefore I only have one question:

**Is there the risk that the user of langium uses another version of vscode than that we re-export?**

I don't see problems regarding _source_ backward compatibility, because we also re-export the types. (_Binary_ backw'compat should be the same as source backw'compat in JS because it is an interpreted language).

However, what about _semantic_ backw'compat in the way that the re-exported vs-code dependency behaves different that the vs-code runtime of the user who uses langium?

This is just a theoretical question. I think it shouldn't be a problem in practice if Microsoft does not break anything.

This PR looks good!

spoenemann:
I'm skeptical about this change. Can't we solve the original issue simply by adding explicit dependencies to the generator-langium template?

msujew:
@spoenemann @danieldietrich I'm inclined to agree, we shouldn't export our dependencies.

I changed the PR to include the dependencies in the `generator-langium` project.

spoenemann (approved):
Looks good. We can still get back to the idea of reexporting later.

One question: How likely is it that users will need to import something from `chevrotain` in their code? Aren't we hiding most of its API in Langium?

msujew:
@spoenemann I believe the `TokenBuilder` exposes some API from Chevrotain, so I think it's a good idea to have that in the `generator-langium` build.

danieldietrich:
Hi @msujew, I just recognized that `test.js` and `test.d.ts` are checked in at _top level_ of the `langium` package and also bundled. Is that intended?

msujew:
@danieldietrich Yep, definitely intended. It allows us to reference the `langium/test` part of our framework when importing code instead of always using `langium/lib/test`.

danieldietrich:
Ok, it looked a bit unusual but it makes sense. 👍

PR #372, opened 2022-01-18
Show grammar path and line number for parser validation errors

### LINKED ISSUE #201: Trace Chevrotain errors to grammar
Once we have done #196, it would be super nice to add tracing information to errors reported by Chevrotain.

Idea: first print the Chevrotain message, then print file:line:column positions of the relevant grammar parts.

spoenemann:
Closes #201. When Chevrotain reports grammar validation errors, they are prefixed with the file path and line number of the corresponding parser rule.

How to verify:
 * Change one of the examples so there is an ambiguity or left-recursion, then run the CLI to see the validation errors.

### DIFF
--- a/packages/langium-cli/src/parser-validation.ts
+++ b/packages/langium-cli/src/parser-validation.ts
@@ -5,15 +5,17 @@
  ******************************************************************************/
 
 import {
-    createDefaultModule, createDefaultSharedModule, createLangiumParser, Grammar, inject,
-    IParserConfig, LangiumGeneratedServices, LangiumGeneratedSharedServices, LangiumServices,
-    LangiumSharedServices, Module
+    createDefaultModule, createDefaultSharedModule, getDocument, Grammar, inject, IParserConfig,
+    isGrammar, isParserRule, LangiumDocuments, LangiumGeneratedServices, LangiumGeneratedSharedServices,
+    LangiumParser, LangiumServices, LangiumSharedServices, Module, ParserRule, prepareLangiumParser
 } from 'langium';
-import { LangiumConfig } from './package';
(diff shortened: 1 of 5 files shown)


packages/langium-cli/src/parser-validation.ts
```diff
@@ -23,20 +25,61 @@ export function validateParser(grammar: Grammar, config: LangiumConfig): Error |
...
+        parser.finalize();
         return undefined;
     } catch (err) {
+        if (parser && parser.definitionErrors.length > 0) {
+            // Construct a message with tracing information
+            let message = 'Parser definition errors detected:';
+            for (const defError of parser.definitionErrors) {
+                message += '\n-------------------------------\n';
+                if (defError.ruleName) {
+                    const rule = findRule(defError.ruleName, grammar, documents);
```
msujew, line 50: Due to #337, Chevrotain rule names have a `:RULE` suffix to prevent clashing with terminal names. The changes don't reflect that. I just noticed that this looks kind of silly in the Chevrotain error messages, so perhaps we should think about changing that.

spoenemann: Yes, it would be great to see the plain rule names in the error messages.

CC @Pluralia

Pluralia: Thank you, I've reopened [the issue](https://github.com/langium/langium/issues/294) and make a new PR with another solution.

spoenemann: Ho to continue? Wait for the other solution or merge this (it shouldn't hurt)?

msujew (approved):
As discussed, we will continue the mentioned issue in another way.

The changes look good to me!

PR #373, opened 2022-01-18
Add descriptions to Yeoman questions

### LINKED ISSUE #258: Add descriptions to requested values
We should print some description of the values requested by our Yeoman generator, preferably through Yeoman API. Users may not know what "extension name" and "language name" mean and what they are used for.

spoenemann:
Closes #258. The descriptions are printed above each question and colored in gray.

I removed the `colors` dependency because it's not used. Yeoman uses `chalk` instead.

### DIFF
--- a/packages/generator-langium/src/index.ts
+++ b/packages/generator-langium/src/index.ts
@@ -6,8 +6,8 @@
 
 import Generator from 'yeoman-generator';
 import _ from 'lodash';
+import chalk from 'chalk';
 import path from 'path';
-import 'colors';
 
 const TEMPLATE_DIR = '../langium-template';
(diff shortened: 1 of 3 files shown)


packages/generator-langium/src/index.ts
```diff
@@ -6,8 +6,8 @@
 
 import Generator from 'yeoman-generator';
 import _ from 'lodash';
+import chalk from 'chalk';
```
msujew, line 9: I assume chalk is imported from `yeoman-generator`? We should add an explicit dependency on that in the `generator-langium` package.

msujew (approved):
Looks good to me!

PR #374, opened 2022-01-19
Deduplicate and sort types

### LINKED ISSUE #363: Sort generated container types
Extracted from https://github.com/langium/langium/pull/346#discussion_r783681429

We should sort the list of `$container` alternatives in order to avoid changes in the generated code when some change in the CLI leads to a different order.

See also #150.

### LINKED ISSUE #367: Type duplication in a AST property type
Type of `elements` infers from `X | Z` for each alternative separately. `X` has type `Y` and `Z` has type `Y`, so, the `elements` type is `Y | Y`, and it looks such way in the generated AST. The `elements` type has to be reduced by removing duplicated types.
```
entry Model:
    elements=(X | Z);

X returns Y:
    'x' name=ID;

Z returns Y:
    'z' name=ID;
```

Langium version: 0.2.0
Package name: `langium-cli`

## Steps To Reproduce

1. Generate an AST with the grammar above
2. Check out the `Model` interface in the AST

## The current behavior
Type of `elements` is `Y | Y`.
```
export interface Model extends AstNode {
    elements: Y | Y
}
```

## The expected behavior
Type of `elements` is `Y`.
```
export interface Model extends AstNode {
    elements: Y
}
```

spoenemann:
Fixes #363, fixes #367.

We now deduplicate type collections using `Set` and sort the resulting arrays before generating.

### DIFF
--- a/examples/arithmetics/src/language-server/generated/ast.ts
+++ b/examples/arithmetics/src/language-server/generated/ast.ts
@@ -8,7 +8,7 @@
 import { AstNode, AstReflection, Reference, isAstNode } from 'langium';
 
 export interface AbstractDefinition extends AstNode {
-    readonly $container: Module | Definition;
+    readonly $container: Definition | Module;
     name: string
 }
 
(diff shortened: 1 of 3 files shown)


msujew (approved):
LGTM!

PR #377, opened 2022-01-19
Extracted workspace initialization into new WorkspaceManager service

### LINKED ISSUE #280: Refactor `IndexManager.skip` into service
Instead of overriding the `skip` method, we should write an `WorkspaceIgnoreService` that checks whether a folder/file should be included in the workspace. One possible default implementation checks the `.gitignore` file and excludes folders/files based on that.

spoenemann:
Closes #280.

The new `WorkspaceManager` service initializes the index from the list of workspace folders reported by the client. This is the main service that needs to be overridden if you want to specialize where the source files are found. In a web context, the service would be completely replaced.

I also simplified the initialize phase of `startLanguageServer`.


dhuebner (approved):
Looks good

sailingKieler:
LGTM, too, I only https://github.com/langium/langium/commit/03995e9ddda9d321aa4278452fb2941831ea5756#r65681876 is not completely clear to me.

Is there any impediment preventing the merge?

PR #380, opened 2022-01-20
Added attach launch config to generator template

spoenemann:
As mentioned in https://github.com/langium/langium/discussions/376#discussioncomment-1997781, this adds an `attach` launch config to generated projects to users can debug their language server.

I also reordered the entries in our main launch config so the most used ones are shown first in the list.


msujew (approved):
LGTM!

Issue #307, opened 2021-11-23
References in other files dont detect name changes


cdietrich:
Given

```a.dmodel
entity A extends B {}
```
and
```b.dmodel
entity B {}
```
when i open b.dmodel and edit B to B2, then the extends statement in A wont be flagged as errorneous unless i open/touch a.dmodel

msujew:
@cdietrich Weird, I can't reproduce this issue with the domain-model example package. Do you have a minimal reproducible grammar at hand?

cdietrich:
@msujew  i am using the domain model example as is with linked ../../packages/langium

https://user-images.githubusercontent.com/238736/143225441-6b78da26-6820-475a-a44d-aa1cc23371f4.mov

```
langium[main] > git log -n 2
commit 49bf240c1996a32b3401650cf70f1335462fdaf9 (HEAD -> main, origin/main, origin/HEAD)
Author: Pluralia <15619772+Pluralia@users.noreply.github.com>
Date:   Fri Nov 19 10:29:11 2021 +0100
```

cdietrich:
```
git diff examples/domainmodel/package.json
diff --git a/examples/domainmodel/package.json b/examples/domainmodel/package.json
index 4b94e71..8cd823c 100644
--- a/examples/domainmodel/package.json
+++ b/examples/domainmodel/package.json
@@ -4,6 +4,7 @@
     "description": "description",
     "version": "0.0.1",
     "private": true,
+    "publisher": "langium",
     "engines": {
         "vscode": "^1.56.0"
     },
@@ -11,17 +12,26 @@
         "Programming Languages"
     ],
     "contributes": {
-        "languages": [{
-            "id": "domain-model",
-            "aliases": ["DomainModel", "domain-model"],
-            "extensions": [".dmodel"],
-            "configuration": "./language-configuration.json"
-        }],
-        "grammars": [{
-            "language": "domain-model",
-            "scopeName": "source.domain-model",
-            "path": "./syntaxes/domain-model.tmLanguage.json"
-        }]
+        "languages": [
+            {
+                "id": "domain-model",
+                "aliases": [
+                    "DomainModel",
+                    "domain-model"
+                ],
+                "extensions": [
+                    ".dmodel"
+                ],
+                "configuration": "./language-configuration.json"
+            }
+        ],
+        "grammars": [
+            {
+                "language": "domain-model",
+                "scopeName": "source.domain-model",
+                "path": "./syntaxes/domain-model.tmLanguage.json"
[...]

PR #386, opened 2022-01-21
References in other files don't detect name changes #307

dhuebner:
Set unlinked document to Parsed state to gather new linked nodes in index.

Alternatively we could update the reference information in index after linking is done. Doing so we need to split the `IndexManager.processDocuments` implementation and call reference description update separately. 
In general I still wonder if the current order of DocumentStates is the right one, or better to say, if we are missing some in-between steps like IndexReferences.

### DIFF
--- a/packages/langium/src/references/linker.ts
+++ b/packages/langium/src/references/linker.ts
@@ -71,7 +71,8 @@ export function getReferenceProperty(referenceId: string): string {
 }
 
 interface DefaultReference extends Reference {
-    _ref?: AstNode | AstNodeDescription | LinkingError;
+    _ref?: AstNode | LinkingError;
+    _nodeDescription?: AstNodeDescription;
 }
 
 export class DefaultLinker implements Linker {
@@ -107,13 +108,15 @@ export class DefaultLinker implements Linker {
             try {
                 const refId = getReferenceId(info.container.$type, info.property);
                 const description = this.getCandidate(info.container, refId, ref);
-                if (!isLinkingError(description) && this.langiumDocuments().hasDocument(description.documentUri)) {
-                    // The target document is already loaded
-                    const linkedNode = this.loadAstNode(description);
-                    ref._ref = linkedNode ?? this.createLinkingError(info, refId, description);
-                } else {
-                    // The target document is not loaded yet, or the target was not found in the scope
+                if (isLinkingError(description)) {
                     ref._ref = description;
+                } else {
+                    ref._nodeDescription = description;
+                    if (this.langiumDocuments().hasDocument(description.documentUri)) {
+                        // The target document is already loaded
+                        const linkedNode = this.loadAstNode(description);
+                        ref._ref = linkedNode ?? this.createLinkingError(info, refId, description);
+                    }
                 }
             } catch (err) {
                 ref._ref = {
@@ -129,6 +132,7 @@ export class DefaultLinker implements Linker {
     unlink(document: LangiumDocument): void {
         for (const ref of document.references) {
             delete (ref as DefaultReference)._ref;
+            delete (ref as DefaultReference)._nodeDescription;
         }
         document.references = [];
     }
@@ -149,20 +153,24 @@ export class DefaultLinker implements Linker {
 
             get ref() {
                 if (isAstNode(this._ref)) {
-                    // Most frequent case: the target is aready resolved.
+                    // Most frequent case: the target is already resolved.
                     return this._ref;
+                } else if (isAstNodeDescription(this._nodeDescription)) {
+                    // A candidate has been found before, but it is not loaded yet.
+                    const linkedNode = linker.loadAstNode(this._nodeDescription);
+                    this._ref = linkedNode ??
+                        linker.createLinkingError({ container: node, property: getReferenceProperty(refId), reference }, refId, this._nodeDescription);
                 } else if (this._ref === undefined) {
                     // The reference has not been linked yet, so do that now.
-                    this._ref = linker.getLinkedNode(node, refId, reference);
-                } else if (isAstNodeDescription(this._ref)) {
-                    // A candidate has been found before, but it is not loaded yet.
-                    const linkedNode = linker.loadAstNode(this._ref);
-                    this._ref = linkedNode ??
-                        linker.createLinkingError({ container: node, property: getReferenceProperty(refId), reference }, refId, this._ref);
+                    const refData = linker.getLinkedNode(node, refId, reference);
+                    this._ref = refData.node ?? refData.error;
+                    this._nodeDescription = refData.descr;
                 }
                 return isAstNode(this._ref) ? this._ref : undefined;
             },
-
+            get $nodeDescription() {
+                return this._nodeDescription;
+            },
             get error() {
                 return isLinkingError(this._ref) ? this._ref : undefined;
             }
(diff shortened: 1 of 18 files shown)


msujew:
@dhuebner I don't think we need to split the `IndexManager#processDocuments` implementation, but rather add a new `DocumentState#IndexedContent` and `Document#IndexedReferences` and check whether the document has reached the state already before indexing the content/references appropriately. Instead of setting the document state to `Parsed` we could instead set it to `IndexedContent`. WDYT?

dhuebner:
I've tried to reflect the state changes that are currently happen on document update. I'm not sure how helpful is it for others, but it helped me to understand what and when is called during document building.
Below the first entry is "Init Workspace" with two files in WS. The second "orange" entry is an update for a document, due to opened dmodel file.
Bildschirmfoto 2022-01-28 um 14 15 09

spoenemann (commented):
I agree with @msujew that we should try splitting the indexing in two phases. Currently the ReferenceDescriptionProvider looks up candidates for all cross-references, and the Linker does that again in a later phase. We could avoid this duplicate work by having the ReferenceDescriptionProvider reuse the information that is collected by the Linker. One idea to do that: the `link` method could return an array of all AstNodeDescription candidates that were collected.

Can we review and merge #377 before making bigger changes to the IndexManager? That would help to avoid conflicts.

dhuebner:
@spoenemann 
> I agree with @msujew that we should try splitting the indexing in two phases. 

But keep the `processDocuments` as one function? Doing so this function will never be called to do both, node and reference indexing. This sounds a bit strange to me, maybe I didn't get the idea behind.

> We could avoid this duplicate work by having the ReferenceDescriptionProvider reuse the information that is collected by the Linker.

I think the best would be to write down the order of state changes and involved players to see if this make sense. Currently we do this in DocumentBuilder:
1. Indexing (Nodes, Refs)
2. Preprocessing
3. Linking
4. Validation

spoenemann:
> But keep the `processDocuments` as one function?

No, that would have to be split in two.

Here's my proposal for the new order of state changes:

 1. LangiumDocuments.getOrCreateDocument ➡️ LangiumDocumentFactory.fromTextDocument ➡️ `Parsed`
 2. IndexManager.updateContent ➡️ `IndexedContent`
 3. ScopeComputation.computeScope ➡️ `Processed` (in languages with a type system, this step might also include type precomputation)
 4. Linker.link ➡️  `Linked`
 5. IndexManager.updateReferences ➡️ `IndexedReferences`
 6. DocumentValidator.validateDocument ➡️ `Validated`

The Linker already stores all outgoing references in the `references` array of the document. The hidden `_ref` property of each Reference points to an AST node, a candidate description from the index, or an error. We just need to expose that information to the IndexManager, and it will be able to update the references index based on it.

WDYT?

dhuebner:
@spoenemann 
I have two additional questions.

We have a call to `initializeWorkspace` on start up, this indexer call is different as we have in DocumentBuilder, because we du not process steps 3. and 4., how to deal with references (local/non-local) and linking there? 

Second, I saw a notifyBuildPhase call just after indexing in the DocumentBuilder `this.notifyBuildPhase(toBeIndexed, DocumentState.Indexed, cancelToken)` do we need to have an additional one for `IndexedReferences`? I'm asking because I don't know the motivation behind this notifications.

spoenemann:
> We have a call to `initializeWorkspace` on start up, this indexer call is different as we have in DocumentBuilder, because we do not process steps 3. and 4., how to deal with references (local/non-local) and linking there? 

Proposal: In `updateReferences` we check the document state. If it's already linked, use the information from the Linker. Otherwise collect references with a tree stream like it's done now. This logic can be extracted to a protected method.

> Second, I saw a notifyBuildPhase call just after indexing in the DocumentBuilder `this.notifyBuildPhase(toBeIndexed, DocumentState.Indexed, cancelToken)` do we need to have an additional one for `IndexedReferences`? I'm asking because I don't know the motivation behind this notifications.

That's for external services (outside of langium core) that need to hook into the build process. I added that for the Sprotty integration in #308, which updates diagrams when the document contents change.

dhuebner:
I didn't change the Implementation of `DefaultDocumentBuilder.shouldValidate` cause I still not sure how to handle it in the initialize call. So please make your suggestions

packages/langium/src/references/linker.ts
```diff
@@ -107,13 +108,16 @@ export class DefaultLinker implements Linker {
...
+                if (isLinkingError(description)) {
                     ref._ref = description;
+                } else {
+                    ref._nodeDescription = description;
+                    if (this.langiumDocuments().hasDocument(description.documentUri)) {
+                        // The target document is already loaded
+                        const linkedNode = this.loadAstNode(description);
+                        ref._ref = linkedNode ?? this.createLinkingError(info, refId, description);
+                    }
+                    // The target document is not loaded yet
```
spoenemann, line 120: This comment doesn't make much sense here (it would if we'd have an `else` block).

packages/langium/src/references/linker.ts
```diff
@@ -149,43 +154,53 @@ export class DefaultLinker implements Linker {
...
+                    // Most frequent case: the target is already resolved.
                     return this._ref;
                 } else if (this._ref === undefined) {
                     // The reference has not been linked yet, so do that now.
-                    this._ref = linker.getLinkedNode(node, refId, reference);
-                } else if (isAstNodeDescription(this._ref)) {
+                    const refData = linker.getLinkedNode(node, refId, reference);
+                    this._ref = refData.node ?? refData.error;
+                    this._nodeDescription = refData.descr;
+                } else if (isAstNodeDescription(this._nodeDescription)) {
```
spoenemann, line 164: This case needs to be checked before the previous one: when `_nodeDescription` is defined, `_ref` is still undefined. Please switch the two `if` blocks.

packages/langium/src/references/linker.ts
```diff
@@ -149,43 +154,53 @@ export class DefaultLinker implements Linker {
...
+                return { error: description };
             } else {
                 const linkedNode = this.loadAstNode(description);
-                return linkedNode ??
-                    this.createLinkingError({ container, property: getReferenceProperty(refId), reference }, refId, description);
+                return {
+                    node: linkedNode,
+                    descr: description,
+                    error:
+                        this.createLinkingError({ container, property: getReferenceProperty(refId), reference }, refId, description)
```
spoenemann, line 193: An error must be created only when `linkedNode` is undefined / falsy.

packages/langium/src/workspace/ast-descriptions.ts
```diff
@@ -80,21 +79,19 @@ export class DefaultAstNodeDescriptionProvider implements AstNodeDescriptionProv
...
     async createDescriptions(document: LangiumDocument, cancelToken = CancellationToken.None): Promise<ReferenceDescription[]> {
         const descr: ReferenceDescription[] = [];
         const rootNode = document.parseResult.value;
         const refConverter = (refInfo: ReferenceInfo): ReferenceDescription | undefined => {
-            const refAstNodeDescr = this.linker.getCandidate(refInfo.container, getReferenceId(refInfo.container.$type, refInfo.property), refInfo.reference);
-            // Do not handle unresolved refs
-            if (isLinkingError(refAstNodeDescr)) {
+            // Do not handle unresolved or not handled refs
+            const refAstNodeDescr = refInfo.reference.$nodeDescription;
+            if (!refAstNodeDescr || isLinkingError(refAstNodeDescr)) {
```
spoenemann, line 94: The `$nodeDescription` property is never set to a LinkingError.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -133,7 +126,7 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
         affected.forEach(e => {
             const linker = this.serviceRegistry.getServices(e.uri).references.Linker;
             linker.unlink(e);
-            e.state = DocumentState.Indexed;
+            e.state = DocumentState.Processed; // need to re-index potentially linked references
```
spoenemann, line 129: Be careful not to "upgrade" documents here:
```suggestion
            e.state = Math.min(e.state, DocumentState.Processed); // need to re-index potentially linked references
```

packages/langium/src/workspace/document-builder.ts
```diff
@@ -157,7 +150,9 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
         await this.runCancelable(documents, DocumentState.Linked, cancelToken, doc =>
             this.serviceRegistry.getServices(doc.uri).references.Linker.link(doc, cancelToken)
         );
-        // 4. Validation
+        // 4. Index references
+        await this.indexManager.updateReferences(documents.filter(e => e.state < DocumentState.IndexedReferences), cancelToken);
```
spoenemann, line 154: Please add a `notifyBuildPhase` call for completeness.

spoenemann (commented):
Thanks, this looks really good.

I'm fine with leaving the validation topic to later, we can solve it in a follow-up PR.

dhuebner:
Thank for the re-view. I will work on this tomorrow

dhuebner:
@spoenemann 
I've applied all the suggested changes

dhuebner:
I will squash  the commits after re-view is done and changes are accepted

spoenemann (approved):
I verified that this change fixes the reported bug :+1:

PR #392, opened 2022-01-27
Fix name clashing of terminal and non-terminal rules

Pluralia:
Closes [#294](https://github.com/langium/langium/issues/294) as described [here](https://github.com/langium/langium/issues/294#issuecomment-1017333601).

### DIFF
--- a/packages/langium/src/parser/langium-parser-builder.ts
+++ b/packages/langium/src/parser/langium-parser-builder.ts
@@ -52,10 +52,7 @@ export function prepareLangiumParser(services: LangiumServices): LangiumParser {
     const grammar = services.Grammar;
     const tokens = new Map<string, TokenType>();
     const buildTokens = services.parser.TokenBuilder.buildTokens(grammar, { caseInsensitive: services.LanguageMetaData.caseInsensitive });
-    buildTokens.forEach(e => {
-        e.name = withKeywordSuffix(e.name);
-        tokens.set(e.name, e);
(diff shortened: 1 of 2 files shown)


msujew (commented):
I don't think it is necessary to add the keyword suffix to terminal rules (I would argue it would be quite misleading). Instead I would change the call for the keyword tokens in particular to use `withKeywordSuffix`, and keep the rest as it is. So instead of calling `withKeywordSuffix` in line 79 of the parser builder, I would call it in 268.

packages/langium/src/parser/langium-parser-builder.ts
```diff
@@ -265,19 +261,23 @@ function buildCrossReference(ctx: RuleContext, crossRef: CrossReference, termina
...
         const keyword = getToken(ctx, terminal.value);
+        keyword.name = withKeywordSuffix(keyword.name);
         return () => ctx.parser.consume(idx, keyword, crossRef);
     }
     else {
         throw new Error('Could not build cross reference parser');
     }
 }
 
+const withKeywordSuffix = (name: string): string => name + (name.endsWith(':KW') ? '' : ':KW');
```
spoenemann, line 272: Minor suggestion:
```suggestion
const withKeywordSuffix = (name: string): string => name.endsWith(':KW') ? name : name + ':KW';
```

spoenemann (approved):

Issue #320, opened 2021-12-10
Allow to return a multi-mode lexer definition in the token builder


msujew:
Currently, the token builder only allows to return an array of tokens. However, this prevents us from using multi-mode lexer definitions.

PR #398, opened 2022-02-01
Add ability to return a multi-mode lexer

Pluralia:
Closes [#320](https://github.com/langium/langium/issues/320).
* Adds ability to return a multi-mode lexer by modification of `TokenBuilder` API (`buildTokens` returns `TokenVocabulary` instead of just `TokenType[]`)
* Refactors `DefaultTokenBuilder` to make it more modular

### DIFF
--- a/packages/langium/src/parser/token-builder.ts
+++ b/packages/langium/src/parser/token-builder.ts
@@ -17,43 +17,22 @@ export interface TokenBuilder {
 
 export class DefaultTokenBuilder implements TokenBuilder {
 
-    // We need suffixes for terminals and keywords which have the same name
-    protected readonly KEYWORD_SUFFIX = '_KEYWORD';
-    protected readonly TERMINAL_SUFFIX = '_TERMINAL';
-
     buildTokens(grammar: Grammar, options?: { caseInsensitive?: boolean }): TokenVocabulary {
-        const tokenMap = new Map<string, TokenType>();
-        const terminalsTokens: TokenType[] = [];
-        const terminals = Array.from(stream(grammar.rules).filter(isTerminalRule)).filter(e => !e.fragment);
-        for (const terminal of terminals) {
-            const token = this.buildTerminalToken(terminal);
-            terminalsTokens.push(token);
-            tokenMap.set(terminal.name + this.TERMINAL_SUFFIX, token);
-        }
+        const terminalTokens: TokenType[] = this.buildTerminalTokens(grammar);
+        const tokens: TokenType[] = this.buildKeywordTokens(grammar, terminalTokens, options);
 
-        const tokens: TokenType[] = [];
-        // We filter by parser rules, since keywords in terminal rules get transformed into regex and are not actual tokens
-        const parserRuleKeywords = grammar.rules.filter(isParserRule).flatMap(rule => streamAllContents(rule).filter(isKeyword).toArray());
-        const keywords = stream(parserRuleKeywords).distinct(e => e.value).toArray()
-            // Sort keywords by descending length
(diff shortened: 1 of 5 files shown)


spoenemann (commented):
Thanks! In general, please write a description of your changes when you submit a PR. This is important for reviewers to understand more quickly and also for long-term documentation of what we have changed.

In particular: With the modified API, how would you implement a multi-mode lexer? Does that require to write your own implementation of the TokenBuilder service interface?

packages/langium/src/parser/token-builder.ts
```diff
@@ -87,20 +79,19 @@ export class DefaultTokenBuilder implements TokenBuilder {
             keyword.value;
     }
 
-    protected findLongerAlt(keyword: Keyword, keywords: Keyword[], terminals: TerminalRule[], tokenMap: Map<string, TokenType>): TokenType[] {
+    protected findLongerAlt(keyword: Keyword, keywordTokens: TokenType[], terminalTokens: TokenType[]): TokenType[] {
```
msujew, line 82: I've though about this for a bit, and came to the conclusion that we don't need the `keywordTokens` parameter here: We already sort keywords by descending length, therefore, no keyword is a longer alternative of another keyword. Shorter keywords can never be matched before a longer keyword anyway.

Removing this would simplify the `buildKeywordTokens` method as well.

packages/langium/src/parser/token-builder.ts
```diff
@@ -4,58 +4,37 @@
...
-            }
-        }
-
+        });
         return tokens;
     }
 
+    protected buildTerminalTokens(grammar: Grammar): TokenType[] {
+        return Array.from(stream(grammar.rules).filter(isTerminalRule)).filter(e => !e.fragment)
+            .map(terminal => this.buildTerminalToken(terminal));
```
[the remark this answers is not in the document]
Pluralia: Yes, it's a nice idea! However, `buildTerminalTokens` and `buildKeywordTokens` can be used in custom implementations of `TokenBuilder`, and it's nice if they both return or streams, or arrays. An array is more natural for `buildKeywordTokens` (moreover, it's a base for the returning result that have to be an array), so, I'd propose to use an array for `buildTerminalTokens` too.

packages/langium/src/parser/token-builder.ts
Pluralia: Unfortunately, we can't remove it from `keywordTokens`.

> Shorter keywords can never be matched before a longer keyword anyway.
Yes, it's true, but we handle keywords in descending order by their length, so we handle longer keywords before shorter ones. As the result, when we handle shorter keywords, they already can have longer alternatives in `keywordTokens`.

For example, keywords `?=` and `?`. `?=` is a longer alternative of `?` that is found only thanks to using `keywordTokens`.

msujew: Yes, but Chevrotain will always try to lex `?=` before attempting `?`. That's why we don't need the `longerAlt` property on keyword tokens for other keywords. There is no scenario where Chevrotain attempts to match `?` even though the next two characters are `?=`.

packages/langium/src/parser/token-builder.ts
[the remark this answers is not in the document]
spoenemann: Ok but even then you can defer array creation to the end to avoid creating intermediate arrays:
```typescript
return stream(grammar.rules).filter(isTerminalRule).filter(e => !e.fragment)
    .map(terminal => this.buildTerminalToken(terminal))
    .toArray();
```

spoenemann (approved):
:+1:

PR #403, opened 2022-02-08
Refactored validation in document builder

spoenemann:
I refactored the validation process in `DocumentBuilder`:

 * Added `BuildOptions` type. Later we can extend this to support partial validation, for example when there are checks that shall be executed only when a document is saved to disk. With the new options, you have full control over which validations to run in a CLI, for example.
 * The `DocumentBuilder` no longer reports diagnostics to the client. This is now done with a build phase listener that is invoked at the end of every rebuild.
 * `IndexManager` methods accept a single document so they fit better in the general phase scheme of `DocumentBuilder`.
 * Renamed `SimpleScope` to `StreamScope`.
 * Added more code comments in the scoping / references area.

### DIFF
--- a/packages/langium/src/workspace/document-builder.ts
+++ b/packages/langium/src/workspace/document-builder.ts
@@ -4,7 +4,7 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { CancellationToken, Connection, Diagnostic } from 'vscode-languageserver';
+import { CancellationToken } from 'vscode-languageserver';
 import { URI } from 'vscode-uri';
 import { ServiceRegistry } from '../service-registry';
 import { LangiumSharedServices } from '../services';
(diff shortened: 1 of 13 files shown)


packages/langium/src/workspace/document-builder.ts
```diff
@@ -14,15 +14,20 @@ import { interruptAndCheck, MaybePromise } from '../utils/promise-util';
 import { IndexManager } from '../workspace/index-manager';
 import { DocumentState, LangiumDocument, LangiumDocuments } from './documents';
 
+export interface BuildOptions {
+    validationChecks?: 'None' | 'All'
```
msujew, line 18: Nitpick: I'm used to seeing lower-case configs, so `'none' | 'all'`, but that's a personal preference perhaps. Independent of how we deal with it in this case, we should be consistent about it.

msujew (approved):
The changes are looking really good to me. Great work on the API improvements!

PR #405, opened 2022-02-09
Implement a file system provider

msujew:
Extract `node.js` related functionality (mostly `fs`) into a `langium/node` directory. Allows for easier bundling of langium projects and different file providers.


packages/langium/src/node/node-file-system-provider.ts
```diff
@@ -0,0 +1,51 @@
...
+ ******************************************************************************/
+
+import fs from 'fs';
+import { URI, Utils } from 'vscode-uri';
+import { FileSystemFilter, FileSystemNode, FileSystemProvider } from '../workspace/file-system-provider';
+
+export class NodeFileSystemProvider implements FileSystemProvider {
+
+    readFile(uri: URI): Promise<string> {
+        return fs.promises.readFile(uri.fsPath, 'utf-8');
```
spoenemann, line 14: Proposal: make the encoding configurable by declaring a public property on this service class, initialized with `utf-8`.

packages/langium/src/node/node-file-system-provider.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    readFileSync(uri: URI): string {
+        return fs.readFileSync(uri.fsPath, 'utf-8');
+    }
+
+    async traverse(root: URI, filter?: FileSystemFilter): Promise<URI[]> {
```
spoenemann, line 21: Proposal: expose a callback function instead of returning an array in the API. That way the method is more flexible, for example you don't need to `flat()` the workspace folder entries in the DefaultWorkspaceManager.

packages/langium/src/workspace/file-system-provider.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export interface FileSystemNode {
+    readonly isFile: boolean;
+    readonly isDirectory: boolean;
+    readonly name: string;
+    readonly container: URI;
+}
+
+export interface FileSystemFilter {
+    (node: FileSystemNode): boolean;
```
spoenemann, line 17: I'd use `type` to declare function types instead of `interface`.

packages/langium/src/workspace/file-system-provider.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export interface FileSystemProvider {
+    /**
+     * Reads a document synchronously from a given URI.
+     * @returns The string content of the file with the specified URI.
```
spoenemann, line 23: You confused this documentation with `readFileSync`.

packages/langium/src/workspace/file-system-provider.ts
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Traverses a directory structure and returns all paths within.
+     * @param root The root path where the traversal starts.
+     * @param filter An optional filter that determines whether a directory or file should be included while traversing.
+     * @returns All URIs which match the specified filter in the directory structure.
+     */
+    traverse(root: URI, filter?: FileSystemFilter): Promise<URI[]>;
+}
+
+export class NullFileSystemProvider implements FileSystemProvider {
```
spoenemann, line 40: What is this good for? Langium will not work when this service implementation is active.

packages/langium/src/node/node-module-context.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright 2022 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { DefaultSharedModuleContext } from '../default-module';
+import { createConnection, ProposedFeatures } from 'vscode-languageserver/node';
+import { NodeFileSystemProvider } from './node-file-system-provider';
+
+export function createNodeModuleContext(context: DefaultSharedModuleContext = {}): DefaultSharedModuleContext {
```
spoenemann, line 11: This method is specifically for a language server context. Its name does not reflect this.

spoenemann (commented):
Thanks, the extraction of file system access is good. But I don't agree with keeping the framework agnostic to Node.js: running in Node.js is by far the most relevant context for Langium. I would prefer designing the API in a way that Node.js is the default, and any other context like browser is a special case that needs more configuration.

Proposal: In `default-module.ts`, create a NodeFileSystemProvider instead of a NullFileSystemProvider. Then we can keep the `main.ts` files of language servers as before. If someone wants to run Langium in a browser, they can configure their bundler such that it ignores `'fs'` imports.

packages/langium/src/workspace/file-system-provider.ts
```diff
@@ -0,0 +1,82 @@
...
+     * @param root The root path where the traversal starts.
+     * @param filter An optional filter that determines whether a directory or file should be included while traversing.
+     * @returns All URIs which match the specified filter in the directory structure.
+     */
+    traverse(root: URI, filter?: FileSystemFilter): Promise<URI[]>;
+}
+
+export class NodeFileSystemProvider implements FileSystemProvider {
+
+    protected readonly encoding: BufferEncoding = 'utf-8';
```
spoenemann, line 41: Proposal: make this public so it can be changed without subclassing.

packages/langium/src/workspace/file-system-provider.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    readFile(uri: URI): Promise<string> {
+        return fs.promises.readFile(uri.fsPath, this.encoding);
+    }
+
+    readFileSync(uri: URI): string {
+        return fs.readFileSync(uri.fsPath, this.encoding);
+    }
+
+    async traverse(root: URI, filter?: FileSystemFilter): Promise<URI[]> {
```
spoenemann, line 51: Proposal: offer a more low-level API here, simply returning the contents of a single folder (as we discussed).

spoenemann (approved):
:+1:

PR #406, opened 2022-02-10
Add types declaration

Pluralia:
* Adds `interface` and `type` syntax
* Adds `ReturnType` & adds it to the scope for cross-references
* Removes properties lifting by replacement interfaces with types in the AST (closes [#408](https://github.com/langium/langium/issues/408))
* Adds references in types as `@` and arrays as `[]` (for example, `@MyType[]` means an array of references on `MyType`)
* Adds types in the AST reflection
* Sorts the AST: interfaces and types / fields / fields types / container types / cases in reflection

### DIFF
--- a/examples/domainmodel/src/language-server/generated/ast.ts
+++ b/examples/domainmodel/src/language-server/generated/ast.ts
@@ -7,9 +7,7 @@
 /* eslint-disable @typescript-eslint/no-empty-interface */
 import { AstNode, AstReflection, Reference, isAstNode } from 'langium';
 
-export interface AbstractElement extends AstNode {
-    readonly $container: Domainmodel | PackageDeclaration;
-}
+export type AbstractElement = PackageDeclaration | Type;
 
 export const AbstractElement = 'AbstractElement';
 
@@ -40,7 +38,7 @@ export function isFeature(item: unknown): item is Feature {
     return reflection.isInstance(item, Feature);
 }
 
-export interface PackageDeclaration extends AbstractElement {
+export interface PackageDeclaration extends AstNode {
     elements: Array<AbstractElement>
     name: QualifiedName
 }
@@ -51,9 +49,7 @@ export function isPackageDeclaration(item: unknown): item is PackageDeclaration 
     return reflection.isInstance(item, PackageDeclaration);
 }
 
-export interface Type extends AbstractElement {
-    name: string
-}
+export type Type = DataType | Entity;
 
 export const Type = 'Type';
 
@@ -61,7 +57,8 @@ export function isType(item: unknown): item is Type {
     return reflection.isInstance(item, Type);
 }
 
-export interface DataType extends Type {
+export interface DataType extends AstNode {
+    name: string
 }
 
 export const DataType = 'DataType';
@@ -70,8 +67,9 @@ export function isDataType(item: unknown): item is DataType {
     return reflection.isInstance(item, DataType);
 }
 
-export interface Entity extends Type {
+export interface Entity extends AstNode {
     features: Array<Feature>
+    name: string
     superType?: Reference<Entity>
 }
 
--- a/examples/domainmodel/src/language-server/generated/ast.ts
+++ b/examples/domainmodel/src/language-server/generated/ast.ts
@@ -15,6 +15,8 @@ export function isAbstractElement(item: unknown): item is AbstractElement {
     return reflection.isInstance(item, AbstractElement);
 }
 
+export type QualifiedName = string;
+
 export type Type = DataType | Entity;
 
 export const Type = 'Type';
@@ -22,8 +24,6 @@ export const Type = 'Type';
 export function isType(item: unknown): item is Type {
     return reflection.isInstance(item, Type);
 }
-
-export type QualifiedName = string;
 
 export interface DataType extends AstNode {
     readonly $container: Domainmodel | PackageDeclaration;
@@ -103,13 +103,13 @@ export class DomainModelAstReflection implements AstReflection {
             return true;
         }
         switch (subtype) {
-            case DataType:
-            case Entity: {
-                return this.isSubtype(Type, supertype);
-            }
             case PackageDeclaration:
             case Type: {
                 return this.isSubtype(AbstractElement, supertype);
+            }
+            case DataType:
+            case Entity: {
+                return this.isSubtype(Type, supertype);
             }
             default: {
                 return false;
--- a/examples/domainmodel/src/language-server/generated/ast.ts
+++ b/examples/domainmodel/src/language-server/generated/ast.ts
@@ -23,6 +23,8 @@ export function isType(item: unknown): item is Type {
     return reflection.isInstance(item, Type);
 }
 
+export type QualifiedName = string;
+
 export interface DataType extends AstNode {
     readonly $container: Domainmodel | PackageDeclaration;
     name: string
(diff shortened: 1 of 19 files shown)


packages/langium/src/grammar/langium-grammar-scope.ts
```diff
@@ -0,0 +1,28 @@
...
+    protected readonly astNodeLocator: AstNodeLocator;
+
+    constructor(services: LangiumServices) {
+        super(services);
+        this.astNodeLocator = services.index.AstNodeLocator;
+    }
+
+    protected processNode(node: AstNode, document: LangiumDocument, scopes: PrecomputedScopes): void {
+        if (isReturnType(node)) return;
+        processTypeNodeWithNodeLocator(this.astNodeLocator)(node, document, scopes);
```
spoenemann, line 25: The function resulting from the call to  `processTypeNodeWithNodeLocator` can be cached in a class property in the constructor.

packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -11,7 +11,43 @@ entry Grammar:
...
+
+fragment SchemaType:
+	'{' attributes+=TypeAttribute* '}' ';'?;
+
+TypeAttribute:
+	name=ID (isOptional?='?')? ':' ^type=AtomType ';'?;
+
+// type ////////////
+Type:
+	'type' name=ID '=' typeAlternatives+=AtomType ('|' typeAlternatives+=AtomType)* ';'?;
```
spoenemann, line 32: Using `AtomType` here is problematic because `Type` is a subtype of `AbstractType`, which is used for cross-reference targets and for supertypes of interfaces, but these may only be object types. We can solve this with validation forbidding the usage of primitive types, references and arrays here.

spoenemann (commented):
Thanks, @Pluralia! At first glance this looks good. I'll have a more in-depth look later, here are a few initial remarks.

examples/domainmodel/src/language-server/generated/ast.ts
```diff
@@ -7,16 +7,35 @@
...
 
 export const AbstractElement = 'AbstractElement';
 
 export function isAbstractElement(item: unknown): item is AbstractElement {
     return reflection.isInstance(item, AbstractElement);
 }
 
+export type QualifiedName = string;
+
+export type Type = DataType | Entity;
```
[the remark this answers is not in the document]
Pluralia: They are sorting already :)

packages/langium/src/grammar/grammar-util.ts
```diff
@@ -55,19 +56,26 @@ function isDataTypeRuleInternal(rule: ast.ParserRule, visited: Set<ast.ParserRul
     return true;
 }
 
-export function findNameAssignment(rule: ast.ParserRule): ast.Assignment | undefined {
-    for (const node of streamAllContents(rule)) {
+export function findNameAssignment(type: ast.AbstractType): ast.Assignment | undefined {
```
msujew, line 59: Calling this function during validation leads to a stack overflow. We need to deal with recursive rulecalls similar to how you dealt with it in the `isDataTypeRule` function.

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -123,17 +107,27 @@ function buildReferenceTypeMethod(interfaces: Interface[]): GeneratorNode {
...
+    for (const typeInterface of astTypes.interfaces) {
+        for (const field of typeInterface.fields.filter(e => e.type.reference)) {
+            crossReferences.push({
+                type: typeInterface.name,
+                feature: field.name,
+                referenceType: field.type.types[0]
+            });
         }
     }
     return crossReferences;
```
msujew, line 127: Suggestion: Should we sort this array as well?

packages/langium-cli/src/generator/type-collector.ts
```diff
@@ -153,51 +269,38 @@ function collectAllParserRules(documents: LangiumDocuments, grammars: langium.Gr
         }
         visited.add(doc.uri);
         for (const rule of grammar.rules) {
-            if (langium.isParserRule(rule) && !rule.fragment && !isDataTypeRule(rule)) {
-                rules.add(rule);
+            if (langium.isParserRule(rule) && !rule.fragment) {
+                isDataTypeRule(rule) ? astResources.datatypeRules.add(rule) : astResources.parserRules.add(rule);
```
msujew, line 273: Minor: Please don't use ternary operations to do control flow. Use `if ... else` instead.

packages/langium-cli/src/generator/type-collector.ts
```diff
@@ -126,25 +167,100 @@ class TypeTree {
     }
 }
 
-export function collectAst(documents: LangiumDocuments, grammars: langium.Grammar[]): Interface[] {
-    const state = createState();
+class StateCollector {
```
msujew, line 170: Suggestion: This name is quite misleading. This class is not a collector of states, but represents the state of the collector and should therefore be named `(Type)CollectorState`.

packages/langium-cli/src/generator/type-collector.ts
```diff
@@ -516,10 +610,10 @@ function sortTypes(interfaces: Interface[]): Interface[] {
...
-                s.push(m);
-            }
+            nodes
+                .filter(e => e.nodes.includes(n))
+                .forEach(m => s.push(m));
         }
     }
     return l.map(e => e.value);
-}
+}
```
msujew, line 619: Nitpick: Add newline again.

msujew (commented):
This is looking fantastic, thanks 👍 

I noticed that highlighting of return types for terminal rules does no longer work as expected. You should replace this code:

https://github.com/langium/langium/blob/e3145e483ab2e9afe9554bb8d6c899d492b6d34a/packages/langium/src/grammar/langium-grammar-semantic-token-provider.ts#L30-L35

With this:

```ts
} else if ((isParserRule(node) && isDataTypeRule(node) || isTerminalRule(node)) && node.type?.name) {
  acceptor({
    node: node.type,
    feature: 'name',
    type: SemanticTokenTypes.type
  });
}
```

packages/langium/src/grammar/langium-grammar.langium
spoenemann: Should be done as part of #415.

packages/langium/src/grammar/grammar-util.ts
```diff
@@ -55,24 +57,41 @@ function isDataTypeRuleInternal(rule: ast.ParserRule, visited: Set<ast.ParserRul
...
-            }
+            return go(node, node.rule.ref);
+        } else if (ast.isAtomType(node) && node.refType.ref) {
+            return go(node, node.refType.ref);
         }
     }
     return undefined;
 }
 
+export function distictAndSorted<T>(list: T[], compareFn?: (a: T, b: T) => number): T[] {
```
spoenemann, line 91: This utility does not relate to the grammar, so `grammar-util` is not the right place for it. I propose moving it directly where you use it and not exporting it.

```suggestion
export function distinctAndSorted<T>(list: T[], compareFn?: (a: T, b: T) => number): T[] {
```

packages/langium/src/grammar/grammar-util.ts
```diff
@@ -55,24 +57,41 @@ function isDataTypeRuleInternal(rule: ast.ParserRule, visited: Set<ast.ParserRul
...
+    return stream(list).distinct().toArray().sort(compareFn);
```
spoenemann, line 92: The intermediate stream does not help here. Proposal:
```suggestion
    return Array.from(new Set(list)).sort(compareFn);
```

spoenemann (approved):
Let's ship this! We need to work on a few basic validations, then release v0.3.0 is ready to go.

PR #407, opened 2022-02-11
Missing JSDoc for services

gfontorbe:
I added a few JSDoc on some services interface where missing

### DIFF
--- a/packages/langium/src/workspace/ast-node-locator.ts
+++ b/packages/langium/src/workspace/ast-node-locator.ts
@@ -8,7 +8,7 @@ import { AstNode } from '../syntax-tree';
 import { LangiumDocument } from './documents';
 
 /**
- * Language-specific service for locating an `AstNode` in a document. 
+ * Language-specific service for locating an `AstNode` in a document.
  */
 export interface AstNodeLocator {
 


packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -30,7 +30,7 @@ export interface WorkspaceManager {
 
 }
 
-export class DefaultWorkspaceManager {
+export class DefaultWorkspaceManager implements WorkspaceManager {
```
dhuebner, line 33: Interesting that nobody complained before :)

spoenemann: This is actually optional in TypeScript because types are compared structurally. This is mostly useful to check that the interface methods are implemented correctly.

msujew (commented):
Looking good so far! Please also add a more general documentation to the interfaces themselves.

msujew (commented):
Looking good so far! Please also add a more general documentation to the interfaces themselves.

packages/langium/src/validation/validation-registry.ts
```diff
@@ -36,6 +36,9 @@ export type ValidationAcceptor = <N extends AstNode>(severity: 'error' | 'warnin
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type ValidationCheck = (node: any, accept: ValidationAcceptor, cancelToken: CancellationToken) => MaybePromise<void>;
 
+/**
+ * Manages a set of `ValidationCheck` defined inside of a `Validator` service.
```
spoenemann, line 40:
```suggestion
 * Manages a set of `ValidationCheck`s to be applied when documents are validated.
```

spoenemann (approved):
Thanks!

PR #419, opened 2022-02-21
Run converter before constructing data rule string

### LINKED ISSUE #418: Parser runs `ValueConverter` too late
Langium version: 0.2
Package name: langium

## Steps To Reproduce

1. Create a data type rule that calls `ID`.
2. Observe that any carets (`^`) are not replaced by the value converter

## The current behavior

The value converter is only called when actually assigning a value to an AST node. However, this is too late in the parsing process and ignores sub-rulecalls.

## The expected behavior

All terminals/data type rulls call `ValueConverter#convert` directly after parsing, not when being assigned.

msujew:
Closes #418 

Runs the value converter when calling the `construct` method of the parser, instead of waiting until the appropriate assignment is performed.

### DIFF
--- a/packages/langium/src/parser/langium-parser.ts
+++ b/packages/langium/src/parser/langium-parser.ts
@@ -6,7 +6,7 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { defaultParserErrorProvider, DSLMethodOpts, EmbeddedActionsParser, ILexingError, IMultiModeLexerDefinition, IOrAlt, IParserErrorMessageProvider, IRecognitionException, IToken, Lexer, TokenType, TokenTypeDictionary, TokenVocabulary } from 'chevrotain';
-import { AbstractElement, Action, Assignment, isAssignment, isCrossReference } from '../grammar/generated/ast';
+import { AbstractElement, Action, Assignment, isAssignment, isCrossReference, isKeyword } from '../grammar/generated/ast';
 import { Linker } from '../references/linker';
 import { LangiumServices } from '../services';
 import { AstNode, CompositeCstNode, CstNode, LeafCstNode } from '../syntax-tree';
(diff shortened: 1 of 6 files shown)


packages/langium/src/parser/langium-parser.ts
```diff
@@ -247,12 +248,19 @@ export class LangiumParser {
...
-            const node = obj.$cstNode;
-            return node.text;
+            return this.constructDataTypeResult(obj.$cstNode);
         }
         return obj;
     }
 
+    private constructDataTypeResult(node: CstNode): string {
+        let text = '';
+        for (const subNode of streamCst(node).filter(isLeafCstNode)) {
```
spoenemann, line 258: Shall we remove hidden tokens here? For example, a sequence `a .b .c` should result in the internal string value `a.b.c` in a language with qualified names.

packages/langium/src/parser/langium-parser.ts
```diff
@@ -247,12 +248,19 @@ export class LangiumParser {
...
+            text += this.converter.convert(subNode.text, subNode);
+        }
+        return text;
```
spoenemann, line 261: Is the datatype result sent to the value converter as well? This is necessary when the return type of a datatype rule is not string. Could you add a test for that, if there is none yet?

Example:
```
Number returns number:
    INT ('.' INT)?;

terminal INT returns number:
    /[0-9]+/;
```

msujew: Thanks, I kind of forgot that data type rules should able be sent to the value converter! Fixed this and added tests for that case 👍

packages/langium/src/parser/langium-parser.ts
msujew: Right, it also filters for hidden tokens now.

packages/langium/src/parser/langium-parser.ts
```diff
@@ -247,12 +248,19 @@ export class LangiumParser {
...
         }
         return obj;
     }
 
+    private constructDataTypeResult(node: CstNode): ValueType {
+        let text = '';
+        for (const subNode of streamCst(node).filter(e => isLeafCstNode(e) && !e.hidden)) {
+            text += this.converter.convert(subNode.text, subNode).toString();
+        }
+        return this.converter.convert(text, node);
```
spoenemann, line 261: Two more remarks on this part 😅:
 * Using `streamCst` does not consider intermediate calls to data type rules (a data type rule can use another data type rule).
 * Keyword tokens should not be passed to the value converter.

Proposal: transform to a recursive function to account for intermediate data type rules.

msujew:
@spoenemann I refactored this PR to better deal with keywords and also perform the value conversion inline. This resolves the issue of having to do it recursively, which incurs some performance penalty (doing it recursively, leads to O(n^2) behavior).

spoenemann (approved):
:+1:

PR #422, opened 2022-02-25
Support parsing of groups with guard conditions

msujew:
Fixes an issue brought up by @gfontorbe. I will add validations in a separate PR.

### DIFF
--- a/packages/langium/src/parser/langium-parser-builder.ts
+++ b/packages/langium/src/parser/langium-parser-builder.ts
@@ -4,7 +4,7 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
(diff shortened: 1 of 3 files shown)


dhuebner (approved):
LGTM

shenghu:
@msujew does guard only accept boolean? e.g. can I pass a string `Element<'en'>` to `Element<lang>`, and then check the condition by `<lang == 'en'>`? It seems not work based my testing. But I want to confirm. Thanks in advance. Or if you can point me where I can find the information, it is appreciated.

msujew:
@shenghu The feature only supports boolean logic, i.e. only `true` and `false` (and their combinations through `&`, `|` and `!`). If you require more advanced expression, please create a feature request.

Issue #415, opened 2022-02-17
Consistency validation of inferred and declared types


pluralia:
We should validate rules that are used a declared type: properties of such rules must match attributes of the declared type. An example below throws a validation error due to `MyRule` has a `ruleValue` property instead of a `typeValue` property.

```
interface MyType {
    typeValue: number;
}

MyRule returns MyType:
    ruleValue=NUMBER;
```

spoenemann:
See https://github.com/langium/langium/pull/406#discussion_r810922786: we should also validate usages of `type`s for consistency. For example, this is not valid because `Foo` is not an AST node type.
```
type Foo = string;

Bar returns Foo: ...
```

Issue #400, opened 2022-02-03
Type-collector tests


pluralia:
It's necessary to have nice covering tests for the existing type-collector before its rewriting.

PR #426, opened 2022-02-28
Add basic validations related to types

Pluralia:
* Validates type names duplication
* Moves type-collector from the `langium-cli` to the `langium` package


packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -5,10 +5,10 @@
  ******************************************************************************/
 
 import {
-    GeneratorNode, Grammar, IndentNode, CompositeGeneratorNode, NL, processGeneratorNode, streamAllContents, isCrossReference, MultiMap, LangiumServices
+    GeneratorNode, Grammar, IndentNode, CompositeGeneratorNode, NL, processGeneratorNode, streamAllContents, isCrossReference, MultiMap, LangiumServices, collectAst
 } from 'langium';
+import { AstTypes } from 'langium/src/typification/types-util';
```
spoenemann, line 10: This is not a valid import path. Use `langium/lib/...' instead.

packages/langium/src/grammar/grammar-util.ts
```diff
@@ -392,11 +396,24 @@ export function processTypeNodeWithNodeLocator(astNodeLocator: AstNodeLocator):
...
+}
+
+export function processActionNodeWithNodeDescriptionProvider(descriptions: AstNodeDescriptionProvider): (node: AstNode, document: LangiumDocument, scopes: PrecomputedScopes) => void {
+    type ContainerType = ast.Alternatives | ast.Assignment | ast.AtomType | ast.CharacterRange | ast.CrossReference | ast.Group | ast.NegatedToken | ast.ParserRule | ast.TerminalAlternatives | ast.TerminalGroup | ast.TerminalRule | ast.UnorderedGroup | ast.UntilToken;
+    return (node: AstNode, document: LangiumDocument, scopes: PrecomputedScopes) => {
+        let container = node.$container;
+        if (container && ast.isAction(node)) {
+            while (container.$container) {
+                container = container.$container as ContainerType;
+            }
```
spoenemann, line 415: We should extract this to a `getRootNode` function in `ast-util`.

packages/langium/src/grammar/langium-grammar-scope.ts
```diff
@@ -4,26 +4,73 @@
...
+                        }
+                    }
+                    scopes.push(stream(
+                        scopesArray.concat(
+                            parserRuleScopesArray.filter(parserRule => !scopesArray.some(e => e.name === parserRule.name))
+                        )
+                    ));
+                }
+                currentNode = currentNode.$container;
+            } while (currentNode);
```
spoenemann, line 48: This can be simplified because we know that all types are present in the root, so we don't need the `do...while` loop navigating all container nodes.

packages/langium/src/index.ts
```diff
@@ -27,6 +27,7 @@ export * from './serializer/json-serializer';
 export * from './service-registry';
 export * from './services';
 export * from './syntax-tree';
+export * from './typification/type-collector';
```
spoenemann, line 30: The type collector is related to the grammar language and is not meant to be used in custom DSLs. Please move it to the `grammar` subfolder.

packages/langium/src/typification/declared-types.ts
```diff
@@ -0,0 +1,60 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
```
spoenemann, line 2: Please use 2022 for new files.

packages/langium/src/typification/inferred-types.ts
```diff
@@ -0,0 +1,405 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import _ from 'lodash';
```
spoenemann, line 7: This is not an explicit dependency of this package. Let's keep dependencies at a minimum and avoid the usage of additional libraries when possible.

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -316,6 +327,15 @@ export class LangiumGrammarValidator {
         }
     }
 
+    checkTypes(grammar: ast.Grammar, accept: ValidationAcceptor): void {
+        for (const typeInconsistency of validateTypes(grammar)) {
+            accept('error', `The inferred type ${typeInconsistency.typeName} is not consistent with its declared version.`, {
```
spoenemann, line 332: This message is not helpful enough. We need to print exactly what properties are inconsistent and why (different type, property not declared).

spoenemann (commented):
Thanks! As far as I can tell, #429 is not addressed here, that's a different validation (and it's ok to handle it separately).

packages/langium/src/typification/inferred-types.ts
spoenemann: Note: it's ok to use lodash and other libraries in `langium-cli` because that is a CLI, not a library.

Pluralia:
> Thanks! As far as I can tell, #429 is not addressed here, that's a different validation (and it's ok to handle it separately).

You are right! I meant #415. I've fixed it in the PR description.

packages/langium/src/grammar/typification/type-validator.ts
```diff
@@ -0,0 +1,142 @@
...
+        } else if (isInterface(declaredType) && isInterface(inferredType.type)) {
+            inconsistencyReasons.push(...checkFieldConsistency(inferredType.type.fields, declaredType.fields));
+            inconsistencyReasons.push(...checkStringElementConsistency(inferredType.type.superTypes, declaredType.superTypes, 'super type'));
+        } else {
+            inconsistencyReasons.push(`Inferred and declared versions of type ${inferredTypeName} have to be both types or interfaces.`);
+        }
+
+        if (inconsistencyReasons.length > 0) {
+            result.push({
+                nodes: inferredType.rules,
```
spoenemann, line 38: Errors related to assignments should be shown on the respective assignment, not the containing grammar rule.

packages/langium/src/grammar/typification/type-validator.ts
```diff
@@ -0,0 +1,142 @@
...
+        .filter(declaredField => !inferredType.some(inferredField => inferredField.name === declaredField.name))
+        .map(e => e.name);
+    inconsistencyReasons.push(...getExtraAndLackError(extra, lack, 'field'));
+
+    for (const comparingFields of commonByName) {
+        const inferred = comparingFields.inferredField;
+        const declared = comparingFields.declaredField;
+        inconsistencyReasons.push(...checkAlternativesConsistency(inferred.typeAlternatives, declared.typeAlternatives, `type alternative of '${inferred.name}' field`));
+        if (inferred.optional && !declared.optional) {
+            inconsistencyReasons.push(`Inferred field '${inferred.name}' is optional, but declared one is mandatory.`);
```
spoenemann, line 128: This check should not be shown for array types. Arrays are always created by the parser, even when no `+=` assignment is processed.

packages/langium/src/grammar/typification/type-validator.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    const lack: string[] = declaredType
+        .filter(declaredField => !inferredType.some(inferredField => inferredField.name === declaredField.name))
+        .map(e => e.name);
+    inconsistencyReasons.push(...getExtraAndLackError(extra, lack, 'field'));
+
+    for (const comparingFields of commonByName) {
+        const inferred = comparingFields.inferredField;
+        const declared = comparingFields.declaredField;
+        inconsistencyReasons.push(...checkAlternativesConsistency(inferred.typeAlternatives, declared.typeAlternatives, `type alternative of '${inferred.name}' field`));
```
spoenemann, line 126: I see two errors when a wrong type is assigned (e.g. declared number, but assigned string):
> There is no 'string' type alternative of 'name' field in declared type.
> Lack of 'number' type alternative of 'name' field in inferred type.

I propose to show a single message like this instead:
> The assigned type 'string' is not compatible with the declared property 'name' of type 'number'.

In general, please use the term "property" (JavaScript) instead of "field" (Java).

packages/langium/src/grammar/typification/type-validator.ts
```diff
@@ -0,0 +1,142 @@
...
+            inconsistencyReasons.push(`Declared field '${inferred.name}' is optional, but inferred one is mandatory.`);
+        }
+    }
+
+    return inconsistencyReasons;
+}
+
+function getExtraAndLackError(extra: string[], lack: string[], errorElementName: string): string[] {
+    const inconsistencyReasons: string[] = [];
+    extra.forEach(e => inconsistencyReasons.push(`There is no '${e}' ${errorElementName ?? ''} in declared type.`));
```
spoenemann, line 139: Please also mention the name of the declared type for the sake of clarity.

msujew (commented):
A general note: Instead of the `typification` folder, I think `type-system` would be more fitting.

I noticed that arrays can be inferred to be optional, however, that's not the runtime works. Array properties are always set with an empty array. This validation error is inconsistent with how the runtime deals with arrays:

image

msujew:
We should also consider refactoring the type inference even more, it is quite slow currently. It needs roughly 50ms for the Langium grammar. Doing this on every change seems quite excessive.

spoenemann:
> We should also consider refactoring the type inference even more, it is quite slow currently. It needs roughly 50ms for the Langium grammar. Doing this on every change seems quite excessive.

Yes, let's do that after the release and after finishing #400.

spoenemann (approved):
Great work!

Please create a new issue about writing code actions as quick fixes for these validations.

packages/langium/src/grammar/type-system/type-validator.ts
```diff
@@ -0,0 +1,257 @@
...
+    for (const errorInfo of errorsInfo) {
+        errorToRuleNodes(`A type '${errorInfo.typeString}' ${errorInfo.errorMessage}`);
+    }
+}
+
+function checkPropertiesConsistency(inferred: Property[], declared: Property[],
+    errorToRuleNodes: (error: string) => void, errorToAssignment: (propertyName: string, error: string) => void): void {
+
+    const baseError = (propertyName: string, foundType: string, expectedType: string) =>
+        `The assigned type ‘${foundType}’ is not compatible with the declared property ‘${propertyName}’ of type ‘${expectedType}’.`;
```
msujew, line 178: Minor: Please use real apostrophs (`'`) here.

msujew (commented):
Looking really good, I just noticed one more issue, which makes using interfaces which are used inside of a `type` impossible:

grafik

Also, the type `A` is not inferred in this case. Generally, `B` is not a "real" super type of `A`, they're only type compatible.

msujew:
I also noticed that the calculation for type alternatives does not work correctly:

grafik

Also, I don't believe it makes sense to force users to make optional properties actually optional in the grammar:

grafik

As an example, we have two rules which return the same type. In one rule, we don't even reference a property (it has to be optional in the type system), while in the other rule its mandatory. This will always lead to a validation error, even though that's a valid use case.

packages/langium/src/grammar/type-system/inferred-types.ts
```diff
@@ -0,0 +1,416 @@
...
+        }]
+    });
+}
+
+function findTypes(terminal: AbstractElement, types: TypeCollection): void {
+    if (isAlternatives(terminal) || isUnorderedGroup(terminal) || isGroup(terminal)) {
+        findInCollection(terminal, types);
+    } else if (isKeyword(terminal)) {
+        types.types.push(`'${terminal.value}'`);
+    } else if (isRuleCall(terminal)) {
```
msujew, line 203:
```suggestion
    } else if (isRuleCall(terminal) && terminal.rule.ref) {
```

packages/langium/src/grammar/type-system/inferred-types.ts
```diff
@@ -0,0 +1,416 @@
...
+        types.types.push(getRuleType(terminal.rule.ref));
+    } else if (isCrossReference(terminal)) {
```
msujew, line 205:
```suggestion
    } else if (isCrossReference(terminal) && terminal.type.ref) {
```

msujew (approved):
Great work, thanks!

I have two minor improvements. These two lines currently lead to errors during validations if the cross references cannot be resolved.

PR #432, opened 2022-03-03
Fixed domainmodel example generator

spoenemann:
URI should always be absolute. Relative paths from the CLI have to be resolved before passing them as URI to any Langium service.


msujew (approved):
Looking good to me!

Issue #382, opened 2022-01-20
Cross-references should be possible to do only on types


Pluralia:
It's necessary to validate cross-references that they work only for types and not for rules.
Look at [the arithmetic example](https://github.com/langium/langium/blob/main/examples/arithmetics/src/language-server/arithmetics.langium#L35): cross-reference on `AbstractDefinition` in the `PrimaryExpression` is correct, because `AbstractDefinition` is a rule and also a type. However, we can replace the cross-reference on `AbstractDefinition` with an `Addition` without any errors, despite the `Addition` is only rule and not a type.

```
AbstractDefinition:
	Definition | DeclaredParameter;

Addition returns Expression:
	Multiplication ({BinaryExpression.left=current} operator=('+' | '-') right=Multiplication)*;

PrimaryExpression returns Expression:
	'(' Expression ')' |
	{NumberLiteral} value=NUMBER |
	{FunctionCall} func=[AbstractDefinition] ('(' args+=Expression (',' args+=Expression)* ')')?;
```

spoenemann:
This affects all usages of types. For example, if you change the `DeclaredParameter` rule to
```
DeclaredParameter returns DP:
	name=ID;
```
its usage in the `AbstractDefinition` type should become invalid.

PR #434, opened 2022-03-04
Cross-references only on types

dhuebner:
Adds a validation as described in:
Cross-references should be possible to do only on types #382

Also added `/packages/langium-vscode/out` to the "Attach to Language Server" launch config to be able to set breakpoints when debugging the Grammar language

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -57,7 +57,8 @@
 				"${workspaceFolder}/packages/langium/lib/**/*.js",
 				"${workspaceFolder}/examples/arithmetics/out/**/*.js",
 				"${workspaceFolder}/examples/domainmodel/out/**/*.js",
-				"${workspaceFolder}/examples/statemachine/out/**/*.js"
+				"${workspaceFolder}/examples/statemachine/out/**/*.js",
+				"${workspaceFolder}/packages/langium-vscode/out/**/*.js"
 			]
 		},
 		{
--- a/packages/langium/src/grammar/langium-grammar-validator.ts
+++ b/packages/langium/src/grammar/langium-grammar-validator.ts
@@ -9,6 +9,7 @@ import { DiagnosticTag } from 'vscode-languageserver-types';
 import { Utils } from 'vscode-uri';
 import { References } from '../references/references';
 import { LangiumServices } from '../services';
+import { Reference } from '../syntax-tree';
 import { getContainerOfType, getDocument, streamAllContents } from '../utils/ast-util';
 import { MultiMap } from '../utils/collections';
 import { stream } from '../utils/stream';
@@ -58,7 +59,11 @@ export class LangiumGrammarValidationRegistry extends ValidationRegistry {
             CrossReference: [
                 validator.checkCrossReferenceSyntax,
                 validator.checkCrossRefNameAssignment,
-                validator.checkCrossRefTerminalType
+                validator.checkCrossRefTerminalType,
+                validator.checkCrossRefType
+            ],
+            AtomType: [
+                validator.checkAtomTypeRefType
             ]
         };
         this.register(checks, validator);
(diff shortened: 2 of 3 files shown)


dhuebner:
Here is an alternative implementation if you like that more:
```ts
checkCrossRefType(reference: ast.CrossReference, accept: ValidationAcceptor): void {
    this.checkReferenceToRuleButNotType(reference.type, { node: reference, property: 'type' }, accept);
}

checkAtomTypeRefType(atomType: ast.AtomType, accept: ValidationAcceptor): void {
    this.checkReferenceToRuleButNotType(atomType.refType, { node: atomType, property: 'refType' }, accept);
}

protected checkReferenceToRuleButNotType<T extends AstNode, P extends Properties<T>>(type: Reference<ast.AbstractType>, info: DiagnosticInfo<T, P>, accept: ValidationAcceptor): void {
    if (type && ast.isParserRule(type.ref) && !isDataTypeRule(type.ref) && type.ref.type) {
        accept('error', `Use the rule type '${type.ref.type.name}' instead of the typed rule '${type.ref.name}' for cross references.`, info);
    }
}
```

.vscode/launch.json
```diff
@@ -57,7 +57,8 @@
 				"${workspaceFolder}/packages/langium/lib/**/*.js",
 				"${workspaceFolder}/examples/arithmetics/out/**/*.js",
 				"${workspaceFolder}/examples/domainmodel/out/**/*.js",
-				"${workspaceFolder}/examples/statemachine/out/**/*.js"
+				"${workspaceFolder}/examples/statemachine/out/**/*.js",
+				"${workspaceFolder}/packages/langium-vscode/out/**/*.js"
```
msujew, line 61: Question: I was wondering for a long time why our source maps for the langium-extension itself didn't work. Did this fix it?

dhuebner: This fix it for me at least. After adding this `out` entry I was able to set breakpoints in langium package files e.g. the `langium-grammar-validator.ts`

msujew: Yeah, that makes sense, thanks 👍

msujew (approved):
Changes are looking good to me, thanks!

dhuebner:
Cool, I just would like to provide a small test for that and then merge

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -438,6 +443,27 @@ export class LangiumGrammarValidator {
...
+    checkAtomTypeRefType(atomType: ast.AtomType, accept: ValidationAcceptor): void {
+        const issue = this.checkReferenceToRuleButNotType(atomType.refType);
+        if (issue) {
+            accept('error', issue, { node: atomType, property: 'refType' });
+        }
+    }
+
+    protected checkReferenceToRuleButNotType(type: Reference<ast.AbstractType>): string | undefined {
+        if(type && ast.isParserRule(type.ref) && !isDataTypeRule(type.ref) && type.ref.type) {
+            return `Use the rule type '${type.ref.type.name}' instead of the typed rule '${type.ref.name}' for cross references.`;
```
Pluralia, line 463: The error message looks to me a bit confusing. I'd propose such one:
`The rule '${type.ref.name}' has a type that doesn't match with its name. Use the type '${type.ref.type.name}' when want to reference on a rule '${type.ref.name}'.`

Pluralia (approved):

packages/langium/src/grammar/langium-grammar-validator.ts
dhuebner: > The rule '${type.ref.name}' has a type that doesn't match with its name.

Is that the real problem why we do the validation? I thought it is more like we need to reference the Type if it is explicitly defined. And we actually do link against the type even if it has the same name.

```js
Definition returns Definition:
	name=ID;
```

dhuebner: @Pluralia 
I will merge the current state if you don't mind. And I also think that someone should look over the messages we are currently producing and check for plausibility, comprehensibility and spelling :)

Pluralia: > Is that the real problem why we do the validation?

Basically, yes.
The real problem is cross-references can be only on types, but not on rules, so, when we want to make a cross-reference on a rule, we have to know a type of the rule. If a rule has an explicitly defined type -- the rule type is obvious. If no explicitly defined type -- the rule type is something, what is inferred from the rule and has the same name that the rule has. So, by the phrase `The rule '${type.ref.name}' has a type that doesn't match with its name.` I wanted to say, that the rule has an explicitly defined type that is not inferred from the rule.

Your example with `Definition` is nice and shows that this phrase is not completely correct. A rule type can really have the same name as the rule when an explicitly defined type has the same name as the rule. However, when we reference on such rule like `Definition`, we use the `Definition` word in cross-reference, and it anyway will be recognized as a reference on the return type, so, the error will not rise. The error will rise exactly in cases when the phrase is correct.

PR #435, opened 2022-03-04
Build vscode extension using esbuild

msujew:
Removes a lot of overhead and improves building times massively.


packages/langium-vscode/esbuild.js
```diff
@@ -0,0 +1,24 @@
...
+    entryPoints: ['src/extension.ts', 'src/language-server/main.ts'],
+    outdir: 'out',
+    bundle: true,
+    loader: { '.ts': 'ts' },
+    external: ['vscode'], // the vscode-module is created on-the-fly and must be excluded.
+    platform: 'node', // VSCode extensions run in a node process
+    sourcemap: true,
+    watch: watch ? {
+        onRebuild(error) {
+            if (error) console.error('Watch build failed')
```
spoenemann, line 17: I don't see any error message in the terminal when a TypeScript source file contains an error.

packages/langium-vscode/esbuild.js
```diff
@@ -0,0 +1,24 @@
...
+            else console.log(success)
```
spoenemann, line 18: Please add a timestamp so one can see when the last rebuild took place.

spoenemann (commented):
This is a lot faster and has fewer dependencies than Webpack, so :+1:

However, the console messages should be improved:

packages/langium-vscode/esbuild.js
msujew: I can't reproduce this issue. I see an error message:

image

It of course only works for `.ts` files in the `langium-vscode` directory, but that is expected behavior (and I believe webpack does the same)

packages/langium-vscode/esbuild.js
msujew: Yep, good idea 👍

packages/langium-vscode/esbuild.js
spoenemann: What I see is that syntax errors are reported, but nothing else. Try adding this to `main.ts`:
```typescript
connection.blabla(1, 2, 3);
let foo = null;
foo.start();
asdf; qwer; 1234;
```

msujew: Ah, I see. Yes, esbuild does not perform any type checking. It basically handles TypeScript as it were JavaScript with type annotations, which it simply ignores. Since we rarely modify the extension itself and usually use it for testing code in the `langium` project, I don't see this as a critical issue.

spoenemann: The language server code is only three lines, but we have much more code in extension.ts, including the template decorations that @danieldietrich once made (but are still in a PoC state AFAIK). Could we at least run `tsc` without generating output (just for type checking) as part of `vscode:prepublish`?

danieldietrich: @kaisalmen and me already use esbuild in some projects and we figured out ways how to best deal with generating type declarations. If you have questions, we can chat.

kaisalmen: vites / vues approach is to perform something like `tsc --noEmit` when building for production as it internally relies on esbuild for dev mode mode. This could be done in prebublish

msujew: Yeah, `tsc --noEmit` does the job well, thanks for the hint 👍

packages/langium-vscode/package.json
```diff
@@ -45,10 +45,10 @@
   "scripts": {
     "prepare": "npm run copy-schema",
     "copy-schema": "cpy ../langium-cli/langium-config-schema.json ./data",
-    "vscode:prepublish": "npm run clean && npm run lint && webpack --mode=production",
+    "vscode:prepublish": "npm run clean && tsc --noEmit && npm run lint && node esbuild.js --minify",
     "clean": "rimraf out",
-    "build": "webpack",
-    "watch": "webpack --watch",
+    "build": "node esbuild.js",
```
spoenemann, line 50: Please also add `tsc --noEmit` to `build` to ensure that the CI build fails when there are type errors.

spoenemann (approved):

PR #436, opened 2022-03-09
Prevent rules from returning type aliases

msujew:
Closes https://github.com/langium/langium/issues/429

Thinking more about this topic, the only valid use case to reference `type Foo = ...` anywhere would be for cases in which it only references a single other type. Since this introduces more complex logic in the runtime part of Langium - the AST building step inside of the parser would need to follow type references - I simply disabled all references to types as return types of rules.


Pluralia:
Extending of interfaces with types is still possible. Could you validate this case too?
```
type Foo = string;
interface Bar extends Foo ...
```

msujew:
@Pluralia Thanks for the hint, they are validated as well now 👍

spoenemann (approved):
Looks good, the only caveat is that it does not work with type aliases defined in other langium files. Let's ignore this problem for now. It will be solved when we switch to real cross-references to declared types.

PR #438, opened 2022-03-10
Add grammar support for `infer`

msujew:


Pluralia (approved):
Looks good!

examples/domainmodel/src/language-server/domain-model.langium
```diff
@@ -22,7 +22,7 @@ Entity:
         (features+=Feature)*
     '}';
 
-Feature:
+Feature infer Feature:
```
spoenemann, line 25: This seems unnecessary.

packages/langium/CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ type Symbol = Entity | PackageDeclaration | DataType | Feature
 ```
 The `interface` form describes the properties of an AST node type. The `@` character used at the `superType` property above denotes a cross-reference to a node of type `Entity`. The `type` form creates _union types_, i.e. an alternative of other declared or inferred types. These are transferred almost identically to TypeScript, where they have [their usual meaning](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html).
 
+Parser rules which use the `returns` directive to infer a new type should now use the `infer` keyword instead. Using `returns` is only valid for explicitly declared types.
```
spoenemann, line 33: Please also mention actions and add a link to this PR.

spoenemann (approved):

PR #439, opened 2022-03-10
Support for backticks in grammars

semperedev:
As the grammar is stored inside a template string delimited by backticks, if a grammar contains backticks the generated `grammar.ts` contains invalid syntax. Escaping the backticks solves the issue.

I couldn't find tests for this package, if they exist let me know to add a check for this situation.


spoenemann (approved):
Thanks for the contribution! You're right, we're still missing tests for that package. You're welcome to add a test for this issue in a separate PR. Just add a file `packages/langium-cli/test/generator/grammar-serializer.test.ts` and call the `serializeGrammar` function directly in a unit test (see `stream.test.ts` for an example of unit tests).

PR #443, opened 2022-03-14
Optional semicolon for imports

msujew:
Also contains a change inside of the `grammar.ts` of the domain-model example, since I forgot to regenerate that one.


spoenemann (approved):
:+1:

PR #444, opened 2022-03-15
Replace `returns` with `return`

pluralia:


spoenemann:
Thanks for the contribution! But after rethinking, I propose to change it the other way around: #445

PR #445, opened 2022-03-16
Change 'infer' to 'infers' for grammar rules

spoenemann:
This is an alternative to #444 that instead of changing `returns` to `return` changes `infer` to `infers` for grammar rules for consistency.

The validations and code actions are not tested yet and probably need to be fixed.

### DIFF
--- a/packages/langium/src/grammar/langium-grammar-code-actions.ts
+++ b/packages/langium/src/grammar/langium-grammar-code-actions.ts
@@ -85,7 +85,7 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
         const data = diagnostic.data as DocumentSegment;
         if (data) {
             return {
-                title: "Correct 'infers' usage",
+                title: "Correct 'infer' usage",
                 kind: CodeActionKind.QuickFix,
                 diagnostics: [diagnostic],
                 edit: {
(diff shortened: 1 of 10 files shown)


packages/langium/src/grammar/langium-grammar-code-actions.ts
```diff
@@ -85,7 +85,7 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
         const data = diagnostic.data as DocumentSegment;
         if (data) {
             return {
-                title: "Correct 'infer' usage",
+                title: "Correct 'infers' usage",
```
pluralia, line 88: It has to still `infer`, because it's applied to actions.

pluralia (changes requested):
Looks like good! The only one moment about a code action for actions. It always adds the `infer` keyword if it was not found. I'd propose to support replacement of `infers` and `returns`:
Screenshot 2022-03-16 at 13 31 36

spoenemann:
The case you are showing is a wrong syntax that leads to a parser error. I would not start supporting quick fixes for such cases.

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -87,7 +87,7 @@ export namespace IssueCodes {
     export const MissingImport = 'missing-import';
     export const UnnecessaryFileExtension = 'unnecessary-file-extension';
     export const InvalidReturns = 'invalid-returns';
-    export const InvalidInfer = 'invalid-infer';
+    export const InvalidInfers = 'invalid-infer';
```
pluralia, line 90: 'invalid-infer' should be replaced with 'invalid-infers'.

pluralia (approved):
Great! Could you submit also updated `grammar.ts` files for langium and arithmetics example?

PR #453, opened 2022-03-18
Fixed two cancellation bugs, added isOperationCancelled

spoenemann:
* `DefaultDocumentValidator` needs to check whether a caught error is the OperationCancelled symbol.
 * `DefaultScopeComputation` had a missing `await` while calling `interruptAndCheck`.
 * Added `isOperationCancelled` to check an error. In case at some point we see problems with the strict symbol equality, we could enhance that function to something like this:
   ```
   export function isOperationCancelled(err: unknown): err is typeof OperationCancelled {
       return err === OperationCancelled || typeof err === 'symbol' && err.description === 'OperationCancelled';
   }
   ```


msujew (approved):
Looks good for me 👍

PR #455, opened 2022-03-18
Fix type alternatives validation

### LINKED ISSUE #454: Wrong type alternatives validation
Langium version: 0.3.0
Package name: langium

## Steps To Reproduce

1. Add this code snippet to the grammar
```
interface X {
    x: string | number
}

RuleX returns X:
    x=ID;
```
2. Take a look at the validation error on `x`

## The current behavior
There is a validation error:
Screenshot 2022-03-18 at 12 49 53

## The expected behavior
No validation errors.

pluralia:
* Fixes #454 
* Improves the error message. "... expected in inferred type X" can confuse when we know, that we use here a declared type `X`. I propose to replace it with "... expected in a rule that returns a type X".
Screenshot 2022-03-17 at 11 14 14

* Fixes error usage of `ref` property in `atomTypeToPropertyType` (it could be undefined)
* Makes `projectName` property in `langium-config.json` UpperCamelCase to avoid problems with generated files that use `projectName`
* Fixes `IParserConfig` import in the generated `module.ts`
* Fixes `UnionType` initialization

### DIFF
--- a/packages/langium-cli/src/langium.ts
+++ b/packages/langium-cli/src/langium.ts
@@ -12,6 +12,7 @@ import { generate, GenerateOptions, GeneratorResult } from './generate';
 import { cliVersion, elapsedTime, getTime, log, schema } from './generator/util';
 import { LangiumConfig, loadConfigs, RelativePath } from './package';
 import path from 'path';
+import _ from 'lodash';
 
 const program = new Command();
 program
@@ -42,7 +43,11 @@ async function forEachConfig(options: GenerateOptions, callback: (config: Langiu
         });
         process.exit(1);
     }
-    const results = await Promise.all(configs.map(config => callback(config, options)));
+    const results = await Promise.all(configs.map(config => {
+        config.projectName = _.camelCase(config.projectName);
+        config.projectName = config.projectName.charAt(0).toUpperCase() + config.projectName.slice(1);
+        return callback(config, options);
+    }));
     const allSuccessful = results.every(result => result === 'success');
     if (options.watch) {
         if (allSuccessful) {
--- a/packages/langium/src/grammar/type-system/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/inferred-types.ts
@@ -295,14 +295,24 @@ function addAction(graph: TypeGraph, parent: TypePart, action: Action): TypePart
 function addAssignment(current: TypePart, assignment: Assignment): void {
     const typeItems: TypeCollection = { types: new Set(), reference: false };
     findTypes(assignment.terminal, typeItems);
-    current.properties.push({
-        name: assignment.feature,
-        optional: isOptional(assignment.cardinality),
-        typeAlternatives: [{
+
+    let typeAlternatives: PropertyType[] = [];
+    if (assignment.operator === '=' && !typeItems.reference) {
+        stream(typeItems.types)
+            .distinct()
+            .forEach(type => typeAlternatives.push({ array: false, types: [type], reference: false }));
+    } else {
+        typeAlternatives = [{
             array: assignment.operator === '+=',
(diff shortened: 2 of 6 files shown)


packages/langium/src/grammar/type-system/inferred-types.ts
```diff
@@ -184,14 +184,23 @@ function addAction(state: TypeCollectorState, action: Action): void {
 function addAssignment(state: TypeCollectorState, assignment: Assignment): void {
     const typeItems: TypeCollection = { types: [], reference: false };
     findTypes(assignment.terminal, typeItems);
-    state.currentType?.properties.push({
```
msujew, line 187: I don't see why this change should be necessary. It works correctly without the change.

pluralia: It improves the error message.

Before this change inferred types group properties types, but declared ones not. For example, here type of `x` in the declared type is an array of 3 `PropertyType`, but type of `x` in the inferred type is the only 1 `PropertyType`. It doesn't allow to match `PropertyType`s correctly and find out that the problem is only in type `number` (it's an array for the declared type but a primitive type for the inferred one).
```
interface X {
    x: string | AbstractDefinition | number[]
}

RuleX returns X:
    x=(ID | NUMBER | AbstractDefinition);
```
We got such error message before this update:
```
... 'AbstractDefinition | number | string' is not expected.
```
And get such one after the update:
```
... 'number' has to be an array.
```

msujew (commented):
@pluralia Thanks for the explanation. You actually made me aware of something which should likely be illegal in the grammar languages:

```
A: name=ID (name+=ID);
```

At least one of these assignments is not valid, since we cannot add the second parsed `ID` to the array `name` (which isn't an array at this point). In the same vein:

```
interface A {
  name: string | string[]
}
```

Should also be illegal. In summary: Mixing array types (`+=`) and normal/reference (`=`) types should be forbidden. @spoenemann WDYT?

packages/langium/src/grammar/type-system/type-validator.ts
```diff
@@ -18,7 +18,7 @@ export function validateTypesConsistency(grammar: Grammar, accept: ValidationAcc
     function applyErrorToRuleNodes(nodes: readonly ParserRule[], typeName: string): (errorMessage: string) => void {
         return (errorMessage: string) => {
             nodes.forEach(node => accept('error',
-                errorMessage + ` in inferred type '${typeName}'.`,
+                errorMessage + ` in a rule that returns a type '${typeName}'.`,
```
msujew, line 21:
```suggestion
                errorMessage + ` in a rule that returns type '${typeName}'.`,
```

msujew (approved):
Thanks, works great. Minor comments, afterwards you can merge it.

packages/langium-cli/src/langium.ts
```diff
@@ -42,7 +43,11 @@ async function forEachConfig(options: GenerateOptions, callback: (config: Langiu
         });
         process.exit(1);
     }
-    const results = await Promise.all(configs.map(config => callback(config, options)));
+    const results = await Promise.all(configs.map(config => {
+        config.projectName = _.camelCase(config.projectName);
+        config.projectName = config.projectName.charAt(0).toUpperCase() + config.projectName.slice(1);
```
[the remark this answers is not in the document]
pluralia: `_.startCase()` adds spaces, what is not suitable for the `projectName`. Compare these cases:
```ts
console.log(_.startCase(_.camelCase('xXx'))); // prints "X Xx"

const x = _.camelCase('xXx')
console.log(x.charAt(0).toUpperCase() + x.slice(1)); // prints "XXx"
```

PR #456, opened 2022-03-18
Enable no-floating-promises eslint check as warning

dhuebner:
Enabled as warning because of #453

Overall 6 Problems detected by this rule.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,7 +7,15 @@
     ],
     "parserOptions": {
         "ecmaVersion": 6,
-        "sourceType": "module"
+        "sourceType": "module",
+        "project": [
+            "./tsconfig.json",
+            "./tsconfig-eslint.json",
+            "./packages/generator-langium/tsconfig.json",
+            "./packages/langium-cli/tsconfig.json",
+            "./packages/langium-sprotty/tsconfig.json",
+            "./packages/langium-vscode/tsconfig.json"
+        ]
     },
     "plugins": [
         "@typescript-eslint",
@@ -69,6 +77,7 @@
         "@typescript-eslint/prefer-for-of": "error",                // prefer for-of loop over arrays
         "@typescript-eslint/prefer-namespace-keyword": "error",     // prefer namespace over module in TypeScript
         "@typescript-eslint/triple-slash-reference": "error",       // ban /// <reference />, prefer imports
-        "@typescript-eslint/type-annotation-spacing": "error"       // consistent space around colon ':'
+        "@typescript-eslint/type-annotation-spacing": "error",      // consistent space around colon ':'
+        "@typescript-eslint/no-floating-promises": "warn"           // Promises returned by functions must be handled appropriately.
     }
 }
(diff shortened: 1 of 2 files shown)


dhuebner:
Feel free to merge after accepted! 

Should we add a label: **accept2ship**  ? :)

packages/langium/tsconfig-eslint.json
```diff
@@ -0,0 +1,15 @@
+{
```
spoenemann, line 1: Why did you add this file? What is is used for?

dhuebner: It is used in .eslintrc.json in `project` section to be when checking the `test` folder. Otherwise running `eslint src test` produces errors like `Can't read tsconfig.json` wenn analyzing the test files.

dhuebner: One could probably reduce the content to this one:
```json
{
  "extends": "./tsconfig.json",
  "include": [
    "test/**/*.ts"
  ]
}

```

spoenemann: Oh I see, it's because of this:

> Error: Error while loading rule '@typescript-eslint/no-floating-promises': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.

dhuebner: Yes, this rule needs Types information, otherwise it is almost useless

.eslintrc.json
```diff
@@ -7,7 +7,15 @@
     ],
     "parserOptions": {
         "ecmaVersion": 6,
-        "sourceType": "module"
+        "sourceType": "module",
+        "project": [
+            "./tsconfig.json",
+            "./tsconfig-eslint.json",
```
spoenemann, line 13: Now I see this error in the editor:

> Parsing error: Cannot read file '/workspace/langium/tsconfig-eslint.json'.

.eslintrc.json
```diff
@@ -7,7 +7,15 @@
...
+            "./packages/generator-langium/tsconfig.json",
+            "./packages/langium-cli/tsconfig.json",
+            "./packages/langium-sprotty/tsconfig.json",
+            "./packages/langium-vscode/tsconfig.json"
```
spoenemann, line 17: Are these necessary? It didn't seem to make a difference when I tried removing these lines.

dhuebner: I just noticed, that after restarting ESlint server linting in editor fails :( 
We will probably need to try something else. It seem to behave differently wenn running eslint in CLI and using it in the editor

.eslintrc.json
dhuebner: Yes, this is kind of a nightmare. I found any way to get everything (IDE, CLI and Tests) working at once. 
Is it possible to move the `test` folder inside the `src` and then exclude it from packaging?

spoenemann:
@dhuebner shall we close this?

dhuebner:
@spoenemann 
Yes, let close it. But the check is still disabled, I didn't managed to get both vscode tooling and CLI working at the same time. :(

PR #460, opened 2022-03-28
Refactor type inference and add tests

msujew:
Closes https://github.com/langium/langium/issues/400
Closes https://github.com/langium/langium/issues/437

Fixes a few issues in the type inference system. Mainly related to optional properties, but also regarding super types, unions etc.


packages/langium/test/grammar/type-system/inferred-types.spec.ts
```diff
@@ -0,0 +1,566 @@
...
+            optional: false,
+            typeAlternatives: [{
+                array: false,
+                reference: false,
+                types: ['string']
+            }]
+        });
+    });
+
+    test('E is inferred with name?:string, value?: number', () => {
```
dhuebner, line 189: Nice!

packages/langium/test/grammar/type-system/inferred-types.spec.ts
```diff
@@ -0,0 +1,566 @@
...
+
+describeTypes('inferred types of simple grammars', `
+    A: name=ID value=NUMBER?;
+    B: name=FQN values+=NUMBER;
+    C: 'c' ref=[A];
+
+    FQN: ID;
+    terminal ID returns string: /string/;
+    terminal NUMBER returns number: /number/;
+`, function() {
```
spoenemann, line 38:
```suggestion
`, () => {
```

spoenemann (approved):
Thanks! Could you re-generate the examples as well (for the updated header comment)?

packages/langium/test/grammar/type-system/inferred-types.spec.ts
msujew: I would love to, but there's a distinct difference between these approaches in how they handle the `this` parameter in TypeScript. `() => {}` expects to stay in the upper `this` context (in this case the file root context), while `function() {}` expects a dedicated `this: ...` parameter.

While both do the same thing in JS code, TypeScript infers the `this` type correctly when using `function() {}` but not when using `() => {}`.

spoenemann: Why don't you pass the container as function argument?

msujew: Actually, we don't even need the container, if we modify our `AstTypes` instance in place. Thanks for the hint 👍

Issue #440, opened 2022-03-11
Use cross-references when using declared types


spoenemann:
We should change the grammar syntax like this:
```
ParserRule:
    (entry?='entry' | fragment?='fragment')?
    RuleNameAndParams
    (wildcard?='*' | 'returns' (returnType=[AbstractType:ID] | dataType=PrimitiveType) | inferredType=InferredType)?
    (definesHiddenTokens?='hidden' '(' (hiddenTokens+=[AbstractRule:ID] (',' hiddenTokens+=[AbstractRule:ID])*)? ')')? ':'
        alternatives=Alternatives ';';

Action infer AbstractElement:
    {infer Action} '{' (type=[AbstractType:ID] | inferredType=InferredType) ('.' feature=FeatureName operator=('='|'+=') 'current')? '}';

InferredType:
    'infer' name=ID;
```

That would greatly simplify the implementation of the grammar language because we'd always have real cross-references whenever we use a declared type.

dhuebner:
@spoenemann 

The new `InferredType` rule should just be like this?
```js
ReturnType:
    name=ID;
```

What should we do with `ReturnType` which is then only used in `TerminalRule`, keep as is? 
```js
ReturnType:
    name=(PrimitiveType | ID);
```

dhuebner:
Merged

PR #461, opened 2022-03-28
Use cross-references when using declared types

dhuebner:
Grammar changes as described in #440
- tests are passing
- generator generates same result as before the change

### DIFF
--- a/packages/langium/src/grammar/langium-grammar-validator.ts
+++ b/packages/langium/src/grammar/langium-grammar-validator.ts
@@ -9,7 +9,7 @@ import { DiagnosticTag } from 'vscode-languageserver-types';
 import { Utils } from 'vscode-uri';
 import { References } from '../references/references';
 import { LangiumServices } from '../services';
-import { AstNode } from '../syntax-tree';
+import { AstNode, Reference } from '../syntax-tree';
 import { getContainerOfType, getDocument, streamAllContents } from '../utils/ast-util';
 import { MultiMap } from '../utils/collections';
 import { toDocumentSegment } from '../utils/cst-util';
@@ -17,7 +17,7 @@ import { stream } from '../utils/stream';
 import { ValidationAcceptor, ValidationCheck, ValidationRegistry } from '../validation/validation-registry';
 import { LangiumDocument, LangiumDocuments } from '../workspace/documents';
 import * as ast from './generated/ast';
-import { findKeywordNode, findNameAssignment, getEntryRule, isDataTypeRule, resolveImport, resolveTransitiveImports, terminalRegex } from './grammar-util';
+import { findKeywordNode, findNameAssignment, getEntryRule, getTypeName, isDataTypeRule, resolveImport, resolveTransitiveImports, terminalRegex } from './grammar-util';
 import type { LangiumGrammarServices } from './langium-grammar-module';
 import { applyErrorToAssignment, collectAllInterfaces, InterfaceInfo, validateTypesConsistency } from './type-system/type-validator';
 
@@ -120,11 +120,11 @@ export class LangiumGrammarValidator {
             } else {
                 accept('error', 'This grammar is missing an entry parser rule.', { node: grammar, property: 'name' });
             }
-        } else if(!grammar.isDeclared && entryRules.length >= 1) {
+        } else if (!grammar.isDeclared && entryRules.length >= 1) {
             entryRules.forEach(rule => accept('error', 'Cannot declare entry rules for unnamed grammars.', { node: rule, property: 'name' }));
         } else if (entryRules.length > 1) {
             entryRules.forEach(rule => accept('error', 'The entry rule has to be unique.', { node: rule, property: 'name' }));
-        } else if (isDataTypeRule(entryRules[0])) {
+        } else if (entryRules.length === 1 && isDataTypeRule(entryRules[0])) {
             accept('error', 'The entry rule cannot be a data type rule.', { node: entryRules[0], property: 'name' });
         }
     }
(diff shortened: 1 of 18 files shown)


dhuebner:
@pluralia 
It would be nice if we could go through the existing type/grammar validator rules and check if they are still needed or working. It also would be good if we could together look into existing grammar-utils functions and revisit the current naming and maybe implementation, adding documentation too. 
So for now it is just a draft :)

packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -50,42 +50,46 @@ GrammarImport:
...
+    (entry?='entry' | fragment?='fragment')?
+    RuleNameAndParams
+    (wildcard?='*' | ('returns' (returnType=[AbstractType:ID] | dataType=PrimitiveType)) | inferredType=InferredType)?
     (definesHiddenTokens?='hidden' '(' (hiddenTokens+=[AbstractRule:ID] (',' hiddenTokens+=[AbstractRule:ID])*)? ')')? ':'
-        alternatives=Alternatives
-    ';';
+    alternatives=Alternatives ';';
+
+InferredType:
+    'infer' name=ID;
```
msujew, line 60: We should keep the syntax the same:

```suggestion
InferredType<imperative>:
    (<imperative> 'infer' | <!imperative> 'infers') name=ID;
```

dhuebner: After all the discussions to this topic I don't really recall what was the final outcome.
I thought it was that we use only `infer` to not have `infer` and `infers` which is easy to mix up and difficult to remember

msujew: We decided on `infer` for actions and `infers` for parser rules. That's how we released Langium 0.3

dhuebner: Okay, this doesn't really affects the outstanding work, needs just yet another round of boostraping orgy

dhuebner:
Really a **Boo**strap x)
https://github.com/langium/langium/pull/461/commits/1172471cb343c210793e4678dadd2037c126d6b2

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -248,33 +248,40 @@ export class LangiumGrammarValidator {
...
         for (const rule of grammar.rules.filter(ast.isParserRule)) {
             const isDataType = isDataTypeRule(rule);
-            if (!isDataType && rule.type?.name && types.has(rule.type.name) === !!rule.infers) {
-                const keywordNode = rule.infers ? findKeywordNode(rule.$cstNode, 'infers') : findKeywordNode(rule.$cstNode, 'returns');
-                accept('error', getMessage(rule.type.name, rule.infers), {
-                    node: rule.type,
+            const isInfers = !!rule.inferredType;
+            const ruleTypeName = getTypeName(rule);
+            if (!isDataType && ruleTypeName && types.has(ruleTypeName) && isInfers) {
+                const keywordNode = isInfers ? findKeywordNode(rule.$cstNode, 'infer') : findKeywordNode(rule.$cstNode, 'returns');
```
pluralia, line 254: Using the variable `isInfers` makes handling the `returns` case impossible: if `isInfers` is false, the evaluation never enter this if-branch and achieves the `findKeywordNode(rule.$cstNode, 'returns')` part.

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -248,33 +248,40 @@ export class LangiumGrammarValidator {
         }
         for (const rule of grammar.rules.filter(ast.isParserRule)) {
             const isDataType = isDataTypeRule(rule);
-            if (!isDataType && rule.type?.name && types.has(rule.type.name) === !!rule.infers) {
-                const keywordNode = rule.infers ? findKeywordNode(rule.$cstNode, 'infers') : findKeywordNode(rule.$cstNode, 'returns');
-                accept('error', getMessage(rule.type.name, rule.infers), {
-                    node: rule.type,
+            const isInfers = !!rule.inferredType;
+            const ruleTypeName = getTypeName(rule);
+            if (!isDataType && ruleTypeName && types.has(ruleTypeName) && isInfers) {
```
pluralia, line 253: Using the variable `isInfers` makes handling the returns case impossible: if `isInfers` is false, the evaluation never enter this if-branch and achieves the `findKeywordNode(rule.$cstNode, 'returns')` part.

dhuebner: I've applied your suggestion

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -248,33 +248,40 @@ export class LangiumGrammarValidator {
...
-            if (!isDataType && rule.type?.name && types.has(rule.type.name) === !!rule.infers) {
-                const keywordNode = rule.infers ? findKeywordNode(rule.$cstNode, 'infers') : findKeywordNode(rule.$cstNode, 'returns');
-                accept('error', getMessage(rule.type.name, rule.infers), {
-                    node: rule.type,
+            const isInfers = !!rule.inferredType;
+            const ruleTypeName = getTypeName(rule);
+            if (!isDataType && ruleTypeName && types.has(ruleTypeName) === isInfers) {
+                const keywordNode = isInfers ? findKeywordNode(rule.$cstNode, 'infer') : findKeywordNode(rule.$cstNode, 'returns');
+                accept('error', getMessage(ruleTypeName, isInfers), {
+                    node: rule.inferredType,
```
msujew, line 256: This code creates a ton of false positives and errors. If the type is not explicitly defined (`types.has` returns false) and no inferred type has been declared (`isInfers = false`), the execution enters this branch, although that's a valid use case. This becomes especially bad if `rule.inferredType` is undefined, since that currently leads to errors inside of the validation collector.

Using `!!rule.inferredType` to check for inference is just in error in general. If it doesn't exist (i.e. no explicit inference has been given), one could still automatically infer a type just from the rule name. Only the existence of `returnType || dataType` makes sure that the type is not inferred.
```suggestion
            const isInfers = !rule.returnType && !rule.dataType;
            const ruleTypeName = getTypeName(rule);
            if (!isDataType && ruleTypeName && types.has(ruleTypeName) === isInfers) {
                const keywordNode = isInfers ? findKeywordNode(rule.$cstNode, 'infer') : findKeywordNode(rule.$cstNode, 'returns');
                accept('error', getMessage(ruleTypeName, isInfers), {
                    node: rule.inferredType ?? rule,
```

msujew (commented):
There's an issue in the validation, aside from that, things are looking fine.

msujew (approved):
👍

PR #469, opened 2022-04-07
Improve AST node initialization

msujew:
Solves three issues:

1. Adds a default value for `boolean` properties
2. Improves handling of array properties, which are now only initialized when needed.
3. Introduces an AST reflection interpreter

### DIFF
--- a/packages/langium/src/grammar/ast-reflection-interpreter.ts
+++ b/packages/langium/src/grammar/ast-reflection-interpreter.ts
@@ -4,22 +4,29 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { AstNode, AstReflection, TypeMandatoryField, TypeMetaData } from '../syntax-tree';
+import { AstReflection, TypeMandatoryField, TypeMetaData } from '../syntax-tree';
 import { isAstNode } from '../utils/ast-util';
 import { MultiMap } from '../utils/collections';
 import { LangiumDocuments } from '../workspace/documents';
-import { Grammar } from './generated/ast';
+import { Grammar, isGrammar } from './generated/ast';
 import { createLangiumGrammarServices } from './langium-grammar-module';
 import { collectAst } from './type-system/type-collector';
-import { AstTypes, Property } from './type-system/types-util';
+import { AstTypes, collectAllProperties, Property } from './type-system/types-util';
 
 let emptyDocuments: LangiumDocuments;
 
-export function interpreteAstReflection(grammar: Grammar, documents?: LangiumDocuments): AstReflection {
-    if (!emptyDocuments && !documents) {
-        emptyDocuments = createLangiumGrammarServices().shared.workspace.LangiumDocuments;
+export function interpreteAstReflection(astTypes: AstTypes): AstReflection;
+export function interpreteAstReflection(grammar: Grammar, documents?: LangiumDocuments): AstReflection;
+export function interpreteAstReflection(grammarOrTypes: Grammar | AstTypes, documents?: LangiumDocuments): AstReflection {
+    let collectedTypes: AstTypes;
+    if (isGrammar(grammarOrTypes)) {
+        if (!emptyDocuments && !documents) {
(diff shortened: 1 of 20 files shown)


spoenemann:
> Introduces an AST reflection interpreter

This is about https://github.com/langium/langium/issues/421#issuecomment-1048653782 ?

msujew:
> This is about https://github.com/langium/langium/issues/421#issuecomment-1048653782 ?

@spoenemann I had that in mind, but I added the `ASTReflection` interpreter for another reason, since it's actually used inside of the parser now, so every parser test needs an `ASTReflection` now. I could've used a stub, but we needed the interpreter anyway, so I wrote that as well.

packages/langium/src/grammar/ast-reflection-interpreter.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export function interpreteAstReflection(grammar: Grammar, documents?: LangiumDocuments): AstReflection {
+    if (!emptyDocuments && !documents) {
+        emptyDocuments = createLangiumGrammarServices().shared.workspace.LangiumDocuments;
+    }
+    const collectedTypes = collectAst(documents ?? emptyDocuments, [grammar]);
+    const allTypes = collectedTypes.interfaces.map(e => e.name).concat(collectedTypes.unions.map(e => e.name));
+    const references = buildReferenceTypes(collectedTypes);
+    const metaData = buildTypeMetaData(collectedTypes);
+    const superTypeMap = buildSupertypeMap(collectedTypes);
```
dhuebner, line 26: Question: How are sub types handled in ast reflection? 
For `interface A extends B, C {}` will there be an entry for `A` with all inherited mandatory entries from `B` and `C`?

dhuebner (approved):

packages/langium/src/grammar/ast-reflection-interpreter.ts
```diff
@@ -0,0 +1,123 @@
...
+import { Grammar, isGrammar } from './generated/ast';
+import { createLangiumGrammarServices } from './langium-grammar-module';
+import { collectAst } from './type-system/type-collector';
+import { AstTypes, collectAllProperties, Property } from './type-system/types-util';
+
+let emptyDocuments: LangiumDocuments;
+
+export function interpreteAstReflection(astTypes: AstTypes): AstReflection;
+export function interpreteAstReflection(grammar: Grammar, documents?: LangiumDocuments): AstReflection;
+export function interpreteAstReflection(grammarOrTypes: Grammar | AstTypes, documents?: LangiumDocuments): AstReflection {
```
spoenemann, line 20:
```suggestion
export function interpretAstReflection(grammarOrTypes: Grammar | AstTypes, documents?: LangiumDocuments): AstReflection {
```

packages/langium/src/grammar/type-system/types-util.ts
```diff
@@ -104,6 +104,27 @@ export type AstResources = {
     types: Set<Type>
 }
 
+/**
+ * Collects all properties of all interface types. Includes super type properties.
+ * @param interfaces A topologically sorted array of interfaces.
+ */
+export function collectAllProperties(interfaces: InterfaceType[]): Map<string, Property[]> {
+    const map = new Map<string, Property[]>();
```
spoenemann, line 112: Use MultiMap?

packages/langium/src/syntax-tree.ts
```diff
@@ -95,10 +95,21 @@ export interface LinkingError extends ReferenceInfo {
...
+    isInstance(node: unknown, type: string): boolean
     isSubtype(subtype: string, supertype: string): boolean
 }
 
+export interface TypeMetaData {
+    name: string
+    mandatory: TypeMandatoryField[]
+}
+
+export interface TypeMandatoryField {
```
spoenemann, line 108: Please add documentation to the new types, especially what 'mandatory' means here.

spoenemann (approved):
Great! 👍

Issue #355, opened 2022-01-13
Hovering for implicit behavior in the grammar language


msujew:
Stuff like

* Which token/data type rule is used for cross references
* TypeScript type definitions for types built from parser rules

spoenemann:
See also #450

PR #473, opened 2022-04-11
Hover provider for crossrefs

gfontorbe:
Solves part of #355 

Added hover behavior when hovering a cross-reference.

1. If the cross-reference references a type alternative, the hover content is the type alternative itself.
```
type T = A | B;

R : prop = [T];
```
Hovering over `[T]` displays `type T = A | B;
`

2. If the cross-reference refers to a parser rule and has an explicit terminal.
```
R: prop = [T:X];
```
Hovering over `[T:X]` displays `[T:X]`

3. If the cross-reference refers to a parser rule and has no explicit terminal
```
R: prop = [T]
```
Hovering over `[T]` displays `[T:X]` with `X` being the terminal associated with the property `name` in the rule T

### DIFF
--- a/packages/langium/src/default-module.ts
+++ b/packages/langium/src/default-module.ts
@@ -8,13 +8,13 @@ import { Connection, TextDocuments } from 'vscode-languageserver';
 import { TextDocument } from 'vscode-languageserver-textdocument';
 import { Module } from './dependency-injection';
 import { createGrammarConfig } from './grammar/grammar-config';
+import { MultilineCommentHoverProvider } from './lsp';
 import { DefaultCompletionProvider } from './lsp/completion/completion-provider';
 import { RuleInterpreter } from './lsp/completion/rule-interpreter';
 import { DefaultDocumentHighlighter } from './lsp/document-highlighter';
 import { DefaultDocumentSymbolProvider } from './lsp/document-symbol-provider';
 import { DefaultFoldingRangeProvider } from './lsp/folding-range-provider';
 import { DefaultGoToResolverProvider } from './lsp/goto';
-import { LangiumGrammarHoverProvider } from './lsp/langium-grammar-hover-provider';
 import { DefaultReferenceFinder } from './lsp/reference-finder';
 import { DefaultRenameHandler } from './lsp/rename-refactoring';
 import { createLangiumParser } from './parser/langium-parser-builder';
@@ -62,7 +62,7 @@ export function createDefaultModule(context: DefaultModuleContext): Module<Langi
                 RuleInterpreter: () => new RuleInterpreter()
             },
             DocumentSymbolProvider: (services) => new DefaultDocumentSymbolProvider(services),
-            HoverProvider: (services) => new LangiumGrammarHoverProvider(services),
+            HoverProvider: (services) => new MultilineCommentHoverProvider(services),
             FoldingRangeProvider: (services) => new DefaultFoldingRangeProvider(services),
             ReferenceFinder: (services) => new DefaultReferenceFinder(services),
             GoToResolver: (services) => new DefaultGoToResolverProvider(services),
--- a/packages/langium/src/default-module.ts
+++ b/packages/langium/src/default-module.ts
@@ -14,7 +14,7 @@ import { DefaultDocumentHighlighter } from './lsp/document-highlighter';
 import { DefaultDocumentSymbolProvider } from './lsp/document-symbol-provider';
 import { DefaultFoldingRangeProvider } from './lsp/folding-range-provider';
 import { DefaultGoToResolverProvider } from './lsp/goto';
-import { MultilineCommentHoverProvider } from './lsp/hover-provider';
+import { LangiumGrammarHoverProvider } from './lsp/langium-grammar-hover-provider';
 import { DefaultReferenceFinder } from './lsp/reference-finder';
 import { DefaultRenameHandler } from './lsp/rename-refactoring';
 import { createLangiumParser } from './parser/langium-parser-builder';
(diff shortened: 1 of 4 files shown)


packages/langium/src/lsp/langium-grammar-hover-provider.ts
```diff
@@ -0,0 +1,74 @@
...
+                const relevantNode = findRelevantNode(cstNode);
+                if(isCrossReference(relevantNode)){
+                    return this.getCrossReferenceHoverContent(relevantNode);
+                } else {
+                    return super.getHoverContent(document, params);
+                }
+            }
+        }
+        return undefined;
+    }
```
msujew, line 39: Suggestion: If you're already using the `MultilineCommentHoverProvide`, you should do yourself a favor and just override the `getAstNodeHoverContent` method.

https://github.com/langium/langium/blob/244442e047424c3c519d1017a81c97225e4545ae/packages/langium/src/lsp/hover-provider.ts#L52

Something like this:

```ts
protected getAstNodeHoverContent(astNode): Hover | undefined {
  if (isCrossReference(astNode) {
    return this.getCrossReferenceHoverContent(astNode);
  }
  return super.getAstNodeHoverContent(astNode);
}
```

packages/langium/src/default-module.ts
```diff
@@ -62,7 +62,7 @@ export function createDefaultModule(context: DefaultModuleContext): Module<Langi
                 RuleInterpreter: () => new RuleInterpreter()
             },
             DocumentSymbolProvider: (services) => new DefaultDocumentSymbolProvider(services),
-            HoverProvider: (services) => new MultilineCommentHoverProvider(services),
+            HoverProvider: (services) => new LangiumGrammarHoverProvider(services),
```
msujew, line 65: Suggestion: We shouldn't use the `LangiumGrammarHoverProvider` as the default hover provider for all services, but only for the langium grammar service. See:

https://github.com/langium/langium/blob/244442e047424c3c519d1017a81c97225e4545ae/packages/langium/src/grammar/langium-grammar-module.ts#L30-L34

packages/langium/src/lsp/langium-grammar-hover-provider.ts
```diff
@@ -0,0 +1,74 @@
...
+                    contents: {
+                        kind: 'markdown',
+                        value: ref.$cstNode!.text
+                    }
+                };
+            }
+            else if(isParserRule(ref)){
+                let terminal;
+                if(node.$cstNode?.text.includes(':')){
+                    terminal = node.$cstNode?.text.split(':')[1].replace(']', '');
```
msujew, line 57: Suggestion: You should use the name of the actually referenced terminal here. Using the text might be misleading, especially if the cross reference to the terminal cannot actually be resolved.

packages/langium/src/lsp/langium-grammar-hover-provider.ts
```diff
@@ -0,0 +1,74 @@
...
+            if(isType(ref)){
+                return {
+                    contents: {
+                        kind: 'markdown',
+                        value: ref.$cstNode!.text
+                    }
+                };
+            }
+            else if(isParserRule(ref)){
+                let terminal;
```
msujew, line 55: Minor: Without a type annotation, this gets automatically typed as `any`, which might lead to errors in the future.

```suggestion
                let terminal: string;
```

packages/langium/src/lsp/langium-grammar-hover-provider.ts
```diff
@@ -0,0 +1,74 @@
...
+        }
+        return undefined;
+    }
+
+    getCrossReferenceHoverContent(node: CrossReference): MaybePromise<Hover | undefined> {
+
+        const ref = node.type.ref;
+
+        if(ref){
+            if(isType(ref)){
```
msujew, line 46: Minor: Formatting (spaces between closing parenthesis and opening curly braces)

packages/langium/src/default-module.ts
gfontorbe: Wrong module...

packages/langium/src/lsp/langium-grammar-hover-provider.ts
gfontorbe: Do we have a config for an auto-formatter?

packages/langium/src/lsp/langium-grammar-hover-provider.ts
```diff
@@ -0,0 +1,83 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
```
msujew, line 5: This files belongs to `src/grammar/lsp` instead of `src/lsp`.

packages/langium/src/lsp/langium-grammar-hover-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+                } else if (isKeyword(terminal)) {
+                    terminalName = terminal.value;
+                } else if (isRuleCall(terminal)) {
+                    terminalName = terminal.rule.ref!.name;
+                }
+
+                return {
+                    contents: {
+                        kind: 'markdown',
+                        value: `[${typeName}:${terminalName}]`
```
msujew, line 75: Suggestion: We should put this in a code block so that it receives correct syntax highlighting
```suggestion
                        value: '```\n' + `[${typeName}:${terminalName}]` + '\n```'
```

packages/langium/src/lsp/langium-grammar-hover-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    getCrossReferenceHoverContent(node: CrossReference): MaybePromise<Hover | undefined> {
+        const ref = node.type.ref;
+        if (ref) {
+            if (isType(ref)) {
+                return {
+                    contents: {
+                        kind: 'markdown',
+                        value: ref.$cstNode!.text
```
msujew, line 54: Suggestion: We should put this in a code block so that it receives correct syntax highlighting
```suggestion
                        value: '```\n' + ref.$cstNode!.text + '\n```'
```

msujew (commented):
Looking mostly good, just some more minor issues.

packages/langium/src/lsp/langium-grammar-hover-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+                    }
+                };
+            } else if (isParserRule(ref)) {
+                const typeName = ref.name;
+                const terminal = node.terminal;
+                let terminalName = '';
+                if (terminal === undefined){
```
msujew, line 61: Nitpick: 
```suggestion
                if (terminal === undefined) {
```

msujew (approved):
Thanks, looks good to me!

PR #474, opened 2022-04-14
Combines watch modes

### LINKED ISSUE #427: Create script that combines watch modes
It would be nice if _watch_ mode would watch for all changes, including ts sources and grammar file(s).

Additional dev dependency:

```sh
npm i -D concurrently
```

Changes to package.json:

```diff
"scripts": {
    "vscode:prepublish": "npm run build && npm run lint",
    "build": "tsc",
+   "build:watch": "tsc -w",
-   "watch": "tsc -w",
+   "watch": "concurrently \"npm:build:watch\" \"npm:langium:watch\"",
    "lint": "eslint src --ext ts",
    "langium:generate": "langium generate",
    "langium:watch": "langium generate --watch"
}
```

gfontorbe:
Closes #427 

The former script `npm run watch` is now started with `npm run build:watch`.

The new `npm run watch` starts both `npm run build:watch` and `npm run langium:watch` concurrently.

Added in the 3 examples, langium, and langium-template

### DIFF
--- a/examples/arithmetics/package.json
+++ b/examples/arithmetics/package.json
@@ -37,10 +37,11 @@
     "scripts": {
         "vscode:prepublish": "npm run build && npm run lint",
         "build": "tsc",
-        "watch": "tsc -watch",
+        "build:watch": "tsc -watch",
         "lint": "eslint src --ext ts",
         "langium:generate": "langium generate",
-        "langium:watch": "langium generate --watch"
+        "langium:watch": "langium generate --watch",
+        "watch": "concurrently \"npm run build:watch\" \"npm run langium:watch\""
     },
     "dependencies": {
         "colors": "^1.4.0",
@@ -54,6 +55,7 @@
         "@types/vscode": "^1.56.0",
         "@typescript-eslint/eslint-plugin": "^4.14.1",
         "@typescript-eslint/parser": "^4.14.1",
+        "concurrently": "^7.0.0",
         "eslint": "^7.19.0",
         "eslint-plugin-header": "^3.1.1",
         "langium-cli": "0.3.0",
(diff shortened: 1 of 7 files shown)


msujew (commented):
The default way of developing Langium is to simply run `npm run watch` inside of the repo root, which compiles the TypeScript code for the whole project. However, this change requires devs to execute this in each sub-project, making it quite difficult to use correctly.

I'm definitely in favor of being able to run the TypeScript compiler and the langium generator side by side, but this isn't the correct approach. Instead I'm thinking of this:

* Add a second `langium:watch` script which operates similar to `tsc --watch` at the [project root](https://github.com/langium/langium/blob/244442e047424c3c519d1017a81c97225e4545ae/package.json#L15)
* For that, we need the ability to provide multiple configs using the `-f` argument to the `langium-cli`, which then compiles/watches multiple projects simultaneously.
* Rename the existing `watch` script to `build:watch`
* Add a new `vscode:watch` script to the root which executes the `watch` script in the `langium-vscode` project
* Create a new `watch` script to the root which calls `build:watch`, `langium:watch` and `vscode:watch` using the `concurrently` package

gfontorbe:
I understand your point in the context of developing Langium.

Would you also suggest this approach in the context of developing **with** Langium? Imho having a script that runs both `build:watch` and `langium:watch` would be quite useful there.

msujew:
@gfontorbe You mean in the yeoman generator? I believe that we should only add conventional stuff there. Having a `concurrently` script in there is definitely not conventional, but downstream devs are always able to write their own script, if they want to do so.

gfontorbe:
Will come back to this later

PR #475, opened 2022-04-14
Automatically add additional documents to LangiumDocuments

### LINKED ISSUE #465: Automatically add library documents to `LangiumDocuments`
When loading library documents/other documents using the `loadAdditionalDocuments` method, downstream users have to manually add them to the `LangiumDocuments` list.

I believe that this could lead to a lot of confusion (see https://github.com/langium/langium/issues/464) and is quite unintuitive. Instead, the collector should automatically add passed documents to the `LangiumDocuments` service.

gfontorbe:
Closes #465 

This is also a problem I encountered while working on a customer project.

With my use-case, it seems that it could be solved by delegating the responsibility to add documents to `LangiumDocuments` to the `collector` instead of the `getOrCreateDocument` method.

There are most likely parts that I overlooked, but we can start from there.

### DIFF
--- a/packages/langium/src/workspace/documents.ts
+++ b/packages/langium/src/workspace/documents.ts
@@ -249,7 +249,6 @@ export class DefaultLangiumDocuments implements LangiumDocuments {
         }
         const textDoc = this.textDocuments.get(uriString) ?? this.textDocumentFactory.fromUri(uri);
         langiumDoc = this.langiumDocumentFactory.fromTextDocument(textDoc, uri);
-        this.documentMap.set(uriString, langiumDoc);
         return langiumDoc;
     }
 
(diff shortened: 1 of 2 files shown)


packages/langium/src/workspace/documents.ts
```diff
@@ -249,7 +249,6 @@ export class DefaultLangiumDocuments implements LangiumDocuments {
         }
         const textDoc = this.textDocuments.get(uriString) ?? this.textDocumentFactory.fromUri(uri);
         langiumDoc = this.langiumDocumentFactory.fromTextDocument(textDoc, uri);
-        this.documentMap.set(uriString, langiumDoc);
```
msujew, line 252: Suggestion: Removing this `set` call here makes the document cache non-functional. We should keep it.

msujew (approved):
LGTM! Thanks 👍

PR #478, opened 2022-04-28
Upgraded TypeScript to latest version

danieldietrich:
This PR fixes a compiler error with the latest TypeScript version 4.6.2. It occurs when a contributor opens Langium with the latest VS Code version.

In some places we used TS 4.1.3, in others we used TS 4.2.2. I have upgraded all versions to the latest (4.6.2):

* packages
* examples
* generator templates

@spoenemann / @msujew / @pluralia do you see any problems with upgrading the TS version in the generator template? This one might need to be thoroughly tested (e.g. in a fresh env using Gitpod). Thanks


spoenemann (approved):
:+1:

PR #481, opened 2022-05-01
Check parser rule `args` for `undefined`

msujew:
Closes https://github.com/langium/langium/issues/480

During parser self analysis, the `args` will always be `undefined`. In this case the `ParameterReference` check returns with an error, which is caught by Chevrotain, stopping the self analysis for this rule. This however, leaves the parser in an invalid state and consequently results in parsing errors.


spoenemann (approved):
Thanks!

PR #485, opened 2022-05-06
Verify type of 'name' property

### LINKED ISSUE #483: Language server crash on 'name' Cross-Ref
Assigning a cross reference to the `name` property in a Rule causes the language server to crash. 

This appears to be due to how cross references on the name property are handled internally. Possibly related to the way that cross references automatically utilize the `name` property to identify the reference.

Langium version: 0.3.0

## Steps To Reproduce

Take the [Hello World grammar](https://langium.org/docs/getting-started/#explaining-the-grammar) in the Getting Started section as an example:

1. Adjust the Greeting rule in `hello-world.langium` to use a property of `name` instead of `person` for the **Person** cross ref.
```
Greeting:
    'Hello' name=[Person:ID] '!';
```
2. Update the usages of this property in your project. In the hello-world case, this would be line **13** in **cli/generator.ts**.
3. Regen w/ Langium and rebuild the project
4. Launching the extension via F5 and viewing a file crashes the language server after a short delay with a msg like so:

```
Debugger listening on ws://127.0.0.1:6009/489f472b-13ac-4366-ba05-5058d0ab6ac5
For help, see: https://nodejs.org/en/docs/inspector
node:internal/child_process/serialization:127
    const string = JSONStringify(message) + '\n';
                   ^

TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property 'elements' -> object with constructor 'Array'
    |     index 0 -> object with constructor 'Object'
    --- property '$container' closes the circle
    at stringify (<anonymous>)
    at writeChannelMessage (node:internal/child_process/serialization:127:20)
    at process._send (node:internal/child_process:819:17)
    at process.send (node:internal/child_process:719:19)
    at IPCMessageWriter.write 
    ..... followed by a stack trace in the vscode-jsonrpc library, and then repeats about 3 times ....
```

I originally came across this on a separate example project, so it *appears* to be independent of the language implementation itself.

Ideally, this should should either not crash, or not be allowed if there's some limitation to rectifying the issue at this time.

montymxb:
This closes #483 by extending an existing check in the`isNamed` function of naming.ts. This verifies that the type of the name property on an **AstNode** is actually a string, as opposed to only checking if it's undefined or not. The check avoids misidentifying an AstNode containing a named cross ref as a regular named one (with a string). 

Before this fix, such an AstNode could sneak past into the underlying jsonrpc lib and would be stringified, producing a TypeError.

We also added a *warning* diagnostic about assigning cross refs to a 'name' property, as it's likely not what was intended.

Tests have been added to check both the behavior of the updated function and the new warning. For the later, we generalized `expectError` into `expecting`, allowing us to use the same function for different diagnostic values.

Both @Lotes and I were working on this, so please feel free to give some feedback for both of us!

### DIFF
--- a/packages/langium/src/references/naming.ts
+++ b/packages/langium/src/references/naming.ts
@@ -12,8 +12,7 @@ export interface NamedAstNode extends AstNode {
 }
 
 export function isNamed(node: AstNode): node is NamedAstNode {
-    return (node as NamedAstNode).name !== undefined
-        && typeof (node as NamedAstNode).name === 'string';
+    return typeof (node as NamedAstNode).name === 'string';
 }
 
 /**
--- a/packages/langium/src/references/naming.ts
+++ b/packages/langium/src/references/naming.ts
@@ -13,7 +13,7 @@ export interface NamedAstNode extends AstNode {
 
 export function isNamed(node: AstNode): node is NamedAstNode {
     return (node as NamedAstNode).name !== undefined
-        && typeof (node as NamedAstNode).name === "string";
+        && typeof (node as NamedAstNode).name === 'string';
 }
 
 /**
--- a/packages/langium/src/references/naming.ts
+++ b/packages/langium/src/references/naming.ts
@@ -12,7 +12,8 @@ export interface NamedAstNode extends AstNode {
 }
 
 export function isNamed(node: AstNode): node is NamedAstNode {
-    return (node as NamedAstNode).name !== undefined;
+    return (node as NamedAstNode).name !== undefined
+        && typeof (node as NamedAstNode).name === "string";
 }
 
 /**
--- a/packages/langium/test/validation/grammar-validation.test.ts
+++ b/packages/langium/test/validation/grammar-validation.test.ts
@@ -53,7 +53,7 @@ describe('Checked Named CrossRefs', () => {
     grammar g
     A: 'a' name=ID;
     B: 'b' name=[A];
-    terminal ID: /[_a-zA-Z][\w_]*/;
+    terminal ID: /[_a-zA-Z][\\w_]*/;
     `.trim();
 
     let validationData: ValidatorData;
--- a/packages/langium/test/validation/grammar-validation.test.ts
+++ b/packages/langium/test/validation/grammar-validation.test.ts
@@ -48,6 +48,31 @@ describe('checkReferenceToRuleButNotType', () => {
 
 });
 
+describe('Validate that cross-references are not named as "name"', () => {
+    const grammar = `
+    grammar HelloWorld
+    entry Model:
+        (persons+=Person | greetings+=Greeting)*;
+    Person:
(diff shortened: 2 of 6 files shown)


packages/langium/test/validation/grammar-validation.test.ts
```diff
@@ -69,7 +63,7 @@ describe('Validate that cross-references are not named as "name"', () => {
     });
 
     test('Named crossReference warning', () => {
-        expectWarning(validationData, 'We recommend not to use the "name" property for cross-references.');
+        expectWarning(validationData, 'The "name" property is not recommended for cross-references.');
```
Lotes, line 66: I do not like to compare with the error message text. Think about rewording the messages. You would have to touch also the test. Error codes are an alternative. But I also see the overhead in maintaining them.
But another argument FOR error codes could be localization/translation into other languages.

packages/langium/src/references/naming.ts
```diff
@@ -12,7 +12,7 @@ export interface NamedAstNode extends AstNode {
 }
 
 export function isNamed(node: AstNode): node is NamedAstNode {
-    return (node as NamedAstNode).name !== undefined;
+    return typeof (node as NamedAstNode).name === 'string';
```
dhuebner, line 15: Will it still be possible to have a `name` property number typed? 
```
Test : 
   name: TEST_NUMBER
 ;
terminal TEST_NUMBER returns number: /[0-9]+(\.[0-9])?/;
```

montymxb: Just checked some things. You can still type the `name` prop as `number`, that's ok.

As for a number as an identifying cross reference, I tested something out on this PR and main. Seems in either case numbers don't bind as cross refs. I'm not sure if this is also the case on the 0.3 release itself, but that part seems to be consistent.

However, what this does change is interesting. Whereas before numbers values don't show up as an identifier for a symbols in VSCode before, they can with this PR like any other 'name' value.

Screen Shot 2022-05-06 at 14 06 12

But, when I extended the `isNamed` check to also factor in number types directly, it still works, but it sometimes leads to reads to a render error. In this case, only when where was a preceding statement.
Screen Shot 2022-05-06 at 14 18 34

I like the idea of accounting for numbers directly, but I'm not yet sure how to proceed given the render error.

spoenemann: If you want numbers as names for cross-references, you should provide a custom NameProvider that converts the numbers to strings.

packages/langium/test/validation/grammar-validation.test.ts
spoenemann: What I like about using the actual diagnostic message in the test is that the test becomes well readable. I understand the concern about changing the message text, but I think adapting test expectations is ok in that case.

spoenemann (approved):
Thanks!

PR #487, opened 2022-05-09
Add validation for usage of rule fragments in types

### LINKED ISSUE #466: Validate usage of rule fragments in types
We should add validation of rule fragments usage in types. Currently, it's possible to write such grammar:
```
type Symbol = ... | SymbolName;

fragment SymbolName:
    name=SymbolID;
```
It generates a wrong AST without `SymbolName`:
```
export type Symbol = ... | SymbolName;
```
`SymbolName` is highlighted as an error: "Cannot find name 'SymbolName'."

claaspetersen:
This closes #466 by adding validation for the usage of rule fragments in type definitions.

Before it was possible to use a rule fragment as part of a type like this:

```
type TypeTest = RuleFragment
fragment RuleFragment:
    name=Test;
```

This led to errors in the AST generation.

### DIFF
--- a/packages/langium/src/grammar/langium-grammar-validator.ts
+++ b/packages/langium/src/grammar/langium-grammar-validator.ts
@@ -53,7 +53,8 @@ export class LangiumGrammarValidationRegistry extends ValidationRegistry {
                 validator.checkGrammarTypeUnions,
                 validator.checkGrammarTypeInfer,
                 validator.checkTypesConsistency,
-                validator.checkPropertyNameDuplication
+                validator.checkPropertyNameDuplication,
+                validator.checkFragmentsInTypes
             ],
             GrammarImport: validator.checkPackageImport,
             CharacterRange: validator.checkInvalidCharacterRange,
@@ -306,6 +307,18 @@ export class LangiumGrammarValidator {
         if (grammar.definesHiddenTokens) {
             accept('error', 'Hidden terminals are declared at the terminal definition.', { node: grammar, property: 'definesHiddenTokens', code: IssueCodes.HiddenGrammarTokens });
         }
+    }
+
+    checkFragmentsInTypes(grammar: ast.Grammar, accept: ValidationAcceptor): void {
+        grammar.types.map(e => {
+            e.typeAlternatives.map(e => {
+                if (ast.isParserRule(e.refType?.ref)) {
+                    if ((e.refType?.ref as ast.ParserRule).fragment) {
+                        accept('error', 'Cannot use rule fragments in as cross references in types.', { node: e, property: 'refType'});
+                    }
+                }
+            });
+        });
     }
 
     checkHiddenTerminalRule(terminalRule: ast.TerminalRule, accept: ValidationAcceptor): void {


packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -308,6 +309,18 @@ export class LangiumGrammarValidator {
         }
     }
 
+    checkFragmentsInTypes(grammar: ast.Grammar, accept: ValidationAcceptor): void {
+        grammar.types.map(e => {
+            e.typeAlternatives.map(e => {
+                if (ast.isParserRule(e.refType?.ref)) {
+                    if ((e.refType?.ref as ast.ParserRule).fragment) {
+                        accept('error', 'Cannot use rule fragments in as cross references in types.', { node: e, property: 'refType'});
```
pluralia, line 317: There is a typo in the error message 'Cannot use rule fragments in as cross references in types.'. Also, I'd suggest to replace it with 'Cannot use rule fragments in types'.

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -308,6 +309,18 @@ export class LangiumGrammarValidator {
         }
     }
 
+    checkFragmentsInTypes(grammar: ast.Grammar, accept: ValidationAcceptor): void {
+        grammar.types.map(e => {
```
pluralia, line 313: Both `map`s here can be replaced with `forEach` since we don't need a "mapped value".

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -308,6 +309,18 @@ export class LangiumGrammarValidator {
...
+            e.typeAlternatives.map(e => {
+                if (ast.isParserRule(e.refType?.ref)) {
```
pluralia, line 315: These `if`s can be unioned by the `&&` operator.

pluralia (commented):
Looks like nicely! Do you think to add tests?

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -308,6 +309,16 @@ export class LangiumGrammarValidator {
         }
     }
 
+    checkFragmentsInTypes(grammar: ast.Grammar, accept: ValidationAcceptor): void {
```
spoenemann, line 312: Please register this validation check for the `AtomType` type instead of `Grammar`, then you can write it like this:
```typescript
    checkFragmentsInTypes(type: ast.AtomType, accept: ValidationAcceptor): void {
        if (ast.isParserRule(...
    } 
```

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -308,6 +309,16 @@ export class LangiumGrammarValidator {
...
+        grammar.types.forEach(e => {
+            e.typeAlternatives.forEach(e => {
+                if (ast.isParserRule(e.refType?.ref) && (e.refType?.ref as ast.ParserRule).fragment) {
```
spoenemann, line 315: The type cast via `as` is not necessary here because it has already been checked with `isParserRule`.

spoenemann (approved):
:+1:

PR #489, opened 2022-05-10
Fix Type Validation w/ Optional Params

### LINKED ISSUE #488: Declared types validation for optional parameters and inheritance
## Current behavior:
* In a declared type, if a parameter is declared as 'optional', a validation error is returned if the optional parameter is not present in a parser rule.
Screenshot 2022-05-10 at 09 50 41

* declared types that extend other declared types do not take the inheritance into account
Screenshot 2022-05-10 at 10 12 08
Screenshot 2022-05-10 at 10 12 18

## Expected behavior:
* No error should be returned if an optional parameter is not present in the parser rule.

* Inherited interface should also validate parent types

## Steps to reproduce
Optional parameters:
```
interface MyType {
    name: string
    count?: number
}

X returns MyType : name=ID;
```

Inheritance:
```
interface Foo {
    name: string
}

interface Event extends Foo {}

Event returns Event:
    name=ID;
```

montymxb:
Closes #488. Addresses the primary issue , but *not* the additional inheritance case. This allows rules to omit parameters that are declared as optional by their declared type. Previously, whether a parameter was optional or not was ignored while checking for property consistency.

Also adds a test case to verify that optional properties can be present (or not) in different rules of the same grammar.

### DIFF
--- a/packages/langium/src/grammar/type-system/type-validator.ts
+++ b/packages/langium/src/grammar/type-system/type-validator.ts
@@ -24,16 +24,29 @@ export function validateTypesConsistency(grammar: Grammar, accept: ValidationAcc
         };
     }
 
+    // Report missing assignments for required properties in offending nodes
+    function applyMissingAssignmentErrorToRuleNodes(nodes: readonly ParserRule[], accept: ValidationAcceptor): (propertyName: string, errorMessage: string) => void {
+        return (propertyName: string, errorMessage: string) => {
+            nodes.forEach(node => {
+                const assignments = extractAssignments(node.alternatives);
+                if (assignments.find(a => a.feature === propertyName) === undefined) {
+                    accept('error', errorMessage, {node, property: 'parameters'});
+                }
+            });
+        };
+    }
+
     const validationResources = collectValidationResources(grammar);
     for (const [typeName, typeInfo] of validationResources.entries()) {
         if (!isInferredAndDeclared(typeInfo)) continue;
         const errorToRuleNodes = applyErrorToRuleNodes(typeInfo.nodes, typeName);
+        const errorToInvalidRuleNodes = applyMissingAssignmentErrorToRuleNodes(typeInfo.nodes, accept);
         const errorToAssignment = applyErrorToAssignment(typeInfo.nodes, accept);
 
         if (isType(typeInfo.inferred) && isType(typeInfo.declared)) {
             checkAlternativesConsistency(typeInfo.inferred.union, typeInfo.declared.union, errorToRuleNodes);
         } else if (isInterface(typeInfo.inferred) && isInterface(typeInfo.declared)) {
-            checkPropertiesConsistency(typeInfo.inferred.properties, typeInfo.declared.properties, errorToRuleNodes, errorToAssignment);
+            checkPropertiesConsistency(typeInfo.inferred.properties, typeInfo.declared.properties, errorToRuleNodes, errorToAssignment, errorToInvalidRuleNodes);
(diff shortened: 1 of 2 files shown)


montymxb:
A second issue (#489) has been created to separately address the inheritance case mentioned above. Just to reiterate, this PR is only for the optional param case.

gfontorbe (commented):
Just tested quickly the PR.

The problem is resolved for this grammar:
```
interface A {
    name:string
    count?:number
}

X returns A: name=ID;

Y returns A: name=ID count=NUMBER;
```

However there is a new problem, which I think is not related to the validation itself.

With the following grammar:
```
interface A {
    name:string
    count?:number
}

X returns A: name=ID;

Y returns A: count=NUMBER;
```

I would expect a validation error to be thrown on 'Y' because Y doesn't have a `name` assignement. Instead the error is thrown on the `name` assignment in the 'X' rule.

Screenshot 2022-05-11 at 09 58 49

montymxb:
I've looked this over a bit, and this other incorrect validation instance seems to have been present before the change here. I'll keep looking into it and see if I can find a way to incorporate it as well. If it gets a bit deep, it may be worth splitting off into another issue to fix separately.

montymxb:
Moved the validation error onto the rule, but because it's during type validation it does show for *all* rules returning the same type. 

When I was looking at the validation amongst inferred & declared types, it did seem a bit strange that the error for a required (but missing) property would be reported for a property assignment; which would not exist in that case. I don't know if there is such a practical case for that particular error, but I suspect it may have been a bug? 

Either way, I've guarded the newly enforced behavior with some more tests, assuming this is correct 🤔 .

Screen Shot 2022-05-12 at 14 28 23

packages/langium/src/grammar/type-system/type-validator.ts
```diff
@@ -193,18 +193,18 @@ function checkPropertiesConsistency(inferred: Property[], declared: Property[],
...
             errorToAssignment(foundProperty.name, `A property '${foundProperty.name}' is not expected.`);
         }
     }
 
     // detects lack of properties
-    for (const foundProperty of declared) {
-        const expectedProperty = inferred.find(e => foundProperty.name === e.name);
-        if (!expectedProperty) {
-            errorToRuleNodes(`A property '${foundProperty.name}' is expected`);
+    for (const expectedProperty of declared) {
```
montymxb, line 204: Just a note, I fixed these namings as they appeared to be inconsistent. Expected should be declared, and inferred being found, matching what's done above in the same function.

spoenemann:
We need to track the individual grammar rules that contribute to a type so we can associate error messages correctly. @pluralia could you help with this so we can cover the example mentioned by @gfontorbe?

We could also fix that case in a separate PR, though.

montymxb:
@spoenemann Like this? I can also add more context to the error message as needed.

Screen Shot 2022-05-17 at 11 19 44

spoenemann (approved):
Looks good, but I'd propose to add detail to the message:

> The non-optional property 'name' declared in type 'B' must be assigned.

gfontorbe:
> Looks good, but I'd propose to add detail to the message, along the lines of:
> 
> > The non-optional property 'name' declared in type 'B' must be assigned in this rule.

If we want to align with the error thrown in Typescript, it could look something like this:

Property 'name' is missing in parser rule 'Y' but required in type 'B'.

montymxb:
Opted to go with a slightly shortened version of the 2nd suggestion.
```
Property 'name' is missing in rule 'Y', but is required by type 'B'.
```

PR #490, opened 2022-05-10
Fix type inference for assigned actions

### LINKED ISSUE #470: Incorrect type inference for action-related rule calls
Normally, a rule call should inherit a calling rule, but this is not the case when this rule call is used for an action assignment. For example, the `IdRule` rule call is part of assignment to the `value` property of the `A` and `B` actions, and it shouldn't be inherited from `Y`. Incorrect inheritance gives the wrong type for the `value` property.

```
IdRule:
    'id' name=ID;

X infers Y:
    IdRule (
         {infer A.value=current} 'A' AProperty=NUMBER
    |    {infer B.value=current} 'B' BProperty=ID
    );
```

Langium version: 0.3.0
Package name: langium

## The current behavior
The property `value` of the generated interface `A` has type `Y`.
```
    value: Y
    AProperty: number
}
```

## The expected behavior
The property `value` of the generated interface `A` has type `IdRule`.
```
export interface Z extends AstNode {
    value: IdRule
    AProperty: number
}
```

pluralia:
* Closes #470.
* Adds a test from the issue. 
* Adds updated generated grammars after previous PRs
* Adds a new method `entriesGroupedByKey` to the `MultiMap` collection that returns pairs [key, value[]] to replace the pattern:
```
for (const key of multiMap.keys()) {
    const values = multiMap.get(key);
    ...
}
```
with
```
for (const [key, values] of multiMap.entriesGroupedByKey()) {
    ...
}

### DIFF
--- a/packages/langium/src/grammar/type-system/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/inferred-types.ts
@@ -410,7 +410,7 @@ function calculateAst(alternatives: TypeAlternative[]): InterfaceType[] {
 }
 
 function flattenTypes(alternatives: TypeAlternative[]): TypeAlternative[] {
-    const nameToAlternatives = alternatives.reduce((acc, e) => { acc.add(e.name, e); return acc; }, new MultiMap<string, TypeAlternative>());
+    const nameToAlternatives = alternatives.reduce((acc, e) => acc.add(e.name, e), new MultiMap<string, TypeAlternative>());
     const types: TypeAlternative[] = [];
 
     for (const [name, namedAlternatives] of nameToAlternatives.entriesGroupedByKey()) {
--- a/packages/langium/src/grammar/type-system/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/inferred-types.ts
@@ -201,14 +201,14 @@ function buildSuperUnions(interfaces: InterfaceType[]): UnionType[] {
             allSupertypes.add(superType, interfaceType.name);
         }
     }
-    for (const superType of allSupertypes.keys()) {
+    for (const [superType, types] of allSupertypes.entriesGroupedByKey()) {
         if (!interfaces.some(e => e.name === superType)) {
             unions.push(new UnionType(
                 superType,
                 [{
                     array: false,
                     reference: false,
-                    types: [...allSupertypes.get(superType)]
+                    types
                 }],
                 { reflection: true }
             ));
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: packages/langium/src/grammar/type-system/inferred-types.ts [hunk 1]
```diff
@@ -410,14 +410,13 @@ function calculateAst(alternatives: TypeAlternative[]): InterfaceType[] {
 }
 
 function flattenTypes(alternatives: TypeAlternative[]): TypeAlternative[] {
-    const names = new Set<string>(alternatives.map(e => e.name));
+    const nameToAlternatives = alternatives.reduce((acc, e) => { acc.add(e.name, e); return acc; }, new MultiMap<string, TypeAlternative>());
```
montymxb, line 413: It looks like you can just return `acc.add(e.name, e)` directly, as it seems to return instance of itself. Could probably shorten it to 
`(acc, e) => acc.add(e.name, e), ...`.

packages/langium/src/utils/collections.ts
```diff
@@ -148,4 +148,11 @@ export class MultiMap<K, V> {
             .flat() as any;
     }
 
+    /**
+     * Returns a stream of key, value set pairs for every key in the map.
+     */
+    entriesGroupedByKey(): Stream<[K, V[]]> {
+        return stream(this.map.entries());
+    }
```
montymxb, line 156: Nett 👍

spoenemann (commented):
Now I get invalid types with this grammar:
```
A infers TypeA:
    B ({infer TypeC.left=current} name=ID)*;

B infers TypeB:
    '(' A ')' | {infer TypeD} value=NUMBER;
```

leads to:
```typescript
export type TypeA = TypeB | TypeC;
export type TypeB = TypeA | TypeD;
```

If you prefer to work on this separately, I'm ok with creating a new issue.

spoenemann:
☝️ possible solution: merge the types and create an alias
```typescript
export type TypeA = TypeC | TypeD;
export type TypeB = TypeA;
```

pluralia:
> If you prefer to work on this separately, I'm ok with creating a new issue.

The issue is not related to actions: this grammar gives similar problems:
```
A infers TypeA:
    B C*;

B infers TypeB:
    '(' A ')' | D;
```
Should I open a new issue about it?

spoenemann:
Yes, please include both examples.

spoenemann (approved):

PR #495, opened 2022-05-13
Change TypeScript target to ES2017

### LINKED ISSUE #452: Generate ES2017 code
Our current tsconfig specifies ES6 as generation target. `async` / `await` are transformed to JS code without those constructs. We should upgrade to ES2017 to generate native `async` / `await` code.

spoenemann:
Closes #452. The main difference to the previous setting (ES6 === ES2015) is the native support for `async` functions. As result, the generated js code is much more readable and closer to the ts source.

Async functions are supported by all major browsers (https://caniuse.com/?search=ES2017). Are there any other concerns against this upgrade?

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,7 @@
 
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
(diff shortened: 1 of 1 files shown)


montymxb:
It seems like a [vast majority of browsers from a report 2 years ago (~95%) support ES2017 features](https://web.dev/publish-modern-javascript/), according to this info collected by Google. And for ES2017 it's specifically native async function support. Unless there's a special case why we cannot, it seems okay to me.

msujew (approved):
LGTM!

PR #496, opened 2022-05-13
Create TextDocument lazily, deprecate TextDocumentFactory

### LINKED ISSUE #279: Consider separating TextDocuments from LangiumDocuments
Since #275, we no longer need the TextDocument to get line / column positions from text offsets of CST nodes. We should consider decoupling LangiumDocument from TextDocument to simplify our API. We will still need the TextDocument in some LSP handlers like completion, but we can use the TextDocuments service for that.

spoenemann:
Closes #279. I did not remove the `textDocument` property from `LangiumDocument` because it's being used in numerous LSP related services. Instead I changed that property to a lazily evaluated getter. As result, a `TextDocument` is no longer created for documents that are loaded in the background (by the `WorkspaceManager`), but not opened in an editor. As soon as the user opens a DSL file in the editor, the client sends a `didOpen` notification, which triggers the creation of a `TextDocument` in the server, so the resulting `LangiumDocument` has a direct reference to that `TextDocument`.

As a side effect, this change deprecates `TextDocumentFactory` because we no longer create those documents eagerly.

### DIFF
--- a/packages/langium/src/grammar/grammar-util.ts
+++ b/packages/langium/src/grammar/grammar-util.ts
@@ -4,7 +4,6 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { TextDocument } from 'vscode-languageserver-textdocument';
 import { URI, Utils } from 'vscode-uri';
 import { TypeResolutionError } from '..';
 import * as ast from '../grammar/generated/ast';
@@ -17,7 +16,7 @@ import { streamCst } from '../utils/cst-util';
 import { escapeRegExp } from '../utils/regex-util';
 import { AstNodeDescriptionProvider } from '../workspace/ast-descriptions';
 import { AstNodeLocator } from '../workspace/ast-node-locator';
-import { documentFromText, LangiumDocument, LangiumDocuments, PrecomputedScopes } from '../workspace/documents';
+import { LangiumDocument, LangiumDocuments, PrecomputedScopes } from '../workspace/documents';
 import { createLangiumGrammarServices } from './langium-grammar-module';
 
 export type Cardinality = '?' | '*' | '+' | undefined;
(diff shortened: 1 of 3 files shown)


packages/langium/src/grammar/grammar-util.ts
```diff
@@ -398,12 +397,7 @@ export function loadGrammar(json: string): ast.Grammar {
         throw new Error('Could not load grammar from specified json input.');
     }
     const grammar = astNode as Mutable<ast.Grammar>;
-    const textDocument = TextDocument.create('memory://grammar.langium', 'langium', 0, '');
-    const document = documentFromText(textDocument, {
-        lexerErrors: [],
-        parserErrors: [],
-        value: grammar
-    });
+    const document = services.shared.workspace.LangiumDocumentFactory.fromModel(grammar, URI.parse('memory://grammar.langium'));
```
Lotes, line 400: What is 'memory://grammar.langium'? What is its purpose? Is this the grammar address in VSCode?

Another thought: Looks like a magic string. If we have that string in multiple places it becomes hard to change them, when necessary. A shared constant could be a solution.

msujew (approved):
I'm happy with the changes. LGTM!

packages/langium/src/grammar/grammar-util.ts
msujew: It has no real purpose, it actually makes this "file" non-addressable, by the rest of the system. The idea is that we load a model of the user-grammar during runtime to operate on it for parsing and LSP services. However, since everything operates on `LangiumDocument` objects, we need to wrap the grammar in a document and give it a bogus URI.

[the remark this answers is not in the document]
spoenemann: It's also used for setting the `precomputedScopes` property.

PR #498, opened 2022-05-16
Fix for Unknown Type during Type Resolution

### LINKED ISSUE #493: TypeResolutionError: Unknown type
A grammar that contains
```
A:
    'a' name=ID (otherA=[A])?;
```
rises an error: `TypeResolutionError: Unknown type`.

The error occurs in the `getTypeName` function [here](https://github.com/langium/langium/blob/main/packages/langium/src/grammar/grammar-util.ts#L297), when the input is not `AbstractType`. Handling of the example above produces a call of `getTypeName` on the `InferredType` argument.

Langium version: 0.3.0
Package name: `langium`

## Steps To Reproduce

1. Inject the `A` rule in your grammar
2. Execute `npm run langium:generate`

montymxb:
Closes #493 by introducing a couple of small adjustments, all with a focus on working with **InferredType** instances that sneak in where **AbstractType** should be.
- Adds validation in `getTypeName` for **InferredTypes**, which were previously unaccounted for
- Adds validation in `findNameAssignment` for **InferredTypes**
- Adds tests for both cases

As a side note, in the long run, it seems we should either prevent **InferredType** propagating this far, or address it by updating all corresponding functions.

### DIFF
--- a/packages/langium/test/grammar/grammar-util.test.ts
+++ b/packages/langium/test/grammar/grammar-util.test.ts
@@ -4,9 +4,38 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { createLangiumGrammarServices, findNameAssignment, getEntryRule, Grammar, isDataTypeRule, isParserRule, isTerminalRule, ParserRule, stream, terminalRegex, TerminalRule } from '../../src';
+import { createLangiumGrammarServices, findNameAssignment, getEntryRule, getTypeName, Grammar, InferredType, isDataTypeRule, isParserRule, isTerminalRule, ParserRule, stream, terminalRegex, TerminalRule } from '../../src';
 import { LangiumGrammarGrammar } from '../../src/grammar/generated/grammar';
 import { parseHelper } from '../../src/test';
+
+describe('Verify type resolution', () => {
+
+    //const services = createLangiumGrammarServices();
+
+    test('able to get type name of InferredType w/out error', async () => {
+        // previous attempt to trip the error, tmp for reference...
+        // const g =`
+        // grammar G
+        // A infers B: 
+        //     {infer q=A} name=ID;
+        // X: 'a' A;
+        // terminal ID: /[a-zA-Z_][a-zA-Z0-9_]*/;
+        // `.trim();
+
+        // const document = await parseDocument(services.grammar, g);
+        // const diagnostics = await services.grammar.validation.DocumentValidator.validateDocument(document);
+        // expect(diagnostics).toHaveLength(0);
+
+        const typeName = 'Parameter';
+        const type = {
+            name: typeName,
+            $type: 'InferredType'
+        };
+        expect(getTypeName(type as InferredType)).toBe(typeName);
+
+    });
+
+});
 
 describe('Data type rules', () => {
 
--- a/packages/langium/test/grammar/grammar-util.test.ts
+++ b/packages/langium/test/grammar/grammar-util.test.ts
@@ -4,35 +4,42 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { createLangiumGrammarServices, findNameAssignment, getEntryRule, getTypeName, Grammar, InferredType, isDataTypeRule, isParserRule, isTerminalRule, ParserRule, stream, terminalRegex, TerminalRule } from '../../src';
+import { createDefaultModule, createDefaultSharedModule, createLangiumGrammarServices, createLangiumParser, findNameAssignment, getEntryRule, Grammar, inject, IParserConfig, isDataTypeRule, isParserRule, isTerminalRule, LangiumGeneratedServices, LangiumGeneratedSharedServices, LangiumParser, LangiumServices, LangiumSharedServices, Module, ParserRule, stream, terminalRegex, TerminalRule } from '../../src';
 import { LangiumGrammarGrammar } from '../../src/grammar/generated/grammar';
 import { parseHelper } from '../../src/test';
 
+const grammarServices = createLangiumGrammarServices().grammar;
+const helper = parseHelper<Grammar>(grammarServices);
+
 describe('Verify type resolution', () => {
 
-    //const services = createLangiumGrammarServices();
+    test('verify no TypeError for InferredType in scope', async () => {
+        const c = `
+        type T = A;
+        B returns A: name=ID;
+        hidden terminal WS: /\\s+/;
+        terminal ID: /[a-zA-Z_][a-zA-Z0-9_]*/;
+        `.trim();
 
-    test('able to get type name of InferredType w/out error', async () => {
-        // previous attempt to trip the error, tmp for reference...
-        // const g =`
-        // grammar G
-        // A infers B: 
-        //     {infer q=A} name=ID;
-        // X: 'a' A;
-        // terminal ID: /[a-zA-Z_][a-zA-Z0-9_]*/;
-        // `.trim();
+        const g = (await helper(c)).parseResult.value;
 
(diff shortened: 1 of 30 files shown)


packages/langium/test/parser/langium-parser-builder.test.ts
```diff
@@ -259,6 +259,42 @@ describe('Parser calls value converter', () => {
...
 });
 
+describe('Parsing with Inferred Types', () => {
+
+    const services = createLangiumGrammarServices();
+    const parser = parseHelper<Grammar>(services.grammar);
+
+    test('verify no TypeError for InferredType in scope', async () => {
+        const c = `
+        type T = A;
```
Lotes, line 269: Where is A defined?

montymxb: That's a good point! It's not defined in this case. It can be added, however this alone was enough to throw a type error. It could be defined for this case, but the result would be the same.

Lotes: Ah! Semantic checks come later, I guess ^^*. It is just a little bit confusing to see a broken example passing the test

montymxb: Yeah it is a little strange still 😅 . I've seen some other tests that do something similar, dropping the `ID` terminal or something like that.

packages/langium/test/grammar/grammar-util.test.ts
```diff
@@ -86,6 +87,12 @@ describe('Find name assignment in parser rules', () => {
         expect(nameAssignment).toBe(undefined);
     });
 
+    test('Should be able to infer data type rule for cross ref', () => {
+        const a = rules[3];
+        const nameAssigment = findNameAssignment(a);
+        expect(nameAssigment?.feature).toBe('name');
+    });
+
 });
```
pluralia, line 97: I'd propose to add tests of `getTypeName` in `grammar-util.test.ts` too that it would have unit-tests. Then you can add "integration" tests in `inferred-types.spec.ts` + `type-validator.test.ts`.

packages/langium/test/parser/langium-parser-builder.test.ts
```diff
@@ -259,6 +259,42 @@ describe('Parser calls value converter', () => {
     });
 });
 
+describe('Parsing with Inferred Types', () => {
```
pluralia, line 262: It may be better to move these tests to `type-validator.test.ts`, since they are more related to type inference than to the parser. Here you can check existing of errors using this:
`expect(diagnostics.filter(d => d.severity === DiagnosticSeverity.Error)).toHaveLength(2);`

montymxb: So these tests actually are tied into parser generation. Generating a previously problematic grammar produces no diagnostics, but when you try to build a parser it would fail.

packages/langium/test/grammar/grammar-util.test.ts
```diff
@@ -174,4 +210,4 @@ describe('TerminalRule to regex', () => {
         }
         return terminal;
     }
-});
+});
```
pluralia, line 213: Could you add a new line?

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -78,4 +78,20 @@ describe('validate params in types', () => {
...
+        hidden terminal WS: /\\s+/;
+        terminal ID: /[a-zA-Z_][a-zA-Z0-9_]*/;
+        `.trim();
+
+        const document = await parseDocument(grammarServices, prog);
+        const diagnostics: Diagnostic[] = await grammarServices.validation.DocumentValidator.validateDocument(document);
+        expect(diagnostics.filter(d => d.severity === DiagnosticSeverity.Error)).toHaveLength(0);
+
+    });
 });
```
pluralia, line 97: Could you add a newline?

packages/langium/test/grammar/grammar-util.test.ts
```diff
@@ -86,12 +87,47 @@ describe('Find name assignment in parser rules', () => {
         expect(nameAssignment).toBe(undefined);
     });
 
+    test('Should be able to infer data type rule for cross ref', () => {
```
pluralia, line 90: I'd propose to name this test "Should be able to find a named assignment by Inferred Type".

packages/langium/test/grammar/type-system/inferred-types.spec.ts
```diff
@@ -564,6 +564,20 @@ describeTypes('inferred types with common names and actions', `
 
 });
 
+describeTypes('inferred types with common names and actions', `
```
pluralia, line 567: I'd propose to name this test group "inferred types that are used by the grammar", because the ability to use inferred rule types at the grammar type level is checked here.

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -78,4 +78,20 @@ describe('validate params in types', () => {
         expect(diagnostics.filter(d => d.severity === DiagnosticSeverity.Error)).toHaveLength(0);
     });
 
+});
+
+describe('Inferred Types Validation', () => {
```
pluralia, line 83: To keep consistency with the previous test group, I suggest to call it "validate inferred types"

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -78,4 +78,20 @@ describe('validate params in types', () => {
...
+
+    test('Should validate InferredType', async () => {
```
pluralia, line 85: Could you give more information about the test case? For example, "using inferred types in cross-reference should not throw errors"

packages/langium/test/grammar/grammar-util.test.ts
montymxb: I'll get those put back.

packages/langium/test/grammar/type-system/inferred-types.spec.ts
montymxb: Name was copied from the describe block above 😅

packages/langium/test/grammar/type-system/type-validator.test.ts
montymxb: Yep, can do

packages/langium/test/grammar/type-system/type-validator.test.ts
montymxb: Good point. Going w/ `inferred type in cross-reference should not produce an error`, given that this test is verifying the absence of error diagnostics.

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -78,4 +78,20 @@ describe('validate params in types', () => {
...
+        hidden terminal WS: /\\s+/;
+        terminal ID: /[a-zA-Z_][a-zA-Z0-9_]*/;
+        `.trim();
+
+        const document = await parseDocument(grammarServices, prog);
+        const diagnostics: Diagnostic[] = await grammarServices.validation.DocumentValidator.validateDocument(document);
+        expect(diagnostics.filter(d => d.severity === DiagnosticSeverity.Error)).toHaveLength(0);
+
+    });
+});
```
pluralia, line 97: Could you add a newline?

pluralia (approved):
It looks nice!

packages/langium/test/grammar/type-system/type-validator.test.ts
montymxb: The diff isn't showing it, but there is a newline present at the end 👍 .

Issue #462, opened 2022-04-05
Improve the error handling in langium-parser-builder.ts


sailingKieler:
Please improve the error handling in
https://github.com/langium/langium/blob/bc75d2866559224e96d632b50585624f1dffd803/packages/langium/src/parser/langium-parser-builder.ts#L142-L159

at
https://github.com/langium/langium/blob/bc75d2866559224e96d632b50585624f1dffd803/packages/langium/src/parser/langium-parser-builder.ts#L157

e.g. like `throw new Error(ruleCall.rule.error?.message ?? '<default msg>');`

In my case the linking failed because of an old node.js version (10.6) not providing `Array.prototype.flat()`.

PR #501, opened 2022-05-17
Improve error handling

Lotes:
* Added a helper function for asserting type emptiness (never)
* Added better error message when was needed

### DIFF
--- a/packages/langium/src/utils/errors.ts
+++ b/packages/langium/src/utils/errors.ts
@@ -1,4 +1,10 @@
-import { CstNode } from "..";
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { CstNode } from '..';
 
 export class ErrorWithLocation extends Error {
     constructor(node: CstNode|undefined, message: string) {
@@ -6,6 +12,6 @@ export class ErrorWithLocation extends Error {
     }
 }
 
-export function assertUnreachable(x: never): never {
-    throw new Error("Error! Please fulfill the logic also for the other types in the inheritence hierarchy.");
+export function assertUnreachable(_: never): never {
+    throw new Error('Error! Please fulfill the logic also for the other types in the inheritence hierarchy.');
 }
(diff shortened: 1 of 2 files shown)


Lotes:
See #462

packages/langium/src/utils/errors.ts
```diff
@@ -0,0 +1,17 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { CstNode } from '..';
```
msujew, line 7: Minor: Please don't import from the index (could easily lead to cyclic dependency issues), but from the actual file.

packages/langium/src/utils/errors.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export class ErrorWithLocation extends Error {
+    constructor(node: CstNode|undefined, message: string) {
+        super(`${message} at ${node?.range.start.line || 0}:${node?.range.start.character || 0}`);
+    }
+}
+
+export function assertUnreachable(_: never): never {
+    throw new Error('Error! Please fulfill the logic also for the other types in the inheritence hierarchy.');
```
msujew, line 16: Minor: The error message is a bit too specific for my taste. This error can happen anytime we get an unexpected/unhandled value in an `if-else/switch` block, so I imagine something like this:

```suggestion
    throw new Error('Error! The input value was not handled.');
```

Lotes: `assertUnreachable` is a compile-time trick and will lead to an error, when the parameter is not the empty type. So actually we never reach this point.

packages/langium/src/utils/errors.ts
```diff
@@ -0,0 +1,17 @@
...
+}
```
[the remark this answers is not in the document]
msujew: GitHub showed me that the file didn't end on a newline. But you seemed to have fixed it already 👍

packages/langium/src/utils/errors.ts
msujew: I see, alright then.

msujew (approved):
Thanks, LGTM 👍

PR #502, opened 2022-05-17
Catch errors during module resolution

### LINKED ISSUE #463: Add error handling in dependency injection impl
If
```javascript
value(injector)
```

of
https://github.com/langium/langium/blob/bc75d2866559224e96d632b50585624f1dffd803/packages/langium/src/dependency-injection.ts#L86

in
https://github.com/langium/langium/blob/bc75d2866559224e96d632b50585624f1dffd803/packages/langium/src/dependency-injection.ts#L83-L87

fails, `obj[prop] === __requested__` holds afterwards, and in a subsequent attempt the following error is thrown, which can be quite misleading:
https://github.com/langium/langium/blob/bc75d2866559224e96d632b50585624f1dffd803/packages/langium/src/dependency-injection.ts#L79-L81

I suggest to replace
https://github.com/langium/langium/blob/bc75d2866559224e96d632b50585624f1dffd803/packages/langium/src/dependency-injection.ts#L86

by
```javascript
    try {
        obj[prop] = (typeof value === 'function') ? value(injector) : _inject(value, injector);
    } catch (error) {
        obj[prop] = undefined;
        console.error(error);
        throw error;
    }
```

Lotes:
@sailingKieler I copy-pasted your solution.

Closes #463

Remark: the affected file is really hard to read.

### DIFF
--- a/packages/langium/src/dependency-injection.ts
+++ b/packages/langium/src/dependency-injection.ts
@@ -77,7 +77,7 @@ const __requested__ = Symbol();
 function _resolve<I, T>(obj: any, prop: string | symbol | number, module: Module<I, T>, injector: I): T[keyof T] | undefined {
     if (prop in obj) {
         if (obj[prop] instanceof Error) {
-            throw new Error("Construction failure. Please make sure that your dependencies are constructable.", {cause: obj[prop]});
+            throw new Error('Construction failure. Please make sure that your dependencies are constructable.', {cause: obj[prop]});
         }
         if (obj[prop] === __requested__) {
             throw new Error('Cycle detected. Please make "' + String(prop) + '" lazy. See https://langium.org/docs/di/cyclic-dependencies');
--- a/packages/langium/src/dependency-injection.ts
+++ b/packages/langium/src/dependency-injection.ts
@@ -76,6 +76,9 @@ const __requested__ = Symbol();
  */
 function _resolve<I, T>(obj: any, prop: string | symbol | number, module: Module<I, T>, injector: I): T[keyof T] | undefined {
     if (prop in obj) {
+        if (obj[prop] instanceof Error) {
+            throw new Error("Construction failure. Please make sure that your dependencies are constructable.", {cause: obj[prop]});
+        }
         if (obj[prop] === __requested__) {
             throw new Error('Cycle detected. Please make "' + String(prop) + '" lazy. See https://langium.org/docs/di/cyclic-dependencies');
         }
@@ -86,8 +89,7 @@ function _resolve<I, T>(obj: any, prop: string | symbol | number, module: Module
         try {
             obj[prop] = (typeof value === 'function') ? value(injector) : _inject(value, injector);
         } catch (error) {
-            obj[prop] = undefined;
-            console.error(error);
+            obj[prop] = error;
             throw error;
         }
         return obj[prop];
--- a/packages/langium/src/dependency-injection.ts
+++ b/packages/langium/src/dependency-injection.ts
@@ -83,7 +83,13 @@ function _resolve<I, T>(obj: any, prop: string | symbol | number, module: Module
     } else if (prop in module) {
         const value: Module<I, T[keyof T]> | ((injector: I) => T[keyof T]) = module[prop as keyof T];
         obj[prop] = __requested__;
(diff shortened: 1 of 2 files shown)


sailingKieler:
LGTM
@Lotes did you try out implementation by letting some services crash on purpose during the initialization, e.g. the name provider in the domain model example: https://github.com/langium/langium/blob/2392ddd9854aca45047a459c85a3cf9eb61e24bd/examples/domainmodel/src/language-server/domain-model-module.ts#L22-L34

Lotes:
Yes. I added some crashes by throwing errors.

sailingKieler (approved):
Fine for me

Lotes:
@sailingKieler Sorry. But I realized that there is a way to test this scenario. In the case of an error, I am saving the error for the after service. If then some other service is requesting it in a follow-up call, we get a "construction failure" where the cause is the saved error.

packages/langium/src/dependency-injection.ts
```diff
@@ -76,14 +76,22 @@ const __requested__ = Symbol();
...
         }
         return obj[prop];
     } else if (prop in module) {
         const value: Module<I, T[keyof T]> | ((injector: I) => T[keyof T]) = module[prop as keyof T];
         obj[prop] = __requested__;
-        obj[prop] = (typeof value === 'function') ? value(injector) : _inject(value, injector);
+        try {
+            obj[prop] = (typeof value === 'function') ? value(injector) : _inject(value, injector);
+        } catch (error) {
+            obj[prop] = error;
```
spoenemann, line 92: Please add an `instanceof Error` check here as well (theoretically you can throw anything in JS).

Lotes: ok

spoenemann (approved):

packages/langium/test/dependency-injection.test.ts
```diff
@@ -59,7 +59,7 @@ describe('A non-cyclic dependency', () => {
     });
 
     test('should be result on an idempotent api call', () => {
-        const api = inject({ dep: () => ({ }) });
+        const api = inject({ dep: () => ({}) });
         expect(api.dep).toBe(api.dep);
```
sailingKieler, line 63: This is a tautology, isn't it? Does that make any sense?

packages/langium/src/dependency-injection.ts
```diff
@@ -76,14 +76,22 @@ const __requested__ = Symbol();
...
         return obj[prop];
     } else if (prop in module) {
         const value: Module<I, T[keyof T]> | ((injector: I) => T[keyof T]) = module[prop as keyof T];
         obj[prop] = __requested__;
-        obj[prop] = (typeof value === 'function') ? value(injector) : _inject(value, injector);
+        try {
+            obj[prop] = (typeof value === 'function') ? value(injector) : _inject(value, injector);
+        } catch (error) {
+            obj[prop] = error instanceof Error ? error : undefined;
+            throw error;
```
sailingKieler, line 93: With this solution the "Construction failure."-Message would be thrown on the subsequent attempt after a previous failure, right?

Why not just:
```suggestion
            obj[prop] =  undefined;
            if (error instanceof Error)
                throw new Error('Construction failure. Please make sure that your dependencies are constructable.', {cause: obj[prop]});
            else
                throw error;
```

packages/langium/test/dependency-injection.test.ts
Lotes: I think it is meant as: two subsequent calls, return the same result.

packages/langium/src/dependency-injection.ts
Lotes: Exactly! Imagine you would throw the original one: You would have the old stack trace, not the new one. It is another trace leading to this construction attempt.
Do you think there is a better option?

packages/langium/test/dependency-injection.test.ts
sailingKieler: At the first glance I expected a test expression like
```
expect(api.dep()).toBe(api.dep());
```
... i.e. `api.dep()` instead of `api.dep`.

I overlooked the fact that `api` is the proxy object created by `inject(...)`, and that the supplier function `dep` is called internally once the property `dep` is requested ... what this PR is more or less about ... my bad!

PR #503, opened 2022-05-17
Added tests for type consistency validations

### LINKED ISSUE #447: Add type consistency validation for actions
It's necessary to add validation for actions that use declared types. For example, interface `X` has a property `name`, but type `X` from action has a property `count` -- validation should forbid it.
```
interface X {
    name: string
}

MyXRule:
    {X} count=NUMBER;
```

Lotes:
Closes #447. The problem was already solved, which means there are validation errors. I added some tests instead.


Lotes:
Ah! I see it is not solved for Actions ^^* Let me fix that...

Issue #425, opened 2022-02-28
Add editor features for property assignments


spoenemann:
In cases like
```
MyRule returns A: ...
```
or
```
MyRule: {A} ...
```
where `A` is either a declared type (`type` or `interface`) or a rule (generating an inferred type `A`), we should offer editor features for the grammar language:
 * Go to definition
 * Document Highlight
 * Document Symbols
 * Find References
 * More?

These require custom implementations because we don't have cross-references here, so the default services don't work.

spoenemann:
This won't be necessary after doing #440.

spoenemann:
Actually we still need to add the editor features mentioned above for property assignments in case the assigned property is declared in an `interface`.

spoenemann:
Fixed with https://github.com/langium/langium/pull/505

PR #505, opened 2022-05-18
Added GoTo solver for properties in parser rules with return type

gfontorbe:
Partially addresses #425 

Adds GoTo functionality for properties in a parser rule referring to a property declared in an interface (only on parser rules that explicitly return a declared type).

How to test:
```
interface X {
    name: string
}

A returns X : name=ID;
```
Cmd+Click on `name` in the parser rule 'A' should send the cursor to the `name` property in interface X.

@msujew I finished the implementation before seeing your new commit with the new CstUtils. Maybe we could make use of the getParentChain function here.


packages/langium/src/grammar/langium-grammar-goto.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    constructor(services: LangiumServices) {
+        super(services);
+    }
+
+    protected findTargetNode(sourceCstNode: CstNode, targetCstNodes: Array<{ source: CstNode; target: CstNode; targetDocument: LangiumDocument; }>): void {
+        if (isAssignment(sourceCstNode.element)){
+            const parserRule = findParentParserRuleCstNode(sourceCstNode);
+            // check that the parser rule has a return type
+            if (parserRule && isParserRule(parserRule.element) && parserRule.element.returnType) {
```
msujew, line 24: Suggestion: I believe that this can be greatly simplified. You can just use `getContainerOfType(sourceCstNode.element, isParserRule)` to get the parent parser rule of the assignment. I think that eliminates the need for `findParentParserRuleCstNode` as well.

packages/langium/src/lsp/goto.ts
```diff
@@ -60,6 +54,17 @@ export class DefaultGoToResolverProvider implements GoToResolver {
...
+
+    protected findTargetNode(sourceCstNode: CstNode, targetCstNodes: Array<{ source: CstNode, target: CstNode, targetDocument: LangiumDocument }>): void {
+        const targetNode = this.references.findDeclaration(sourceCstNode);
+        if (targetNode?.element) {
+            const targetDoc = getDocument(targetNode?.element);
+            if(targetNode && targetDoc) {
+                targetCstNodes.push({ source: sourceCstNode, target: targetNode, targetDocument: targetDoc });
+            }
+        }
+    }
```
msujew, line 66: Suggestion: I think that this should return a target node, rather than pushing it to the input parameter. Would be a bit cleaner design. Also, you should probably extract this type into an interface:
```ts
export interface GoToLink { 
  source: CstNode
  target: CstNode
  targetDocument: LangiumDocument 
}
```

```suggestion
    protected findLink(source: CstNode): GoToLink | undefined {
        const target= this.references.findDeclaration(source);
        if (target?.element) {
            const targetDocument = getDocument(targetNode?.element);
            if (target && targetDocument) {
                return { source, target, targetDocument};
            }
        }
        return undefined;
    }
```

packages/langium/src/utils/cst-util.ts
```diff
@@ -6,6 +6,7 @@
 
 import { IToken } from '@chevrotain/types';
 import { Range } from 'vscode-languageserver';
+import { isParserRule } from '..';
```
msujew, line 9: Minor: Don't import from the index.

packages/langium/src/grammar/langium-grammar-goto.ts
```diff
@@ -0,0 +1,47 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { findParentParserRuleCstNode, isAssignment, isInterface, isParserRule } from '..';
```
msujew, line 7: Minor: Don't import from the index.

packages/langium/src/lsp/goto.ts
gfontorbe: Makes a lot of sense

packages/langium/src/grammar/lsp/langium-grammar-goto.ts
```diff
@@ -0,0 +1,43 @@
...
+import { DefaultGoToResolverProvider, GoToLink } from '../../lsp';
+import { LangiumServices } from '../../services';
+import { CstNode } from '../../syntax-tree';
+import { getContainerOfType, getDocument } from '../../utils/ast-util';
+
+export class LangiumGrammarGoToResolver extends DefaultGoToResolverProvider {
+
+    constructor(services: LangiumServices) {
+        super(services);
+    }
```
msujew, line 17: Minor: This constructor isn't actually necessary. The constructor of the parent type will be inherited, if nothing else is specified.

packages/langium/src/grammar/lsp/langium-grammar-goto.ts
```diff
@@ -0,0 +1,43 @@
...
+                const returnType = parserRule.returnType;
+                const returnTypeCstNode = returnType.ref?.$cstNode;
+                if (returnTypeCstNode && isInterface(returnTypeCstNode.element)){
+                    const sourcePropertyName = source.element.feature;
+                    returnTypeCstNode.element.attributes.forEach(attribute => {
+                        // find attribute corresponding to the property
+                        if (attribute.name === sourcePropertyName) {
+                            const target = attribute.$cstNode!;
+                            const targetDocument = getDocument(returnTypeCstNode.element);
+                            goToLink = { source, target, targetDocument };
```
msujew, line 34: Suggestion: We should target the name cst node directly.
```suggestion
                            const target = findNodeForFeature(attribute.$cstNode, 'name');
                            if (target) {
                                const targetDocument = getDocument(returnTypeCstNode.element);
                                goToLink = { source, target, targetDocument };
                            }
```

msujew (commented):
Looking quite good already. Just another two remarks an then it's ready to go 👍

packages/langium/src/grammar/lsp/langium-grammar-goto.ts
```diff
@@ -0,0 +1,41 @@
...
+                if (returnTypeCstNode && isInterface(returnTypeCstNode.element)){
+                    const sourcePropertyName = source.element.feature;
+                    returnTypeCstNode.element.attributes.forEach(attribute => {
+                        // find attribute corresponding to the property
+                        if (attribute.name === sourcePropertyName) {
+                            const target = findNodeForFeature(attribute.$cstNode, 'name');
+                            if (target) {
+                                const targetDocument = getDocument(returnTypeCstNode.element);
+                                goToLink = { source, target, targetDocument };
+                            }
```
Lotes, line 32: Five nested IFs are too much. I recommend smaller functions. A rule of thumb could be to put the commented chunks into their own methods. The name is equal to the content of the comment.

Lotes:
Are there no tests? How can we be sure that it works after some changes in the future?

montymxb:
Checked out the behavior, looks nice!
jump-to-ref

packages/langium/src/grammar/lsp/langium-grammar-goto.ts
pluralia: As an alternative solution to reduce the nested level, you can replace the construction
```
if (condition) {
  return X;
}
return Y;
```
with
```
if (!condition) return Y;
return X;
```
The method reduces 2 levels in case of `findLink`.

packages/langium/src/lsp/goto.ts
```diff
@@ -44,12 +50,9 @@ export class DefaultGoToResolverProvider implements GoToResolver {
             const cst = rootNode.$cstNode;
```
spoenemann, line 50: Two lines above (can't comment there directly):
```typescript
const targetCstNodes: Array<GoToLink> = [];
```

Lotes:
@gfontorbe I do not understand it completely, because some semantics are still missing in my mind, but the code is definetly more readable now. Before this change I had no chance to understand anything, because there were a lot of variables. Short functions with good names are a win for the understandibility.

packages/langium/test/lsp/go-to/goto-definition.test.ts
```diff
@@ -8,6 +8,12 @@ import { createLangiumGrammarServices } from '../../../src';
 import { expectGoToDefinition } from '../../../src/test';
 import { expectFunction } from '../../fixture';
 
+/**
+ * Represents a grammar file
+ *
+ * `index` <|> represents the position of the curser where the GoTo Request is executed
```
Lotes, line 14: it is cursOr, typo.

msujew (approved):
Thanks, I'm happy with the changes!

PR #506, opened 2022-05-19
A small test framework for checking diagnostics

### LINKED ISSUE #414: Validation Test Helper
### Discussed in https://github.com/langium/langium/discussions/413

<div type='discussions-op-text'>

<sup>Originally posted by **MathiasSJacobsen** February 17, 2022</sup>
Is there an example that exists to demonstrate how to test the custom validation rules written for the language? Could only find #340, but that was only testing the parsing. 
</div>

Like in xText where utility class ValidationTestHelper allows to test the custom validation rules written for the language. 

```java
@Inject ParseHelper<Domainmodel> parseHelper
 
 @Inject ValidationTestHelper validationTestHelper
 
 @Test
 def testValidModel() {
     val entity = parseHelper.parse(
         "entity MyEntity {
             parent: MyEntity
         }")
     validationTestHelper.assertNoIssues(entity)
 } 
 
 @Test
 def testNameStartsWithCapitalWarning() {
     val entity = parseHelper.parse(
         "entity myEntity {
             parent: myEntity
         }")
     validationTestHelper.assertWarning(entity,
         DomainmodelPackage.Literals.ENTITY,
         DomainmodelValidator.INVALID_NAME,
         "Name should start with a capital"
     )
 }
```

Should be nice to have this in Langium as well.

Lotes:
I saw the task about `ValidationTestHelper` named #414 .
I used the verb `expect` instead of `assert`, since we use Jest with `expect()`.
The filters for diagnostic determination can be extended step by step. I have one for the message and one for the severity.

Closes #414 . @msujew what do you think about this approach?

### DIFF
--- a/packages/langium/src/test/langium-test.ts
+++ b/packages/langium/src/test/langium-test.ts
@@ -5,7 +5,7 @@
  ******************************************************************************/
 
 import {
-    CompletionItem, DocumentSymbol, MarkupContent, Range, TextDocumentIdentifier, TextDocumentPositionParams
+    CompletionItem, Diagnostic, DiagnosticSeverity, DocumentSymbol, MarkupContent, Range, TextDocumentIdentifier, TextDocumentPositionParams
 } from 'vscode-languageserver';
 import { URI } from 'vscode-uri';
 import { LangiumServices } from '../services';
@@ -187,4 +187,61 @@ function replaceIndices(base: ExpectedBase): { output: string, indices: number[]
     }
 
     return { output: input, indices, ranges: ranges.sort((a, b) => a[0] - b[0]) };
+}
+
+export function validationHelper<T extends AstNode = AstNode>(services: LangiumServices): (input: string) => Promise<Diagnostic[]> {
+    const parse = parseHelper(services);
+    return async (input) => {
+        const document = await parse(input);
+        return await services.validation.DocumentValidator.validateDocument(document);
+    };
+}
+
+export type Predicate<T> = (arg: T) => boolean;
+
+export interface DiagnosticFilters {
+    severity: DiagnosticSeverity;
+    message: RegExp | string;
+}
+
+export type DiagnosticFilterOptions = Partial<DiagnosticFilters>;
+
+function filterByOptions(diagnostics: Diagnostic[], filterOptions?: DiagnosticFilterOptions) {
+    const options = filterOptions || {};
+    const filters: Predicate<Diagnostic>[] = [];
+    if (options.severity) {
+        filters.push(d => d.severity === options.severity);
+    }
+    if (options.message) {
+        if (typeof options.message === 'string') {
+            filters.push(d => d.message === options.message);
+        } else if (options.message instanceof RegExp) {
+            const regexp = options.message as RegExp;
+            filters.push(d => regexp.test(d.message));
+        }
+    }
+    return diagnostics.filter(diag => filters.every(holdsFor => holdsFor(diag)));
+}
+
+export function expectNoIssues(diagnostics: Diagnostic[], filterOptions?: DiagnosticFilterOptions) {
+    const filtered = filterByOptions(diagnostics, filterOptions);
+    expect(filtered).toHaveLength(0);
+}
+
+export function expectIssue(diagnostics: Diagnostic[], filterOptions?: DiagnosticFilterOptions) {
+    const filtered = filterByOptions(diagnostics, filterOptions);
+    expect(filtered).not.toHaveLength(0);
+}
+
+export function expectError(diagnostics: Diagnostic[], message: string | RegExp) {
+    expectIssue(diagnostics, {
+        message,
+        severity: DiagnosticSeverity.Error
+    });
+}
+export function expectWarning(diagnostics: Diagnostic[], message: string | RegExp) {
+    expectIssue(diagnostics, {
+        message,
+        severity: DiagnosticSeverity.Warning
+    });
 }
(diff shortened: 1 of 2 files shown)


msujew:
@Lotes Looks like a good start. The only issue I have is that this method to test for diagnostics is ambiguous. Only testing for the message and severity leaves us open to multiple AST nodes triggering an validation, without being able to check which AST node really triggered the validation:

```ts
const x = 42;
const x = 1337;
```

Should show a validation on both `x` nodes (due to being duplicate names), but we can only check that there's a diagnostic, not to which node it applies. You should be able to determine the AST node by looking at the range of the diagnostic and use the `findLeafNodeAtOffset` method.

Lotes:
@msujew True, but I have no example test for this use case yet. But I could add a condition to the existing ones, of course. How should the check look like? I have some ideas from Xtext.
```typescript
expectIssue(diagnostics, {
  // by AST type
  $astType: string;

  // by text offset+length
  offset: number;
  length: number;

  //or by range object
  within: Range; //=TextRange.searchFirst("x = 42")
                 //or =TextRange.line(1)
  // by diagnostic code
  code: string;
});
```

msujew:
@Lotes I would have designed the `ExpectDiagnosticOptions` something like this:

```ts
export type ExpectDiagnosticOptions<T extends AstNode = AstNode> = { 
    severity?: Severity
    code?: string
    message?: string | RegExp 
} & (ExpectDiagnosticAstOptions<T>  | ExpectDiagnosticRangeOptions | ExpectDiagnosticOffsetOptions);

export interface ExpectDiagnosticAstOptions<T extends AstNode> {
    node: T
    property?: { name: Properties<T>, index?: number } // Optionally, which property of the node shows the diagnostic, and the index if it is an array
}

export interface ExpectDiagnosticRangeOptions {
    range: Range // The exact range the diagnostic appears
}

export interface ExpectDiagnosticOffsetOptions {
    offset: number
    length: number
}
```

packages/langium/src/test/langium-test.ts
```diff
@@ -188,3 +189,123 @@ function replaceIndices(base: ExpectedBase): { output: string, indices: number[]
...
+
+function isLeftOrEqualFrom(lhs: Position, rhs: Position) {
+    return lhs.line < rhs.line || lhs.line === rhs.line && lhs.character <= rhs.character;
+}
+
+function containsRange(outer: Range, inner: Range): boolean {
+    const start = isLeftOrEqualFrom(outer.start, inner.start);
+    const end = isLeftOrEqualFrom(inner.end, outer.end);
+    return start && end;
+}
```
msujew, line 242: Suggestion: I would replace the `containsRange` function with a simple `isRangeEqual` function. The issue is that we can't safely rely on range, if the node in question (or range, or offset) contains another node with a validation error. For example in the domain model example, we could have something like this:

```
entity A { // We want to check for a validation error on this element
  entity B { // But the check also completes successfully, if the error appears on this element
  }
}
```

This is the reason why we need the `property` property of the interface in the first place; to correctly identify the **exact** range of the diagnostic.

msujew (commented):
I really like this one! There's one issue which I would like to resolve before merging this (see below), but it looks quite good already 👍

packages/langium/src/test/langium-test.ts
```diff
@@ -231,14 +231,11 @@ export interface ExpectDiagnosticOffsetOptions {
...
-function isLeftOrEqualFrom(lhs: Position, rhs: Position) {
-    return lhs.line < rhs.line || lhs.line === rhs.line && lhs.character <= rhs.character;
-}
-
-function containsRange(outer: Range, inner: Range): boolean {
-    const start = isLeftOrEqualFrom(outer.start, inner.start);
-    const end = isLeftOrEqualFrom(inner.end, outer.end);
-    return start && end;
+function isRangeEqual(lhs: Range, rhs: Range): boolean {
+    return lhs.start.character == rhs.start.character
```
msujew, line 235: You should install the `eslint` extension for vscode to catch those during dev-time.

packages/langium/src/test/langium-test.ts
Lotes: I could not find any deep comparison utility, so I implemented it.

packages/langium/src/test/langium-test.ts
```diff
@@ -188,3 +189,120 @@ function replaceIndices(base: ExpectedBase): { output: string, indices: number[]
...
+        && lhs.end.character === rhs.end.character
+        && lhs.end.line === rhs.end.line;
+}
+
+function filterByOptions<T extends AstNode = AstNode, N extends AstNode = AstNode>(validationResult: ValidationResult<T>, options: ExpectDiagnosticOptions<N>) {
+    const filters: Array<Predicate<Diagnostic>> = [];
+    if ('node' in options) {
+        const node = 'property' in options
+            ? findNodeForFeature(options.node.$cstNode, options.property!.name, options.property!.index)
+            : options.node.$cstNode;
```
msujew, line 246: This can actually fail, hence why TypeScript complains about `property` being possibly undefined: `'x' in { x: undefined }` returns true, since a property `x` exists, even though it is undefined.

Also, naming this also `node` is quite ambiguous, since we have one ast node and a cst node.

```suggestion
        const cstNode = options.property
            ? findNodeForFeature(options.node.$cstNode, options.property.name, options.property.index)
            : options.node.$cstNode;
```

msujew (approved):
Thanks, looks great!

PR #510, opened 2022-05-20
Validate interfaces structurally instead of nominally

msujew:
Closes https://github.com/langium/langium/issues/492

Simply apply the same logic as TypeScript to interfaces. Everything is checked structurally, instead of nominally.

During testing I noticed other issues with the type validation. I will fix them later, as these issues are move involved.

### DIFF
--- a/packages/langium/src/grammar/type-system/type-validator.ts
+++ b/packages/langium/src/grammar/type-system/type-validator.ts
@@ -25,7 +25,7 @@ export function validateTypesConsistency(grammar: Grammar, accept: ValidationAcc
     }
 
     // Report missing assignments for required properties in offending nodes
-    function applyMissingAssignmentErrorToRuleNodes(nodes: readonly ParserRule[], typeName: string, accept: ValidationAcceptor): (propertyName: string, errorMessage: string) => void {
+    function applyMissingAssignmentErrorToRuleNodes(nodes: readonly ParserRule[], typeName: string): (propertyName: string, errorMessage: string) => void {
         return (propertyName: string, errorMessage: string) => {
             nodes.forEach(node => {
                 const assignments = extractAssignments(node.alternatives);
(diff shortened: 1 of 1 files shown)


packages/langium/src/grammar/type-system/type-validator.ts
```diff
@@ -217,26 +340,14 @@ function checkPropertiesConsistency(inferred: Property[], declared: Property[],
...
-            errorToRuleNodes(`A property '${expectedProperty.name}' is expected`);
+    for (const [property, expectedProperties] of declared.entriesGroupedByKey()) {
+        const foundProperty = inferred.get(property);
+        if (foundProperty.length === 0 && !expectedProperties.some(e => e.optional)) {
+            errorToRuleNodes(`A property '${property}' is expected`);
         }
     }
 }
 
-function checkSuperTypesConsistency(inferred: string[], declared: string[], errorToRuleNodes: (error: string) => void): void {
```
pluralia, line 228: Removing super types consistency check doesn't allow to notice cases, when parser rules hierarchy doesn't repeat the types hierarchy. The issue example has the types hierarchy: `Foo` <- `Event` and the rules hierarchy: `Event`.
```
interface Foo {
    name: string
}

interface Event extends Foo {}

Event returns Event:
    name=ID;
```
Maybe we don't always want to have a "full match," but it might be a good idea to leave this check and enable it with some kind of flag.

pluralia, line 228: Removing super types consistency check doesn't allow to notice cases, when parser rules hierarchy doesn't repeat the types hierarchy. The issue example has the types hierarchy: `Foo` <- `Event` and the rules hierarchy: `Event`.
```
interface Foo {
    name: string
}

interface Event extends Foo {}

Event returns Event:
    name=ID;
```
Maybe we don't always want to have a "full match," but it might be a good idea to leave this check and enable it with some kind of flag.

msujew: Yes, this is actually the idea behind this PR. We don't want to restrict users to have a static type hierarchy which is closely linked to the parser rule hierarchy. Instead, they should be free to structure their meta model how they want. The only limitation should be that the resulting parsed object matches the meta model structure.

packages/langium/src/grammar/type-system/type-validator.ts
```diff
@@ -25,7 +25,7 @@ export function validateTypesConsistency(grammar: Grammar, accept: ValidationAcc
     }
 
     // Report missing assignments for required properties in offending nodes
-    function applyMissingAssignmentErrorToRuleNodes(nodes: readonly ParserRule[], typeName: string, accept: ValidationAcceptor): (propertyName: string, errorMessage: string) => void {
+    function applyMissingAssignmentErrorToRuleNodes(nodes: readonly ParserRule[], typeName: string): (propertyName: string, errorMessage: string) => void {
```
montymxb, line 28: Ah, good catch! I had this function defined outside of `validateTypesConsistency` to start with, and forgot to remove the redundant param once it was back in.

spoenemann (approved):

PR #516, opened 2022-05-31
Enable semantic highlighting for root element

msujew:
Closes https://github.com/langium/langium/issues/515

Since `streamAllContents` does not include the root element, we have to highlight it "manually".


spoenemann (approved):
General question: should the tree iterator include the root? I see this "apply to root, then content" scheme quite often.

spoenemann:
Alternative: offer a `streamAST` function that prepends the root to `streamAllContents`.

msujew:
@spoenemann I noticed that as well. Maybe something for another PR.

spoenemann:
I'll make a proposal.

PR #517, opened 2022-06-01
Add type consistency validation for actions

### LINKED ISSUE #447: Add type consistency validation for actions

Lotes:
Closes #447 .

It is one attempt. I tried to reuse the existing as much as possible. I am happy to see suggestions :D.

I am not totally convinced that this covers all possible cases.

I marked this PR as draft FTM.

### DIFF
--- a/packages/langium/src/grammar/type-system/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/inferred-types.ts
@@ -9,6 +9,7 @@ import { AbstractElement, Action, Alternatives, Assignment, Group, isAction, isA
 import { stream } from '../../utils/stream';
 import { AstTypes, distictAndSorted, Property, PropertyType, InterfaceType, UnionType } from './types-util';
 import { MultiMap } from '../../utils/collections';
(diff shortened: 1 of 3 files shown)


msujew (commented):
I'm not too sure about this approach in general, due to the inherent complexity of some grammar rules. For example, rules such as these:

```
X: name=ID ({Y} count=number)?;

X: name=ID ({Y} y=STRING | {Z} z=ID);
```

won't be correctly validated by your approach.

For it to work correctly, you will need to use the `allTypes` array that is constructed in the `collectInferredTypes` method (you will likely need to add the rule that constructed each `TypeAlternative` object to the object itself). That way you can gather all properties that are assigned to an element of a specific type per rule, which you can then match against the expected properties.

Lotes:
Let me close this. I will keep the branch locally (and remotely for a while).

PR #519, opened 2022-06-02
Configurable folder excludes

### LINKED ISSUE #344: Introduce runtime language configuration and UI settings
Some aspects of a language and its runtime behavior need to be configured by the user depending on the particular runtime environment and/or project setup.
One use-case is e.g. `DefaultIndexManager.skipFolder` where we need to exclude some folders from traversing during indexing.

The idea is to have a service that provides configurations for special property or a key (compared to Xtext PreferenceStoreAccess) at runtime.

When using in an IDE e.g. VSCode extension, we will need a corresponding implementation. In case of VS Code this can be user or workspace specific:
Bildschirmfoto 2022-01-07 um 12 56 04

For langium-vscode it would be nice to have a boilerplate code generated if possible.

dhuebner:
Introduces a new configuration service.
Configuration service listens for changes of all registered Langium languages. Configurations are updated when user changes the settings (user/workspace). Other services can request configurations using:
`getConfiguration(language: string, configuration: string)` 

Each language will have his own settings section, section name is the `<languageId>-settings`. 
 
Exclude pattern syntax is handled by node-ignore module [kaelzhang/node-ignore](https://github.com/kaelzhang/node-ignore)

For Langium grammar language it allows configuration of excluded folders using vscode settings.

Fixes #344


packages/langium/src/grammar/workspace/grammar-workspace-manager.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    protected includeEntry(entry: FileSystemNode, fileExtensions: string[]): boolean {
+        if (this.matcher) {
+            const parentWsFolder = this.folders?.map(folder => folder.uri).find(folderUri => entry.uri.toString().startsWith(folderUri));
+            let entryPath = entry.uri.path;
+            if (parentWsFolder) {
+                // create path relative to workspace folder root: /user/foo/workspace/entry.txt -> entry.txt
+                const wsFolderPath = URI.parse(parentWsFolder).path + '/';
+                entryPath = entryPath.replace(wsFolderPath, '');
```
spoenemann, line 49: The string operations `startsWith` and `replace` seem unsafe to me: they can lead to false matches. Following my recommendation of moving this file to `sprotty-vscode`, you can use the Node.js `path` module for path operations.

spoenemann (commented):
Great, thanks! I have a few proposals for reorganizing.

packages/langium/src/grammar/workspace/grammar-workspace-manager.ts
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * gitignore separated exclusion patterns, separated by comma
+     */
+    ignorePatterns: string
+}
+
+export class LangiumGrammarWorkspaceManager extends DefaultWorkspaceManager {
+
+    protected readonly configurationProvider: ConfigurationProvider;
+    protected matcher: Ignore | null;
```
spoenemann, line 28: Usually we avoid the usage of `null` and use `undefined` instead.

msujew:
@dhuebner There's a compilation error in your latest changes

dhuebner:
@msujew 
Yeah.. There were some other issues as well :-(

packages/langium/src/grammar/workspace/grammar-workspace-manager.ts
dhuebner: Changed

packages/langium/package.json
```diff
@@ -39,7 +39,8 @@
     "chevrotain": "^9.1.0",
     "vscode-languageserver": "^8.0.1",
     "vscode-languageserver-textdocument": "^1.0.5",
-    "vscode-uri": "^3.0.2"
+    "vscode-uri": "^3.0.2",
+    "ignore":"~5.2.0"
```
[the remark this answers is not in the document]
dhuebner: Moved to langium-vscode

packages/langium/src/lsp/language-server.ts
```diff
@@ -291,6 +292,15 @@ export function addSemanticTokenHandler(connection: Connection, services: Langiu
         services
     ));
 }
+export function addConfigurationChangeHandler(connection: Connection, services: LangiumSharedServices): void {
+    connection.onDidChangeConfiguration(change => {
+        if (change.settings) {
+            Object.keys(change.settings).forEach(section => {
+                services.workspace.ConfigurationProvider.updateConfiguration(section, change.settings[section]);
```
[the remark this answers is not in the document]
dhuebner: Now `updateConfiguration ` gets the whole `change`

packages/langium-vscode/src/language-server/grammar-workspace-manager.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    protected includeEntry(entry: FileSystemNode, fileExtensions: string[]): boolean {
+        if (this.matcher) {
+            const parentWsFolder = this.folders?.map(folder => folder.uri).find(folderUri => entry.uri.toString().startsWith(folderUri));
+            let entryPath = entry.uri.path;
+            if (parentWsFolder) {
+                // create path relative to workspace folder root: /user/foo/workspace/entry.txt -> entry.txt
+                const wsFolderPath = URI.parse(parentWsFolder).path + '/';
+                entryPath = entryPath.replace(wsFolderPath, '');
```
spoenemann, line 49: See my [previous comment](https://github.com/langium/langium/pull/519#discussion_r944211799). Could you try using the [path module](https://nodejs.org/dist/latest-v16.x/docs/api/path.html) instead of `startsWith` and `replace` string operations?

packages/langium/src/workspace/configuration.ts
```diff
@@ -0,0 +1,94 @@
...
+import { LangiumSharedServices } from '../services';
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export interface ConfigurationProvider {
+    /**
+    * Returns a configuration value stored for the given language.
+    *
+    * @language the language id
+    * @configuration configuration name
```
spoenemann, line 18:
```
    * @param language The language id
    * @param configuration Configuration name
```

packages/langium/src/workspace/configuration.ts
```diff
@@ -0,0 +1,94 @@
...
+     */
+    updateConfiguration(change: DidChangeConfigurationParams): void;
+}
+
+export class DefaultConfigurationProvider implements ConfigurationProvider {
+
+    protected settings: Record<string, Record<string, any>> = {};
+    protected workspaceConfig = false;
+    protected initialized = false;
+    protected readonly sharedServices: LangiumSharedServices;
```
spoenemann, line 36: I think it's better to store only the actual dependencies (`ServiceRegistry` and `lsp.Connection`) instead of the whole DI container as a class field. This makes it easier to keep track of the service dependencies.

packages/langium/src/workspace/configuration.ts
```diff
@@ -0,0 +1,94 @@
...
+        this.settings[section] = configuration;
+    }
+
+    async getConfiguration(language: string, configuration: string): Promise<any> {
+        if (!this.initialized) {
+            await this.initialize();
+        }
+        const sectionName = this.toSectionName(language);
+        if (!this.settings[sectionName]) {
+            this.settings[sectionName] = {};
```
spoenemann, line 85: Instead of initializing all languages, could we initialize only the requested language here?

spoenemann (commented):
Thanks for the restructuring!

packages/langium-vscode/src/language-server/grammar-workspace-manager.ts
```diff
@@ -0,0 +1,57 @@
...
+import { ConfigurationProvider } from 'langium/src/workspace/configuration';
+import { FileSystemNode } from 'langium/src/workspace/file-system-provider';
+import { DefaultWorkspaceManager } from 'langium/src/workspace/workspace-manager';
+import { LangiumGrammarLanguageMetaData } from 'langium/src/grammar/generated/module';
+
+const CONFIG_KEY = 'build';
+
+interface WorkspaceManagerConf {
+    /**
+     * gitignore separated exclusion patterns, separated by comma
```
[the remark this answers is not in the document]
dhuebner: Changed. Thanks!

packages/langium/src/workspace/configuration.ts
dhuebner: Changed. Thanks!

packages/langium/src/workspace/configuration.ts
dhuebner: Changed. Thanks!

packages/langium/src/workspace/configuration.ts
dhuebner: Changed. Thanks!

packages/langium-vscode/src/language-server/grammar-workspace-manager.ts
dhuebner: Well, it doesn't matter which API we would use, we can't recover the missing information about the WorkspaceFolder we are operating on. That means that we can't create a WSFolder relative path. 
I think the best is to pass over the WorkspaceFolder, I changed the implementation accordingly

spoenemann (approved):
:+1:

PR #523, opened 2022-06-03
Added streamAst utility function

spoenemann:
- Added an option to TreeStreamImpl to include the root node
 - Added `streamAst` function as alternative to `streamAllContents` to include the root node
 - Use the new function where it makes sense
 - Changed `streamCst` to include the root node to be consistent with the new function (the name is almost the same)


msujew (approved):
LGTM, thanks!

msujew (changes requested):
Nvm, noticed that the semantic token provider still uses the old method, right?

spoenemann:
Right, but I think in that case it actually makes sense to handle the root node separately. The `comparedRange` check is not necessary for the root node.

msujew (approved):
@spoenemann Right, that makes sense 👍

PR #525, opened 2022-06-03
Moved a few services from 'index' to 'workspace' group

spoenemann:
We renamed the `index` folder to `workspace` some time ago, but forgot to rename the service group as well.


msujew (approved):
👍

PR #528, opened 2022-06-07
Find References for declared types

gfontorbe:
Adds the 'Find All References' feature for declared type.

I think the new service can most likely be simplified but I looked way too many times at it and can't see where to improve it anymore...

## Grammar
```
grammar test hidden(WS)

entry Model: value=RuleA;

interface A {
    name: string
}

interface B extends A {
    assignmentB: string
}

type C = A | B;

RuleA returns A: name=ID;

RuleB returns B: name=ID assignmentB=ID;

RuleC returns C: name=ID assignmentB=ID;

ActionRule: {A} 'A' name=ID | {B} name=ID assignmentB=ID | {C} name=ID assignmentB=ID;

terminal ID: /\\w+/;
terminal WS: /\\s+/;
```

## Behavior
* Calling 'Find All References' from interface property should find all references in parser rules returning the type or one of its children types
* Calling 'Find All References' from an assignment in a parser rule returning a declared interface should find the property in return type (or one of its super types) and all assignments
*  Calling 'Find All References' from an assignment in a parser rule returning a union type should find the property in atom type (or one of its super types) and all assignments
* Calling 'Find All References' from an assignment in an Action should find the property in the Action type (or super type) and all assignments

## Side effects
* It is now possible to rename a property in an interface and propagate the renaming to all references in assignments. Do not work with renaming from assignment in parser rules (not implemented).

## Other
* `ParserHelper` method in `langium-test.ts`now adds parsed document to `LangiumDocuments` service
* New method `CleanUpTestDocuments` in `langium-test.ts` to remove documents from the workspace during tests.
* Refactored test in `domainmodel` example to comply with new `ParserHelper` method.


packages/langium/src/test/langium-test.ts
```diff
@@ -306,3 +327,8 @@ export function expectWarning<T extends AstNode = AstNode, N extends AstNode = A
         ...content,
     });
 }
+
+function CleanUpTestDocuments(services: LangiumServices) {
```
msujew, line 331: Suggestion: I believe we should export this, so that downstream users can easily clean up their tests, even when they're not using our testing functions.

```suggestion
export function clearDocuments(services: LangiumServices) {
```

examples/domainmodel/test/refs-index.test.ts
```diff
@@ -62,7 +60,7 @@ function nodeRefString(from: AstNode, to: AstNode): string {
 }
 
 function createPath(node: AstNode): string {
-    return services.workspace.AstNodeLocator.getAstNodePath(node);
+    return services.index.AstNodeLocator.getAstNodePath(node);
```
msujew, line 63: There might've been an issue during rebasing:
```suggestion
    return services.workspace.AstNodeLocator.getAstNodePath(node);
```

packages/langium/src/grammar/lsp/langium-grammar-reference-finder.ts
```diff
@@ -0,0 +1,196 @@
...
+        const interfaceNode = getContainerOfType(typeAttributeNode, isInterface);
+        if (interfaceNode) {
+            const collectedTypes = this.collectChildrenTypes(interfaceNode);
+            collectedTypes.add(interfaceNode);
+
+            let referencesToTypes: Set<ReferenceDescription> = new Set<ReferenceDescription>();
+            collectedTypes.forEach(collectedType => {
+                const refs = this.references.findReferences(collectedType).toSet();
+                referencesToTypes = new Set<ReferenceDescription>([...referencesToTypes, ...refs]);
+            });
```
msujew, line 119: Minor: I think there might be a lot of unnecessary computation in there (constructing multiple sets again and again).

```suggestion
            const referencesToTypes = new Set<ReferenceDescription>();
            collectedTypes.forEach(collectedType => {
                const refs = this.references.findReferences(collectedType);
                for (const ref of refs) {
                    referencesToTypes.add(ref);
                }
            });
```

packages/langium/src/grammar/lsp/langium-grammar-reference-finder.ts
```diff
@@ -0,0 +1,196 @@
...
+
+            const parserRules: Set<ParserRule | Action> = this.findParserRulesFromReferences(referencesToTypes);
+
+            parserRules.forEach(rule => {
+                const assignment = isParserRule(rule) ? extractAssignments(rule.alternatives).find(assignment => assignment.feature === selectedNode.text) : extractAssignments(getContainerOfType(rule, isGroup)!).find(assignment => assignment.feature === selectedNode.text);
```
msujew, line 124: Minor: Add some linebreaks, this line's quite difficult to read.

packages/langium/src/lsp/reference-finder.ts
```diff
@@ -42,11 +42,22 @@ export class DefaultReferenceFinder implements ReferenceFinder {
         if (!rootNode) {
             return [];
         }
-        const refs: Array<{ docUri: URI, range: Range }> = [];
+        // const refs: Array<{ docUri: URI, range: Range }> = [];
```
msujew, line 45: Minor: You can safely remove this line.

packages/langium/src/test/langium-test.ts
```diff
@@ -124,6 +125,26 @@ export function expectGoToDefinition(services: LangiumServices, expectEqual: Exp
...
+        const document = await parseDocument(services, output);
+        const referenceFinder = services.lsp.ReferenceFinder;
+        const references = await referenceFinder.findReferences(document, referenceParams(document, indices[expectedFindReferences.index], expectedFindReferences.includeDeclaration)) ?? [];
+        CleanUpTestDocuments(services);
+        expectEqual(references.length, expectedFindReferences.referencesCount);
+    };
+}
+
+function referenceParams(document: LangiumDocument, offset: number, includeDeclaration: boolean): ReferenceParams {
+    return { textDocument: { uri: document.textDocument.uri }, position: document.textDocument.positionAt(offset), context: {includeDeclaration: includeDeclaration} };
```
msujew, line 146: Minor: Formatting

```suggestion
    return { 
        textDocument: { uri: document.textDocument.uri },
        position: document.textDocument.positionAt(offset),
        context: { includeDeclaration }
    };
```

packages/langium/src/test/langium-test.ts
```diff
@@ -306,3 +327,8 @@ export function expectWarning<T extends AstNode = AstNode, N extends AstNode = A
         ...content,
     });
 }
+
+function CleanUpTestDocuments(services: LangiumServices) {
+    const allDocs = services.shared.workspace.LangiumDocuments.all.map(x => x.uri).toArray();
+    services.shared.workspace.DocumentBuilder.update([], allDocs);
+}
```
msujew, line 334: Minor: Missing newline

packages/langium/test/lsp/find-references.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    test ('Must find all references to property name from Action C', async () => {
+        await findReferences({
+            text,
+            index: 8,
+            referencesCount: 7,
+            includeDeclaration: true
+        });
+    });
+});
```
msujew, line 132: Minor: Missing newline

packages/langium/src/grammar/lsp/langium-grammar-reference-finder.ts
```diff
@@ -0,0 +1,196 @@
...
+            const collectedTypes = this.collectChildrenTypes(interfaceNode);
+            collectedTypes.add(interfaceNode);
+
+            let referencesToTypes: Set<ReferenceDescription> = new Set<ReferenceDescription>();
+            collectedTypes.forEach(collectedType => {
+                const refs = this.references.findReferences(collectedType).toSet();
+                referencesToTypes = new Set<ReferenceDescription>([...referencesToTypes, ...refs]);
+            });
+
+            const parserRules: Set<ParserRule | Action> = this.findParserRulesFromReferences(referencesToTypes);
```
msujew, line 121: Minor: This type can be inferred
```suggestion
            const parserRules = this.findParserRulesFromReferences(referencesToTypes);
```

packages/langium/src/grammar/lsp/langium-grammar-reference-finder.ts
```diff
@@ -0,0 +1,196 @@
...
+                const assignment = isParserRule(rule) ? extractAssignments(rule.alternatives).find(assignment => assignment.feature === selectedNode.text) : extractAssignments(getContainerOfType(rule, isGroup)!).find(assignment => assignment.feature === selectedNode.text);
+
+                if (assignment) {
+                    const cstNode = assignment.$cstNode;
+                    if (cstNode) {
+                        const leaf = findLeafNodeAtOffset(cstNode, cstNode.offset);
+                        if (leaf) {
+                            const rootNode = extractRootNode(leaf.element);
+                            if (rootNode) {
+                                refs.push({docUri: rootNode.$document!.uri, range: leaf.range});
```
msujew, line 133: Minor: Formatting
```suggestion
                                refs.push({
                                    docUri: rootNode.$document!.uri,
                                    range: leaf.range
                                });
```

packages/langium/src/grammar/lsp/langium-grammar-reference-finder.ts
```diff
@@ -0,0 +1,196 @@
...
+            const astNode = this.astNodeLocator.getAstNode(doc, ref.sourcePath);
+            if (isParserRule(astNode) || isAction(astNode)) {
+                parserRules.add(astNode);
+            }
+        });
+        return parserRules;
+    }
+
+    collectChildrenTypes(interfaceRule: Interface): Set<Interface | Type> {
+        let childrenTypes: Set<Interface | Type> = new Set<Interface | Type>();
```
msujew, line 156: Minor: Type can be inferred
```suggestion
        let childrenTypes = new Set<Interface | Type>();
```

packages/langium/src/grammar/lsp/langium-grammar-reference-finder.ts
```diff
@@ -0,0 +1,196 @@
...
+        const refs = this.references.findReferences(interfaceRule).toArray();
```
msujew, line 157: Minor: Calling `toArray()` here is not necessary, as forEach is also available on streams.

packages/langium/src/grammar/lsp/langium-grammar-reference-finder.ts
```diff
@@ -0,0 +1,196 @@
...
+
+        refs.forEach(ref => {
+            const doc = this.langiumDocuments.getOrCreateDocument(ref.sourceUri);
+            const astNode = this.astNodeLocator.getAstNode(doc, ref.sourcePath);
+            if (isInterface(astNode)) {
+                childrenTypes.add(astNode);
+                childrenTypes = new Set([...childrenTypes, ...this.collectChildrenTypes(astNode)]);
```
msujew, line 164: Suggestion: Again, don't reconstruct the set multiple times.

packages/langium/src/grammar/lsp/langium-grammar-reference-finder.ts
```diff
@@ -0,0 +1,196 @@
...
+                if (isInterface(superType.ref)) {
+                    superTypes.add(superType.ref);
+                    superTypes = new Set([...superTypes, ...this.collectSuperTypes(superType.ref)]);
+                }
+            });
+        } else if (isType(ruleNode)) {
+            ruleNode.typeAlternatives.forEach(typeAlternative => {
+                if (typeAlternative.refType?.ref) {
+                    if (isInterface(typeAlternative.refType.ref) || isType(typeAlternative.refType.ref)) {
+                        superTypes = new Set<Interface>([...superTypes, ...this.collectSuperTypes(typeAlternative.refType!.ref!)]);
```
msujew, line 187: Suggestion: Same, see above.

msujew (commented):
Looks quite

packages/langium/test/lsp/find-references.test.ts
```diff
@@ -0,0 +1,132 @@
...
+terminal WS: /\\s+/;
+
+`;
+
+const grammarServices = createLangiumGrammarServices().grammar;
+const findReferences = expectFindReferences(grammarServices, expectFunction);
+
+describe('findReferences', () => {
+    test('Must find all references to interface A from interface\'s name including declaration', async () => {
+        await findReferences({
```
Lotes, line 44: I know this pattern from goto and hover. I have some doubts to this form, because it hides what is tested. Sure, I can go to the definition of the `text` variable or to this `findReferences` definition. But you have to do much effort to find out what is done.
Suggestions:
* break out every `findReferences` into three parts (3 A's):
  * one for preparing the situation (arrange)
  * one for doing the action that is important for the test case (act)
  * one for asserting the state after this action (assert)
* also split the giant grammar text into small texts in each test case
  * because to understand one case now you will have to scroll up and count down the index of your range, which is really annoying I would say...

packages/langium/test/lsp/find-references.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+interface B extends A {
+    assignmentB: string
+}
+
+type C = A | B;
+
+RuleA returns A: na<|>me=ID;
+
+RuleB returns B: na<|>me=ID assignmentB=ID;
```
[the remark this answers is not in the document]
dhuebner: @Lotes 
I think there is at least an example for `<|>A`, however I would say it would be better to test the `findLeafNodeAtOffset(node: CstNode, offset: number): LeafCstNode | undefined `  in a separate test file if it not already done. https://github.com/langium/langium/blob/02276b565f66a053fe299b8602c08ab761504d14/packages/langium/src/utils/cst-util.ts#L83

packages/langium/src/test/langium-test.ts
```diff
@@ -124,6 +125,30 @@ export function expectGoToDefinition(services: LangiumServices, expectEqual: Exp
...
+    referencesCount: number,
+    includeDeclaration: boolean
+}
+
+export function expectFindReferences(services: LangiumServices, expectEqual: ExpectFunction): (expectedFindReferences: ExpectedFindReferences) => Promise<void> {
+    return async expectedFindReferences => {
+        const { output, indices } = replaceIndices(expectedFindReferences);
+        const document = await parseDocument(services, output);
+        const referenceFinder = services.lsp.ReferenceFinder;
+        const references = await referenceFinder.findReferences(document, referenceParams(document, indices[expectedFindReferences.index], expectedFindReferences.includeDeclaration)) ?? [];
```
Lotes, line 139: Looks a bit suspicious. Why does it return NULL/UNDEFINED when it is empty?

Lotes, line 139: Minor: Put the subexpression `referencParams(...)` into its own variable.

packages/langium/test/lsp/find-references.test.ts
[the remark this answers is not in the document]
Lotes: Yes, actually this is a reason for a separate PR. This PR assumes that the underlying function is already tested and correct.

packages/langium/src/grammar/lsp/langium-grammar-reference-finder.ts
```diff
@@ -0,0 +1,212 @@
+/******************************************************************************
```
Lotes, line 1: You put a lot of effort in here: about 200 LOC. Wow.

As I mentioned [in Slack (internal)](https://typefox.slack.com/archives/C01M3T0BQ7L/p1654760192140229), we should think about ways of minimizing the duplicated code. Because that is a kind of risk. If you add another feature XY where type names are also part of, you are likely in a position where you could forget 1 of 3-4 places.

I think we can make this code much shorter (later), like
```typescript
graph.nodes.getByName(typeName) //every type has a node
           .getIncomingEdges()  //with a bunch of incoming or defining edges
           .filter(e => e.type == 'typedef') //filter out the right edges, if needed
           .flatMap(e => e.astNodes) //get the ast nodes that define the type
```

This is just an example.

packages/langium/src/test/langium-test.ts
Lotes: Thanks for clarifying

Lotes: Ah wait i read necessary. Reopening

packages/langium/test/lsp/find-references.test.ts
gfontorbe: After our discussion, I assigned a grammar text for each test. This (I hope) makes the test easier to read.

The tests now check that the length of the returned array is correct **and** that the range of the reference is also correct.

gfontorbe:
I added a different grammar for each example, this way I hope it is easier to follow what each individual test is addressing

packages/langium/test/lsp/find-references.test.ts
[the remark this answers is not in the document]
Lotes: The function is already tested, resolved

packages/langium/test/lsp/find-references.test.ts
```diff
@@ -0,0 +1,505 @@
...
+
+const grammarServices = createLangiumGrammarServices().grammar;
+const findReferences = expectFindReferences(grammarServices);
+
+describe('findReferences', () => {
+    test('Must find references to parent interface in another interface declaration. Including parent interface declaration', async () => {
+        const grammar = `grammar test hidden (WS)
+        terminal ID: /\\w+/;
+
+        interface <|<|>A|> {
```
Lotes, line 18: Much better to read. Thank you!

One thing that is not clear on first sight is, what is meant? Like "AHA! There is an index and two ranges. So, it might be the case that the test picks out the index and points to that position to find all references which are equal to the ranges."...

For the moment it is good like this, I believe. The only thing we could do better is giving these ranges and indices names as a kind of handles.

But as I said, it is already a huge improvement

packages/langium/src/lsp/reference-finder.ts
```diff
@@ -60,14 +76,11 @@ export class DefaultReferenceFinder implements ReferenceFinder {
                     refs.push({ docUri: document.uri, range: reference.$refNode.range });
                 } else {
                     const range = reference.segment.range;
-                    refs.push({ docUri: reference.sourceUri, range });
+                    refs.push({ docUri: reference.sourceUri, range: range });
```
Lotes, line 79: Minor: just `range` is enough, or `range: reference.segment.range`

packages/langium/src/lsp/reference-finder.ts
```diff
@@ -28,6 +28,11 @@ export interface ReferenceFinder {
     findReferences(document: LangiumDocument, params: ReferenceParams, cancelToken?: CancellationToken): MaybePromise<Location[]>;
 }
 
+export interface ReferenceLocation {
```
Lotes, line 31: You only need this intermediate interface because your URIs are not compatible. Is it possible to use the one from VSCode? Or is there a reason?

packages/langium/src/grammar/lsp/langium-grammar-reference-finder.ts
Lotes: Not the right moment, but I will add an issue...

packages/langium/src/lsp/reference-finder.ts
gfontorbe: Afaik, the VSCode URI doesn't have information about the range

packages/langium/src/lsp/reference-finder.ts
gfontorbe: Personal taste, I find it "cleaner" to have an explicit name for properties when declaring an object.

Lotes: I will not cry, it is not that important to me. It is just sugar from TypeScript.

packages/langium/src/test/langium-test.ts
```diff
@@ -124,6 +125,39 @@ export function expectGoToDefinition(services: LangiumServices, expectEqual: Exp
     };
 }
 
+export interface ExpectedFindReferences extends ExpectedBase {
+    index: number,
```
msujew, line 129: Suggestion: I'm not sure whether having the index in there is actually useful. Let's take the following example:

```
// Note the <|> in this line
interface <|<|>A|> {
  name: string
}
// Note the <|> in this line
ruleA returns <|<|>A|>: name=ID;
```

I believe we should be able to test this in one call, as there's no benefit to calling it like this:

```ts
await findReferences({
  text: grammar,
  index: 0,
  includeDeclaration: true
});
await findReferences({
  text: grammar,
  index: 1,
  includeDeclaration: true
});
```

Since we need to test all of them anyway, we should do this in a single call. Alternatively, you can add a `rangeIndices` property here, which can be used to test multiple references using different calls. That's what we do for the `goto` tests for example:

```ts
await findReferences({
  text: grammar,
  index: 0,
  rangeIndices: [0, 1],
  includeDeclaration: true
});
await findReferences({
  text: grammar,
  index: 1,
  rangeIndices: [2, 3]
  includeDeclaration: true
});
```

However, that would make these tests really convoluted, so I'm just in favor of removing the `index` property.

packages/langium/test/lsp/find-references.test.ts
```diff
@@ -0,0 +1,505 @@
...
+
+        await findReferences({
+            text: grammar,
+            index: 0,
+            includeDeclaration: true
+        });
+    });
+
+    test('Must find references to parent interface in another interface declaration. Excluding parent interface declaration', async () => {
+        const grammar = `grammar test hidden (WS)
```
msujew, line 33: Minor: The grammars in this file use outdated syntax

```suggestion
        const grammar = `grammar test
        hidden terminal WS: /\\s+/;
```

msujew (commented):
Manual testing seems fine, although the `document-highlighter` could use some changes as well.

I have a few comments regarding the testing API. It's almost there, just minor change would make it perfect 👍

packages/langium/src/lsp/reference-finder.ts
gfontorbe: That will be removed because the `Location` type as a `create` function

msujew (approved):
Great, looks good to me! I agree with everything @Lotes said about the verbosity of the code, but we can refactor this feature later.

PR #530, opened 2022-06-08
Remove node `path` dependency

msujew:
Also adds some more uri utility functions, which replace the functionality of the `path` dependency.


spoenemann (approved):
:+1:

PR #533, opened 2022-06-09
Refactor reference ID sourcing

msujew:
Due to issues in fragment cross reference resolution, I had to refactor how all cross ref IDs are calculated. We should merge this before 0.4.

For reference: When creating a cross reference in a fragment, the cross reference can never be resolved (thanks to @gfontorbe for figuring this out)

```
Test: 'test' Frag;
fragment Frag: ref=[Item]; // Never resolves
```

### DIFF
--- a/examples/arithmetics/src/language-server/generated/ast.ts
+++ b/examples/arithmetics/src/language-server/generated/ast.ts
@@ -115,7 +115,7 @@ export function isNumberLiteral(item: unknown): item is NumberLiteral {
 
 export type ArithmeticsAstType = 'AbstractDefinition' | 'BinaryExpression' | 'DeclaredParameter' | 'Definition' | 'Evaluation' | 'Expression' | 'FunctionCall' | 'Module' | 'NumberLiteral' | 'Statement';
 
-export type ArithmeticsAstReference = 'FunctionCall:func';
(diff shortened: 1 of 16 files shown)


examples/arithmetics/src/language-server/generated/ast.ts
```diff
@@ -115,7 +115,7 @@ export function isNumberLiteral(item: unknown): item is NumberLiteral {
 
 export type ArithmeticsAstType = 'AbstractDefinition' | 'BinaryExpression' | 'DeclaredParameter' | 'Definition' | 'Evaluation' | 'Expression' | 'FunctionCall' | 'Module' | 'NumberLiteral' | 'Statement';
 
-export type ArithmeticsAstReference = 'FunctionCall:func';
+export type ArithmeticsAstReference = 'PrimaryExpression:func';
```
spoenemann, line 118: So the name preceding the colon is now the parser rule, not the type?

msujew: Yes, exactly. This makes cross references unambiguous in regards to the actual type of the object we're passing around.

msujew:
We have chosen another approach to tackle this issue.

PR #541, opened 2022-06-14
Make CST checking less strict

msujew:
Closes https://github.com/langium/langium/issues/451

Previously, we've thrown an error every time we encountered an empty CST composite node.

Since there are some valid and some invalid use cases for empty composite nodes, I've decided to remove the error in favor of returning `0` in case it is truly empty.


spoenemann:
Thanks. Could you explain briefly how this situation can happen?

msujew:
@spoenemann No, no idea how this happens. None of my tests are able to reproduce this issue :)

spoenemann (approved):
It's a bit unsatisfying to be in the dark here, but let's move on.

PR #542, opened 2022-06-15
Fix extract destination and name process in cli-util template.

### LINKED ISSUE #526: Improve extract destination and name process in cli-util template
Here, if a relative path given, the process of obtaining the file path is blank.

e.g: `'./path/to/test.hello'.replace(/\..*$/, '')` -> `''`.

https://github.com/langium/langium/blob/f6fc15eee6a05f361d7bf9a78f173b10dee47273/packages/generator-langium/langium-template/src/cli/cli-util.ts#L46

Relative paths can be used with the following modification.

```ts
filePath = path.basename(filePath, path.extname(filePath)).replace(/[.-]/g, '')
```

mikoto2000:
Closes #526.

Fixed extract destination and name process in cli-util template and examples.

### DIFF
--- a/examples/domainmodel/src/cli/cli-util.ts
+++ b/examples/domainmodel/src/cli/cli-util.ts
@@ -63,7 +63,7 @@ interface FilePathData {
 }
 
 export function extractDestinationAndName(filePath: string, destination: string | undefined): FilePathData {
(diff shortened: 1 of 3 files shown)


msujew (approved):
Thanks that's definitely an improvement 👍 

In the long run we should probably aim to implement https://github.com/langium/langium/issues/239, but that's good for now.

PR #543, opened 2022-06-15
Added code action to add new parser rule

### LINKED ISSUE #540: Grammar code action: Create new rule
When I write a new grammar, I usually do it top-down starting with the entry rule. This very often leads me to writing rules like
```
Computation:
    'comp' name=ID '(' params+=Parameter* ')' '=' expression=Expression;
```
while `Parameter` and `Expression` have not been written yet. It would be great to have a code action for such an unresolved reference that creates a new rule:
```
Parameter:
    /* cursor position here, if possible */;
```

spoenemann:
Closes #540. The new rule contain an action `{infer RuleName}` so they start without errors, but that's not optimal.

### DIFF
--- a/packages/langium/src/validation/document-validator.ts
+++ b/packages/langium/src/validation/document-validator.ts
@@ -44,7 +44,7 @@ export class DefaultDocumentValidator implements DocumentValidator {
 
         await interruptAndCheck(cancelToken);
 
-        // Process lexer errors
+        // Process lexing errors
         for (const lexerError of parseResult.lexerErrors) {
             const diagnostic: Diagnostic = {
                 severity: DiagnosticSeverity.Error,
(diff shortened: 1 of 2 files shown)


packages/langium/src/validation/document-validator.ts
```diff
@@ -168,3 +178,15 @@ export function toDiagnosticSeverity(severity: 'error' | 'warning' | 'info' | 'h
             throw new Error('Invalid diagnostic severity: ' + severity);
     }
 }
+
+export namespace DocumentValidator {
+    export const LexingError = 'lexing-error';
+    export const ParsingError = 'parsing-error';
+    export const LinkingError = 'linking-error';
+}
```
msujew, line 186: Since the `code` of diagnostics appears in the frontend, we should have something like a `ValidationCodeProvider` which can be easily overriden by users. They likely want to have more control over which code exactly they want to show. But that's something for another PR.

msujew (approved):
I'm happy with the changes. I have a remark, but that's actually a new feature so we should implement that in another PR.

PR #545, opened 2022-06-16
Don't validate grammar rules when empty

spoenemann:
When you start typing a grammar rule
```
MyRule:
    ;
```
the editor used to show a whole bunch of validation errors, which is annoying because you know that you're in the middle of writing something. This change reduces the shown errors for the above case to just the syntax error.

I also changed the grammar AST along the way such that both ParserRule and TerminalRule have their definition in a `definition` property instead of `alternatives` and `terminal`.


spoenemann:
Caveat: when you don't write the semicolon at the end, the rule is not detected as empty.
```
MyRule:
    

OtherRule: ...
```
Here `OtherRule` is parsed as it were a rule call within the definition of `MyRule`.

PR #546, opened 2022-06-17
Allow to create `LangiumServices` from a grammar string

msujew:
Closes https://github.com/langium/langium/issues/421

Allows us to delete quite a lot of code, as we needed a lot of async wrapping previously. The new method runs sync.

### DIFF
--- a/packages/langium-cli/src/parser-validation.ts
+++ b/packages/langium-cli/src/parser-validation.ts
@@ -4,11 +4,7 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import {
-    createDefaultModule, createDefaultSharedModule, getDocument, Grammar, inject, interpretAstReflection, IParserConfig,
-    isGrammar, isParserRule, LangiumDocuments, LangiumGeneratedServices, LangiumGeneratedSharedServices,
-    LangiumParser, LangiumServices, LangiumSharedServices, Module, ParserRule, prepareLangiumParser
-} from 'langium';
+import { createServicesForGrammar, getDocument, Grammar, IParserConfig, isGrammar, isParserRule, LangiumDocuments, LangiumParser, LanguageMetaData, ParserRule, prepareLangiumParser } from 'langium';
 import { getFilePath, LangiumConfig, LangiumLanguageConfig } from './package';
 
 export function validateParser(grammar: Grammar, config: LangiumConfig, grammarConfigMap: Map<Grammar, LangiumLanguageConfig>,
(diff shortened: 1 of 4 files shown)


packages/langium/src/grammar/grammar-util.ts
```diff
@@ -409,20 +413,51 @@ function resolveTransitiveImportsInternal(documents: LangiumDocuments, grammar:
     return Array.from(grammars);
 }
 
+export function createServicesForGrammar(grammar: string, module?: Module<LangiumServices>, sharedModule?: Module<LangiumSharedServices>): LangiumServices {
```
spoenemann, line 416: Proposal: wrap `module` and `sharedModule` into a `config` object and add more optional properties to make these customizable / reusable: `grammarServices`, `parserConfig`, `languageMetaData`. When not provided, fall back to the defaults you specified here.

msujew: Great idea, that allows us to easily reuse this function in the `langium-cli` for parser validation 👍

packages/langium-cli/src/parser-validation.ts
```diff
@@ -18,19 +14,11 @@ export function validateParser(grammar: Grammar, config: LangiumConfig, grammarC
...
-            ParserConfig: () => parserConfig
-        }
-    };
-    const shared = inject(createDefaultSharedModule(), generatedSharedModule);
-    const services = inject(createDefaultModule({ shared }), generatedModule);
+    const services = createServicesForGrammar({
+        grammar,
+        languageMetaData: languageConfigToMetaData(grammarConfigMap.get(grammar)!),
+        parserConfig
+    });
```
spoenemann, line 21: We could also pass in the grammar services here so we don't have to recreate them for every language (relevant when the CLI runs for multiple languages).

packages/langium/src/grammar/grammar-util.ts
```diff
@@ -409,20 +414,59 @@ function resolveTransitiveImportsInternal(documents: LangiumDocuments, grammar:
...
+    const grammarNode = typeof config.grammar === 'string' ? grammarServices.parser.LangiumParser.parse<ast.Grammar>(config.grammar).value : config.grammar;
+    prepareGrammar(grammarServices, grammarNode);
+
+    const parserConfig = config.parserConfig ?? {
+        skipValidations: false
+    };
+    const languageMetaData = config.languageMetaData ?? {
+        caseInsensitive: false,
+        fileExtensions: ['.test'],
+        languageId: 'test'
```
spoenemann, line 435: Idea: derive language id and file extensions from the parsed grammar name.

spoenemann (approved):

PR #547, opened 2022-06-17
Adds Code Actions for Types

### LINKED ISSUE #442: Various code actions about types
Write some code actions as quick fixes for validations from #426:
* type names duplication
* consistency validation of interred and declared types (#415)
* validation of property name duplication (#430)

montymxb:
Closes #442

This one addresses 3/5 of the points noted in #442, so this does not fully address the issue. The reason why I did not fully cover all points is due to some type related complexities, which are elaborated below. What would be helpful here is to see:
- whether the points addressed so far are Ok (like removing a duplicate type)
- whether the reason why some points are not addressed seems rational 

I'm open to feedback to revise the added diagnostics or actions 🙂.

The points I did address are:
- [x] type names duplication -> removes a duplicated type
- [x] validation of property name duplication (https://github.com/langium/langium/issues/430) -> removes a duplicated property
- [x] adds return T to rule for general error The type 'T' is already explicitly declared and cannot be inferred. From https://github.com/langium/langium/issues/449.

Of these, I'm still unsure about the first point. It results in situations like this, where removing either type is probably not desirable. I thought about merging them together, but in order to do so it requires pretty printing an interface from it's internal representation. Such a solution would also align with the goal we have in (https://github.com/langium/langium/issues/441), where we would also be generating declared types by pretty printing them from our internal representation. This feels like the base solution for both of these is a separate issue that can be addressed first, and then used to build both of the complete solutions.

```ts
interface T1 { a : number } // <-- remove this?
interface T1 { b : string } // <-- or this?

// in either case, we lose something about T1
```

The second point feels ok. It leaves the exact property to remove to fix the duplication up to the user.
```ts
interface T1 { dup: string } // < -- could drop this dup property
interface T2 extends T1 { dup: string } // <-- or this one

// Choice is up to the user, who ideally knows more about where the property should be
```

The last point is pretty handy in my opinion, I've already used it several times myself.
Gif of returns being injected, and infers converted to returns

 The remaining 2 points are a bit of an issue:

- [ ] adds a declared type by inferred type name
- [ ] replaces a dummy rule with a declared type

In both cases I decided not to include fixes for these, as they require further work on the AST types. 

For both issues, in order to generate a declared type from an inferred type, we have to process the whole grammar first to be certain that our type is complete. Since this information is not preprocessed we would have to recompute it for each instance. Even then we would still need a way to pretty print that declared type, which is the same issue I noted above. So in both cases I feel like having our types validated/computed in advance, and accessible via something like the document object would greatly help with this.


montymxb:
@spoenemann question about this. Would it make more sense to break this PR up into 2 cases fixing the 2nd and 3rd point, and dropping the first. I'm considering this because the first one feels too problematic to be a reasonable code action. The 2nd case I feel better about, and the 3rd I believe is the strongest.

montymxb:
Closing in favor of sub-dividing into 2 separate PRs.

PR #548, opened 2022-06-17
Type Usage Consistency (2)

### LINKED ISSUE #429: Consistency of type usage
See https://github.com/langium/langium/pull/406#discussion_r810922786: we should validate usages of `type`s for consistency. For example, this is not valid because `Foo` is not an AST node type.
```
type Foo = string;

Bar returns Foo: ...
```
The same applies to actions and extended interfaces.

montymxb:
Closes #429 (again) by extending the same super type checks done for declared types to inferred super types, extending upon the existing checks that were first applied for this issue. These checks include a warning against extending declared types by inferred types. For more details on why this is an issue, see the discussion in the original issue, which has some good conversation on the issue.

This was done by tapping into the inferred type collection already in the type system, but is only applied to a singular ParserRule. This gives enough information to be able to determine whether a type union would be ultimately generated, and in such a case we can give a validation error. In the case where no type union is generated, we can assume that an interface would be generated, and can instead apply a warning.

Hopefully this achieves the original goals without introducing *too* much overhead by pulling in some of the type validation logic. Ideally, it would be nice to keep those two steps more separate, but this seemed better than trying to duplicate some of the raw AST type inference logic in the grammar validation phase.


packages/langium/test/grammar/langium-grammar-validator.test.ts
```diff
@@ -30,4 +30,35 @@ describe('Langium grammar validation', () => {
             property: {name: 'terminal'}
         });
     });
+
+    test('Interfaces should only be able to extend inferred interfaces', async () => {
+        const validationResult = await validate(`
+        grammar G
```
Lotes, line 36: I would vote for speaking names, it is hard to read

montymxb:
@Lotes updated that test message you pointed out, should be clearer now.

packages/langium/test/grammar/langium-grammar-validator.test.ts
```diff
@@ -30,4 +30,36 @@ describe('Langium grammar validation', () => {
...
+
+        entry SA: S1 | S2;
+        SB: S1;
+
+        S1: 's1' s1=ID;
+        S2: 's2' s2=ID;
+
+        interface DA extends SA {}
+        interface D2 extends S2 {}
+        interface DB extends SB {}
```
msujew, line 47: I believe @Lotes meant the names of these elements ;)

msujew (approved):
LGTM 👍

packages/langium/test/grammar/langium-grammar-validator.test.ts
```diff
@@ -30,4 +30,51 @@ describe('Langium grammar validation', () => {
...
+
+        // generates an inferred union of 2 other inferred interfaces
+        entry InferredUnion: InferredI1 | InferredI2;
+
+        // for the case where an inferred interface extends another inferred interface
+        InferredI0: InferredI1;
+
+        // just for the sake of generating a pair of inferred interfaces
+        InferredI1: 'InferredI1' InferredI1=ID;
+        InferredI2: 'InferredI2' InferredI2=ID;
```
Lotes, line 47: Confusing… rule name equal to assignment property?

packages/langium/test/grammar/langium-grammar-validator.test.ts
```diff
@@ -30,4 +30,51 @@ describe('Langium grammar validation', () => {
...
+
+        // should fail...
+        interface DeclaredExtendsUnion extends InferredUnion {}
```
Lotes, line 50: I prefer to have one test fixture per scenario. Also confusing when you want to refactor one test but accidently break another

packages/langium/test/grammar/langium-grammar-validator.test.ts
```diff
@@ -30,4 +30,51 @@ describe('Langium grammar validation', () => {
...
+
+        // we can extend an inferred type that has no parent
+        interface DeclaredExtendsInferred1 extends InferredI2 {}
+
+        // we can extend an inferred type that extends other inferred types
```
Lotes, line 55: Comments could be the name of the new test.

Lotes (commented):
A huge improvement around readibility. I skipped the initial version because it would have taken some effort, to understand what is important. Nice!

packages/langium/test/grammar/langium-grammar-validator.test.ts
Lotes: Comments have to explain, why something is what it is. I think we should form could so that it is easy to read without checking comments… Comments can outdate, very fast.

packages/langium/test/grammar/langium-grammar-validator.test.ts
montymxb: Implemented 👍

Issue #379, opened 2022-01-20
Don't block the initialize request while indexing


spoenemann:
Currently we `await` the WorkspaceManager (#377) to finish indexing the workspace folders within the initialize request handler. This can potentially lead to problems like timeouts on the client side, so we should continue indexing in the background.

What would be the consequences? Any `didOpen` or `didChange` notification would have to wait to be processed, but that's easy because we already have a mechanism to wait for the previous notification processing to finish. What about other requests? The easiest solution would be if they just return nothing until the workspace is ready.

This possibly depends on #273 because we need to load workspace files asynchronously. Currently we use `readFileSync` to ensure that we can load documents synchronously during linking.

PR #549, opened 2022-06-17
Handle workspace changes using a mutex

msujew:
Closes https://github.com/langium/langium/issues/379

This took me a whole friday afternoon, I hope it's working perfectly: The `ReadWriteMutex` allows us to better hide some of the promise magic that's happening inside of the `language-server.ts` file. The `write` method performs a **single non-parallel** write action on the workspace (for workspace changes), while `read` allows **multiple parallel** reads on the workspace (usually for answering LSP requests).

`read` will always wait for all `write` calls to be executed before executing itself. `write` will only wait for the **current** `read` call and then execute itself. Any subsequent `write` calls will automatically move up in priority, even if some `read` call happened before that. See the different test cases for more context on how this exactly works.

This change addresses the issue by wrapping the `initializeWorkspace` in a `write` call, which will be performed in the background, while we send back the answer for the initialize request.

### DIFF
--- a/packages/langium/src/lsp/language-server.ts
+++ b/packages/langium/src/lsp/language-server.ts
@@ -55,12 +55,10 @@ export function startLanguageServer(services: LangiumSharedServices): void {
             }
         };
 
-        try {
-            if (params.workspaceFolders) {
-                await services.workspace.WorkspaceManager.initializeWorkspace(params.workspaceFolders);
-            }
-        } catch (err) {
-            console.error(err);
+        if (params.workspaceFolders) {
+            const folders = params.workspaceFolders;
+            const mutex = services.workspace.ReadWriteMutex;
+            mutex.write(() => services.workspace.WorkspaceManager.initializeWorkspace(folders));
         }
         return result;
     });
@@ -77,7 +75,7 @@ export function startLanguageServer(services: LangiumSharedServices): void {
     addCodeActionHandler(connection, services);
     addRenameHandler(connection, services);
     addHoverHandler(connection, services);
-    addSemanticHighlighting(connection, services);
+    addSemanticTokenHandler(connection, services);
 
     // Make the text document manager listen on the connection for open, change and close text document events.
     const documents = services.workspace.TextDocuments;
@@ -89,33 +87,23 @@ export function startLanguageServer(services: LangiumSharedServices): void {
 
 export function addDocumentsHandler(connection: Connection, services: LangiumSharedServices): void {
     const documentBuilder = services.workspace.DocumentBuilder;
+    const mutex = services.workspace.ReadWriteMutex;
     let changeTokenSource: AbstractCancellationTokenSource;
-    let changePromise: Promise<void> | undefined;
 
-    async function onDidChange(changed: URI[], deleted?: URI[]): Promise<void> {
+    function onDidChange(changed: URI[], deleted: URI[]): void {
         changeTokenSource?.cancel();
-        if (changePromise) {
-            await changePromise;
-        }
         changeTokenSource = startCancelableOperation();
-        changePromise = documentBuilder
-            .update(changed, deleted ?? [], changeTokenSource.token)
-            .catch(err => {
-                if (!isOperationCancelled(err)) {
-                    console.error('Error: ', err);
-                }
-            });
+        const token = changeTokenSource.token;
+        mutex.write(() => documentBuilder.update(changed, deleted, token));
     }
 
     const documents = services.workspace.TextDocuments;
     documents.onDidChangeContent(change => {
-        // eslint-disable-next-line @typescript-eslint/no-floating-promises
-        onDidChange([URI.parse(change.document.uri)]);
+        onDidChange([URI.parse(change.document.uri)], []);
     });
     connection.onDidChangeWatchedFiles(params => {
         const changedUris = params.changes.filter(e => e.type !== FileChangeType.Deleted).map(e => URI.parse(e.uri));
         const deletedUris = params.changes.filter(e => e.type === FileChangeType.Deleted).map(e => URI.parse(e.uri));
-        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         onDidChange(changedUris, deletedUris);
     });
 }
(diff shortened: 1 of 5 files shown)


packages/langium/src/lsp/language-server.ts
```diff
@@ -89,33 +87,23 @@ export function startLanguageServer(services: LangiumSharedServices): void {
...
         changeTokenSource = startCancelableOperation();
-        changePromise = documentBuilder
-            .update(changed, deleted ?? [], changeTokenSource.token)
-            .catch(err => {
-                if (!isOperationCancelled(err)) {
-                    console.error('Error: ', err);
-                }
-            });
+        const token = changeTokenSource.token;
+        mutex.write(() => documentBuilder.update(changed, deleted, token));
```
spoenemann, line 97: First thought before looking deeper into this: You now lock the whole workspace until the update is completely done, which includes validation. That last phase of the document builder can potentially take very long and thus block all LSP services.

It would be good to verify this: Implement an async validation check that uses `setTimeout` to simulate a long-running process.

Potential solution: extract the validation phase such that it's performed in a _read_ lock after the _write_ is finished.

spoenemann:
This PR does much more than what was requested in #379. I'm currently questioning whether the added complexity is justified by the gain in user experience.

What is the difference from the user perspective?
 - Previously: LSP requests could potentially fail to obtain the required information if the DocumentBuilder is not done yet. This would likely result in empty responses (to be verified).
 - With this change: LSP requests are held back until the DocumentBuilder is done. Responses are delivered with a delay.

Potential problem: if the delay between LSP request and the corresponding response is too high, the client will either (A) discard the response or (B) show it anyway. Case A means we have no advantage from the added complexity. Case B is worse because it would show something to the user although they're already moved on and busy with something else.

In general I think it's important to verify the assumptions and compare the behavior with and without this change.

packages/langium/src/utils/promise-util.ts
```diff
@@ -73,7 +73,119 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
...
+            this.writeLock = new Deferred();
+        }
+        this.writeCount++;
+    }
+
+    /**
+     * Call this to unlock this mutex after a write operation.
+     */
+    private unlockRead(): void {
+        this.writeCount--;
```
montymxb, line 136: I believe there's a potential issue here if `unlockRead` is called when no previous writeLock has been acquired. That would drop the writeCount below zero, and result in some misbehavior with subsequent locks. It would be a good idea to do something to guard against this, either throwing something or silently preventing such cases. This depends on the implementation of course, and a good usage of it may never trip the issue, but it would still be prudent to take care of.

packages/langium/src/utils/promise-util.ts
```diff
@@ -73,7 +73,119 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
...
+            this.readLock = new Deferred();
+        }
+        this.readCount++;
+    }
+
+    /**
+     * Call this to lock this mutex after a read operation.
+     */
+    private unlockWrite(): void {
+        this.readCount--;
```
montymxb, line 172: Same issue here as above with `unlockRead`, it would be best to do something explicit in the case that no previous lock has been acquired.

packages/langium/src/utils/promise-util.ts
```diff
@@ -73,7 +73,119 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
...
+    /**
+     * Performs a read action on the language server. Multiple read actions can be performed at a time.
+     *
+     * If a read action is queued up while a write action is being performed, the read action will wait until the write action has completed.
+     */
+    async read<T>(action: () => Promise<T>): Promise<T> {
+        // Wait one tick for any write request to arrive
+        await delayNextTick();
+        // Await that writing has completed
+        await this.writeLock.promise;
```
montymxb, line 151: If repeated write requests were sent one after the other, couldn't this lead to the read requests never being handled? The idea of blocking all subsequent reads after a write request makes sense, but at some point it would be nice to come back and handle them. I think it may be helpful to go with a queue design instead of a priority queue. Sequenced reads can be still be performed in parallel, but writes block only the next sequence of operations, rather than blocking only subsequent reads.

dhuebner: Writing happens (at least for now) only when user is typing. With auto save I could be quite frequent. With each new write on the Document the old state becomes outdated and therefor obsolet. I  think the idea is to prevent  e.g. an outline view or a validator from rendering or checking the old state. I also don't like the idea of prio ordering that much, at least not by default. I could imaging having a special request read type that is marked as low prio.

msujew: @dhuebner Saving is unrelated to this, since the file has to be kept in sync with the editor (on every keystroke), not the file system.

dhuebner: @msujew 
I see, I thought the discussion is around the Document access :)

packages/langium/src/utils/promise-util.ts
```diff
@@ -73,7 +73,46 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
...
+        // Append the new action to the previous action. We usually don't have to wait for long, as the previous action
+        // 1. has either completed
+        // 2. has been cancelled due to the new write request
+        return this.previousAction = this.previousAction.then(() =>
+            action().catch(err => {
+                if (!isOperationCancelled(err)) {
+                    console.error('Error: ', err);
+                }
+            })
+        );
```
spoenemann, line 102: Suggestion:
```typescript
        return this.previousAction = this.previousAction.then(
            action,
            err => {
                if (!isOperationCancelled(err)) {
                    console.error('Error: ', err);
                }
            }
        );
```

spoenemann (approved):
Thanks! Let's get back to your idea of a read-write lock when we see actual problems with readers (LSP requests) while documents are being built.

packages/langium/src/utils/promise-util.ts
msujew: This actually breaks the error recovery, since this isn't the same as `action.catch(err)`

PR #551, opened 2022-06-20
Validates actions do not redefine declared types

### LINKED ISSUE #514: Actions can rewrite declared types
Actions can rewrite declared types, effectively erasing/changing their previously declared properties. This leads to an inability to reason about declared types as a fixed source of information about the resulting AST that is generated.

An example grammar in Langium that causes this:

```typescript
// 1. define some interface
interface A {
    val: string
}

// X inherits correctly from A via a code action, leaving A unchanged
X: 'x' {A} val=ID;

// but Y redefines A entirely
Y: 'y' {infer A} q='broken';
```
*assumes the ID terminal is declared*

This generates the following interface, which is not correct:
```typescript
    q: 'broken'
}
```
Langium version: 0.3.0 (dev)

Instead:
- declared types should be immutable, their properties should not be open to changing, erasure, or addition.
- validation should indicate something like `A is a declared type and cannot be inferred`
- optional: add a code action to rewrite `{infer A}` to `{A}` (could potentially be in a separate PR)

montymxb:
Closes #514 by adding an additional check in the **langium-grammar-validator.ts**. This check ensures that an action that infers a type with the same name as an explicitly declared type is invalid. Additionally, this provides a small code action that offers to correct the validation error by removing the `infers` keyword from the action. A subsequent test is put in place to verify that the validation is tripped, but the code action itself is untested.

Also added a small utility `getProperty` function in the **inferred-types.spec.ts** file to abstract property retrieval from interfaces. Previously I was using this in my test design, but I ended up dropping it. Still, I found it was nice to abstract away the two instances of looking for a property via `find`.


packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -291,6 +291,17 @@ export class LangiumGrammarValidator {
...
                     });
+                } else if(actionType && types.has(typeName) && isInfers) {
+                    // error: action infers type that is already defined
+                    if(action.$cstNode) {
+                        accept('error', `${typeName} is a declared type and cannot be redefined.`, {
+                            node: action,
+                            property: 'type',
+                            code: IssueCodes.SuperfluousInfer,
+                            // use the same data, stripping the 'infer' keyword out
+                            data: action.$cstNode.text.replace(/\s*infer\s+/, '')
```
montymxb, line 302: A specific call for feedback here. I was having some trouble deciding how to go about removing the infers. Initially I thought I was going to dig down into the keyword `infer` node in the Cst, and remove it entirely. However this ended up not working as expected. This approach feels cleaner than what I was doing previously, but I'm open to hear any alternatives that can more directly apply for removing the `infer` keyword.

packages/langium/test/grammar/type-system/inferred-types.spec.ts
```diff
@@ -595,13 +595,18 @@ function getType(types: AstTypes, name: string): InterfaceType | UnionType | und
     return interfaceType || unionType;
 }
```
Lotes, line 596: I think the file should be named with the suffix .test.ts

montymxb: I have no issue with that 👍 , it is the odd one out in this folder.

msujew (commented):
This works in all cases, but leaves some formatting to be desired, since it never strips the whitespace between the `infer` keyword and the type name. I believe it'd be fine to use the `{` keyword from the action as a starting marker and simply remove everything until the type name. There could be an edge case with removed comments in there, but it's reasonably unlikely, and can always be reinstated using the undo command.

montymxb:
Added a commit for using end of opening curly brace range, and start of type name range as bounds for removing the `infer` keyword in an action. Everything else remains the same as before.

msujew (approved):
The changes look good to me. Do you mind rebasing the branch?

montymxb:
@msujew yep I can do that later tonight, and bring it in 👍 .

PR #553, opened 2022-06-20
Inititialize LSP services with client capabilities

msujew:
Closes https://github.com/langium/langium/issues/552

This PR became larger than expected:

1. Fixes the issue described in https://github.com/langium/langium/issues/552 by keeping a token builder for each open document.
2. Gives each LSP service an `initialize` method that can be used to optimize behavior based on frontend (editor) capabilities.
3. Based on 2, if the frontend does not support multi line tokens, split each semantic token into multiple tokens, one for each line. Addresses an issue brought up in https://github.com/langium/langium/discussions/524

### DIFF
--- a/packages/langium/src/services.ts
+++ b/packages/langium/src/services.ts
@@ -33,7 +33,7 @@ import type { ScopeComputation, ScopeProvider } from './references/scope';
 import type { JsonSerializer } from './serializer/json-serializer';
 import type { ServiceRegistry } from './service-registry';
 import type { AstReflection } from './syntax-tree';
-import type { MutexLock } from './utils/promise-util';
+import type { MaybePromise, MutexLock } from './utils/promise-util';
 import type { DocumentValidator } from './validation/document-validator';
 import type { ValidationRegistry } from './validation/validation-registry';
 import type { AstNodeDescriptionProvider, ReferenceDescriptionProvider } from './workspace/ast-descriptions';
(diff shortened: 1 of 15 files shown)


spoenemann (approved):
Great, thanks!

packages/langium/src/services.ts
```diff
@@ -75,6 +76,15 @@ export type LangiumLspServices = {
     Formatter?: Formatter
 }
 
+export interface InitializableService<T extends object> {
+    /**
+     * Initializes the service with client (IDE) related capabilities.
+     *
+     * This method is only executed if Langium is running as part of a language server.
+     */
+    initialize?(clientCapabilities?: T): MaybePromise<void>
```
Lotes, line 85: A naive question: When there is setup code, do we also need teardown code for removing the created infrastructure or whatever was used? Then this interface could be a life cycle service.

msujew: We can maybe add something like this later, but for now there's no need for it. We don't really "setup" anything here, but rather notify the LSP service of any existing client capabilities. The implementation is usually quite simplistic:

https://github.com/langium/langium/blob/245629e79008aea422c0cde36d0cb018453fbd2c/packages/langium/src/lsp/semantic-token-provider.ts#L163-L165

Lotes: It feels like a code smell to have this optional `initialize` method everywhere. Would it make sense to approach this with a life cycle event, like subscribing to the event when client capabilities are ready and then pick the information out of it?

PR #558, opened 2022-06-21
Make workspace initialization cancellable

msujew:
Closes https://github.com/langium/langium/issues/555

The `MutexLock#lock` method now gives each caller a cancellation token that they can use to abort execution. The token is cancelled whenever another action requests the lock.

### DIFF
--- a/packages/langium/src/utils/promise-util.ts
+++ b/packages/langium/src/utils/promise-util.ts
@@ -84,23 +84,36 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
 export class MutexLock {
 
     private previousAction = Promise.resolve();
+    private previousTokenSource = new CancellationTokenSource();
 
     /**
      * Performs a single async action, like initializing the workspace or processing document changes.
      * Only one action will be executed at a time.
+     *
+     * When another action is queued up, the token provided for the action will be cancelled.
+     * Assuming the action makes use of this token, the next action only has to wait for the current action to finish cancellation.
      */
-    async lock(action: () => Promise<void>): Promise<void> {
+    lock(action: (token: CancellationToken) => Promise<void>): Promise<void> {
+        this.cancel();
+        const tokenSource = new CancellationTokenSource();
+        this.previousTokenSource = tokenSource;
         // Append the new action to the previous action. We usually don't have to wait for long, as the previous action
         // 1. has either completed
-        // 2. has been cancelled due to the new write request
+        // 2. has been cancelled
         return this.previousAction = this.previousAction.then(
-            action,
-            err => {
+            () => action(tokenSource.token).catch(err => {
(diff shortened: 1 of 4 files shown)


packages/langium/src/utils/promise-util.ts
```diff
@@ -84,22 +84,25 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
...
     private previousAction = Promise.resolve();
+    private previousTokenSource = new CancellationTokenSource();
 
     /**
      * Performs a single async action, like initializing the workspace or processing document changes.
      * Only one action will be executed at a time.
      */
-    async lock(action: () => Promise<void>): Promise<void> {
+    lock(action: (token: CancellationToken) => Promise<void>): Promise<void> {
+        this.previousTokenSource.cancel();
```
Lotes, line 94: You are cancelling the current task in progress? Or how to understand this?

packages/langium/src/utils/promise-util.ts
```diff
@@ -84,22 +84,25 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
...
 
     private previousAction = Promise.resolve();
+    private previousTokenSource = new CancellationTokenSource();
 
     /**
      * Performs a single async action, like initializing the workspace or processing document changes.
      * Only one action will be executed at a time.
      */
-    async lock(action: () => Promise<void>): Promise<void> {
+    lock(action: (token: CancellationToken) => Promise<void>): Promise<void> {
```
Lotes, line 93: Testing this would be great. Also for documenting how it is used...

msujew: I've added some additional tests for this. The `lock` function is already tested, but the cancellation wasn't yet.

packages/langium/src/utils/promise-util.ts
msujew: Yep, that the idea. We want to stop processing the current workspace change in favor of processing the newest changes instead.

Lotes: I see a bit a contradiction. Because you can use a Mutex everywhere. Does this mean if two asynchronous processes acquire the lock, the last one will always cancel the first one? I do not see any relation to a workspace HERE in this file. Because HERE is very generic and not WORKSPACE only.

msujew: I don't really see a contradiction here:

1. We don't force to cancel the execution (by some v8 nodejs magic). We simply notify that a new action has arrived, and that the action can return from execution. They don't have to listen to this cancellation if they don't want to.
2. You can create mutexes for any use case you want, we use a single mutex for workspace writing activities. Another use case would use a separate mutex, which has it's own execution context.

spoenemann (approved):
Great solution!

PR #568, opened 2022-06-27
Enable browser code bundling without polyfills

msujew:
This change removes the dependency on the `os` package and the `setImmediate` function (which is only available in node).

It also splits Langium into a browser and node package area. Both export a createDefaultSharedModule function which should be used in the respective environment (browser can be used anywhere, node only in node.js). This enables us to bundle Langium for the browser without needing to alias the fs library or any other polyfill.


kaisalmen (commented):
I was just wondering what the meaning of the default export now is? Should the default target be node and the browser just be a second export?

Is this the time to change to ESM as target as well?

packages/langium/package.json
```diff
@@ -19,6 +19,10 @@
   "files": [
     "lib",
     "src",
+    "browser.js",
+    "browser.d.ts",
+    "node.js",
+    "node.d.ts",
```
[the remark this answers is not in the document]
kaisalmen: It could become this also this:
```
  "main": "lib/node/index.js",
  "types": "lib/node/index.d.ts",
    ".": {
      "types": "./lib/node/index.d.ts",
      "default": "./lib/node/index.js"
    },
    "./browser": {
      "types": "./lib/browser/index.d.ts",
      "default": "./lib/browser/index.js"
    }
  },
  "typesVersions": {
    "*": {
      "browser": [ "lib/browser" ]
    }
  },
```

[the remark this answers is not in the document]
msujew: > then the default exports links directly to node

I don't think that's a good idea, since we want to import from `langium` for runtime independent code, that we can later reuse for `langium/node` and `langium/browser` based language servers.

But otherwise, I agree 👍

packages/langium/package.json
```diff
@@ -18,14 +18,43 @@
   "license": "MIT",
   "files": [
     "lib",
-    "src",
-    "test.js",
-    "test.d.ts"
+    "src"
   ],
```
kaisalmen, line 22: Shouldn't you now add:
```
"type": "module",
```
[new type in package.json](https://www.typescriptlang.org/docs/handbook/esm-node.html#type-in-packagejson-and-new-extensions). I guess we/I should have done that in other projects as well, but we didn't. According to the doc it should be the case.

kaisalmen (dismissed):
LGTM when the type=module question is clarified.

kaisalmen:
It seems we need CommonJS builds for VSCode extensions for now. So, either we stay with `module=node`  for now or we follow a ESM + CJS output strategy with different exports as long as CJS modules are still required.

spoenemann (commented):
The `browser` module implies that we could use browser API there, but we can't because it's actually "common" code and is even used internally by the main framework. And I agree with the remark of @kaisalmen that the meaning of the default import is now unclear.

I propose the following:
 - Move the `fileSystemProvider` property from `CommonSharedModuleContext` to `DefaultSharedModuleContext`.
 - Move `createDefaultSharedModule` from `browser-module.ts` back to `default-module.ts`.
 - Move `createLangiumGrammarServices` back to `langium-grammar-module.ts`.
 - Add this to `file-system-provider.ts`:
   ```typescript
   export const emptyFileSystem = {
        fileSystemProvider: () => new EmptyFileSystemProvider()
    }
   ```
 - Export this in `node-module.ts`:
   ```typescript
   export const nodeFileSystem = {
        fileSystemProvider: () => new NodeFileSystemProvider()
    }
   ```
 - Don't reexport `../index` in the `node` module, remove the `browser` module.

Now we decide whether to use the Node file system when calling the `createMyLanguageServices` function. All language server `main.ts` files would have to be rewritten like this:
```typescript
const connection = createConnection(ProposedFeatures.all);
const { shared } = createLangiumGrammarServices({ ...nodeFileSystem, connection });
startLanguageServer(shared);
```

WDYT?

examples/statemachine/src/language-server/main-browser.ts
```diff
@@ -0,0 +1,18 @@
+/******************************************************************************
+ * Copyright 2022 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { startLanguageServer, EmptyFileSystem } from 'langium';
+import { createConnection, ProposedFeatures } from 'vscode-languageserver/node';
```
spoenemann, line 8: Shouldn't this be imported from `'vscode-languageserver/browser'`?

packages/langium/src/default-module.ts
```diff
@@ -97,14 +97,25 @@ export function createDefaultModule(context: DefaultModuleContext): Module<Langi
...
+    /**
+     * Represents an abstract language server connection
+     */
     connection?: Connection;
+    /**
+     * Factory function to create a {@link FileSystemProvider}.
+     *
+     * Langium exposes an `EmptyFileSystem` and `NodeFileSystem`, exported through `langium/node`.
+     * When running Langium as part of a vscode language server, using the `NodeFileSystem` is recommended,
+     * the `EmptyFileSystem` in every other use case.
```
spoenemann, line 109: NodeFileSystem is also required in Node.js CLIs, e.g. to resolve cross-references to other files.

packages/langium/src/utils/promise-util.ts
```diff
@@ -8,13 +8,20 @@ import { AbstractCancellationTokenSource, CancellationToken, CancellationTokenSo
 
 export type MaybePromise<T> = T | Promise<T>
 
+/**
+ * This is a function that can be promisified to delay execution to the next tick of the event loop
+ * In case we are running in a non-node environment, `setImmediate` isn't available.
+ * Using `setTimeout` of the browser API accomplishes the same result.
+ */
+const delayImmediate = setImmediate || setTimeout;
```
spoenemann, line 16:
```suggestion
const delayImmediate = setImmediate ?? setTimeout;
```

spoenemann (commented):
Looks good!

tsconfig.json
```diff
@@ -6,7 +6,7 @@
     // "incremental": true,                   /* Enable incremental compilation */
     "target": "ES2017",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
     "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
-    "lib": ["ESNext"],                        /* Specify library files to be included in the compilation. */
+    "lib": ["ESNext", "WebWorker"],           /* Specify library files to be included in the compilation. */
```
spoenemann, line 9: Do we need this here?

spoenemann (approved):

tsconfig.json
msujew: Yes, the `self` in `main-browser` isn't available otherwise (the constant is only available in a webworker context).

spoenemann (approved):

PR #570, opened 2022-06-28
Browser context: Detect if process is undefined

kaisalmen:
Just saw this happening with a new worker. Good news: With the fix it looks good with local builds.

Side question: Should we set a formatter for vscode (see auto-corrected spaces)?


msujew (approved):
LGTM 👍 

> Side question: Should we set a formatter for vscode (see auto-corrected spaces)?

Yeah, we should likely activate some more eslint rules to catch these issues.

PR #575, opened 2022-07-01
Refactored References service and related LSP services

gfontorbe:
This PR refactors the `References` service to give it more responsibilities, enabling out-of-the-box improvement to LSP services relying on references.

* `findDeclaration` finds the location of the declaration of a cross-reference, it now also find the type attribute in an interface related to an assignment in a parser rule
*  `findReferences` finds references to an AstNode globally or locally.

This enables some LSP services to work (with some minor adjustments) without the need to have a custom service. This concerns:
* `GoToSolver`
* `ReferenceFinder`
* `DocumentHighlighter`
* `RenameHandler`
* `HoverProvider`

There were some hurdles in the refactoring so there's potentially improvements to be made and some better organizing of the code.

I will create tests for the related LSP services later on in a separate PR.


packages/langium/src/references/references.ts
```diff
@@ -71,15 +77,239 @@ export class DefaultReferences implements References {
...
+                if (ref) {
+                    refs.push(ref);
+                }
+            }
+            targetRules.forEach(rule => {
+                const assignment = isParserRule(rule) ?
+                    extractAssignments(rule.definition).find(a => a.feature === target.name)
+                    : getContainerOfType(rule, isGroup)!.elements.find(el => isAssignment(el) && el.feature === target.name);
+                if (assignment?.$cstNode) {
+                    const leaf = findLeafNodeAtOffset(assignment.$cstNode!, assignment.$cstNode!.offset);
```
msujew, line 179: Suggestion: Using the `$cstNode` of the whole assignment to get the offset of the assigned property isn't recommended here. Instead, you should find the correct node by calling `findNodeForFeature(assignment.$cstNode, 'feature')`.
```suggestion
                    const leaf = findNodeForFeature(assignment.$cstNode, 'feature');
```

msujew (commented):
This is already going in the right direction, good job 👍 

There are some architectural issues in there, but they can be easily resolved. There's one unexpected behavior which I couldn't pinpoint in the code: When placing the cursor on an assignment which doesn't belong to some interface, the assignment isn't highlighted at all. It seems like some code in the `findDeclaration` method makes the method return `undefined` when on top of an assignment.

packages/langium/src/references/references.ts
```diff
@@ -32,18 +36,20 @@ export interface References {
      *
      * @param targetNode Specified target node whose references should be returned
      */
-    findReferences(targetNode: AstNode): Stream<ReferenceDescription>;
+    findReferences(target: AstNode, onlyLocal: boolean, includeDeclaration: boolean): Stream<ReferenceDescription>;
 }
```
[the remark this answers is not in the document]
gfontorbe: Should we make all of the parameters part of `options`?

I would tend to leave `target` outside of it and make something like this:
```ts
findReferences(target: AstNode, options: FindReferencesOptions): Stream<ReferenceDescription>;

export interface FindReferencesOptions {
    onlyLocal?: boolean
    includeDeclaration?: boolean
}
```

packages/langium/src/references/references.ts
```diff
@@ -71,15 +77,239 @@ export class DefaultReferences implements References {
                         || nameNode && nameNode.offset <= sourceCstNode.offset
                         && nameNode.offset + nameNode.length > sourceCstNode.offset) {
                         return nameNode;
+                    } else if (isAssignment(nodeElem)) {
+                        return this.findAssignmentDeclaration(nodeElem);
```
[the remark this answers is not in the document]
gfontorbe: Refactoring on its way

gfontorbe:
@msujew 
* Bug with highlight of assignment in parser rule with no return type was due to missing logic in the `findAssignmentDeclaration` function. 
* I made a new interface `FindReferencesOptions` without the target AstNode inside (can be changed)
* Moved most (all?) of the code to a new file
* Replaced `findLeafNodeAtOffset` occurrences where possible

packages/langium/src/grammar/references/langium-grammar-references.ts
```diff
@@ -0,0 +1,248 @@
...
+        }
+        return super.findDeclaration(sourceCstNode);
+    }
+
+    findReferences(target: AstNode, options: FindReferencesOptions): Stream<ReferenceDescription> {
+        if (options.onlyLocal) {
+            return this.findLocalReferences(target, options.includeDeclaration);
+        } else {
+            return this.findGlobalReferences(target, options.includeDeclaration);
+        }
```
msujew, line 41: Suggestion: As mentioned above, the default implementation should respect the `onlyLocal` property already and ideally provide a template method to modify the behavior of both execution flows.

packages/langium/src/lsp/reference-finder.ts
```diff
@@ -56,13 +56,8 @@ export class DefaultReferenceFinder implements ReferenceFinder {
...
         const targetAstNode = this.references.findDeclaration(selectedNode)?.element;
         if (targetAstNode) {
-            if (params.context.includeDeclaration) {
-                const declDoc = getDocument(targetAstNode);
-                const nameNode = this.findNameNode(targetAstNode, selectedNode.text);
-                if (nameNode)
-                    refs.push(Location.create(declDoc.uri.toString(), nameNode.range));
-            }
-            this.references.findReferences(targetAstNode).forEach(reference => {
+            const options = {includeDeclaration: params.context.includeDeclaration};
```
msujew, line 59: Nitpick: Formatting
```suggestion
            const options = { includeDeclaration: params.context.includeDeclaration };
```

msujew (commented):
Looks already way better. The default behavior of the `onlyLocal` option property needs to be improved, but this is already close to a satisfying solution 👍

gfontorbe:
@msujew I think we are getting closer to the expected code structure

packages/langium/src/grammar/references/langium-grammar-references.ts
```diff
@@ -0,0 +1,182 @@
...
+                if (ref) {
+                    refs.push(ref);
+                }
+            }
+            targetRules.forEach(rule => {
+                const assignment = isParserRule(rule) ?
+                    extractAssignments(rule.definition).find(a => a.feature === target.name)
+                    : getContainerOfType(rule, isGroup)!.elements.find(el => isAssignment(el) && el.feature === target.name);
+                if (assignment?.$cstNode) {
+                    const leaf = findLeafNodeAtOffset(assignment.$cstNode!, assignment.$cstNode!.offset);
```
msujew, line 75: Suggestion: This doesn't return correct results. It does when we try to find the node that belongs to the `feature` property of the assignment.
```suggestion
                    const leaf = findNodeForFeature(assignment.$cstNode, 'feature');
```

packages/langium/src/lsp/document-highlighter.ts
```diff
@@ -49,14 +49,13 @@ export class DefaultDocumentHighlighter implements DocumentHighlighter {
...
-                const nameNode = this.findNameNode(targetAstNode);
-                if (nameNode) {
-                    refs.push([nameNode, this.getHighlightKind(nameNode)]);
+            const includeDeclaration = equalURI(getDocument(targetAstNode).uri, document.uri);
+            const options = {onlyLocal: true, includeDeclaration};
+            this.references.findReferences(targetAstNode, options).forEach(ref => {
+                const leaf = findLeafNodeAtOffset(rootNode, ref.segment.offset);
+                if (leaf) {
+                    refs.push([leaf, this.getHighlightKind(leaf)]);
                 }
```
msujew, line 58: Suggestion: Trying to find the leaf of the offset here seems a bit usual. The old API required that, but the new doesn't. Something like this is more appropriate here:

```suggestion
                refs.push([ref.segment.range, this.getHighlightKind(ref)]);
```

Note that you'll have to change some signatures for that. `getHighlightKind` should only be as simple as:

```ts
/**
 * Override this method to determine the highlight kind of the given reference.
 */
protected getHighlightKind(reference: ReferenceDescription): DocumentHighlightKind {
  return DocumentHighlightKind.Read;
}
```

packages/langium/src/grammar/references/langium-grammar-references.ts
```diff
@@ -0,0 +1,182 @@
...
+            interfaces.forEach(interf => {
+                const rules = this.findRulesWithReturnType(interf);
+                targetRules.push(...rules);
+            });
+            targetRules.forEach(rule => {
+                const assignment = isParserRule(rule) ?
+                    extractAssignments(rule.definition).find(a => a.feature === target.name)
+                    : getContainerOfType(rule, isGroup)!.elements.find(el => isAssignment(el) && el.feature === target.name);
+                if (assignment?.$cstNode) {
+                    const leaf = findLeafNodeAtOffset(assignment.$cstNode!, assignment.$cstNode!.offset);
```
msujew, line 111: Suggestion: Same here
```suggestion
                    const leaf = findNodeForFeature(assignment.$cstNode, 'feature');
```

packages/langium/src/grammar/references/langium-grammar-references.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    protected findReferencesToTypeAttribute(target: TypeAttribute): Stream<ReferenceDescription> {
+        const refs: ReferenceDescription[] = [];
+        const interfaceNode = getContainerOfType(target, isInterface);
+        if (interfaceNode) {
+            const ref = this.getReferenceToSelf(target);
+            if (ref) {
+                refs.push(ref);
+            }
```
msujew, line 99: Minor: I believe this should only be executed when `includeDeclaration` is `true`. right?

packages/langium/src/grammar/references/langium-grammar-references.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    protected findGlobalReferences(target: AstNode, includeDeclaration = false): Stream<ReferenceDescription> {
+        if (isTypeAttribute(target)) {
+            return this.findReferencesToTypeAttribute(target);
+        } else {
+            return super.findGlobalReferences(target, includeDeclaration);
+        }
+    }
+
+    protected findLocalReferencesToTypeAttribute(target: TypeAttribute, rootNode: AstNode): Stream<ReferenceDescription> {
```
msujew, line 54: This also doesn't respect the `includeDeclaration` parameter.

packages/langium/src/lsp/rename-refactoring.ts
```diff
@@ -79,7 +82,7 @@ export class DefaultRenameHandler implements RenameHandler {
             }
             const isCrossRef = this.references.findDeclaration(leafNode);
             // return range if selected CstNode is the name node or it is a crosslink which points to a declaration
-            if (isCrossRef || this.isNameNode(leafNode)) {
+            if (isCrossRef || isAssignment(leafNode.element) || this.isNameNode(leafNode)) {
```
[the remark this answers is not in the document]
gfontorbe: You are correct, we don't need this anymore with the refactored `findDeclaration`. Most likely some leftover code from my previous attempt to rework the renameHandler

packages/langium/src/lsp/document-highlighter.ts
gfontorbe: Seems like we don't need the `getHighlightKind` function anymore

msujew: It's a template method, that allows downstream users to easily modify some behavior, even if the default behavior is just to return `DocumentHighlightKind.Read`. Removing that would also make it way harder to override this minor (but important) detail.

gfontorbe:
Thanks you @msujew for taking care of the the action assignments. This is something I completely overlooked in the initial PR. 

I checked the behavior manually and it all seem fine to me.

msujew (approved):
Thanks @gfontorbe the changes look good to me!

PR #579, opened 2022-07-05
Support MultiMode Lexing

montymxb:
This PR makes a tiny change in Langium to support multimode lexing. This was a matter of swapping the order of the lexer & ChevrotrainWrapper initialization. The underlying problem was a subtle issue in Chevrotain that is related to the order that these are initialized in.

This also adds tests to reinforce expected multimode lexing behavior, and to hopefully provide a simple example for how this can be done in Langium for other projects.


msujew (commented):
Looking quite good, I just have a minor comment regarding the `createServicesForGrammar` usage.

montymxb:
Changes noted before have been added :octocat:

msujew (approved):
LGTM, thanks for taking care of this 👍

PR #586, opened 2022-07-08
Take imported types into account when validating types

### LINKED ISSUE #507: Interface defines in a Langium file isn't accessible from another
When I try to separate my grammar and my model in two differents Langium File and add an import, I cannot return an element of one of my defined types. If I paste one of my interfaces in the grammar file, an error is thrown because the interface already exists on an imported file.

image
image

I don't really know if this is a bug or an expected behaviour, but if it's intended, it's strange in this situation to throw error for an interface declared in another file.

Langium version: 0.3.0

## Steps To Reproduce

1. Create Hello World project
2. Replace the grammar with this (I keep default rules to avoid errors with Langium CLI):
```
grammar HelloWorld

import 'foo'

entry Model returns Foo:
    'model' name=ID;

Model2 infers Model:
    (persons+=Person | greetings+=Greeting)*;

Person:
    'person' name=ID;

Greeting:
    'Hello' person=[Person:ID] '!';

hidden terminal WS: /\s+/;
terminal INT returns number: /[0-9]+/;
terminal STRING: /"[^"]*"|'[^']*'/;

```
3. Create a new file `foo.langium` in the same directory than the existing grammar and paste this:
```
interface Foo {
    name: string
}
```

Link to code example:
https://github.com/theogiraudet/Minimal-Langium-Model-Error.git

## The current behavior
The interface is imported but cannot be used as return type.

## The expected behavior
The interface can either be used as a return type or be undetected, or display a custom message indicating that it is not possible.

dhuebner:
Fixes #507
Also changed min node version from 12 to 14 in CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ Thank you for your interest in the Langium project! The following is a set of gu
 
 ## Prerequisites
 
-For developing Langium you require at least the Maintenance LTS version of Node.js (currently 12) and at least npm version 7.7.0 (`npm@^7.7.0`) to be able to use npm workspaces. With Node.js versions below 16 you have to install the correct version of npm with `npm install -g npm@7`.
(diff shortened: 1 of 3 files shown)


msujew (approved):
Thanks, looks good to me, just one minor issue 👍

PR #590, opened 2022-07-08
Separate inject package

danieldietrich:


spoenemann:
`inject` is already taken: https://www.npmjs.com/package/inject

Maybe a variation like `ginject`?

PR #592, opened 2022-07-11
Support `workspace/executeCommand` LSP request

msujew:
Introduces support for the `workspace/executeCommand` which allows us to call into the language server, execute a provided command, and return the result to the caller.

### DIFF
--- a/packages/langium/src/lsp/language-server.ts
+++ b/packages/langium/src/lsp/language-server.ts
@@ -63,6 +63,7 @@ export class DefaultLanguageServer implements LanguageServer {
         const formattingOnTypeOptions = languages.map(e => e.lsp.Formatter?.formatOnTypeOptions).find(e => !!e);
         const hasCodeActionProvider = languages.some(e => e.lsp.CodeActionProvider !== undefined);
         const hasSemanticTokensProvider = languages.some(e => e.lsp.SemanticTokenProvider !== undefined);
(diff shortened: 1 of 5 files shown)


packages/langium/src/lsp/language-server.ts
```diff
@@ -34,6 +35,9 @@ export function startLanguageServer(services: LangiumSharedServices): void {
                         supported: true
                     }
                 },
+                executeCommandProvider: commandNames && {
```
dhuebner, line 38: Q: Would it only assign `{commands: commandNames}` if `commandNames ` is defined?

msujew:
@dhuebner Could be possible, but I imagine simply restarting the language server by restarting the language-client is a better idea, as it would restart the process. That can be done from the extension.

dhuebner:
>  by restarting the language-client is a better idea, as it would restart the process. That can be done from the extension.

I'm thinking about a command like "Developer: Restart TS server",  it is useful when e.g. working on workspace manager like changing excluded folders from indexing

packages/langium/src/lsp/language-server.ts
msujew: A: Exactly, the `&&` is a short-circuit operator in this case. The whole operation returns `undefined` if `commandNames` is undefined, and `{commands: commandNames}` otherwise.

PR #593, opened 2022-07-11
Use LSP lifecycle events instead of methods

msujew:
Closes https://github.com/langium/langium/issues/557

Also introduces a way to eagerly load all service instances, since we register the lifecycle events in the constructor. As the constructor could be called after the lifecycle event is fired, we need a way to make sure all services are loaded.


Lotes:
Actually nothing to lift the finger. LGTM. Maybe you can find a way in the dependency injection not to use any type…

Lotes (approved):

PR #607, opened 2022-07-19
Always perform workspace initialization

msujew:
Closes https://github.com/langium/langium/issues/606

Performs workspace initialization even if `workspaceFolders` is `null`. This ensures that `loadAdditionalDocuments` is always called. Also adds a bit more comments on the concurrency model of the `initializeWorkspace` method.


montymxb (approved):
Changes look good. For a moment was wondering about the empty promise for `loadAdditionalDocuments`, but I see this is just the default.

PR #610, opened 2022-07-25
Various improvements to CST behavior

msujew:
Improves the behavior of different LSP services, which operate directly on the CST

* Improves comment finding capabilities if the comment is appended to the root of the document
* Align some off-by-one errors in LSP services to expectations of vscode. When working with names/IDs, vscode expects that they are still found when the cursor is at the end of the token. However, in Langium, we tried to work with the next token, which lead to some unexpected behavior.
  * document highlight
  * renaming
  * go to declaration
  * hover
  * reference finding

### DIFF
--- a/packages/langium/src/utils/cst-util.ts
+++ b/packages/langium/src/utils/cst-util.ts
@@ -6,8 +6,7 @@
 
 import { IToken } from '@chevrotain/types';
 import { Range } from 'vscode-languageserver';
-import { DatatypeSymbol } from '../parser/langium-parser';
-import { AstNode, CstNode, CompositeCstNode, isCompositeCstNode, isLeafCstNode, LeafCstNode } from '../syntax-tree';
+import { CstNode, CompositeCstNode, isCompositeCstNode, isLeafCstNode, LeafCstNode, isRootCstNode } from '../syntax-tree';
 import { DocumentSegment } from '../workspace/documents';
 import { Stream, TreeStream, TreeStreamImpl } from './stream';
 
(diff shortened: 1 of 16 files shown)


gfontorbe:
I was looking at how the new CST utils would behave in combination with #575.

Maybe I did something wrong when combining the 2 PRs (something that is clearly not impossible) but I get some strange behavior on the renaming requests.

Let's consider the following:

```
interface A {
    foo: string[]
}

RuleA returns A: 
    foo+=ID;
```

Let's assume we trigger a rename request from the cursor positioned between `+` and `=`. The renaming text that is proposed is `+=` but what is renamed is `foo`.

I haven't fully figured out yet what is causing this, but I think we either should change what is being returned by the method `renameNodeRange` in `rename-refactoring.ts` or change the behavior of `findDeclaration` in the References service so that it does not return the CstNode containing `foo`.

msujew:
@gfontorbe I wouldn't worry too much about that, I planned to merge this only after rebasing on #575, since it modifies quite heavily what we do while resolving references in LSP services.

msujew:
@gfontorbe I rebased my changes and addressed the issue you were experiencing 👍

packages/langium/src/utils/cst-util.ts
```diff
@@ -57,29 +56,51 @@ export function toDocumentSegment(node: CstNode): DocumentSegment {
...
+ *
+ * For LSP services, users expect that the declaration of an element is available if the cursor is directly after the element.
+ */
+export function findDeclarationNodeAtOffset(cstNode: CstNode | undefined, offset: number): LeafCstNode | undefined {
+    if (cstNode) {
+        const localOffset = offset - cstNode.offset;
+        const textAtOffset = cstNode.text.substring(localOffset, localOffset + 1);
+        // The \p{L} regex matches any unicode letter character, i.e. characters from non-english alphabets
+        // Together with \w it matches any kind of character which can commonly appear in IDs
+        if (/^[^\w\p{L}]$/u.test(textAtOffset)) {
```
spoenemann, line 71: How do we handle languages with name patterns containing non-word characters, like `kebab-case` identifiers? Do they need to override all services using this function?

msujew: I agree, this would be limited to rare cases such as `function(identifier-|)` (with `|` being the cursor position), but that's still highly language dependent. Instead I will refactor this function to operate on a terminal+keyword condition.

msujew: I've made the function a bit better configurable. Users can now override the `GrammarConfig` DI object to control the behavior of the regexp.

spoenemann (approved):
:+1:

PR #611, opened 2022-07-27
Prevent stack overflow in completion

msujew:
Adds a `visited` set that keeps track of already seen grammar features. The stack overflow can be experienced in some parts of the langium-lox language example.


packages/langium/src/lsp/completion/follow-element-computation.ts
```diff
@@ -68,27 +73,27 @@ export function findNextFeaturesInternal(featureStack: ast.AbstractElement[], ca
  * @returns A list of features that could be the first feature of the given `AbstractElement`.
  * These features contain a modified `cardinality` property. If the given `feature` is optional, the returned features will be optional as well.
```
montymxb, line 74: Could we also update the function header to capture the new changes?

msujew:
Closing in favor of a more intricate refactoring, see https://github.com/langium/langium/pull/623.

PR #618, opened 2022-08-03
Implemented go to type service

gfontorbe:
This PR adds a service to resolve GoTo Types Requests.

It implements an abstract class `AbstractGoToTypeProvider` with an abstract method `collectGoToTypeLocationLinks` which should be overridden by the end-user to implement the logic specific to their specific language.


packages/langium/src/lsp/goto-type.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export abstract class AbstractGoToTypeProvider implements GoToTypeProvider {
+
+    goToTypeDefinition(document: LangiumDocument, params: TypeDefinitionParams): MaybePromise<LocationLink[] | undefined> {
+        const rootNode = document.parseResult.value;
+        if (rootNode.$cstNode) {
+            const sourceCstNode = findLeafNodeAtOffset(rootNode.$cstNode, document.textDocument.offsetAt(params.position));
+            if (sourceCstNode) {
+                return this.collectGoToTypeLocationLinks(sourceCstNode.element);
```
msujew, line 30: Suggestion: Since the leaf node at the specified position actually might be a reference to something like a variable, using `References#findDeclaration` would be preferable here.

packages/langium/src/lsp/goto-type.ts
```diff
@@ -5,6 +5,8 @@
  ******************************************************************************/
 
 import { CancellationToken, LocationLink, TypeDefinitionParams } from 'vscode-languageserver';
+import { LangiumServices } from '..';
```
msujew, line 8: Minor: Please don't import from the index.

packages/langium/src/lsp/goto-type.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    goToTypeDefinition(document: LangiumDocument, params: TypeDefinitionParams): MaybePromise<LocationLink[] | undefined> {
+        const rootNode = document.parseResult.value;
+        if (rootNode.$cstNode) {
+            const sourceCstNode = findLeafNodeAtOffset(rootNode.$cstNode, document.textDocument.offsetAt(params.position));
+            if (sourceCstNode) {
+                const nodeDeclaration = this.references.findDeclaration(sourceCstNode);
+                if (nodeDeclaration) {
+                    return this.collectGoToTypeLocationLinks(nodeDeclaration.element);
```
msujew, line 40: Suggestion: We should support the `CancellationToken` in here:
```suggestion
    goToTypeDefinition(document: LangiumDocument, params: TypeDefinitionParams, cancelToken = CancellationToken.None): MaybePromise<LocationLink[] | undefined> {
        const rootNode = document.parseResult.value;
        if (rootNode.$cstNode) {
            const sourceCstNode = findLeafNodeAtOffset(rootNode.$cstNode, document.textDocument.offsetAt(params.position));
            if (sourceCstNode) {
                const nodeDeclaration = this.references.findDeclaration(sourceCstNode);
                if (nodeDeclaration) {
                    return this.collectGoToTypeLocationLinks(nodeDeclaration.element, cancelToken);
```

msujew (approved):
Great, looks good to me 👍

PR #620, opened 2022-08-05
Adds Monarch Grammar File Generation

### LINKED ISSUE #582: Native Monarch support for Langium
By default, the Monaco editor only has support for the `monarch` syntax definition. We should add a generator to the `langium-cli` for this type of syntax definition in addition to the existing textmate grammar support.

montymxb:
Closes #582 by adding an option to generate a Monarch grammar file. This supplements our existing generation for TextMate grammar files, so we can generate one, both, or neither of them as we choose. 

As an example I've updated the statemachine example's `langium-config` to construct a Monarch grammar, and I've added the resulting file.

Currently I've been comparing the results by using Microsoft's [Monarch Editor](https://microsoft.github.io/monaco-editor/monarch.html) (playground), and it looks good. 

The shows is marked as having quite a few errors in VSCode but the format should be ok when used in a Monaco instance. I haven't tried setting this up directly, but that would be the next step.

### DIFF
--- a/packages/langium-cli/src/generator/highlighting/monarch-generator.ts
+++ b/packages/langium-cli/src/generator/highlighting/monarch-generator.ts
@@ -1,0 +1,476 @@
+/******************************************************************************
+ * Copyright 2022 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import * as langium from 'langium';
+import { getTerminalParts, isCommentTerminal, isRegexToken, isTerminalRule, terminalRegex } from 'langium';
+import { LangiumLanguageConfig } from '../../package';
+import { collectKeywords } from '../util';
+
+/**
+ * Monarch Language Definition, describes aspects & token categories of target language
+ */
+interface LanguageDefinition {
+    readonly name: string;
+    readonly keywords: string[];
+    readonly operators: string[];
+    readonly symbols: string[];
+    readonly tokenPostfix: string;
+}
+
+/**
+ * Monarch Tokenizer, consists of an object that defines states.
+ */
+interface Tokenizer {
+    states: State[]
+}
+
+/**
+ * Name of a State
+ */
+type StateName = string;
+
+/**
+ * Each state is defined as an array of rules which are used to match the input
+ * Rules can be regular, or other States whose rules we should include in this state
+ */
+interface State {
+    name: StateName
+    rules: Array<Rule | State>
+}
+
+/**
+ * A rule that matches input. Can have either an action, or an array of cases.
+ */
+interface Rule {
+    regex: RegExp | string;
+    action: Action | Case[];
+}
+
+/**
+ * A case that selects a specific action by matching a guard pattern
+ */
+interface Case {
+    guard: string;
+    action: Action;
+}
+
+/**
+ * Determines whether a given object is a Rule instance
+ * @param obj Object to check
+ * @returns Whether this object is a Rule
+ */
+function isRule(obj: State | Rule): obj is Rule {
+    return (obj as Rule).regex !== undefined && (obj as Rule).action !== undefined;
+}
+
+/**
+ * Name of a token type, such as 'string'
+ */
+type Token = string;
+
+/**
+ * Token class to be used for CSS rendering, such as 'keyword', 'component', or 'type.identifer'
+ */
+type TokenClass = string;
+
+/**
+ * Next state that proceeds from an action, can also be a pop or a push of the current state (like for nested block comments)
+ */
+type NextState = StateName | '@pop' | '@push';
+
+/**
+ * An action performed when a rule (or a case) matches token.
+ * It can determine the token class, as well whether to push/pop a tokenizer state
+ */
+interface Action {
+    token?: Token
+    tokenClass?: TokenClass
+    next?: NextState
+    // other more advanced states omitted...
+}
+
+/**
+ * Abstract representation of a Monarch grammar file
+ */
+interface MonarchGrammar {
+    readonly languageDefinition: LanguageDefinition;
+    readonly tokenizer: Tokenizer;
+}
+
+/**
+ * Generates a Monarch highlighting grammar file's contents, based on the passed Langium grammar
+ * @param grammar Langium grammar to use in generating this Monarch syntax highlighting file content
+ * @param config Langium Config to also use during generation
+ * @returns Generated Monarch syntax highlighting file content
+ */
+export function generateMonarch(grammar: langium.Grammar, config: LangiumLanguageConfig): string {
+
+    const symbols = getSymbols(grammar);
+    const regex = /[{}[\]()]/;
+    const operators = symbols.filter(s => !regex.test(s));
+
+    // build absract monarch grammar representation
+    const monarchGrammar: MonarchGrammar = {
+        languageDefinition: {
+            name:       config.id,  // identifier for generating the grammar export
+            keywords:   getKeywords(grammar),
+            operators,
+            symbols,
+            tokenPostfix: '.' + config.id, // category appended to all tokens
+        },
+        tokenizer: {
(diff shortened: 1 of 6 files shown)


packages/langium-cli/src/generator/highlighting/monarch-generator.ts
```diff
@@ -0,0 +1,476 @@
...
+ * Generates an entry for a language definition, given a name (token category) and values
+ * @param name Category of language definition to add
+ * @param values Values to add under the given category
+ * @param fmt Formatter to keep things indented
+ * @returns A string of this language def entry, for use in a monarch file
+ */
+function genLanguageDefEntry(name: string, values: string[], fmt: Formatter): string {
+    return [
+        `${name}: [`,
+        '\t' + values.map(v => `'${v}'`).join(','),
```
msujew, line 224: Suggestion: Please use the `GeneratorNode` infrastructure like the other generator files (except for the `TextMate` stuff since it just uses `JSON.stringify`)

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
```diff
@@ -0,0 +1,476 @@
...
+        // add language definitions
+        prettyPrintLangDef(monarchGrammar.languageDefinition),
+
+        // add folding
+        '\tfolding: {',
+        '\t\tmarkers: {',
+        '\t\t\tstart: new RegExp(\'^\\s*//\\s*#?region\\b\'),',
+        '\t\t\tend: new RegExp(\'^\\s*//\\s*#?endregion\\b\')',
+        '\t\t}',
+        '\t},\n',
```
msujew, line 205: Suggestion: I don't believe we need folding to be available at the syntax level. We already provide a folding provider in the language server part of Langium.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
```diff
@@ -0,0 +1,476 @@
...
+/**
+ * Pretty prints a monarch grammar into it's concrete form, suitable for writing to a file
+ * @param monarchGrammar Grammar to pretty print
+ * @returns Monarch grammar in concrete form, suitable for writing to a file
+ */
+function prettyPrint(monarchGrammar: MonarchGrammar): string {
+    const name = monarchGrammar.languageDefinition.name;
+    return ([
+        `// Monarch syntax highlighting for the ${name} language.`,
+        `export const language${name} = {\n`,
```
msujew, line 194: Suggestion: It might make sense to just make this a default export.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
```diff
@@ -0,0 +1,476 @@
...
+        return '{ cases: { ' + prettyCases + ' }}';
+    }
+}
+
+/**
+ * Convert a deafult Langium token names to a monarch one
+ * @param name Token name to convert
+ * @returns Returns the equivalent monarch name, or the original token name
+ */
+function getMonarchTokenName(name: string): string {
```
msujew, line 310: Suggestion: These names are quite specific. We perform a more sophisticated analysis of what a comment or string is in the texmate Generator. Especially when it comes to `STRING`, `INT` and `Bigint` we can simply look at the return types of the terminal rules.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
```diff
@@ -0,0 +1,476 @@
...
+                    action: {token: getMonarchTokenName(rule.name) }
+                });
+            }
+        }
+    }
+    return rules;
+}
+
+/**
+ * Gets comment state rules from the Langium grammar. Accounts for nested multi-line comments.
```
msujew, line 360: We might not actually want to account for nested multi-line rules, since Chevrotain can't parse those.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
```diff
@@ -0,0 +1,476 @@
...
+
+/**
+ * Retrieves non-comment terminal rules, creating associated actions for them
+ * @param grammar Grammar to get non-comment terminals from
+ * @returns Array of Rules to add to a Monarch tokenizer state
+ */
+function getTerminalRules(grammar: langium.Grammar): Rule[] {
+    const rules: Rule[] = [];
+    for (const rule of grammar.rules) {
+        if (isTerminalRule(rule) && !isCommentTerminal(rule) && rule.name !== 'WS' && isRegexToken(rule.definition)) {
```
msujew, line 415: Suggestion: Instead of checking for `name === 'WS'` we should rather check whether the terminal regex can match a whitespace character. We do this in other places of the codebase as well.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
montymxb: I can omit this in the next commit then.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
montymxb: For `prettyPrint`? If we do that we also need to export the `MonarchGrammar` interface as well, is that what you're thinking? Personally, figured it would be good to only expose the `generateMonarch` function, similar to what we do for TextMate.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
montymxb: Initially I was using the original ML_COMMENT regex, but found some issues with the nested case. If I'm not mistaken, isn't this similar to the approach we're using in TextMate (sans the stack) by specifying our start/stop sequence? We could perhaps still mirror that solution, but w/out the stack?

montymxb:
Thanks for the feedback @msujew . Next commit of changes should be up sometime later tomorrow.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
msujew: I mean the `export const language${name}` part. That way you don't need to specify a name for the element. Just use `export default { ... }`.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
msujew: Yeah. I'm not sure whether textmate uses a stack. Anyway, we should always try that the monarch/textmate highlighting is in sync with what's actually getting parsed.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
montymxb: Seems we just check `rule.name` for string (lower cased) in TM. I've done the same here, using the type name when possible, and falling back on the rule name when that isn't available.

packages/langium-cli/src/generator/highlighting/monarch-generator.ts
montymxb: Seems like [TextMate supports nesting of patterns for the begin/end rule keys](https://macromates.com/manual/en/language_grammars#rule_keys). Although to keep it the same as what our current TM grammar generation is, I've settled just for simple start -> end match, with no nesting for comments.

montymxb:
New changes are up 👍

packages/langium-cli/src/generator/highlighting/textmate-generator.ts
```diff
@@ -1,13 +1,13 @@
 /******************************************************************************
- * Copyright 2021 TypeFox GmbH
+ * Copyright 2022 TypeFox GmbH
```
spoenemann, line 2: Note: if we update the copyright year, we should add a range `2021-2022`.

msujew (commented):
Looks quite good. Just a few minor things, see below.

montymxb:
Last review points added in.

msujew (approved):
Looks good to me 👍

PR #622, opened 2022-08-09
Renamed 'preprocess' phase to clarify document lifecycle

spoenemann:
In the initial draft of the document lifecycle, I named the second phase "preprocessing" because I thought users could extend this with additional processing such as type computation. But this does not work: if you extended the `process` method in DefaultDocumentBuilder, you could resolve local references by calling the `super` implementation first, but there was no guarantee that resolving references in _other_ documents worked because they might not have run through this phase yet (all documents requiring a rebuild are processed sequentially in each phase).

Additional processing should be implemented on user side by adding build listeners to the document builder. We might also consider adding an explicit type computation phase sitting between scope computation and linking.


msujew (approved):
Looks good to me, thanks 👍

PR #623, opened 2022-08-09
Completion service refactoring

msujew:
Since the change requires an addition parser for completion, I also refactored the parser builder to accept more generic parsers.


packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -32,84 +32,123 @@ export interface CompletionProvider {
...
-                        return match === 'partial' || match === 'both';
-                    });
-                    const notMatchingFeatures = possibleFeatures.filter(e => !partialMatches.includes(e));
-                    features.push(...partialMatches);
-                    features.push(...notMatchingFeatures.flatMap(e => findNextFeatures([e])));
+
+        const node = findLeafNodeAtOffset(cst, offset);
+
+        if (!node) {
+            const parserRule = getEntryRule(this.grammar)!;
```
montymxb, line 70: The case where `parserRule` is undefined could cause problems here with `!`.

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -32,84 +32,123 @@ export interface CompletionProvider {
...
+    protected backtrackToTokenStart(text: string, offset: number): number {
+        const wordRegex = /\w/;
+        const whiteSpaceRegex = /\s/;
+        let lastCharacter = text.charAt(offset - 1);
+        if (whiteSpaceRegex.test(lastCharacter)) {
+            return offset;
+        }
+        while (wordRegex.test(lastCharacter) && offset > 0) {
+            offset--;
+            lastCharacter = text.charAt(offset - 1);
```
montymxb, line 144: Wouldn't this cause issues if `offset` is ever `1`. This would end up retrieving a char from `-1` in such a case.

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -32,84 +32,123 @@ export interface CompletionProvider {
...
+        }
+        return offset;
+    }
+
     protected completionForRule(astNode: AstNode | undefined, rule: ast.AbstractRule, acceptor: CompletionAcceptor): void {
         if (ast.isParserRule(rule)) {
-            const features = findFirstFeatures(rule.definition);
+            const features = findFirstFeatures(rule.definition, new Map(), new Set());
```
montymxb, line 151: If a fresh map & set are passed, maybe these could be decent defaults for the last 2 params of `findFirstFeatures` instead?

packages/langium/src/lsp/completion/follow-element-computation.ts
```diff
@@ -4,25 +4,34 @@
...
 
 /**
  * Calculates any features that can follow the given feature stack.
  * This also includes features following optional features and features from previously called rules that could follow the last feature.
  * @param featureStack A stack of features starting at the entry rule and ending at the feature of the current cursor position.
  * @returns Any `AbstractElement` that could be following the given feature stack.
  */
-export function findNextFeatures(featureStack: ast.AbstractElement[]): ast.AbstractElement[] {
-    return findNextFeaturesInternal(featureStack, new Map<ast.AbstractElement, Cardinality>());
+export function findNextFeatures(featureStack: ast.AbstractElement[], unparsedTokens: IToken[]): ast.AbstractElement[] {
```
montymxb, line 17: Block comment above should be updated to match new signature.

packages/langium/src/lsp/completion/follow-element-computation.ts
```diff
@@ -103,25 +108,88 @@ export function findFirstFeatures(feature: ast.AbstractElement | undefined, card
...
     } while (firstFeature);
     return features;
 }
+
+interface InterpretationContext {
+    tokens: IToken[]
+    stacks: ast.AbstractElement[][]
+}
+
+function interpreteTokens(context: InterpretationContext): void {
```
montymxb, line 136: Seems like a small typo, interpret**e**Tokens

packages/langium/src/lsp/completion/follow-element-computation.ts
```diff
@@ -4,25 +4,34 @@
...
-    return findNextFeaturesInternal(featureStack, new Map<ast.AbstractElement, Cardinality>());
+export function findNextFeatures(featureStack: ast.AbstractElement[], unparsedTokens: IToken[]): ast.AbstractElement[] {
+    const context: InterpretationContext = {
+        stacks: [featureStack],
+        tokens: unparsedTokens
+    };
+    interpreteTokens(context);
+    const nextStacks = findNextFeatureStacks(context.stacks);
+    // We only need the last element of each stack
+    return nextStacks.map(e => e[e.length - 1]);
```
montymxb, line 25: Personal preference, `e => e.pop()`

packages/langium/src/lsp/completion/follow-element-computation.ts
```diff
@@ -103,25 +108,88 @@ export function findFirstFeatures(feature: ast.AbstractElement | undefined, card
...
+
+function findNextFeatureStacks(stacks: ast.AbstractElement[][], token?: IToken): ast.AbstractElement[][] {
+    const newStacks: ast.AbstractElement[][] = [];
+    for (const stack of stacks) {
+        newStacks.push(...interpreteStackToken(stack, token));
+    }
+    return newStacks;
+}
+
+function interpreteStackToken(stack: ast.AbstractElement[], token?: IToken): ast.AbstractElement[][] {
```
montymxb, line 152: Same interpret**e** typo here

packages/langium/src/lsp/completion/follow-element-computation.ts
```diff
@@ -103,25 +108,88 @@ export function findFirstFeatures(feature: ast.AbstractElement | undefined, card
...
+    tokens: IToken[]
+    stacks: ast.AbstractElement[][]
+}
+
+function interpreteTokens(context: InterpretationContext): void {
+    while (context.tokens.length > 0) {
+        const token = context.tokens.pop()!;
+        const nextFeatureStacks = findNextFeatureStacks(context.stacks, token);
+        context.stacks = nextFeatureStacks;
+    }
```
montymxb, line 141: Is this while loop creating some effect by calling `findNextFeatureStacks`? I'm not quite sure if there's a throwable check or something in there, but if not you could just evaluate the last token instead of the loop.

packages/langium/src/parser/langium-parser.ts
```diff
@@ -42,52 +43,101 @@ type Args = Record<string, boolean>;
...
+    abstract rule(rule: ParserRule, impl: RuleImpl): RuleResult;
+    abstract consume(idx: number, tokenType: TokenType, feature: AbstractElement): void;
+    abstract subrule(idx: number, rule: RuleResult, feature: AbstractElement, args: Args): void;
+    abstract action($type: string, action: Action): void;
+    abstract construct(): unknown;
+
+    isRecording(): boolean {
+        return this.wrapper.IS_RECORDING;
+    }
+    get unorderedGroups(): Map<string, boolean[]> {
```
montymxb, line 102: small nit, spacing between the functions here

packages/langium/src/parser/langium-parser.ts
```diff
@@ -403,6 +423,123 @@ export class LangiumParserErrorMessageProvider implements IParserErrorMessagePro
...
+        }
+        return ruleMethod;
+    }
+
+    private resetState(): void {
+        this.elementStack = [];
+        this.lastElementStack = [];
+        this.lastToken = undefined;
+        this.stackSize = 0;
+        this.nextTokenIndex = 0;
```
montymxb, line 479: This doesn't reset `tokens`, as well as `mainRule`, is that intentional?

packages/langium/src/parser/parser-builder-base.ts
```diff
@@ -0,0 +1,419 @@
+/******************************************************************************
+ * Copyright 2022 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
montymxb, line 7: I think this can be dropped now.

packages/langium/src/parser/parser-builder-base.ts
```diff
@@ -0,0 +1,419 @@
...
+            subrule: 1,
+            many: 1,
+            or: 1
+        };
+        ctx.rules.set(rule.name, parserContext.parser.rule(rule, buildRuleContent(ctx, rule)));
+    }
+}
+
+function buildRuleContent(ctx: RuleContext, rule: ParserRule): Method {
+    const method = buildElement(ctx, rule.definition);
```
montymxb, line 68: Can just return

packages/langium/src/parser/parser-builder-base.ts
```diff
@@ -0,0 +1,419 @@
...
+            // Since we handle the guard condition in the alternative already
+            // We can ignore the group guard condition inside
+            ALT: buildElement(ctx, element, true)
+        };
+        const guard = getGuardCondition(element);
+        if (guard) {
+            predicatedMethod.GATE = buildPredicate(guard);
+        }
+        methods.push(predicatedMethod);
+    }
```
montymxb, line 208: The for loop above is almost identical to the one in `buildAlternatives`. Perhaps these could be factored out into a common function `buildMethods` that takes the necessary elements and context?

packages/langium/src/parser/parser-builder-base.ts
```diff
@@ -0,0 +1,419 @@
...
+        }
+    };
+}
+
+function buildGroup(ctx: RuleContext, group: Group): Method {
+    const methods: Method[] = [];
+
+    for (const element of group.elements) {
+        methods.push(buildElement(ctx, element));
+    }
```
montymxb, line 260: Could combine 256 - 260 here as a single map.
```ts
const methods: Method[] = group.elements.map(e => buildElement(ctx, e))
```

montymxb (commented):
Left some comments, questions, and suggestions.

packages/langium/src/lsp/completion/completion-provider.ts
msujew: Since we know that the grammar in question is validated (it is only used after it has passed the validation), we know that there's always an entry rule.

packages/langium/src/lsp/completion/follow-element-computation.ts
msujew: I feel like `pop` implies that we are interested in the other part of the array as well (due to its side effects). However, we only want the last element, so I'll keep it like this. `pop` is also slower.

packages/langium/src/lsp/completion/follow-element-computation.ts
msujew: Since we always reassign `context.stacks`, there is a side effect for each iteration of the loop. So we definitely have to look at all tokens in the list. You've actually made me aware of an issue in this implementation, since it just should be a normal `for-each` loop instead of using `pop`.

packages/langium/src/parser/langium-parser.ts
msujew: Yes, that's intentional. `mainRule` is set once during parser initialization and marks the entry rule of the grammar. `tokens` is reset anyway once we perform our lexing routine.

spoenemann (commented):
Thank you, this looks good so far.

There's still a problem with parser rules starting with a cross-reference. Example - write the following in the statemachine example:
```
state sdf
    |
```
Here it fails with
```

spoenemann (commented):
Excellent work! I noticed only one defect: sometimes too many keywords are offered.

Example: open trafficlight.statemachine and do completion at the end of the file. It offers `commands`, `initialState` and `state`, but only `state` is valid here because the other two keywords can only be used before the list of state declarations.

spoenemann (commented):
Update: with the current state, I always get the `statemachine` keyword as proposal, but no longer `state`.

msujew:
@spoenemann Yep, didn't check after merging. The binary cst node search has broken completion support for any position after the last token. I've fixed this now 👍

spoenemann (commented):
Sorry, it stills shows odd results for me: I get the `statemachine` keyword at most positions in the example file, except after the last state, where it's correctly offering `state`.

msujew:
@spoenemann Thanks for looking into it again. The new cst search behavior broke more in the completion provider than anticipated. It will now correctly search for the last node of the current cursor position.

spoenemann (approved):
Great!

I noticed one further thing, but this might be improved in a follow-up PR (please create an issue in that case): when doing completion within a parser rule in a grammar file, for example, right after the colon, it seems to list _all_ keywords of the grammar language, including those that are not valid at that position.

msujew:
@spoenemann This is a bit more difficult than one would assume since Chevrotain mangles our output that we get, as it sees the input that we give it as invalid. It's missing the mandatory closing `;`, so it performs it's error recovery. Seems like in that case it isn't perfectly able to return to a valid state.

PR #628, opened 2022-08-11
Extend validation options

msujew:
Brought up in discussion https://github.com/langium/langium/discussions/458#discussioncomment-3364456. If we apply `Properties<AstNode>` the resulting type will be `never`, which isn't really helpful. This change makes it return `string` if we find that the node type given is `AstNode`, so that the user is free to enter any property.

Also extends the validation options to include the `keyword` option. This allows to show validations on keywords without needing to wrap them in an assignment like `keyword?='hello'`.


spoenemann (approved):
:+1: Thanks!

PR #633, opened 2022-08-14
Fix unused terminal issue

msujew:
The validation checking for unused rules currently stops at a terminal rule and doesn't inspect which terminal rules are called within that terminal. This change lets the validation also respect terminals referenced only by other terminals. Also adds tests for the validation in the first place.


spoenemann (approved):
:+1:

PR #639, opened 2022-08-17
Fix terminal regex builder

msujew:
Fixes an issue brought up in https://github.com/langium/langium/discussions/638. The issue was that we weren't surrounding some items with parenthesis, so that mixed regexes sometimes didn't work as expected. The new approach is a bit more conversative (adds parenthesis even if they might not be necessary) in favor of being also always correct.


spoenemann (approved):
:+1:

PR #641, opened 2022-08-19
Reworked scope provider API

spoenemann:
This changes the input arguments of the ScopeProvider service to using ReferenceInfo, enabling us to get rid of the `referenceId` concept. Instead of generating strings with colon-separated information, we use a proper object.

@msujew please look carefully into the completion provider. I guess that needs more adaptation to make it consistent with the expected API.

### DIFF
--- a/examples/arithmetics/src/language-server/generated/ast.ts
+++ b/examples/arithmetics/src/language-server/generated/ast.ts
@@ -5,7 +5,7 @@
 
 /* eslint-disable @typescript-eslint/array-type */
 /* eslint-disable @typescript-eslint/no-empty-interface */
-import { AstNode, AstReflection, Reference, isAstNode, TypeMetaData } from 'langium';
+import { AstNode, AstReflection, Reference, ReferenceInfo, isAstNode, TypeMetaData } from 'langium';
 
 export type AbstractDefinition = DeclaredParameter | Definition;
 
@@ -115,8 +115,6 @@ export function isNumberLiteral(item: unknown): item is NumberLiteral {
 
 export type ArithmeticsAstType = 'AbstractDefinition' | 'BinaryExpression' | 'DeclaredParameter' | 'Definition' | 'Evaluation' | 'Expression' | 'FunctionCall' | 'Module' | 'NumberLiteral' | 'Statement';
 
-export type ArithmeticsAstReference = 'FunctionCall:func';
-
 export class ArithmeticsAstReflection implements AstReflection {
 
     getAllTypes(): string[] {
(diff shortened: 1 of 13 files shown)


msujew (approved):
Looks good to me. I'll perform the necessary changes for the completion provider after merging https://github.com/langium/langium/pull/623 👍

spoenemann:
Use `context.container.$type` and `context.property` to determine the exact reference. You can build a reference id with `context.container.$type + ':' + context.property` if you want to switch over a string.

sailingKieler:
> You can build a reference id with context.container.$type + ':' + context.property if you want to switch over a string.

But how does that work in the content assist case when `context.container.$type` is does not define `context.property`?
Imagine you have something like a class with contained declarations, and a declaration starts with a TypeReference whose only property is a cross reference named `type`, like

```
Class:
   declarations += Declaration
;
Declaration:
    type=TypeReference ...
;
TypeReference:
    type=[Type]
;
```

sailingKieler:
In other words: I think the property defining type is required in `context` as well, isn't it?

spoenemann:
The trick is, we adapt the completion provider so it always passes a container object that matches the cross-reference owner. If it does not exist in the parser yet, we create such an object synthetically. This will simplify a lot of things in general when implementing scope providers, and it makes the API contract much clearer. The way how that is handled in Xtext is quite ugly, though we've somehow got used to it.

sailingKieler:
> The trick is, we adapt the completion provider so it always passes a container object that matches the cross-reference owner.

And that object has proper `$container` references etc?
The `context.container.$type + ':' + context.property` thing is still to be used with caution if you have
```
interface A {
    $type: typeof 'A';
    prop foo;
}
interface B extends A {
    $type: typeof 'B';
}
```
and `context.container` gets you a `B`

The only generally proper approach is to combine `isSubType(...)` and a distinction of `context.poperty`.

Just thinking while writing this:
Did you think about a making `context.poperty` an object like `{ name: string, definingType: string }`?

spoenemann:
This boils down to the fact that we don't have a reflection level like EMF has (except for a few utility methods in AstReflection). The subtype case might be a problem, but that was already present before this change: we always used the type of the current object to create the "reference id".

In your concrete DSL code, a subtype situation can be solved with multiple `case` patterns leading to the same code within a `switch`. What you propose about enriching the property with more "metamodel" information is problematic as we don't have that information at hand. I would like to avoid reintroducing a full-fledged runtime metamodel. Let's see how the current approach works out in practice, maybe it's just a matter of documenting these details well (https://github.com/langium/langium-website/issues/58).

spoenemann:
I just realized there's now a much simpler way to solve the subtype case:

```

PR #643, opened 2022-08-22
Implement CallHierarchy LSP service

### LINKED ISSUE #599: Support call hierarchy related LSP requests
There are three of them in total:

- [`textDocument/prepareCallHierarchy`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareCallHierarchy)
- [`callHierarchy/incomingCalls`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#callHierarchy_incomingCalls)
- [`callHierarchy/outgoingCalls`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#callHierarchy_outgoingCalls)

All of these should ideally be covered by one single service.

gfontorbe:
First draft for the implementation of the CallHierarchy LSP service.

It contains a default implementation that always return a `null` object since the implementation is highly dependent on the language grammar. We provide two methods to be overridden by the user to implement their own logic `getIncomingCalls` and `getOutgoingCalls`.

I also made a first iteration of the implementation of the CallHierarchy for Langium where CallHierarchy requests can be made on parser rules and rule calls. This implementation is functional as is but we can most likely improve it and discuss what we want to be available as source and targets in the incoming and outgoing calls.

Closes #599


packages/langium/src/lsp/language-server.ts
```diff
@@ -326,6 +333,64 @@ export function addSignatureHelpHandler(connection: Connection, services: Langiu
     ));
 }
 
+function addCallHierarchyHandler(connection: Connection, services: LangiumSharedServices): void {
```
msujew, line 336: Minor: This should be exported.

packages/langium/src/lsp/language-server.ts
```diff
@@ -326,6 +333,64 @@ export function addSignatureHelpHandler(connection: Connection, services: Langiu
...
+            if (services.lsp.CallHierarchyProvider) {
+                return services.lsp.CallHierarchyProvider.outgoingCalls(params, cancelToken);
+            }
+            return new ResponseError<void>(0, errorMessage);
+        },
+        services
+    ));
+}
+
+function createCallHierarchyRequestHandler<P extends CallHierarchyIncomingCallsParams | CallHierarchyOutgoingCallsParams, R, PR, E = void>(
```
msujew, line 369: Minor: This should be exported.

packages/langium/src/lsp/language-server.ts
```diff
@@ -326,6 +333,64 @@ export function addSignatureHelpHandler(connection: Connection, services: Langiu
...
+        const uri = URI.parse(params.item.uri);
+        const language = serviceRegistry.getServices(uri);
+        if (!language) {
+            console.error(`Could not find service instance for uri: '${uri.toString()}'`);
+            throw new Error();
+        }
+        const document = documents.getOrCreateDocument(uri);
+        if (!document) {
+            throw new Error();
+        }
```
msujew, line 385: Suggestion: Retrieving the document here is not necessary.

packages/langium/src/lsp/language-server.ts
```diff
@@ -326,6 +333,64 @@ export function addSignatureHelpHandler(connection: Connection, services: Langiu
     ));
 }
 
+function addCallHierarchyHandler(connection: Connection, services: LangiumSharedServices): void {
+    const errorMessage = 'No call hierarchy provider registered';
+    connection.languages.callHierarchy.onPrepare(createServerRequestHandler(
+        (services, document, params, cancelToken) => {
+            if (services.lsp.CallHierarchyProvider) {
+                return services.lsp.CallHierarchyProvider.prepareCallHierarchy(document, params, cancelToken);
```
msujew, line 341: Suggestion: Instead of returning `null` (which we don't do anywhere else in the framework), we should instead return `undefined` from our service and do the following here. The same applies to the other service methods as well.
```suggestion
                return services.lsp.CallHierarchyProvider.prepareCallHierarchy(document, params, cancelToken) ?? null;
```

packages/langium/src/default-module.ts
```diff
@@ -69,7 +70,8 @@ export function createDefaultModule(context: DefaultModuleContext): Module<Langi
             ReferenceFinder: (services) => new DefaultReferenceFinder(services),
             GoToResolver: (services) => new DefaultGoToResolverProvider(services),
             DocumentHighlighter: (services) => new DefaultDocumentHighlighter(services),
-            RenameHandler: (services) => new DefaultRenameHandler(services)
+            RenameHandler: (services) => new DefaultRenameHandler(services),
+            CallHierarchyProvider: (services) => new DefaultCallHierarchyProvider(services),
```
msujew, line 74: Suggestion: I don't believe the default implementation has real value for most languages, especially since it only supports one direction (incoming calls). Instead we should have an `AbstractCallHierarchyProvider` instead, which isn't injected by default.

packages/langium/src/lsp/call-hierarchy-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        const references = this.references.findReferences(targetNode.element, {includeDeclaration: false, onlyLocal: false}).toArray();
+        return this.getIncomingCalls(references);
+    }
+
+    /**
+     * Override this method to collect the incoming calls for your language
+     */
+    protected getIncomingCalls(_references: ReferenceDescription[]): CallHierarchyIncomingCall[] | null {
```
msujew, line 87: Suggestion: To create more fine-grained results, the original source node should be passed in here as well.
```suggestion
    protected abstract getIncomingCalls(node: AstNode, references: ReferenceDescription[]): CallHierarchyIncomingCall[] | undefined {
```

packages/langium/src/lsp/call-hierarchy-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+        if (!targetNode) {
+            return null;
+        }
+        return this.getOutgoingCalls(targetNode.element);
+    }
+
+    /**
+     * Override this method to collect the outgoing calls for your language
+     */
+    protected getOutgoingCalls(_node: AstNode): CallHierarchyOutgoingCall[] | null {
```
msujew, line 104: Minor: In line with the previous suggestions, this should be made abstract.
```suggestion
    protected abstract getOutgoingCalls(node: AstNode): CallHierarchyOutgoingCall[] | undefined {
```

packages/langium/src/lsp/call-hierarchy-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+        const nameNode = this.nameProvider.getNameNode(targetNode);
+        if (!nameNode) {
+            return [];
+        }
+
+        return [{
+            kind: SymbolKind.Method,
+            name: nameNode.text,
+            range: nameNode.range,
+            selectionRange: nameNode.range,
```
msujew, line 67: Suggestion: The `selectionRange` contains the name, while the `range` usually contains the whole element (just so vscode knows where to scroll to).
```suggestion
            range: targetNode.$cstNode.range,
            selectionRange: nameNode.range,
```

packages/langium/src/lsp/call-hierarchy-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+        if (!declarationNode) {
+            return null;
+        }
+
+        return this.getCallHierarchyItems(declarationNode?.element, document);
+    }
+
+    protected getCallHierarchyItems(targetNode: AstNode, document: LangiumDocument<AstNode>): CallHierarchyItem[] {
+        const nameNode = this.nameProvider.getNameNode(targetNode);
+        if (!nameNode) {
```
msujew, line 59: Suggestion: in line with the previous comment:
```suggestion
        if (!nameNode || !targetNode.$cstNode) {
```

packages/langium/src/lsp/call-hierarchy-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+        if (!targetNode) {
+            return null;
+        }
+
+        const declarationNode = this.references.findDeclaration(targetNode);
+        if (!declarationNode) {
+            return null;
+        }
+
+        return this.getCallHierarchyItems(declarationNode?.element, document);
```
msujew, line 54: Minor:
```suggestion
        return this.getCallHierarchyItems(declarationNode.element, document);
```

packages/langium/src/grammar/lsp/grammar-call-hierarchy.ts
```diff
@@ -0,0 +1,106 @@
...
+                    kind: SymbolKind.Method,
+                    name: nodes.nameNode.text,
+                    range: nodes.nameNode.range,
+                    selectionRange: nodes.nameNode.range,
+                    uri: nodes.docUri
+                },
+                fromRanges: nodes.targetNodes.map(node => node.range)
+            });
+        });
+        return incomingCalls;
```
msujew, line 59: Minor: This can be mapped to a single `map` function call. (instead of `forEach` and `push`)

packages/langium/src/grammar/lsp/grammar-call-hierarchy.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    protected getOutgoingCalls(node: AstNode): CallHierarchyOutgoingCall[] | null {
+        if (isParserRule(node)) {
+            const ruleCalls = streamAllContents(node).toArray().filter(n => isRuleCall(n)) as RuleCall[];
```
msujew, line 64: Minor:
```suggestion
            const ruleCalls = streamAllContents(node)filter(isRuleCall).toArray();
```

packages/langium/src/grammar/lsp/grammar-call-hierarchy.ts
```diff
@@ -0,0 +1,106 @@
...
+                        kind: SymbolKind.Method,
+                        name: rule.to.text,
+                        range: rule.to.range,
+                        selectionRange: rule.to.range,
+                        uri: rule.docUri
+                    },
+                    fromRanges: rule.from
+                });
+            });
+            return outgoingCalls;
```
msujew, line 102: Minor: Same simplification as above

msujew (commented):
From a functionality perspective, this is already working really good. I have a few thoughts on the API of the call hierarchy provider. The initial draft is going in a great direction though.

packages/langium/src/lsp/call-hierarchy-provider.ts
gfontorbe: VSCode still jumps to the name, this is the same behavior as in TypeScript LS

packages/langium/src/grammar/lsp/grammar-call-hierarchy.ts
gfontorbe: I couldn't find a way to use a `map` function on a `Map` object, so I changed the map to an array and introduced an interface to help with the structure of the objects in the array

msujew: You can do `Array.from(uniqueRules.values()).map(...)` instead.

gfontorbe: Went an other way, but can change it back if it's better

msujew (approved):
Great looks good to me. Please rebase the changes and merge them afterwards 👍

PR #656, opened 2022-08-29
Run GitHub Actions on all pull requests

msujew:
Right now, GH actions are only executed for pushes directly on our repo. In order for actions to run on pull requests from forks, some modification in our workflows file is needed.


spoenemann (approved):
:+1: first-time contributors will require approval by a committer before workflows are run.

PR #657, opened 2022-08-29
Find optimization

snarkipus:
Added test case - hopefully I did this correctly.

### DIFF
--- a/packages/langium/src/utils/cst-util.ts
+++ b/packages/langium/src/utils/cst-util.ts
@@ -109,12 +109,19 @@ export function findLeafNodeAtOffset(node: CstNode, offset: number): LeafCstNode
     if (isLeafCstNode(node)) {
         return node;
     } else if (isCompositeCstNode(node)) {
-        const children = node.children.filter(e => e.offset <= offset).reverse();
-        for (const child of children) {
-            const result = findLeafNodeAtOffset(child, offset);
-            if (result) {
-                return result;
+        let firstChild = 0;
+        let lastChild = node.children.length - 1;
+        while (firstChild <= lastChild) {
+            let middleChild = Math.floor((firstChild + lastChild) / 2);
+            const n = node.children[middleChild];
+            if (n.offset > offset) {
+                lastChild = middleChild - 1;
+            } else if (n.end <= offset) {
+                firstChild = middleChild + 1;
+            } else {
+                return findLeafNodeAtOffset(n, offset);
             }
+            middleChild = Math.floor((firstChild + lastChild) / 2);
         }
     }
     return undefined;
--- a/packages/langium/src/utils/cst-util.ts
+++ b/packages/langium/src/utils/cst-util.ts
@@ -112,7 +112,7 @@ export function findLeafNodeAtOffset(node: CstNode, offset: number): LeafCstNode
         let firstChild = 0;
         let lastChild = node.children.length - 1;
         while (firstChild <= lastChild) {
-            let middleChild = Math.floor((firstChild + lastChild) / 2);
+            const middleChild = Math.floor((firstChild + lastChild) / 2);
             const n = node.children[middleChild];
             if (n.offset > offset) {
                 lastChild = middleChild - 1;
@@ -121,7 +121,6 @@ export function findLeafNodeAtOffset(node: CstNode, offset: number): LeafCstNode
             } else {
                 return findLeafNodeAtOffset(n, offset);
             }
-            middleChild = Math.floor((firstChild + lastChild) / 2);
         }
     }
     return undefined;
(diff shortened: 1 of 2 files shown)


packages/langium/src/utils/cst-util.ts
```diff
@@ -109,12 +109,19 @@ export function findLeafNodeAtOffset(node: CstNode, offset: number): LeafCstNode
...
+            let middleChild = Math.floor((firstChild + lastChild) / 2);
+            const n = node.children[middleChild];
+            if (n.offset > offset) {
+                lastChild = middleChild - 1;
+            } else if (n.end <= offset) {
+                firstChild = middleChild + 1;
+            } else {
+                return findLeafNodeAtOffset(n, offset);
             }
+            middleChild = Math.floor((firstChild + lastChild) / 2);
```
spoenemann, line 124: `middleChild` is already set at the beginning of the iteration, no need to set it again.

packages/langium/test/utils/cst-utils.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    test('Find Leaf Node at Offset', async () => {
+        const text = `
+        grammar test
+        Main: value=AB;
+        terminal fragment Frag: 'B';
+        terminal AB: 'A' Frag;
+        `;
+
+        const grammar = await parser(text);
+        const offset = grammar.parseResult.value.$document?.textDocument.getText().indexOf('AB');
```
spoenemann, line 24:
```suggestion
        const offset = grammar.textDocument.getText().indexOf('AB');
```

spoenemann (commented):
Thanks, almost good to go.

packages/langium/test/utils/cst-utils.test.ts
```diff
@@ -0,0 +1,28 @@
...
+        const leafnode = c.findLeafNodeAtOffset(grammar.parseResult.value.$cstNode!, offset!);
```
msujew, line 25:
```suggestion
        const leafnode = c.findLeafNodeAtOffset(grammar.$cstNode!, offset!);
```

packages/langium/test/utils/cst-utils.test.ts
```diff
@@ -0,0 +1,28 @@
+/******************************************************************************
+ * Copyright 2022 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+import { createLangiumGrammarServices, EmptyFileSystem, Grammar } from '../../src';
+import { parseHelper } from '../../src/test';
+import * as c from '../../src/utils/cst-util';
```
msujew, line 8: Minor: The name is a bit non-descriptive.
```suggestion
import * as cstUtil from '../../src/utils/cst-util';
```

packages/langium/test/utils/cst-utils.test.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const services = createLangiumGrammarServices(EmptyFileSystem);
+const parser = parseHelper<Grammar>(services.grammar);
+
+describe('CST Utils', () => {
+
+    test('Find Leaf Node at Offset', async () => {
```
msujew, line 15: Suggestion: It might be interesting to test edge case behavior in here as well. Your test covers the case where the cursor is at `<|>AB;`, but `A<|>B` and `AB<|>;` (which should return the leaf node for `;`) are also very important.

packages/langium/test/utils/cst-utils.test.ts
snarkipus: fixed

packages/langium/test/utils/cst-utils.test.ts
snarkipus: From what I could tell, `$cstNode` is only available on the `parseResult`. Am I missing something?

packages/langium/test/utils/cst-utils.test.ts
snarkipus: fixed

packages/langium/test/utils/cst-utils.test.ts
snarkipus: added add'l test cases

packages/langium/test/utils/cst-utils.test.ts
msujew: Yeah, sorry got confused, since I expected `grammar` to be an AST element, not a document. It's all good.

spoenemann (approved):

spoenemann:
Unrelated to this change: we don't check the offset if the given node is already a leaf node. @msujew could there be cases where this is problematic?

Issue #277, opened 2021-11-01
Domainmodel example: Renaming with qualified name does not work


spoenemann:
Langium version: #275
Package name: langium

## Steps To Reproduce

1. Open `datatypes.dmodel`
2. Select `Date`
3. Press F2
4. Enter a new name

## The current behavior

The datatype name is changed in `blog.dmodel`, but it loses the `complex.` prefix.

## The expected behavior

The new name should include the package name.

dhuebner:
@spoenemann 
Should supporting renaming of a package e.g. `complex` also be a part of this issue?

dhuebner:
What is also strange, you can exec `Rename Symbol` for every QName CSTNode part, which doesn't make much sense:
<img width="375" alt="Bildschirmfoto 2022-03-02 um 15 28 53" src="https://user-images.githubusercontent.com/454322/156381467-b89795e4-45f0-40aa-b18d-14c304e5f436.png">
<img width="370" alt="Bildschirmfoto 2022-03-02 um 15 28 23" src="https://user-images.githubusercontent.com/454322/156381477-74bc9159-fac5-4170-8d8e-54eab10112c1.png">
<img width="366" alt="Bildschirmfoto 2022-03-02 um 15 28 12" src="https://user-images.githubusercontent.com/454322/156381483-6949e14d-3477-40fb-968a-2af682e8e0cf.png">

Every such a rename request will end up in rename service with a selected QName part (`comlex` or `.` or `Date`) as `newText`

spoenemann:
> Should supporting renaming of a package e.g. complex also be a part of this issue?

Ideally yes (needs special handling to rename all references to contained classes).

> What is also strange, you can exec Rename Symbol for every QName CSTNode part, which doesn't make much sense

That seems to be a client side behavior. Don't know what we can do to fix it.

insafuhrmann:
The issue seems to be no longer present. I cannot reproduce with current main state.

spoenemann:
?? I can still reproduce following the steps described at the top.

insafuhrmann:
Yes, today I also can (with todays state of main). Not clear why it worked yesterday for me on the main branch. However, I'll update the fix.

PR #660, opened 2022-09-02
Support rename refactoring for domain-model qualified names (#277).

insafuhrmann:
Closes https://github.com/langium/langium/issues/277. Bugfix for the loss of the prefix in changes of qualified names. Added a DomainModelRenameHandler.


spoenemann (commented):
Thanks! I see a problem when renaming in `qualified-names.dmodel`:
 - Renaming `baz.E3` results in an error message "Rename failed to apply edits"

insafuhrmann:
Thanks for reviewing @spoenemann! The bug you addressed in your comment has been addressed now.

spoenemann (approved):
:+1:

PR #664, opened 2022-09-05
Renamed main indexing method and moved to ScopeComputation service

spoenemann:
I propose moving and renaming `AstNodeDescriptionProvider.createDescriptions` to `ScopeComputation.getExports`. The former name does not explain what this is about: gathering the exported symbols of a document. I think ScopeComputation is a better home for that service method because it's conceptually similar to the scope computation phase. We can see that these two service methods often need to be customized together in the domainmodel example.

I also renamed `ScopeComputation.computeScope` to `computeScopes` to avoid confusion with `ScopeProvider.getScope`.

This change heavily breaks API and needs to be documented in the change log.


spoenemann:
Q: Shall we further rename `computeScopes` to `computeLocalScopes`?

msujew (approved):
Looks good to me. We will also have to change the documentation on the document lifecycle afterwards.

Anyway, changing `computeScopes` -> `computeLocalScopes` also sounds good to me.

spoenemann:
Ok renamed that as well, and I also separated ScopeProvider and ScopeComputation into different source files. @msujew please have another look (and double-check that I didn't break anything).

sailingKieler (approved):
LGTM and works fine for me. I have a suggestion for another small improvement, I'll file another PR.

PR #665, opened 2022-09-05
Added utility method to create a scope from AST nodes

spoenemann:
This makes the process of customizing a scope provider simpler: you can get a scope just by calling a method with a collection of elements you got from the AST.

### DIFF
--- a/packages/langium/src/utils/stream.ts
+++ b/packages/langium/src/utils/stream.ts
@@ -126,6 +126,12 @@ export interface Stream<T> extends Iterable<T> {
     filter(predicate: (value: T) => unknown): Stream<T>;
 
     /**
+     * Returns the elements of the stream that are _non-nullable_, which means they are neither `undefined`
+     * nor `null`.
+     */
+    nonNullable(): Stream<NonNullable<T>>;
+
+    /**
      * Calls the specified callback function for all elements in the stream. The return value of the
      * callback function is the accumulated result, and is provided as an argument in the next call to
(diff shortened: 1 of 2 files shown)


msujew (approved):
Looks good to me 👍

packages/langium/src/utils/stream.ts
```diff
@@ -125,6 +125,12 @@ export interface Stream<T> extends Iterable<T> {
     filter<S extends T>(predicate: (value: T) => value is S): Stream<S>;
     filter(predicate: (value: T) => unknown): Stream<T>;
 
+    /**
+     * Returns the elements of the stream that are _non-nullable_, which means they are neither `undefined`
+     * nor `null`.
+     */
+    nonNullable(): Stream<NonNullable<T>>;
```
pluralia, line 132: I'd propose to call this method `filterNullable` or `filterNull`

spoenemann: I prefer `nonNullable` because it's immediately clear what it does: give you the non-nullable elements. With `filterNullable` or `filterNull` it's not so clear because it's missing the negation. `filterNonNullable` would work, but IMO that's too long.

What do others think @langium/committers?

sailingKieler: All options are fine me as long as the documentation is clear, which is the case.

spoenemann: Moving forward with this for now.

PR #666, opened 2022-09-05
Fix extention of DataTypes by Declared types

### LINKED ISSUE #559: Declared types cannot extend DataTypes
In Langium version: 0.4.0

Declared types are unable to incorporate a `DataType` for their properties. As an example:

Screen Shot 2022-06-21 at 16 51 32

However, when inferring, this works just fine.

Screen Shot 2022-06-21 at 16 52 21

The produced AST when inferred is as follows:
```typescript
export type DT = string;

...

export interface R1 extends AstNode {
    val: DT
}
```
It seems that the way inferred vs. declared types are processed is different with regards to how a DataType is resolved. In the declared case, it appears to stop at the DataType. In the inferred case, it appears to resolve to the base type of the DataType. In both cases, these need to be the same to avoid such errors.

To solve this:
- we should be able to write a declared type that directly utilizes the DT datatype, just like the inferred case.
- the check here should resolve correctly when a DataType is essentially the same as another type (in this example, that would be `string`), https://github.com/langium/langium/blob/25e3902abd6bf2b77b20f26413c7a8e8596f63b5/packages/langium/src/grammar/type-system/type-validator.ts#L322

claaspetersen:
This closes #559 by adding a new function `getExplicitRuleTypeName` that returns the names of DataTypes and modifying the function `getExplicitRuleType` to return the correct type for DataTypes which in turn fixes the extension of DataTypes by Declared types and the AST generation. 

Having:
```
interface R1 { 
    val: DT 
}

R1 returns R1: 'r1' val=DT;
DT returns string: 'dt';
```

now returns the expected AST:
```typescript
export type DT = string;

export interface R1 extends AstNode { val: DT }
```

This is also the same behavior as if (in this case R1) would be inferred


msujew (commented):
Hey @claaspetersen,

while this successfully addresses the issue. I'm not sure whether we should add more complexity to the already quite complex structure that we've build here. On the contrary, I believe that the complexity that we've introduced is the source of the original issue. Instead of adding on top of it, we should try to reduce it and make it clear when one should use which function to get want they need.

spoenemann:
I agree we should rather work to reduce the code in grammar-util.ts rather than add more to it.

@claaspetersen simple solution: move the dataType case to its usage:
```typescript
export function getTypeName(type: ast.AbstractType | ast.InferredType): string {
    if (ast.isParserRule(type)) {
        if (rule.dataType) {
            return rule.name;
        }
        return getExplicitRuleType(type) ?? type.name;
    } else ...
```

packages/langium/src/grammar/grammar-util.ts
```diff
@@ -308,7 +308,7 @@ function withCardinality(regex: string, cardinality?: string, wrap = false): str
 
 export function getTypeName(type: ast.AbstractType | ast.InferredType): string {
     if (ast.isParserRule(type)) {
-        return getExplicitRuleType(type) ?? type.name;
+        return getExplicitRuleTypeName(type) ?? type.name;
```
montymxb, line 311: Referencing the comment below, here would be a great spot to pull such a check up to. See below for details.

packages/langium/src/grammar/grammar-util.ts
```diff
@@ -322,6 +322,25 @@ export function getTypeName(type: ast.AbstractType | ast.InferredType): string {
     throw new TypeResolutionError('Cannot get name of Unknown Type', type.$cstNode);
 }
 
+export function getExplicitRuleTypeName(rule: ast.ParserRule): string | undefined {
+    if (rule.inferredType) {
+        return rule.inferredType.name;
+    } else if (rule.dataType) {
+        return rule.name;
```
montymxb, line 329: Seems like this is the only change here. If that's the case, we can pull this out like what's done below:
https://github.com/langium/langium/blob/17f062935c2915889cf5a1def5bbddf90fa3fea2/packages/langium/src/grammar/grammar-util.ts#L376
And in such a case, that change could be lifted up to the portion noted above in the previous comment, and this function could be dropped -- especially since it duplicates most of what's already present in `getExplicitRuleType` below.

montymxb (commented):
I agree with @msujew & @spoenemann . The fix here is good, just validated it myself, but it would be good to apply some changes to the implementation.

packages/langium/src/grammar/grammar-util.ts
claaspetersen: I used your example and fitted the check into the code using a ternary operator. It works and should be less complex than before now.

spoenemann (approved):
:+1:

montymxb (approved):
Agree with miro, this looks great now! Would still be nice to have a test or two, but this is okay to take as is.

PR #667, opened 2022-09-06
Changed semantic token for fragments

### LINKED ISSUE #602: Use a different (semantic token) color for fragment calls than for rule calls
I wouldn't change the syntax because a parser rule can be without assignment as well (unassigned rule call, see also #205). But using a special color using semantic tokens sounds good.

_Originally posted by @spoenemann in https://github.com/langium/langium/discussions/567#discussioncomment-3039927_

gfontorbe:
Changed the semantic token for fragments inside of a Parser Rule.

I chose `SemanticTokenTypes.type` to have something similar to the color we have in `type A = B  |  C`

Closes #602 

Screenshot 2022-09-06 at 10 09 46
Screenshot 2022-09-06 at 10 09 59


Lotes:
Can you show a normal non-terminal rule call as difference?

gfontorbe:
Sure, I didn't touch the tokens for rule calls that are not fragments. They still appear in white (in dark theme) similar to terminal rule calls

Screenshot 2022-09-06 at 10 35 34

Lotes (approved):
Looks good to me.

PR #668, opened 2022-09-06
Split find declaration method for AST and CST

msujew:
Performs a few changes:

1. Splits `findDeclaration` into `findDeclaration` (returns `AstNode`) and `findDeclarationNode` (returns `CstNode`). We need this to perform `findDeclaration` calls on synthetic/in-memory AST nodes. Useful for LSP requests such as Hover, which do nothing as of now, if the referenced node is synthetic.
2. Renames a bunch of `GoTo` services so that they all align to a single naming schema.
3. Makes all LSP services optional, so that they can be disabled if the downstream dev wants to.
4. Introduces a `GenericAstNode` interface, which removes the need for some `any` casts.


spoenemann (approved):
Thanks!

PR #670, opened 2022-09-09
bug fix in interface inference for rules with actions referencing a declared type

sailingKieler:
This change fixes the type inference for parser rules like:
```
MemberAccess infers Expression:
    PrimaryExpression (
        {infer MemberAccess.receiver=current} '.' member=[Symbol:ID]
    )+
;

interface SuperMemberAccess extends MemberAccess {}
SuperMemberAccess infers Expression:
    PrimaryExpression (
        {SuperMemberAccess.receiver=current} '.' member=[Symbol:'super']
    )+
;
```
as well as the combination in a single rule, like

```
interface SuperMemberAccess extends MemberAccess {}
MemberAccess infers Expression:
    PrimaryExpression (
        {ThisMemberAccess.receiver=current} '.' member=[Symbol:'super'] |
        {infer MemberAccess.receiver=current} '.' member=[Symbol:ID]
    )+
;
```

### DIFF
--- a/packages/langium/test/grammar/type-system/inferred-types.test.ts
+++ b/packages/langium/test/grammar/type-system/inferred-types.test.ts
@@ -4,7 +4,7 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { AstTypes, createLangiumGrammarServices, Grammar, InterfaceType, isDataTypeRule, isParserRule, Property, PropertyType, UnionType, EmptyFileSystem } from '../../../src';
+import { AstTypes, createLangiumGrammarServices, Grammar, InterfaceType, isDataTypeRule, isParserRule, Property, PropertyType, UnionType, EmptyFileSystem, collectAllAstResources, collectAst, s } from '../../../src';
 import { collectInferredTypes } from '../../../src/grammar/type-system/inferred-types';
 import { parseHelper } from '../../../src/test';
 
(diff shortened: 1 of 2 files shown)


packages/langium/test/grammar/type-system/inferred-types.test.ts
```diff
@@ -578,6 +578,135 @@ describeTypes('inferred types that are used by the grammar', `
...
+            export interface MemberAccess extends AstNode {
+                readonly $container: MemberAccess;
+                member: Reference<Symbol>
+                receiver: PrimaryExpression
+            }
+        `);
+
+        const allInterfaces = collectAst(undefined!, [grammar]).interfaces;
+        const allInterfacesString = allInterfaces.sort(sortByName).map(toSubstring).join('\n').trim();
+        expect(allInterfacesString).toBe(s`
```
msujew, line 687: Suggestion: I'm not sure whether this kind of assertion will survive the next refactoring of the type inference. Can you refactor the string based comparison to the one that's used in the rest of the tests? What do you think about that?

msujew (commented):
The change looks good, but I have a comment on the tests.

packages/langium/test/grammar/type-system/inferred-types.test.ts
sailingKieler: Basically I like the string-based assertion checking a lot, and I explicitly advocate that.

Relying on the "generated code"-based `toString()` representation can of course be questioned.
It was just convenient and easy to understand, once one is familiar the generated AST representation, I currently have the time to implement a dedicated text rep.

I suggest to implement a beneficial `toString()` rep of the internal type objects during the refactoring,
and suggest to consider the string based comparison an inspiration, as I personally prefer the human readable diffs in case of failures a lot over a certain diff in
```
function expectProperty(interfaceType: InterfaceType, property: Property | string): void {
    if (typeof property === 'string') {
        const prop = getProperty(interfaceType, property)!;
        expect(prop).toBeDefined();
        expect(prop.optional).toStrictEqual(false);
    } else {
        const prop = getProperty(interfaceType, property.name)!;
        expect(prop).toBeDefined();
        expect(prop.optional).toStrictEqual(property.optional);
        expect(prop.typeAlternatives.length).toStrictEqual(property.typeAlternatives.length);
        for (let i = 0; i < prop.typeAlternatives.length; i++) {
            const actualType = prop.typeAlternatives[i];
            const expectedType = property.typeAlternatives[i];
            expect(actualType.types).toEqual(expectedType.types);
            expect(actualType.array).toEqual(expectedType.array);
            expect(actualType.reference).toEqual(expectedType.reference);
        }
    }
}
```
where first have to investigate the context of the failure.

msujew (approved):
Alright, I can agree with that line of reasoning. Looks good to me 👍

Issue #671, opened 2022-09-10
langium:generate fails with imported grammars


goto40:
When importing a file (`import './grammar'`) 'langium:generate' produces an error: `TypeError: Cannot read properties of undefined (reading '$container')`.

Langium version: 8c5d43d26a2afd9dadacff02328ff619d74df30b
Package name: langium

## Steps To Reproduce

1. checkout langium @ 8c5d43d26a2afd9dadacff02328ff619d74df30b
2. `npm ci`
3. add file `dummy.langium` with `terminal Dummy: '*dummy';` inside at `examples/domainmodel/src/language-server`.
4. add `import './dummy'` in `domain-model.langium` 
5. `npm run langium:generate -w example/domainmodel`

## The current behavior

```
npm run langium:generate -w examples/domainmodel/

> langium-domainmodel-dsl@0.4.0 langium:generate
> langium generate

Reading config from langium-config.json
src/language-server/domain-model.langium:34:10 - This rule is declared but never referenced.
src/language-server/domain-model.langium:35:10 - This rule is declared but never referenced.
TypeError: Cannot read properties of undefined (reading '$container')
npm ERR! Lifecycle script `langium:generate` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: langium-domainmodel-dsl@0.4.0 
npm ERR!   at location: /home/pierre/internet/langium/examples/domainmodel 
```

## The expected behavior

no such error.

goto40:
I also debugged a little bit but I do not have enough knowledge/experience to find my way at the current state of my experiments...

It seems to happen in the constructor of `LangiumParser` `langium-parser.ts`, while assigning `this.linker = services.references.Linker;`.

msujew:
I can confirm the issue. This seems to be a side effect of https://github.com/langium/langium/pull/546, which is why this issue doesn't appear in version 0.4. However the fix is luckily very simple, see https://github.com/langium/langium/pull/672.

Thanks for catching this 👍

PR #672, opened 2022-09-10
Remove grammar imports on rule embedding

msujew:
Closes https://github.com/langium/langium/issues/671

This is an issue which has accidentally been introduced in https://github.com/langium/langium/pull/546. When trying to validate the parser of a grammar with imports, the AST reflection interpreter will try to resolve the imports of the grammar. However, this is not necessary, since all imports have already been resolved at this point. Instead, it will just crash the langium-cli process.

This change removes all imports once they have been embedded, making this a non-issue.


goto40 (approved):
Looks good! Thank you for the very fast response to #671, @msujew!

pluralia (approved):

PR #675, opened 2022-09-12
Update language-server and testing libraries

msujew:
It seems like the `8.0.2` update of the language-server-protocol library contains a breaking change, see https://github.com/langium/langium/discussions/674. Users downloading that update will encounter a breaking change in Langium.

Also updates jest to `29.0.0`, that should also resolve the issue brought up in https://github.com/langium/langium/discussions/653.

### DIFF
--- a/packages/langium-vscode/package.json
+++ b/packages/langium-vscode/package.json
@@ -64,8 +64,7 @@
   },
   "dependencies": {
     "langium": "0.4.0",
-    "vscode-languageclient": "^8.0.1",
-    "vscode-languageserver": "^8.0.1",
+    "vscode-languageserver": "^8.0.2",
     "ignore":"~5.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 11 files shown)


packages/langium-vscode/package.json
```diff
@@ -64,8 +64,7 @@
   },
   "dependencies": {
     "langium": "0.4.0",
-    "vscode-languageclient": "^8.0.1",
-    "vscode-languageserver": "^8.0.1",
+    "vscode-languageserver": "^8.0.2",
```
spoenemann, line 67: Why did you remove the client lib here? We use it in extension.ts

spoenemann (approved):

spoenemann:
Test failure:

> packages/langium-cli/test/generator/module-generator.test.ts:21:34 - error TS2339: Property 'not' does not exist on type 'Matchers<string, {}>'.

packages/langium-vscode/package.json
spoenemann: @msujew what about this? Seems to work anyway because of hoisted dependencies, but it seems unclean.

msujew: Ups, forgot about this. I'll create another PR.

Issue #655, opened 2022-08-28
expectError() Behavior


msujew:
### Discussed in https://github.com/langium/langium/discussions/654

<div type='discussions-op-text'>

<sup>Originally posted by **snarkipus** August 27, 2022</sup>
Not sure if this is an issue or not, but the behavior doesn't quite seem right.  If needed, I can try to come up with a minimal test case.

### Validator Test Case:
```typescript
...
    it('Should detect method invocation on a field', () => {
        const mark = validationResult.document.textDocument.getText().lastIndexOf('(');
        expectError(validationResult, "Method invocation on a field",
        {
            offset: mark,
            length: 2,
        });
    });
...
```
Note: Test passes with length of 2, but fails with length of 1 (off by one in range)

### Code [langium-tests.ts](https://github.com/langium/langium/blob/4af91a15ff4275b434281e90330d5f1ebaacc0a5/packages/langium/src/test/langium-test.ts#L318-L324)
```typescript
...
    if ('offset' in options) {
        const outer = {
            start: validationResult.document.textDocument.positionAt(options.offset),
            end: validationResult.document.textDocument.positionAt(options.offset + options.length - 1)
        };
        filters.push(d => isRangeEqual(outer, d.range));
    }
...
```
Note: wondering if the `end` decrement `options.offset + options.length - 1` is correct?
### Validator in action:
<img width="742" alt="image" src="https://user-images.githubusercontent.com/3688989/187049253-9a489fea-fba8-4335-b425-c1e43228e9fb.png">

I don't believe that this would have been caught in the tests since it's uncovered in the test report:
<img width="626" alt="image" src="https://user-images.githubusercontent.com/3688989/187049493-9bfcdaf8-502e-4b22-8ec6-318d89878474.png">

</div>

snarkipus:
Asserting expectError() in validator unit tests using offset & length does not work as expected (see discussion).

Langium version: 0.4
Package name: langium/test

## Steps To Reproduce

1. Update default validator - I'm sure there's a cleaner way to do this, but it works ([here](https://github.com/snarkipus/langium-dev/blob/error-behavior/examples/hello-world/src/language-server/hello-world-validator.ts))
```typescript
    checkPersonStartsWithCapital(person: Person, accept: ValidationAcceptor): void {
        if (person.name) {
            const firstChar = person.name.substring(0, 1);
            if (firstChar.toUpperCase() !== firstChar) {
[...]

PR #677, opened 2022-09-15
expectError() off by one fix

snarkipus:
Fixes Issue #655 

Apologies, I had completely forgotten about this until it came up during today's discussion.


msujew (commented):
Thanks for the contribution. Do you mind adding a regression test for that, so we won't run into this issue again in the future?

snarkipus:
Hopefully that works. I just tucked the regression test into the existing test code, although I didn't explicitly reference the issue for regression.

packages/langium/test/validation/grammar-validation.test.ts
```diff
@@ -155,7 +155,9 @@ describe('Unordered group validations', () => {
         const validation = await validate(text);
         expect(validation.diagnostics).toHaveLength(1);
         const range = { start: { character: 18, line: 7 }, end: { character: 45, line: 7 } };
+        const offset = validation.document.textDocument.getText().indexOf('("edition" version=STRING)?');
         expectError(validation, 'Optional elements in Unordered groups are currently not supported', { range, code: IssueCodes.OptionalUnorderedGroup } );
+        expectError(validation, 'Optional elements in Unordered groups are currently not supported', { offset: offset, length: 27, code: IssueCodes.OptionalUnorderedGroup } );
```
msujew, line 160: Minor: Since the second assertion does the same thing as the first, I believe we can just use one of them. Also, hardcoding the length seems like a plausible cause of errors in the future. So we should just use the length of the target string.
```suggestion
        const errorText = '("edition" version=STRING)?';
        const offset = validation.document.textDocument.getText().indexOf(errorText);
        expectError(validation, 'Optional elements in Unordered groups are currently not supported', { offset: offset, length: errorText.length, code: IssueCodes.OptionalUnorderedGroup } );
```

msujew (commented):
Great, just one minor thing, see below.

packages/langium/test/validation/grammar-validation.test.ts
snarkipus: Makes sense to me. 

However, it was intentionally duplicative since I couldn't see anywhere else the `ExpectDiagnosticRangeOptions` variant of the `filterOptions` argument was asserted in langium code. I was/am hesitant to convert any of the `ExpectDiagnosticAstOptions<T>` (`node`) based tests over to `range`/`offset` since I'm not sure how much you guys really want me mucking with the core code base. :)

That's probably the right thing to do though since the `filterOptions` argument and associated logic are tied into all of the `ExpectXXX()` functions. I'm happy to do it, just wanted to check with you guys first.

msujew (approved):
LGTM, thanks 👍

PR #678, opened 2022-09-15
additional customization hook in 'DefaultScopeComputation'

sailingKieler:
supplement on #664

Simplifies customization of indexing and re-use of the routine for creating `AstNodeDescriptions`.

### DIFF
--- a/packages/langium/src/references/scope-computation.ts
+++ b/packages/langium/src/references/scope-computation.ts
@@ -58,11 +58,15 @@ export interface ScopeComputation {
 }
 
 /**
- * The default scope computation gathers all AST nodes that have a name (according to the `NameProvider`
- * service) and makes them available in their container node. As a result, from every cross-reference in
- * the AST, target elements from the same level and further up towards the root are visible. Elements that
- * are nested inside lower levels are not visible by default, but that can be changed by customizing this
- * service.
+ * The default scope computation creates and collectes descriptions of the AST nodes to be exported into the
+ * _global_ scope from the given document. By default those are the document's root AST node and its directly
+ * contained child nodes.
+ *
+ * Besides, it gathers all AST nodes that have a name (according to the `NameProvider` service) and includes them
+ * in the local scope of their particular container nodes. As a result, for every cross-reference in the AST,
+ * target elements from the same level (siblings) and further up towards the root (parents and siblings of parents)
+ * are visible. Elements being nested inside lower levels (children, children of siblings and parents' siblings)
+ * are _invisible_ by default, but that can be changed by customizing this service.
  */
 export class DefaultScopeComputation implements ScopeComputation {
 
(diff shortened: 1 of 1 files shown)


packages/langium/src/references/scope-computation.ts
```diff
@@ -75,11 +75,15 @@ export class DefaultScopeComputation implements ScopeComputation {
     }
 
     async computeExports(document: LangiumDocument, cancelToken = CancellationToken.None): Promise<AstNodeDescription[]> {
+        return this.computeExportsForNode(document.parseResult.value, document, undefined, cancelToken);
+    }
+
+    async computeExportsForNode(parentNode: AstNode, document: LangiumDocument<AstNode>, children?: (root: AstNode) => Iterable<AstNode>, cancelToken: CancellationToken = CancellationToken.None ): Promise<AstNodeDescription[]> {
```
msujew, line 81: Suggestion: You can improve this signature to simplify the ternary inside.
```suggestion
    async computeExportsForNode(parentNode: AstNode, document: LangiumDocument<AstNode>, children: (root: AstNode) => Iterable<AstNode> = streamContents, cancelToken: CancellationToken = CancellationToken.None ): Promise<AstNodeDescription[]> {
```

sailingKieler: I know and I had it like this before.
However, the debugging experience (the crazy bouncing of the debugger) convinced me to add the ternary explicitly...

sailingKieler: Just gave it another try and it feels ok, maybe I had it like
`children: (root: AstNode) => Iterable<AstNode> = () => streamContents(root)` earlier.

Wrt. to readability I'm not sure whether
```
async computeExportsForNode(parentNode: AstNode, document: LangiumDocument<AstNode>, children: (root: AstNode) => Iterable<AstNode> = streamContents, cancelToken: CancellationToken = CancellationToken.None ): Promise<AstNodeDescription[]> {
        const exports: AstNodeDescription[] = [];

        this.exportNode(parentNode, exports, document);
        for (const node of children(parentNode)) {
            await interruptAndCheck(cancelToken);
            this.exportNode(node, exports, document);
        }
}
```
is preferable over
```
async computeExportsForNode(parentNode: AstNode, document: LangiumDocument<AstNode>, children?: (root: AstNode) => Iterable<AstNode>, cancelToken: CancellationToken = CancellationToken.None ): Promise<AstNodeDescription[]> {
        const exports: AstNodeDescription[] = [];

        this.exportNode(parentNode, exports, document);
        const childrenIterable = children ? children(parentNode) : streamContents(parentNode); // Here we navigate only the top-level AST contents - nested nodes are not exported by default
        for (const node of childrenIterable) {
            await interruptAndCheck(cancelToken);
            this.exportNode(node, exports, document);
        }
}
```

esp. wrt. the fact that we should make the default behavior
`// Here we navigate only the top-level AST contents - nested nodes are not exported by default`
as transparent as possible.

What do you guys think?

msujew: I would put this comment as a JSDoc for the `children` parameter of the method. We should document these internal methods a bit better anyway. Also, should we make this `protected`?

sailingKieler: I left it public on purpose, as it shall be callable from outside, e.g. for setting up a built-in libraries that should _not_ be visible via the global index.

sailingKieler: That way you will end up with code clones that is also questionable.
And currently I have both cases: Call the method from out side for producing AstNodeDescriptions, and specializing the children to be visited while performing the regular indexing.

spoenemann: > That way you will end up with code clones that is also questionable.

My experience so far has shown that the dependencies introduced by code reuse tend to be more problematic than cloning code. And I've seen this observation being confirmed by many others (on Twitter etc.)

But I'm ok with your approach in this case. Could you add documentation about how to use the new method? I'd also move the inline comment about top-level AST contents into the method doc.

sailingKieler: I'll also extend the doc of https://github.com/langium/langium/blob/a7017f45924ade3cca23e985cb04d4a65ba3e9a9/packages/langium/src/references/scope-computation.ts#L60-L67
wrt. the exports computation.
Agreed?

sailingKieler: doc update is done.

packages/langium/src/references/scope-computation.ts
```diff
@@ -75,11 +79,26 @@ export class DefaultScopeComputation implements ScopeComputation {
     }
 
     async computeExports(document: LangiumDocument, cancelToken = CancellationToken.None): Promise<AstNodeDescription[]> {
+        return this.computeExportsForNode(document.parseResult.value, document, undefined, cancelToken);
+    }
+
+    /**
+     * Creates {@link AstNodeDescription AstNodeDescriptions} for the given @link AstNode parentNode} and its children.
```
spoenemann, line 86: These `@link` tags don't seem to work in VS Code. I'd propose to use `` `code-style` `` instead.

spoenemann (approved):

packages/langium/src/references/scope-computation.ts
sailingKieler: For me they worked actually, otherwise I wouldn't have used them ... I have a quite brand new version

sailingKieler: image

Ok, they click experience is not ultimately perfect yet...

sailingKieler: Just missed a `{` ... damn.

PR #684, opened 2022-09-19
Pass target type into reference

msujew:
In a few cases we had issues resolving to the expected type of a reference. This is due to `AstReflection.getReferenceType` not always working correctly in cases such as fragments.

This change stores the target type on the reference when it is created, removing the need for `getReferenceType` altogether.


spoenemann:
This change has a cost: it adds an additional property to every instance of a reference, which can mean a lot for large ASTs.

Could you identify a specific grammar example where reference resolution fails with the previous approach?

spoenemann:
Can this be closed in favor of #822?

PR #685, opened 2022-09-19
Prevent unused token creation

msujew:
When creating a language from multiple grammars, we often have unused rules within then. Right now, the token builder and other associated services always create _all tokens_ that can be found within the grammar. This change addresses this by only looking at rules and terminals which are actually used either directly or transitively by the entry rule.


montymxb (approved):
Left some comments, but functionality looks good to add 👍 . Verification checks all terms, but it's nice not to include those later on.

PR #687, opened 2022-09-19
Enable debugging Langium as downstream user

msujew:
Closes https://github.com/langium/langium/issues/572

Also changes from the deprecated `pwa-node` to `node`.


gfontorbe (approved):
Tested on several projects, works like a charm!

montymxb (approved):
Cool, switch from progressive to just stable node looks good 👍

PR #688, opened 2022-09-19
Support document link requests

msujew:
The document link requests are handled on a shared language level, as we don't know what language a link belongs to when handling the `resolveDocumentLink` request. Users can implement their own language specific solutions if they absolutely need to.


packages/langium/src/lsp/language-server.ts
```diff
@@ -71,6 +71,7 @@ export class DefaultLanguageServer implements LanguageServer {
         const hasCodeActionProvider = this.hasService(e => e.lsp.CodeActionProvider);
         const hasSemanticTokensProvider = this.hasService(e => e.lsp.SemanticTokenProvider);
         const commandNames = this.services.lsp.ExecuteCommandHandler?.commands;
+        const documentLinkProvider = this.services.lsp.DocumentLinkProvider;
```
gfontorbe, line 74: In order to be in line with other providers, would it make sense to also call the `hasService` method here?
```suggestion
        const hasDocumentLinkProvider = this.hasService(e => e.shared.lsp.DocumentLinkProvider);
```

packages/langium/src/lsp/language-server.ts
```diff
@@ -117,6 +118,9 @@ export class DefaultLanguageServer implements LanguageServer {
                 implementationProvider: hasGoToImplementationProvider,
                 callHierarchyProvider: hasCallHierarchyProvider
                     ? {}
+                    : undefined,
+                documentLinkProvider: documentLinkProvider
+                    ? { resolveProvider: !!documentLinkProvider.resolveDocumentLink }
                     : undefined
```
gfontorbe, line 124: This would also change if above is accepted
```suggestion
                documentLinkProvider: hasDocumentLinkProvider
                    ? { resolveProvider: !!this.services.lsp.DocumentLinkProvider?.resolveDocumentLink }
                    : undefined
```

gfontorbe (approved):
Very minor comment, otherwise looks good to me

packages/langium/src/lsp/language-server.ts
gfontorbe: Makes sense

PR #689, opened 2022-09-19
Clarified public API of grammar utils

spoenemann:
This splits the grammar utils into two files `grammar/internal-grammar-util.ts` (not exported to index) and `utils/grammar-util.ts` (exported to index). We can of course discuss which functions should go where, this is just a proposal. The main reason for this split is that many utility functions are used by the implementation of the grammar language, but are almost certainly not required when you implement your own language.

I added comments to the "public API" utility functions and cleaned up a few other things along the way.


msujew (approved):
Thanks, the new API (especially the names) looks much cleaner to me 👍

PR #690, opened 2022-09-19
Vitest

snarkipus:
No issue per se - just a possible enhancement for consideration.


vite.config.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/// <reference types="vitest" />
+import { defineConfig } from 'vite';
+
+export default defineConfig({
+    test: {
+        globals: true,
+        coverage: {
+            provider: 'c8',
+            reporter: ['text', 'html']
```
msujew, line 15: Just a minor comment: We're only looking for coverage in our core framework package (as of now). Also, we're excluding generated files.

```suggestion
            provider: 'c8',
            reporter: ['text', 'html'],
            include: ['packages/langium/src'],
            exclude: ['**/generated']
```

tsconfig.json
```diff
@@ -48,7 +48,7 @@
     // "rootDirs": [],                        /* List of root folders whose combined content represents the structure of the project at runtime. */
     // "typeRoots": [],                       /* List of folders to include type definitions from. */
     "types": [
-      "jest",
+      "vitest/globals",
```
msujew, line 51: For some reason, vscode doesn't seem to pick these types up, when navigating files in the `test` folder. Can you add this line to the `fixture.ts` file?

```ts
import 'vitest/globals';
```

vite.config.ts
```diff
@@ -0,0 +1,18 @@
...
+        }
+    }
+});
```
msujew, line 18: Nitpick: Missing newline.

.vscode/launch.json
```diff
@@ -111,33 +111,25 @@
...
+            "args": ["run", "--no-color", "--no-coverage", "--no-watch"],
+            "smartStep": true,
             "console": "integratedTerminal",
-            "internalConsoleOptions": "neverOpen"
         },
         {
-            "name": "Jest: Run Selected File",
-            "type": "node",
+            "name": "Vitest: Run Selected File",
+            "type": "pwa-node",
```
msujew, line 125: Minor:
```suggestion
            "type": "node",
```

.vscode/launch.json
```diff
@@ -111,33 +111,25 @@
             ]
         },
         {
-            "name": "Jest: Run All",
-            "type": "node",
+            "name": "Vitest: Run All",
+            "type": "pwa-node",
```
msujew, line 115: Minor:
```suggestion
            "type": "node",
```

msujew (commented):
This looks great, thanks a lot! Just some minor comments and then it's already good to go 👍

.vscode/launch.json
snarkipus: fixed

.vscode/launch.json
snarkipus: fixed

vite.config.ts
snarkipus: fixed

tsconfig.json
snarkipus: This one is interesting - somehow it wasn't appearing in my environment due to an 'unknown to me' magical node_modules package cache living in my home directory. It was just resolving to the @types/jest that lived there.

Adding this resolves the vscode linting ... but breaks all of the tests which call `expectFunction()` from `fixture.ts`.

```term
 FAIL  packages/langium/test/lsp/completion-provider.test.ts [ packages/langium/test/lsp/completion-provider.test.ts ]
 FAIL  packages/langium/test/lsp/document-symbol.test.ts [ packages/langium/test/lsp/document-symbol.test.ts ]
 FAIL  packages/langium/test/lsp/folding-range.test.ts [ packages/langium/test/lsp/folding-range.test.ts ]
 FAIL  packages/langium/test/lsp/goto-definition.test.ts [ packages/langium/test/lsp/goto-definition.test.ts ]
 FAIL  packages/langium/test/lsp/hover.test.ts [ packages/langium/test/lsp/hover.test.ts ]
Error: No known conditions for "./globals" entry in "vitest" package
```

So, I'm at a bit of a loss - I went ahead and committed the code with the import masked incase someone had an idea. I'll keep poking at it to see if I can figure it out.

msujew (commented):
Hm, I see the issue. It's very likely related to the fact that vitest is an ESM only test engine. This confuses both TypeScript and Vitest itself for some reason. Do you mind reverting your latest two commits and instead add the types of jest to the root package json?

```
"@types/jest": "^29.0.3",
"@types/jest-expect-message": "^1.0.4",
```

And also re-add `jest` instead of `vitest/globals` to the typescript configuration.

snarkipus:
Ok - I believe that should check the boxes. I'm just not savvy enough to understand the full implications of vitest being 'ESM-first'.

msujew (approved):
Thanks, looks good to me 👍

PR #694, opened 2022-09-22
Copy all grammar elements for multi-dsl projects

msujew:
Declared types and interfaces from other grammars currently aren't copied to their target grammar in the langium-cli. This leads to errors during generation and runtime. This PR addresses this oversight.

### DIFF
--- a/packages/langium-cli/src/generate.ts
+++ b/packages/langium-cli/src/generate.ts
@@ -93,6 +93,8 @@ function embedReferencedGrammar(grammar: Grammar, map: Map<Grammar, GrammarEleme
                 grammar.types.push(shallowCopy);
             } else if (isInterface(shallowCopy)) {
                 grammar.interfaces.push(shallowCopy);
+            } else {
+                throw new Error('Received invalid grammar element while generating project with multiple languages');
             }
         }
     }
--- a/packages/langium-cli/src/generate.ts
+++ b/packages/langium-cli/src/generate.ts
@@ -6,8 +6,8 @@
 
 import fs from 'fs-extra';
 import {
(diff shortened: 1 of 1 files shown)


packages/langium-cli/src/generate.ts
```diff
@@ -53,41 +55,51 @@ function eagerLoad(document: LangiumDocument, uris: Set<string> = new Set()): UR
...
+            if (isParserRule(shallowCopy)) {
+                shallowCopy.entry = false;
+            }
+            if (isAbstractRule(shallowCopy)) {
+                grammar.rules.push(shallowCopy);
+            } else if (isType(shallowCopy)) {
+                grammar.types.push(shallowCopy);
+            } else if (isInterface(shallowCopy)) {
+                grammar.interfaces.push(shallowCopy);
+            }
```
montymxb, line 96: Should we be handling the additional case if something slips through that's not one of these 3? It shouldn't happen, but if it did there's no reporting.

packages/langium-cli/src/generate.ts
```diff
@@ -53,41 +55,51 @@ function eagerLoad(document: LangiumDocument, uris: Set<string> = new Set()): UR
...
     for (const importedGrammar of allGrammars) {
-        const rules = map.get(importedGrammar);
-        if (rules) {
-            grammar.rules.push(...rules);
+        const grammarElements = map.get(importedGrammar) ?? [];
+        for (const element of grammarElements) {
+            const shallowCopy = { ...element };
+            // Deactivate copied entry rule
+            if (isParserRule(shallowCopy)) {
+                shallowCopy.entry = false;
```
montymxb, line 88: Right, have to clear flag on an entry rule from an included grammar.

montymxb (approved):
Looks good!

PR #696, opened 2022-09-23
Removed inferred type union validation

### LINKED ISSUE #680: Inferred Type Union Validation is Incorrect
Langium version: pre-0.5.0 (commit 52ab0856ca2f1a7c71f45555783c504da3bb594d)

I added a PR back in July (#548) to help address the issue of declared types extending type unions, when they were inferred. The initial fix works in most cases, but is lacking a critical element.

Attempting to determine whether an inferred type will be a union or interface cannot be done on the basis of a single deriving rule. There is no way to ensure that further rules will not alter the inferred type to change it. The only case that comes up is when a union type becomes an inferred type (but not the other way around).

For example, the following is correctly caught, where the inferred type `Rule` is a type union.
ex1

But this is incorrectly flagged due to the type being incomplete when derived from only one contributing rule. Here `Rule2` takes what would be inferred as a type union, and adds properties to it, changing it to an interface. 
ex2

To amend this, I would suggest we revert this change, and drop the validation until we can do it efficiently (not recomputing complete inferred types on every check). Ideally this could come back after the upcoming type system work, which should remove this problem by pre-computing types before validation that relies on them.

gfontorbe:
Closes #680 

The validation for inferred type union was not working as expected.

For now I disabled it by commenting out the necessary lines in order to not lose @montymxb 's work. This should be reimplemented when we are through with the type system modifications.

### DIFF
--- a/packages/langium/src/grammar/langium-grammar-validator.ts
+++ b/packages/langium/src/grammar/langium-grammar-validator.ts
@@ -22,7 +22,6 @@ import * as ast from './generated/ast';
 import { isParserRule, isRuleCall } from './generated/ast';
 import { getTypeName, isDataTypeRule, isOptionalCardinality, resolveImport, resolveTransitiveImports, terminalRegex } from './internal-grammar-util';
 import type { LangiumGrammarServices } from './langium-grammar-module';
-import { collectInferredTypes } from './type-system/inferred-types';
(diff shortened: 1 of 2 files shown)


montymxb (approved):
:octocat: approved! Double checked and the validations are no longer present, but nice to keep for when they are ready to be re-introduced.

Issue #268, opened 2021-10-22
Handle keywords in completion


spoenemann:
Currently we show all keywords in completion, even if they consist only of a single character. This is rather unusual if you compare it to other language servers.

How could we handle this better?
 * Hide all keywords?
 * Hide keywords that consist of non-alphabetic characters?
 * Hide keywords if we haven't found any other proposals?

We should also consider priority levels of proposals. Cross-references should have higher priority than keywords. This affects the order and possibly also filtering (see above).

PR #697, opened 2022-09-23
Hide non alphabetic tokens

gfontorbe:
This is a proposal for addressing #268 

It adds a new boolean entry in `langium-config.ts` named "showNonAlphabeticKeywords". It allows for the completion provider to skip keywords that are composed of a single non-alphabetic character. If not present in the config file, it defaults to `false`

With "showNonAlphabeticKeywords" set to `true`:
Screenshot 2022-09-23 at 14 47 47

With "showNonAlphabeticKeywords" set to `false`:
Screenshot 2022-09-23 at 14 48 23

One thing I didn't manage to achieve is to add support for this new entry in `langium-config.ts`:
Screenshot 2022-09-23 at 14 48 50
The json validation shows an error but the module is generated correctly.

PS: I didn't adapt the completion tests yet. Will do in the same PR if the approach is accepted

### DIFF
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -9,7 +9,6 @@ import { TextDocument, TextEdit } from 'vscode-languageserver-textdocument';
 import * as ast from '../../grammar/generated/ast';
 import { GrammarConfig } from '../../grammar/grammar-config';
 import { getExplicitRuleType } from '../../grammar/internal-grammar-util';
-import { LanguageMetaData } from '../../grammar/language-meta-data';
 import { LangiumCompletionParser } from '../../parser/langium-parser';
 import { NameProvider } from '../../references/naming';
 import { ScopeProvider } from '../../references/scope-provider';
@@ -45,7 +44,6 @@ export class DefaultCompletionProvider implements CompletionProvider {
     protected readonly grammar: ast.Grammar;
     protected readonly nameProvider: NameProvider;
     protected readonly grammarConfig: GrammarConfig;
-    protected readonly languageMetadata: LanguageMetaData;
 
     constructor(services: LangiumServices) {
         this.scopeProvider = services.references.ScopeProvider;
@@ -53,7 +51,6 @@ export class DefaultCompletionProvider implements CompletionProvider {
         this.completionParser = services.parser.CompletionParser;
         this.nameProvider = services.references.NameProvider;
         this.grammarConfig = services.parser.GrammarConfig;
-        this.languageMetadata = services.LanguageMetaData;
     }
 
     async getCompletion(document: LangiumDocument, params: CompletionParams): Promise<CompletionList | undefined> {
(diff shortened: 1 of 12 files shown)


packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -239,6 +242,9 @@ export class DefaultCompletionProvider implements CompletionProvider {
     }
 
     protected completionForKeyword(keyword: ast.Keyword, context: AstNode | undefined, acceptor: CompletionAcceptor): MaybePromise<void> {
+        if (keyword.value.match(/^[\W]/) && keyword.value.length === 1 && !this.languageMetadata.showNonAlphabeticKeywords) {
```
Lotes, line 245: Hard to read... what is happening?

Lotes: Starts with a non word character of length 1?

gfontorbe: This checks if the the keyword start with a non-alphabetic token and is a single character.
I don't think hiding keywords such as `_keyword` from completion would be correct.

It also checks if the option `showNonAlphabeticKeywords ` is set to false.

For better readability, I guess we could first check for the option in an other if. What do you think?

Lotes, line 245: The name `showNonAlphabeticKeywords` makes it a bit hart to think about it:
1. `show` -> ok
2. `non-alphabetic-keywords`
   1. `keyword` makes sense
   2. `alphabethic`, mhhh ok like ABC
   3. `non` = negating the term before
3. now try to read `!non-alphabetic-keywords` , what is it?

Proposal: `showOnlyAlphabeticKeywords`... ok, when NOTTING it is a bit the same. ^^*... so even I am not sure about this solution...

gfontorbe: That would remove the need for negation in the if statement

Lotes: Yes, maybe a nested IF or a predicate function for the other two condition...
```js
function isNonAlphabeticToken() {
  return keyword.value.match(/^[\W]/) && keyword.value.length === 1;
}
```

spoenemann (commented):
Proposal: Omit the new configuration property and hard-code the change to `completionForKeyword` as the new default behavior. If someone wants to change this, they can simply override the method in a subclass of the CompletionProvider service.

Advantages:
 - Keeps the configuration format simpler
 - More flexible: users can change the regular expression to something else if they want to fine-tune which keywords are shown

gfontorbe:
I went with @spoenemann approach and hardcoded the new behavior in the completion provider

gfontorbe:
I changed the regex to check if the keyword does not contain any word character, this way we also filter out keywords like `>=` `&&` ...

I also modified 3 tests that were expecting keywords with no word characters inside.

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -239,6 +239,10 @@ export class DefaultCompletionProvider implements CompletionProvider {
     }
 
     protected completionForKeyword(keyword: ast.Keyword, context: AstNode | undefined, acceptor: CompletionAcceptor): MaybePromise<void> {
+        // Filter out keywords that do not contain any word character
+        if (!keyword.value.match(/[\w]+/)) {
+            return Promise.resolve();
```
spoenemann, line 244: I think you can just return nothing here because `MaybePromise` does not force you to create a Promise.

spoenemann (approved):
Thanks!

packages/langium/src/lsp/completion/completion-provider.ts
gfontorbe: Changed

PR #698, opened 2022-09-23
Code Action: Adds Missing 'return' to Parser Rule

montymxb:
Closes one of the points of #442.

Adds return T to rule for general error The type 'T' is already explicitly declared and cannot be inferred. From https://github.com/langium/langium/issues/449.

Sep-23-2022 15-09-18

### DIFF
--- a/packages/langium/src/grammar/langium-grammar-validator.ts
+++ b/packages/langium/src/grammar/langium-grammar-validator.ts
@@ -97,6 +97,7 @@ export namespace IssueCodes {
     export const InvalidReturns = 'invalid-returns';
     export const InvalidInfers = 'invalid-infers';
     export const MissingInfer = 'missing-infer';
+    export const MissingReturns = 'missing-returns';
     export const SuperfluousInfer = 'superfluous-infer';
     export const OptionalUnorderedGroup = 'optional-unordered-group';
 }
@@ -269,13 +270,31 @@ export class LangiumGrammarValidator {
             const isInfers = !rule.returnType && !rule.dataType;
             const ruleTypeName = getTypeName(rule);
             if (!isDataType && ruleTypeName && types.has(ruleTypeName) === isInfers) {
-                const keywordNode = isInfers ? findNodeForKeyword(rule.$cstNode, 'infer') : findNodeForKeyword(rule.$cstNode, 'returns');
-                accept('error', getMessage(ruleTypeName, isInfers), {
-                    node: rule.inferredType ?? rule,
-                    property: 'name',
-                    code: isInfers ? IssueCodes.InvalidInfers : IssueCodes.InvalidReturns,
-                    data: keywordNode && toDocumentSegment(keywordNode)
-                });
+                if(rule.inferredType === undefined) {
+                    // diagnostic should go on the rule name, since there is no explicit 'infers'
+                    accept('error', getMessage(ruleTypeName, isInfers), {
+                        node: rule,
+                        property: 'name',
+                        code: isInfers ? IssueCodes.MissingReturns : IssueCodes.InvalidReturns,
+                        data: {
+                            suggestion: `${ruleTypeName} returns ${ruleTypeName}`,
+                            ruleTypeName: ruleTypeName
+                        }
+                    });
+
+                } else {
+                    // diagnostic should go on the explict infers, which will be rewritten as returns
+                    accept('error', getMessage(ruleTypeName, isInfers), {
+                        node: rule,
+                        property: 'inferredType',
+                        code: isInfers ? IssueCodes.InvalidInfers : IssueCodes.InvalidReturns,
+                        data: {
+                            suggestion: `returns ${ruleTypeName}`,
+                            ruleTypeName: ruleTypeName
+                        }
+                    });
+
+                }
             } else if (isDataType && isInfers) {
                 const inferNode = findNodeForKeyword(rule.$cstNode, 'infer');
                 accept('error', 'Data type rules cannot infer a type.', {
--- a/packages/langium/src/grammar/lsp/grammar-code-actions.ts
+++ b/packages/langium/src/grammar/lsp/grammar-code-actions.ts
@@ -48,6 +48,8 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
                 return this.fixMissingImport(diagnostic, document);
             case IssueCodes.UnnecessaryFileExtension:
                 return this.fixUnnecessaryFileExtension(diagnostic, document);
+            case IssueCodes.MissingReturns:
+                return this.fixMissingReturns(diagnostic, document);
             case IssueCodes.InvalidInfers:
             case IssueCodes.InvalidReturns:
                 return this.fixInvalidReturnsInfers(diagnostic, document);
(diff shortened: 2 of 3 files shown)


packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -269,13 +270,31 @@ export class LangiumGrammarValidator {
...
             const ruleTypeName = getTypeName(rule);
             if (!isDataType && ruleTypeName && types.has(ruleTypeName) === isInfers) {
-                const keywordNode = isInfers ? findNodeForKeyword(rule.$cstNode, 'infer') : findNodeForKeyword(rule.$cstNode, 'returns');
-                accept('error', getMessage(ruleTypeName, isInfers), {
-                    node: rule.inferredType ?? rule,
-                    property: 'name',
-                    code: isInfers ? IssueCodes.InvalidInfers : IssueCodes.InvalidReturns,
-                    data: keywordNode && toDocumentSegment(keywordNode)
-                });
+                if(rule.inferredType === undefined) {
```
msujew, line 273: Suggestion: I believe there is quite a bit of overlap in these branches. This can be trimmed down by wrapping the `suggestion` property in a ternary.

packages/langium/src/grammar/lsp/grammar-code-actions.ts
```diff
@@ -7,6 +7,7 @@
 import { CodeActionKind, Diagnostic } from 'vscode-languageserver';
 import { CodeActionParams } from 'vscode-languageserver-protocol';
 import { CodeAction, Command, Position, TextEdit } from 'vscode-languageserver-types';
+import { AstNode } from '../..';
```
msujew, line 10: Minor: Please don't import from the index.

packages/langium/src/grammar/lsp/grammar-code-actions.ts
```diff
@@ -66,6 +69,52 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
...
+    /**
+    * Returns whether an AST node of a given type exists within the range of a diagnostic
+    * Helpful to prevent code actions being suggested in the incorrect context
+    *
+    * @param diagnostic To use to determine the range to search for a node
+    * @param document To search within
+    * @param astTypePredicate To select the desired node
+    * @returns Whether or not the node exists
+    */
+    private astNodeExistsAtDiagnostic<T extends AstNode>(diagnostic: Diagnostic, document: LangiumDocument, astTypePredicate: (n: AstNode) => n is T): boolean {
```
msujew, line 81: Question: Does this method accomplish something except for catching errors from the validator? I believe the method only returns `false` if the validator returns incorrect diagnostic ranges, right?

packages/langium/src/grammar/langium-grammar-validator.ts
montymxb: There is, I still want to go back over and rework this. Most of it is a slice from the pre-existing PR, and much has changed since then.

packages/langium/src/grammar/lsp/grammar-code-actions.ts
montymxb: Yep, it was intended as a sanity check during development, where I had accidentally caught other nodes in the diagnostic. That was some months ago now, and looking back at it now...it can be dropped 🗑️

montymxb:
There are a few cases that get mixed up in here which need a little more work to be separated cleanly. Missing returns & invalid returns are coupled together a bit, and invalid returns is in-turn coupled along with code that handles invalid infers. Plan is to come back to this at the start of Oct at the top of my list 📚 , and to clean this up a bit.

packages/langium/src/grammar/langium-grammar-validator.ts
montymxb: Updated to consolidate

packages/langium/src/grammar/lsp/grammar-code-actions.ts
montymxb: Removed this

packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -276,14 +277,23 @@ export class LangiumGrammarValidator {
...
                 accept('error', getMessage(ruleTypeName, isInfers), {
-                    node: rule.inferredType ?? rule,
-                    property: 'name',
-                    code: isInfers ? IssueCodes.InvalidInfers : IssueCodes.InvalidReturns,
-                    data: keywordNode && toDocumentSegment(keywordNode)
+                    node: rule,
+                    property: implicitInfers ? 'name' : 'inferredType',
+                    code: isInfers ? (implicitInfers ? IssueCodes.MissingReturns : IssueCodes.InvalidInfers) : IssueCodes.InvalidReturns,
+                    data: {
+                        suggestion: implicitInfers ? `${ruleTypeName} returns ${ruleTypeName}` : `returns ${ruleTypeName}`,
```
msujew, line 292: This breaks for some reason, when the type is explicitly inferred:

image

<details>
<summary>Repro</summary>

```
grammar Arithmetics

entry Module:
    'module' name=ID
    (statements+=Statement)*;

interface Evaluation {
    expression: Expression
}

Statement:
    Definition | Evaluation;

Definition:
    'def' name=ID ('(' args+=DeclaredParameter (',' args+=DeclaredParameter)* ')')?
    ':' expr=Expression ';';

DeclaredParameter:
    name=ID;

type AbstractDefinition = Definition | DeclaredParameter;

Evaluation infers Evaluation:
    expression=Expression ';';

Expression:
    Addition;

Addition infers Expression:
    Multiplication ({infer BinaryExpression.left=current} operator=('+' | '-') right=Multiplication)*;

Multiplication infers Expression:
    PrimaryExpression ({infer BinaryExpression.left=current} operator=('*' | '/') right=PrimaryExpression)*;

PrimaryExpression infers Expression:
    '(' Expression ')' |
    {infer NumberLiteral} value=NUMBER |
    {infer FunctionCall} func=[AbstractDefinition] ('(' args+=Expression (',' args+=Expression)* ')')?;

hidden terminal WS: /\s+/;
terminal NUMBER returns number: /[0-9]+(\.[0-9]*)?/;

```

</details>

montymxb: Hmm, I'll take another look, and add a test case to cover this.

montymxb:
Alright, reworked this and it's now down to:
- code action to add missing 'returns'
- code action to correct invalid 'infers' (added due to some proximity)
- same tests as before

Shaves down some of the changes for `InvalidInfers` by passing along the range of the appropriate CST node.

packages/langium/src/grammar/langium-grammar-validator.ts
montymxb: The `data` prop was not set correctly for the rule. It seems this PR feel far enough behind that the mechanics changed internally a bit.  I've rebased and rewritten it a bit to clean it up to help with this.

dhuebner:
@montymxb 
Looks really nice! Is the same for Actions already planned.

montymxb:
@dhuebner thanks! The original plan was to take care of the cases for parser rules, but it would be ideal to set that up for actions too. I believe once we have the code actions we want for parser rules, the same approach can be adapted to actions as well.

msujew (approved):
Cool, looks good to me 👍

PR #701, opened 2022-09-27
Fixed terminal rule for NUMBER in arithmetics example

gfontorbe:
Fixed an error in the Regex for the the terminal rule NUMBER in the arithmetics example.

The previous regex would only match numbers with a maximum of 1 decimal:
Screenshot 2022-09-27 at 15 32 54

 The new regex matches numbers with any amount of decimals. Also matches no decimals even if the dot is present (ex: `1.`)


Lotes (approved):
Simple little fix :D

PR #702, opened 2022-09-27
getServices problem with more than one language; quickfix, TBD

goto40:
I had a problem with langium 0.4.0 and can reproduce it with the current HEAD version:

* I have multiple languages registered (like in https://github.com/langium/langium/pull/676) and I want to load built-in models (like in https://github.com/langium/langium/discussions/391).
* For the configuration with only one language everything worked fine for the build-in models (loaded via `factory.fromString` as described in https://github.com/langium/langium/discussions/391).
* For two languages I get strange errors: `Error: The service registry contains no services for the extension ''.` (empty extension).

I could track this down to `Utils.extname(uri);` not working correctly for my URI `memory://built_in_0.item`. This patch seems to help...

I am not 100% sure why it did not work - do you have an idea?

### DIFF
--- a/packages/langium/src/service-registry.ts
+++ b/packages/langium/src/service-registry.ts
@@ -4,8 +4,9 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { URI, Utils } from 'vscode-uri';
+import { URI } from 'vscode-uri';
 import { LangiumServices } from './services';
+import * as path from 'path';
 
 /**
  * The service registry provides access to the language-specific services. These are resolved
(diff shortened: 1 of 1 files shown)


msujew:
Hey @goto40,

two issues with this:
1. We don't import Node.js dependencies into common Langium code, as that would complicate/prevent browser bundling.
2. The `memory://built_in_0.item` URI is actually malformed: The part after the `//` of a URI is always the `authority`, not the `path`. In order to parse the `built_in_0.item` part as the `path`, you need to add another `/` to make it `memory:///built_in_0.item`

goto40:
>     2. The `memory://built_in_0.item` URI is actually malformed: The part after the `//` of a URI is always the `authority`, not the `path`. In order to parse the `built_in_0.item` part as the `path`, you need to add another `/` to make it `memory:///built_in_0.item`. (Note that I recommended to use just `//` in the past, but I was made aware of the `authority` issue in another project recently).

Oh! my fault. I have overseen it! Thank you, @msujew - I was debugging quite a time to locate my error...
Maybe we could add an URI validator to throw an exception on malformed URIs?

Anyway: thank you for the prompt and competent response!

PR #703, opened 2022-09-28
Refactored grammar AST export, renamed naming.ts

spoenemann:
This PR further cleans up our public API.

 * Instead of exporting all grammar AST elements directly to the package index, I put them into the `GrammarAST` namespace.
 * Removed a few files from the package index because they are not required by most downstream projects.
 * Renamed `naming.ts` to `name-provider.ts` for consistency.


packages/langium-cli/src/generate.ts
```diff
@@ -6,8 +6,7 @@
 
 import fs from 'fs-extra';
 import {
-    AbstractRule, createLangiumGrammarServices, getDocument, Grammar, Interface, isAbstractRule, isGrammar, isInterface, isParserRule, isType,
-    LangiumDocument, linkContentToContainer, Type
+    createLangiumGrammarServices, getDocument, Grammar, GrammarAST, LangiumDocument, linkContentToContainer
 } from 'langium';
```
Lotes, line 10: I like the prefixing idea 'GrammarAST.<...>'. But the naming is not very 1:1. I mean `LangiumAST` would be more direct, whereas `GrammarAST` does not tell me what grammar is meant, or?

Grammar -> LangiumGrammar
GrammarAST -> LangiumAST
LangiumDocument -> perfect

spoenemann: Right, but then I'd name it `LangiumGrammarAST`. LangiumAST is still ambiguous.

pluralia: Isn't `LangiumGrammarAST` too long?

montymxb (commented):
When trying to build locally this causes the examples to no longer generate. It seems that the imports need to updated in the CLI as well. The specific error I'm seeing is `Cannot find module 'langium/src/grammar/type-system'` in **langium-cli/src/generator/ast-generator.ts**. Taking a look there it seems the import just needs to be updated to reflect these changes.

packages/langium-cli/src/generate.ts
spoenemann: @pluralia you are right, too. So then I'd leave the namespace as `GrammarAST` for now. We should rethink our exported API on a more general level. Ideas so far:

 - Create focused index files in every subfolder, narrow the main index (https://github.com/langium/langium/discussions/155)
 - Create a hierarchy of namespaces like the [vscode API](https://code.visualstudio.com/api/references/vscode-api)
 - Keep the current global index and ensure proper naming of exported things

montymxb (commented):
Build issues corrected with cli ✅ . No issues here with merging, will defer to remaining reviewers.

Lotes (approved):

pluralia (approved):

packages/langium-cli/src/generate.ts
pluralia: Should we put these ideas in a separate issue?

spoenemann: I don't know whether it's really an issue. I'd make a discussion and turn it into an issue if we find problems with the current approach.

PR #705, opened 2022-09-28
Fix types validation of a parser rule with alternatives

### LINKED ISSUE #704: Wrong types validation of a parser rule with alternatives
Langium version: 0.5.0-next
Package name: langium

## Steps To Reproduce

```
type RuleType = A | B;
Rule returns RuleType:
    A | B;

A: 'A' name=ID;
B: 'B' name=ID;
```

## The current behavior
`Rule` has a validation error: `A type 'A | B' is not expected in a rule that returns type 'RuleType'.langium`.

## The expected behavior
No errors

pluralia:
Closes #704.

### DIFF
--- a/packages/langium/src/grammar/type-system/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/inferred-types.ts
@@ -554,16 +554,16 @@ function extractTypes(interfaces: InterfaceType[], unions: UnionType[]): AstType
     for (const interfaceType of interfaces) {
         // the criterion for converting an interface into a type
         if (interfaceType.properties.length === 0 && interfaceType.subTypes.size > 0) {
-            const alternative: PropertyType = {
-                types: [...interfaceType.subTypes].sort(),
+            const alternatives: PropertyType[] = [...interfaceType.subTypes].map(type => { return {
+                types: [type],
                 reference: false,
                 array: false
-            };
+            };});
             const existingUnion = unions.find(e => e.name === interfaceType.name);
             if (existingUnion) {
-                existingUnion.union.push(alternative);
+                existingUnion.union.push(...alternatives);
             } else {
-                const type = new UnionType(interfaceType.name, [alternative], { reflection: true });
+                const type = new UnionType(interfaceType.name, alternatives, { reflection: true });
                 type.superTypes = interfaceType.superTypes;
                 astTypes.unions.push(type);
                 typeNames.add(interfaceType.name);


packages/langium/src/grammar/type-system/inferred-types.ts
```diff
@@ -312,18 +307,11 @@ function addAssignment(current: TypePart, assignment: Assignment): void {
     const typeItems: TypeCollection = { types: new Set(), reference: false };
     findTypes(assignment.terminal, typeItems);
 
-    let typeAlternatives: PropertyType[] = [];
-    if (assignment.operator === '=' && !typeItems.reference) {
```
dhuebner, line 316: What happened with this case here? Is it obsolete?

pluralia: This case has to be covered by new code.

montymxb: I'm also a bit curious why this was dropped. You added this originally as part of #455 , but I'm not exactly sure what this part was added for. Given this an assignment, and this is the `addAssignment` function, I would think we would want to retain this. Is there a reason why not?

montymxb (commented):
Fixes the issue at hand, but I still have a question & a small nit. Nit can be ignored, but I am curious about the question.

packages/langium/src/grammar/type-system/inferred-types.ts
```diff
@@ -483,9 +471,10 @@ function comparePropertyType(a: PropertyType, b: PropertyType): boolean {
 }
 
 function compareLists<T>(a: T[], b: T[], eq: (x: T, y: T) => boolean = (x, y) => x === y): boolean {
-    if (a.length !== b.length) return false;
     const distictAndSortedA = distictAndSorted(a);
-    return distictAndSorted(b).every((e, i) => eq(e, distictAndSortedA[i]));
+    const distictAndSortedB = distictAndSorted(b);
```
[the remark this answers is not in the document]
pluralia: Unfortunately, we can't remove it. `eq` takes 2 elements and says they are equal or not, and we have 2 ways to use it:
1. Distinct and sort both lists + use `eq`: O(n * logn) + O(n) = O(n * logn)
2. Use `eq` directly, what requires comparison of each element of one list with each element of another list: O(n^2)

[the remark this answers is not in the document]
pluralia: However, for small sizes of input the second approach wins: (1) `2 * n * log n + n` vs (2) `n^2`, and as `compareLists` input lists are short most of the time (currently the function `compareLists` is used only in `comparePropertyType`), we really can remove usage of `distictAndSorted`. @montymxb , WDYT?

packages/langium/src/grammar/type-system/inferred-types.ts
pluralia: Previous logic was:
* if we have an array or/and a reference of type alternatives, we pack them in one list
* if we have only type alternatives (not under an array or a reference), we consider each type separately and get many 1-element-lists

Now I've added the `toPropertyType` function, that implements this logic, so we don't need to keep these code anymore.

packages/langium/src/grammar/type-system/inferred-types.ts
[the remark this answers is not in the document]
montymxb: The nit was the spelling 😉 , the function usage is perfectly fine here, no problems with that.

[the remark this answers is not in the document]
pluralia: Oh, I thank you :D However, question about usage of `distinctAndSorted` is relevant. Should I remove it?

packages/langium/src/grammar/type-system/inferred-types.ts
montymxb: Ahh, got it, I see you've lifted it up into the common function. Nice 😎

montymxb (approved):
Looks good :octocat:

PR #717, opened 2022-10-07
Added TS option to enforce explicit override

spoenemann:
With TypeScript 4.3, an option to enforce the `override` keyword was introduced. Reasons for enabling this:

 - When you click on the `override` keyword, the editor jumps to the superclass declaration of that member.
 - When refactoring a method that is overridden somewhere, breaking changes can slip through without noticing if the method is implicitly overridden. Making that explicit reveals the breakage at compile time.

### DIFF
--- a/examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
+++ b/examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
@@ -4,18 +4,18 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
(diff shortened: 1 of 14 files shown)


montymxb (approved):
:octocat: looks great! Nice addition to have.

Lotes:
Today I have learned something new :D... Looks good. I cannot believe that there are so few places with overrides.

spoenemann:
> I cannot believe that there are so few places with overrides.

We want to avoid class hierarchy within the framework. Always choose composition rather than hierarchy. Overriding is only necessary to customize services in a specific language.

PR #721, opened 2022-10-14
Support customizable lexer infrastructure

msujew:
Closes https://github.com/langium/langium/issues/707

This allows users to build their own lexers for more complicated languages such as Python and JavaScript by building a custom lexer. 

Also fixes an issue where we had to append `:KW` to keyword tokens. This is no longer necessary.


spoenemann (approved):
Great!

PR #727, opened 2022-10-19
Only use name-nodes as definitions

msujew:
Closes https://github.com/langium/langium/issues/726

Improves the existing code on two parts:
1. Only provide the original element in case the targeted node is part of the name node
2. Only return in `findAssignment` if the assignment belongs to the current AST node

Also adds some new test cases for this.


spoenemann (approved):
:+1:

PR #731, opened 2022-10-24
Set current user directory dynamically

rhumbertgz:
USER_DIR= '.' doesn't allow to dynamically change the output directory for the template to generate. I ran into this problem by trying to integrate the generator through a third-party app using the 'yeoman-environment'. Since the USER_DIR is hard-coded there is no option to change the output directory. However, by using  `process.chdir('NEW-OUTPUT-DIRECTORY')` in a third-party app and `const USER_DIR = cwd();` we can get that behaviour.

### DIFF
--- a/packages/generator-langium/src/index.ts
+++ b/packages/generator-langium/src/index.ts
@@ -8,9 +8,10 @@ import Generator from 'yeoman-generator';
 import _ from 'lodash';
 import chalk from 'chalk';
 import path from 'path';
+import { cwd } from 'node:process';
 
 const TEMPLATE_DIR = '../langium-template';
-const USER_DIR = '.';
+const USER_DIR = cwd();
 
 const OPEN = '<%= ';
 const CLOSE = ' %>';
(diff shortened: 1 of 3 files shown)


montymxb:
@rhumbertgz thanks for putting this up. I'll get around to taking a look at this later today.

packages/generator-langium/src/index.ts
```diff
@@ -8,9 +8,10 @@ import Generator from 'yeoman-generator';
 import _ from 'lodash';
 import chalk from 'chalk';
 import path from 'path';
+import { cwd } from 'node:process';
```
montymxb, line 11: This import doesn't work from what I can tell. Seems the package.json is missing a dev dep of `@types/node` to get this to work.

montymxb (commented):
Check out the comment about the package.json.

rhumbertgz:
> Check out the comment about the package.json.

Fixed!

montymxb:
@rhumbertgz I took another peek at this and missed a quick simplification for this. Process is implicitly available already, and `cwd` can be accessed by just invoking it under the process object, rather than on its own.
```ts
const USER_DIR = process.cwd();
```
There shouldn't be any need to add deps this way. I believe this will work as a standalone change, but let me know if that works on your end too.

rhumbertgz:
Perfect! I will close this PR then

Issue #596, opened 2022-07-13
Support `textDocument.declaration` LSP request


msujew:
See [the spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_declaration) for more info on the request type.

msujew:
Closed by https://github.com/langium/langium/pull/734

PR #734, opened 2022-10-25
Go to declaration service

gfontorbe:
Solves #596
Added a default implementation of the go to declaration service.

For languages like Langium, for what I understand, go to declaration and go to definition achieve the same thing.

Should we instead provide an abstract implementation?


packages/langium/src/services.ts
```diff
@@ -53,6 +53,7 @@ import type { ImplementationProvider } from './lsp/implementation-provider';
 import type { CallHierarchyProvider } from './lsp/call-hierarchy-provider';
 import type { DocumentLinkProvider } from './lsp/document-link-provider';
 import type { CodeLensProvider } from './lsp/code-lens-provider';
+import { DeclarationProvider } from './lsp/declaration-provider';
```
msujew, line 56: Minor:
```suggestion
import type { DeclarationProvider } from './lsp/declaration-provider';
```

msujew (commented):
I don't believe we need a default service for the declaration provider. An interface declaration for the service is already enough. As the service is highly semantic specific, we can't really do anything here. So it's probably best to just remove the implementation and tests.

msujew (approved):
Looks good to me, thanks 👍

PR #738, opened 2022-10-28
Improve validation check type safety

msujew:
As of now, validations can be written as `Type: (node: NotOfType, acceptor)` which might break validations during runtime. This change improves type safety in the validation registry and also introduces a better `AstTypes` generator.


pluralia (approved):
LGTM!

PR #739, opened 2022-10-28
Improve completion API and functionality

msujew:
This change lets the completion provider make use of fuzzy matching similar to the TypeScript language server. Also improves handling of `textEdit` and `insertText` in case downstream users override this value for their completions.

### DIFF
--- a/packages/langium/langium-config.json
+++ b/packages/langium/langium-config.json
@@ -5,7 +5,10 @@
         "grammar": "src/grammar/langium-grammar.langium",
         "fileExtensions": [
           "langium"
-        ]
+        ],
+        "chevrotainParserConfig": {
+            "maxLookahead": 3
+        }
     }],
     "out": "src/grammar/generated",
     "langiumInternal": true
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -276,7 +276,7 @@ export class DefaultCompletionProvider implements CompletionProvider {
         } else {
             insertText = label;
         }
-        const textEdit = this.buildCompletionTextEdit(document, offset, insertText);
+        const textEdit = this.buildCompletionTextEdit(document, offset, label, insertText);
         if (!textEdit) {
             return undefined;
         }
@@ -284,15 +284,15 @@ export class DefaultCompletionProvider implements CompletionProvider {
         return completionItem;
     }
 
-    protected buildCompletionTextEdit(document: TextDocument, offset: number, completion: string): TextEdit | undefined {
+    protected buildCompletionTextEdit(document: TextDocument, offset: number, label: string, newText: string): TextEdit | undefined {
         const content = document.getText();
         const tokenStart = this.backtrackToTokenStart(content, offset);
         const identifier = content.substring(tokenStart, offset);
-        if (this.charactersFuzzyMatch(identifier, completion.toLowerCase())) {
+        if (this.charactersFuzzyMatch(identifier, label.toLowerCase())) {
             const start = document.positionAt(tokenStart);
             const end = document.positionAt(offset);
             return {
-                newText: completion,
+                newText,
                 range: {
                     start,
                     end
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -22,7 +22,11 @@ import { stream } from '../../utils/stream';
 import { LangiumDocument } from '../../workspace/documents';
 import { findFirstFeatures, findNextFeatures, NextFeature } from './follow-element-computation';
 
-export type CompletionAcceptor = (value: string | AstNode | AstNodeDescription, item?: Partial<CompletionItem>) => void
+export type CompletionAcceptor = (value: CompletionValueItem) => void
+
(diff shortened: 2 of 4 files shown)


packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -240,52 +242,57 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
+            label = item.label;
+        } else if (isAstNode(item.label)) {
+            const name = this.nameProvider.getName(item.label);
             if (!name) {
                 return undefined;
             }
             label = name;
-        } else if (!isAstNode(value)) {
-            label = value.name;
+        } else if (item.label && !isAstNode(item.label)) {
```
spoenemann, line 266: Check with `isAstNodeDescription` in this case.

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -240,52 +242,57 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
-            if (completionLowerCase.startsWith(contentLowerCaseSub) && (i === 0 || !this.isWordCharacterAt(contentLowerCase, offset - i - 1))) {
-                negativeOffset = i;
-                break;
-            }
-        }
-        if (negativeOffset > 0 || offset === 0 || !this.isWordCharacterAt(completion, 0) || !this.isWordCharacterAt(content, offset - 1)) {
-            const start = document.positionAt(offset - negativeOffset);
+        const tokenStart = this.backtrackToTokenStart(content, offset);
+        const identifier = content.substring(tokenStart, offset);
+        if (this.charactersFuzzyMatch(identifier, label.toLowerCase())) {
```
spoenemann, line 291: Proposal: move the `toLowerCase()` call into the `charactersFuzzyMatch` method to support when someone would like to change the behavior to case-sensitive exact match.

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -171,28 +198,29 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
 
-    protected completionForCrossReference(crossRef: NextFeature<ast.CrossReference>, context: AstNode | undefined, acceptor: CompletionAcceptor): MaybePromise<void> {
+    protected completionForCrossReference(context: CompletionContext, crossRef: NextFeature<ast.CrossReference>, acceptor: CompletionAcceptor): MaybePromise<void> {
         const assignment = getContainerOfType(crossRef.feature, ast.isAssignment);
-        if (assignment) {
-            if (crossRef.type && (crossRef.new || context?.$type !== crossRef.type)) {
-                context = {
+        let node = context.node;
+        if (assignment && node) {
+            if (crossRef.type && (crossRef.new || node?.$type !== crossRef.type)) {
```
spoenemann, line 220: The `?` is not necessary due to the condition in the preceding line.

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -97,28 +121,31 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
         }
 
         return CompletionList.create(items, true);
     }
 
+    /**
+     * Determines whether the completion parser will reparse the input at the point of completion.
+     * By default, this returns `false`, indicating that the completion will only look for completion results starting from the token at the cursor position.
+     * Override this and return `true` to indicate that the completion should parse the input a second time.
+     * This might add some missing completions at the cost at parsing the input twice.
```
spoenemann, line 143: Please add a test so we know that we don't accidentally break this functionality (it's disabled by default), or create an issue to add such a test.

spoenemann (approved):
I noticed just one problem while testing: completion does not work with qualified names in the domain model example. It works for a reference if the prefix is `baz`, but not for `baz.` or `baz.nested` or anything else after the first `.`

We don't have to fix that in this PR though, so feel free to create a new issue instead.

packages/langium/src/lsp/completion/completion-provider.ts
msujew: Done see https://github.com/langium/langium/issues/831

packages/langium/langium-config.json
```diff
@@ -5,7 +5,10 @@
         "grammar": "src/grammar/langium-grammar.langium",
         "fileExtensions": [
           "langium"
-        ]
+        ],
+        "chevrotainParserConfig": {
+            "maxLookahead": 3
```
spoenemann, line 10: So you want to use LL(k) for the grammar language?

msujew: It fixes a completion issue, due to very slight differences in lookahead behavior for malformed input.

spoenemann: General question, could we allow a different parser config for completion?

msujew: Not really, since the parser is build the same way, the implementation of the embedded actions are just different.

spoenemann: Ok last question, have you checked whether _k=2_ would work for the grammar language?

msujew: Haven't checked, but it might probably work.

spoenemann: Or maybe even _k=1_? 😁

PR #748, opened 2022-11-01
Optimize subtype lookup

msujew:
This change introduces a major optimization for subtype-supertype lookups of the `isSubtype` method of the `AstReflection` service. It keeps the old recursive algorithm to determine supertype relationships, but caches the resulting values.

In my benchmarks, this optimization has little effect on truthy results within a small hierarchy (i.e. is `Keyword` a subtype of `AbstractElement`?), but increases performance for falsy cases (i.e. is `Keyword` a subtype of `Group`?) by 30% in small hierarchies, with more performance improvements expected with larger type hierarchies. This is due to the fact that the algorithm has to traverse all supertypes for falsy cases.


examples/arithmetics/src/language-server/generated/ast.ts
```diff
@@ -126,20 +125,13 @@ export interface ArithmeticsAstType {
...
 
-    isInstance(node: unknown, type: string): boolean {
-        return isAstNode(node) && this.isSubtype(node.$type, type);
-    }
-
-    isSubtype(subtype: string, supertype: string): boolean {
-        if (subtype === supertype) {
-            return true;
-        }
+    isSubtypeOverride(subtype: string, supertype: string): boolean {
```
spoenemann, line 134: This generated method should be `protected`. About the name: could `computeIsSubtype` or similar be more understandable?

spoenemann (commented):
Thanks, the new approach makes sense.

packages/langium/src/syntax-tree.ts
```diff
@@ -108,6 +130,38 @@ export interface AstReflection {
     isSubtype(subtype: string, supertype: string): boolean
 }
 
+export abstract class AbstractAstReflection implements AstReflection {
```
spoenemann, line 133: Please add a comment about the purpose of this abstract class.

spoenemann (approved):

PR #749, opened 2022-11-01
Prevent usage of reserved words in grammars

msujew:
Closes https://github.com/langium/langium/issues/529

Validates all AST instances which are generated as part of types (i.e. parser rules, inferred types, interfaces, etc). Anything that is either a keyword of TS/JS or a globally available object/function/constant will result in a validation error.


packages/langium/src/grammar/langium-grammar-validator.ts
```diff
@@ -29,21 +29,27 @@ export class LangiumGrammarValidationRegistry extends ValidationRegistry {
...
-                validator.checkRuleParametersUsed
+                validator.checkRuleParametersUsed,
+                validator.checkParserRuleReservedName,
             ],
             TerminalRule: [
                 validator.checkTerminalRuleReturnType,
                 validator.checkHiddenTerminalRule,
                 validator.checkEmptyTerminalRule
             ],
+            InferredType: validator.checkReservedTypeName,
```
pluralia, line 52: Why not `checkTypeReservedName`? It would be more consistent with other names.

pluralia (changes requested):
Property `arguments` gives a validation error:
Screenshot 2022-11-01 at 23 23 34

Issue #385, opened 2022-01-20
The strict mode for types inference


pluralia:
In Langium 0.2.0 types are inferred from rules names implicitly. For example, the following 2 examples will infer a type `MyRule` with the attribute `name`.
```
MyRule:
    name=ID;
```
```
MyRule returns MyRule:
    name=ID;
```
When we [add types](https://github.com/langium/langium/discussions/91#discussioncomment-1264914), type inference like described above will still be allowed by default.

This feature issue proposes to add the strict mode, when implicit type inference is forbidden. So, the way to specify a type of `MyRule` in the strict mode is:
```
type MyRule = {
    name: string
}

MyRule returns MyRule:
    name=ID;
```

Pluralia:
We can generate "a type system template" automatically in the strict mode. So, we just infer types like in the "default" mode and put them in separate `.langium` file. As the result:
1. The strict mode works out of box
2. A user develops the type system by adding and modifying types instead of creating everything from zero

@danieldietrich proposed to add that functionality to the CLI.

spoenemann:
This will be considerably easier to implement with #438 and #440.

PR #754, opened 2022-11-04
langium-cli to generate types definition

dhuebner:
This PR adds:
- a new option `-t, --types  <file>` to langium-cli that only triggers the type definition generation
- a command to the Langium extension that triggers langium-cli to generate type definitions
- a context menu entry to the Langium editor the triggers the command 

generate-langium-types

### DIFF
--- a/packages/langium-cli/src/langium.ts
+++ b/packages/langium-cli/src/langium.ts
@@ -8,7 +8,7 @@ import chalk from 'chalk';
 import fs from 'fs-extra';
 import { Command } from 'commander';
 import { validate } from 'jsonschema';
-import { generate, GenerateOptions, GeneratorResult } from './generate';
+import { ExtractTypesOptions, generate, GenerateOptions, generateTypes, GeneratorResult } from './generate';
 import { cliVersion, elapsedTime, getTime, log, schema } from './generator/util';
 import { LangiumConfig, loadConfigs, RelativePath } from './package';
 import path from 'path';
@@ -27,6 +27,18 @@ program
             process.exit(1);
         });
     });
+program.command('extract-types')
+    .argument('<file>', 'the langium grammar file to generate types for')
+    .option('-o, --output <file>', 'output file name. Default is types.langium next to the grammar file.')
+    .option('-f, --force', 'Force overwrite existing file.')
+    .action((file, options: ExtractTypesOptions) => {
+        options.grammar = file;
+        generateTypes(options).catch(err => {
+            console.error(err);
+            process.exit(1);
+        });
+    })
+    .action;
 
 program.parse(process.argv);
 
--- a/packages/langium-vscode/package.json
+++ b/packages/langium-vscode/package.json
@@ -19,12 +19,18 @@
     "vscode": "^1.53.0"
   },
   "contributes": {
-    "languages": [{
-      "id": "langium",
-      "extensions": [ ".langium" ],
-      "aliases": [ "Langium" ],
-      "configuration": "./data/langium.configuration.json"
-    }],
+    "languages": [
+      {
+        "id": "langium",
+        "extensions": [
+          ".langium"
+        ],
+        "aliases": [
+          "Langium"
+        ],
+        "configuration": "./data/langium.configuration.json"
+      }
+    ],
     "grammars": [
       {
         "language": "langium",
@@ -65,7 +71,7 @@
   "dependencies": {
     "langium": "0.5.0",
     "vscode-languageserver": "^8.0.2",
-    "ignore":"~5.2.0"
+    "ignore": "~5.2.0"
   },
   "devDependencies": {
     "@types/vscode": "^1.53.0",
@@ -78,7 +84,7 @@
   },
   "bugs": "https://github.com/langium/langium/issues",
   "author": {
-      "name": "TypeFox",
-      "url": "https://www.typefox.io"
+    "name": "TypeFox",
+    "url": "https://www.typefox.io"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1349,9 +1349,9 @@
       "integrity": "sha512-dOy+3AuW3a2wNbZHIuMZpTcgjGuLU/uBL/ubcZF9OXbDo8ff4O8yVp5Bf0efS8uEoYo5q4Fx7dY9OgQGXgAsQA=="
     },
     "node_modules/commander": {
-      "version": "8.2.0",
-      "resolved": "https://registry.npmjs.org/commander/-/commander-8.2.0.tgz",
-      "integrity": "sha512-LLKxDvHeL91/8MIyTAD5BFMNtoIwztGPMiM/7Bl8rIPmHCZXRxmSWr91h57dpOpnQ6jIUqEWdXE/uBYMfiVZDA==",
+      "version": "8.3.0",
+      "resolved": "https://registry.npmjs.org/commander/-/commander-8.3.0.tgz",
+      "integrity": "sha512-OkTL9umf+He2DZkUq8f8J9of7yL6RJKI24dVITBmNfZBmri9zYZQrKkuXiKhyfPSu8tUhnVBB1iKXevvnlR4Ww==",
       "engines": {
         "node": ">= 12"
       }
@@ -4579,7 +4579,7 @@
       "license": "MIT",
       "dependencies": {
         "chalk": "^4.1.2",
-        "commander": "^7.2.0",
+        "commander": "^8.0.0",
         "fs-extra": "^9.1.0",
         "jsonschema": "^1.4.0",
         "langium": "~0.5.0",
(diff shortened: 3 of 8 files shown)


dhuebner:
Known issue: wrong grammar is used to escape the names with `^`

packages/langium-cli/src/langium.ts
```diff
@@ -20,12 +20,20 @@ program
     .command('generate')
     .description('Generate code for a Langium grammar')
     .option('-f, --file <file>', 'the configuration file or package.json setting up the generator')
+    .option('-t, --types <file>', 'the langium grammar file to generate types for')
```
spoenemann, line 23: Thanks! I'd propose to define a new command instead of an option for the existing command. What about `extract-types`?

dhuebner: I had problems with adding CLI options with `-` in the name. I will try it with `-e, --extract-types` again

spoenemann: I mean defining a command, not an option:
```typescript
    .command('extract-types')
    .action(() => ...)
```

dhuebner: Oh you mean like `langium extract-types -g mygrammar.langium --target types.langium` ?

spoenemann: Yes, something like that. Either for a specific grammar given as option or for all grammars mentioned in langium-config.json

dhuebner: I've added a new command `extract-types`.
Usage: `langium extract-types <file> [--output <file>]`

Generating types for multiple grammars in one flow can be added later

pluralia:
Could you please add "a button", that replaces `returns` keyword automatically and remove declared types and interfaces from the initial grammar?

pluralia (commented):
Could you add also generation type files automatically for each `.langium` file in the workspace?

One more moment: the types generation is not correct, because of [this issue](https://github.com/langium/langium/issues/756). So, it's not a problem of your PR changes, but we should keep it in mind.

packages/langium-vscode/src/commands.ts
```diff
@@ -0,0 +1,92 @@
...
+        return;
+    }
+    let cliVersion;
+    try {
+        cliVersion = await execute('langium -V', true);
+        // TODO add version compatibility check
+        log('Using langium-cli v' + String(cliVersion).trim());
+    } catch (error) {
+        // not installed
+        const installCmd = 'npm install -g langium-cli';
```
msujew, line 26: I wouldn't expect users to install the `langium-cli` globally. The extension should instead use the package that is in the `node_modules`. TypeScript for example works the same way.

msujew:
> Could you add also generation type files automatically for each .langium file in the workspace?

What do you mean with that?

pluralia:
> > Could you add also generation type files automatically for each .langium file in the workspace?
> 
> What do you mean with that?

If I have some `.langium` files in my workspace and want to have a type file for each of them, currently, I have to start generation for each of them. I'd like to have a flag, that do it automatically.

msujew:
@pluralia Since people will execute this command only rarely for their grammars, I'm fine with not generating all fines using a single click. We can add this additional feature in a separate PR, it's a bit out of scope for this PR.

spoenemann:
@dhuebner @pluralia please keep in mind that VS Code is designed to be terminal- and keyboard-centric. It's not like in Eclipse IDE where you need a menu item for everything. Adding UI elements can be perceived as intrusive by users who want to keep their IDE clean.

In fact I believe Eclipse IDE is a very bad example of what happens when all plug-in developers add their own UI elements: the tool becomes bloated with buttons, views and menu entries, and as result it feels heavyweight.

spoenemann (commented):
I tried the following in `examples/statemachine`:

```
 $ ../../packages/langium-cli/bin/langium.js extract-types src/language-server/statemachine.langium 
Error: ENOENT: no such file or directory, open '/src/language-server/statemachine.langium'
    at Object.openSync (node:fs:590:3)
    at Object.readFileSync (node:fs:458:35)
    at NodeFileSystemProvider.readFileSync (/workspace/langium/packages/langium/lib/node/node-file-system-provider.js:22:29)
    at DefaultLangiumDocuments.getContent (/workspace/langium/packages/langium/lib/workspace/documents.js:180:40)
    at DefaultLangiumDocuments.getOrCreateDocument (/workspace/langium/packages/langium/lib/workspace/documents.js:174:70)
    at generateTypes (/workspace/langium/packages/langium-cli/lib/generate.js:206:32)
    at Command.<anonymous> (/workspace/langium/packages/langium-cli/lib/langium.js:38:34)
    at Command.listener [as _actionHandler] (/workspace/langium/node_modules/commander/lib/command.js:488:17)
    at /workspace/langium/node_modules/commander/lib/command.js:1227:65
    at Command._chainOrCall (/workspace/langium/node_modules/commander/lib/command.js:1144:12) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/src/language-server/statemachine.langium'
}
```

I guess you should resolve the input file against `.` (use `path.resolve` from Node.js), and check whether the file exists before opening it.

Second observation: Please check whether the output file exists before overwriting it. In that case I propose asking the user a yes/no question. You can use the `readline` Node.js module for this, see [this example](https://github.com/eclipse/openvsx/blob/464d6a48e8b81cc6e7526f62a8ae6b9b309a119d/cli/src/util.ts#L161-L185).

Third observation: The Langium editor does not support using types from other files yet: cross-references are broken and validations don't work as expected. Do we have open issues for that?

packages/langium-vscode/package.json
```diff
@@ -47,10 +47,28 @@
           "description": "Specifies the exclusion patterns during initial workspace indexing. You will need to reload your extension afterwards."
         }
       }
+    },
+    "commands": [
```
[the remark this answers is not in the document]
dhuebner: True, I will format the whole file, thanks!

packages/langium-vscode/src/commands.ts
dhuebner: > The extension should instead use the package that is in the node_modules

I disagree here. Langium extension as part of the tooling should **not** use the project setup, but a compatible langium-cli version installed globally. Imagine you have a Langium project in v0.5.0 and Langium extension in v0.6.0, this will not work as the released v0.5.0 langium-cli lacks corresponding CLI command implementation. 

However, I would love to see how TypeScript manages that with `npm_modules` folder, could you point me to the corresponding repository? We could offer this as an option before starting generation.

Btw. we suggest users to install example cli generators globally, is that something we can improve/change?

dhuebner:
@spoenemann 
> I guess you should resolve the input file against . (use path.resolve from Node.js), and check whether the file exists before opening it.

Done.

> Second observation: Please check whether the output file exists before overwriting it. In that case I propose asking the user a yes/no question. You can use the readline Node.js module for this, see [this example](https://github.com/eclipse/openvsx/blob/464d6a48e8b81cc6e7526f62a8ae6b9b309a119d/cli/src/util.ts#L161-L185).

Done. Also added `--force` option to force overwrite existing file.

> Third observation: The Langium editor does not support using types from other files yet: cross-references are broken and validations don't work as expected. Do we have open issues for that?

Yes, I think so.

packages/langium-vscode/package.json
```diff
@@ -47,10 +53,28 @@
...
+          "when": "editorLangId == langium",
+          "group": "z_commands"
+        }
+      ]
     }
   },
   "activationEvents": [
-    "onLanguage:langium"
+    "onLanguage:langium",
+    "langium-vscode.generate-types"
```
spoenemann, line 77:
```suggestion
    "onCommand:langium-vscode.generate-types"
```

packages/langium-vscode/src/commands.ts
spoenemann: > Langium extension as part of the tooling should not use the project setup, but a compatible langium-cli version installed globally.

I strongly disagree. Depending on a globally installed `langium-cli` breaks reproducibility of builds. You should always add it to `devDependencies`, and any tooling should use the local version if available.

spoenemann:
I'm skeptical about the UI contribution, both from a conceptual standpoint (https://github.com/langium/langium/pull/754#issuecomment-1305518558) and a technical standpoint (https://github.com/langium/langium/pull/754#discussion_r1014230447). We shouldn't make too strong assumptions about the user's project setup. I doubt that we can make such a UI contribution work as the user expects in all circumstances (think of different package managers: yarn, pnpm, bun, ...)

@dhuebner please have a look at the [Task Provider API](https://code.visualstudio.com/api/extension-guides/task-provider), this could match our use case better than a command.

My proposal: Remove the UI contribution from this PR and create a separate PR for it after evaluating the Task Provider approach.

dhuebner:
> My proposal: Remove the UI contribution from this PR and create a separate PR for it after evaluating the Task Provider approach.

@spoenemann 
I removed all the UI contributions. Will look into Task API to check if it fits the requirements well and create a new PR.
The new cli functionality is very helpful also without UI, so please review once again :)

spoenemann (approved):
Great! The only thing I'm missing is a console output after generating the types file, printing the path of the generated file so I can cmd-click it to open it right away.

Follow-up work: add an option to modify the grammar file so it uses the generated types instead of inferring them. @dhuebner could you create an issue for that?

Further follow-up work: #385

PR #755, opened 2022-11-04
Enable unbounded parser lookahead

msujew:
Closes https://github.com/langium/langium/issues/202

Note that the normal LL(_k_) lookahead can be enabled in the parser config by specifying a `maxLookahead`.


packages/langium/src/parser/langium-parser.ts
```diff
@@ -518,9 +519,13 @@ class ChevrotainWrapper extends EmbeddedActionsParser {
...
 
     constructor(tokens: TokenVocabulary, config?: IParserConfig) {
+        const useDefaultLookahead = config && 'maxLookahead' in config;
         super(tokens, {
             ...defaultConfig,
-            ...config
+            ...config,
+            lookaheadStrategy: useDefaultLookahead
+                ? new LLkLookaheadStrategy({ maxLookahead: config.maxLookahead })
+                : new LLStarLookaheadStrategy()
```
spoenemann, line 528: Can we swap the new property with the `...config` object so we can explicitly override this strategy?

spoenemann (approved):
FINALLY! 🎉 

I'll publish a `next` version after merging this so we can test it with other languages.

Issue #745, opened 2022-10-31
Cross reference in separate files with rule name same as type name raises error


luan-xiaokun:
The error `Use the rule type 'X' instead of the typed rule name 'X' for cross references.` is raised when

1. The name of the referenced type and the name of its parse rule are the same,
2. The parse rule of the referenced type and the cross reference are placed in different files.

Using different name or placing them in one file can remove this error.

A minimal example:
```
// file debug.langium
grammar Debug

import './foo'

entry Model:
    (b+=B | a+=A)*;

// A infers A:
//     'A' a=ID;

B infers B:
    'B' b=[A:ID];
```

```
// file foo.langium
A infers A:
    'A' a=ID;

terminal ID: /[_a-zA-Z][\w_]*/;
```

Renaming the rule `A` or moving it into debug.langium both resolve the error.

msujew:
For anyone interesting in picking this up: This issue can be addressed by overriding the `Linker` for the grammar language and prioritizing interfaces/types over rules when resolving cross references to `AbstractType`.

luan-xiaokun:
@msujew If I understand it correctly, the scope provider of langium grammar seems to have already prioritized interfaces and types over rules. So the issue here seems to be the type/interface we want to reference (`A` in the example) not being included in the scope.

https://github.com/langium/langium/blob/824f098b1c924e039aa2884ff82d6544ab4b8ee5/packages/langium/src/grammar/references/grammar-scope.ts#L30-L46

Indeed, in this minimal example, I found no type/interface named `A` in the provided scope, but there is a rule named `A`. (I used the console to print all elements in the scope, something like `scopeProvider.getScope(refInfo).getAllElements().toArray().map(...)`, please let me know if there is a better way to debug this).

If we omit the `infers A`, just write `A: 'A' a=ID`, then the error disappears, but in this case, there is still no interface named `A` in the scope. This error is not raised because now the `type.ref.returnType || type.ref.inferredType` is false, so it passes the validation check. If we remove this condition, then the same error is raised again. This means that when we omit the `infers A` to suppress the error, the cross reference is not correctly resolved (but the generated extension works well).

https://github.com/langium/langium/blob/824f098b1c924e039aa2884ff82d6544ab4b8ee5/packages/langium/src/grammar/langium-grammar-validator.ts#L698-L706

Renaming the rule does not help now, but I'm sure that when I posted this issue renaming was a workaround (on Ubuntu platform). Maybe it's because I'm using Windows platform now?

[...]

PR #768, opened 2022-11-14
Added test for #745

spoenemann:
I also added / changed a few APIs along the way.

### DIFF
--- a/packages/langium/src/workspace/ast-node-locator.ts
+++ b/packages/langium/src/workspace/ast-node-locator.ts
@@ -30,7 +30,7 @@ export interface AstNodeLocator {
      * @returns The `AstNode` located under the given path, or `undefined` if the path cannot be resolved.
      * @see AstNodeLocator.getAstNodePath
      */
-    getAstNode(document: LangiumDocument, path: string): AstNode | undefined;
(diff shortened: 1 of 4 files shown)


packages/langium/test/grammar/references/grammar-scoping.test.ts
```diff
@@ -0,0 +1,60 @@
...
+            'B' b=[A:ID];
+        terminal ID: /[_a-zA-Z][\\w_]*/;
+        `);
+        await services.shared.workspace.DocumentBuilder.build([grammar]);
+        const reference = locator.getAstNode<CrossReference>(grammar, 'rules@2/definition/elements@1/terminal');
+        const inferredType = locator.getAstNode<InferredType>(grammar, 'rules@1/inferredType');
+        expect(reference?.type.ref).toBe(inferredType);
+    });
+
+    test('Linking in other file', async () => {
```
msujew, line 40: Do you mind adding another test here, which tests that cross references to type declarations (interfaces) are correctly resolved even in the face of a parser rule with the same name, which returns this interface? Something like this:

```
interface X {
  ...
}
X returns X: ...;

// In a separate file
Y: ref=[X:ID];
```

msujew (approved):
LGTM 👍

PR #774, opened 2022-11-22
Programmatic reference construction

spoenemann:
See https://github.com/langium/langium/discussions/773. The `$refNode` property is made optional so programmatic construction of (partial) ASTs with references is easier.

### DIFF
--- a/packages/langium/src/lsp/references-provider.ts
+++ b/packages/langium/src/lsp/references-provider.ts
@@ -60,7 +60,7 @@ export class DefaultReferencesProvider implements ReferencesProvider {
         if (targetAstNode) {
             const options = { includeDeclaration: params.context.includeDeclaration };
             this.references.findReferences(targetAstNode, options).forEach(reference => {
-                if (isReference(reference)) {
+                if (isReference(reference) && reference.$refNode) {
                     refs.push(Location.create(document.uri.toString(), reference.$refNode.range));
                 } else {
                     refs.push(Location.create(reference.sourceUri.toString(), reference.segment.range));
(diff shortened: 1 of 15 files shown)


sailingKieler (commented):
Looks basically good to me.
Just one thing: `isReference(reference) && reference.$refNode` looks a bit magic.
Can we instead have something like `isParsedReference(reference)` to make difference clear?

packages/langium/src/lsp/references-provider.ts
```diff
@@ -60,7 +60,7 @@ export class DefaultReferencesProvider implements ReferencesProvider {
         if (targetAstNode) {
             const options = { includeDeclaration: params.context.includeDeclaration };
             this.references.findReferences(targetAstNode, options).forEach(reference => {
-                if (isReference(reference)) {
+                if (isReference(reference) && reference.$refNode) {
```
sailingKieler, line 63: Question: In which cases will a `ReferenceDescription` a proper `Reference`. Is this legacy?

spoenemann: Probably yes, I'll remove it.

spoenemann:
> Can we instead have something like `isParsedReference(reference)` to make difference clear?

After removing the code you found above, there was only one use site remaining, so I didn't add this for now.

sailingKieler (approved):
LGTM 👍 

What's the problem with `!!`?

spoenemann:
> What's the problem with `!!`?

It's a pattern to convert something into a Boolean value, but double negation is always harder to understand than a direct statement ("I wouldn't say that noone has done it" vs. "Someone has done it"). `Boolean(...)` yields the same result, but directly says what it does.

PR #787, opened 2022-11-28
Use document paths for json serializer references

msujew:
To facilitate grammar loading without resorting to using the linker, we now use document paths for the json serializer. This makes some functions async which were previously sync, but these were used for testing purposes only anyway.

Also introduces a function to copy AST nodes, as we now need deep copies of AST nodes for the langium-cli to correctly embed referenced grammars.

### DIFF
--- a/packages/langium/src/serializer/json-serializer.ts
+++ b/packages/langium/src/serializer/json-serializer.ts
@@ -71,48 +71,6 @@ export class DefaultJsonSerializer implements JsonSerializer {
         return value;
     }
 
-    protected decycle(object: AstNode, ignore: Set<string>): unknown {
-        const objects = new Set<unknown>(); // Keep references to each unique object
-
-        const replace = (item: unknown) => {
-            // The replace function recurses through the object, producing the deep copy.
-            if (typeof item === 'object' && item !== null) {
-                if (objects.has(item)) {
-                    throw new Error('Cycle in ast detected.');
-                } else {
-                    objects.add(item);
-                }
-                // If it is a reference, transform it into a path
-                if (isReference(item)) {
-                    const refValue = item.ref;
-                    return {
-                        $ref: refValue && this.astNodeLocator.getAstNodePath(refValue)
-                    };
-                }
-                let newItem: Record<string, unknown> | unknown[];
-                // If it is an array, replicate the array.
-                if (Array.isArray(item)) {
-                    newItem = [];
-                    for (let i = 0; i < item.length; i++) {
-                        newItem[i] = replace(item[i]);
-                    }
-                } else {
-                    // If it is an object, replicate the object.
-                    newItem = {};
-                    for (const [name, itemValue] of Object.entries(item)) {
-                        if (!ignore.has(name)) {
-                            newItem[name] = replace(itemValue);
-                        }
-                    }
-                }
-                return newItem;
-            }
-            return item;
-        };
-        const result = replace(object);
-        return result;
-    }
-
     protected linkNode(node: GenericAstNode, root: AstNode, container?: AstNode, containerProperty?: string, containerIndex?: number) {
         for (const [propertyName, item] of Object.entries(node)) {
             if (Array.isArray(item)) {
--- a/packages/langium/src/serializer/json-serializer.ts
+++ b/packages/langium/src/serializer/json-serializer.ts
@@ -4,9 +4,9 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { AstNode, isAstNode, Reference, isReference } from '../syntax-tree';
-import { Linker } from '../references/linker';
 import { LangiumServices } from '../services';
+import { AstNode, isAstNode, isReference } from '../syntax-tree';
+import { AstNodeLocator } from '../workspace/ast-node-locator';
 
 /**
  * Utility service for transforming an `AstNode` into a JSON string and vice versa.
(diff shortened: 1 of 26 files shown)


packages/langium/src/serializer/json-serializer.ts
```diff
@@ -99,9 +114,11 @@ export class DefaultJsonSerializer implements JsonSerializer {
             } else if (typeof value === 'object' && value !== null) {
                 for (const [name, item] of Object.entries(value)) {
                     if (typeof item === 'object' && item !== null) {
-                        if (isReference(item)) {
-                            const reference = this.linker.buildReference(value as AstNode, name, item.$refNode, item.$refText);
-                            value[name] = reference;
+                        if (isIntermediateReference(item)) {
+                            pathActions.push(() => {
+                                const ref = this.evalRef(root, item.$ref);
+                                value[name] = { ref, $refText: '' };
```
spoenemann, line 120: Proposal: include the `$refText` in the intermediate references to retain that information. We could add an option to the serialize method to suppress this when it's not required (e.g. when the JSON data is used outside of Langium).

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -52,9 +60,12 @@ export class DefaultJsonSerializer implements JsonSerializer {
                 } else {
                     objects.add(item);
                 }
-                // If it is a reference, just return the name
+                // If it is a reference, transform it into a path
                 if (isReference(item)) {
-                    return { $refText: item.$refText } as Reference; // surprisingly this cast works at the time of writing, although $refNode is absent
+                    const refValue = item.ref;
+                    return {
+                        $ref: refValue && this.astNodeLocator.getAstNodePath(refValue)
```
spoenemann, line 67: Proposal: prepend the path with `#` so we can later add support for references to other documents in the format `{uri}#{path}`

packages/langium/src/utils/ast-util.ts
```diff
@@ -194,3 +195,52 @@ export function findLocalReferences(targetNode: AstNode, lookup = getDocument(ta
...
     return stream(refs);
 }
+
+/**
+ * Creates a deep copy of the specified AST node.
+ * The resulting copy will only contain semantically relevant information, such as the `$type` property and AST properties.
+ *
+ * References are copied without resolved cross reference. The specified linker is used to resolve them when necessary.
+ */
+export function copyAstNode<T extends AstNode = AstNode>(node: T, linker: Linker): T {
```
spoenemann, line 205: Proposal: accept a `buildReference` function instead of the whole linker (makes it clearer what the state of the references is in the copy).

packages/langium/src/utils/grammar-util.ts
```diff
@@ -270,28 +271,33 @@ function findNameAssignmentInternal(type: ast.AbstractType, cache: Map<ast.Abstr
  */
 export function loadGrammarFromJson(json: string): ast.Grammar {
     const services = createLangiumGrammarServices(EmptyFileSystem).grammar;
-    const astNode = services.serializer.JsonSerializer.deserialize(json);
-    if (!ast.isGrammar(astNode)) {
-        throw new Error('Could not load grammar from specified json input.');
-    }
-    return prepareGrammar(services, astNode);
+    const astNode = services.serializer.JsonSerializer.deserialize(json) as Mutable<ast.Grammar>;
+    const document = services.shared.workspace.LangiumDocumentFactory.fromModel(astNode, URI.parse('memory://grammar.langium'));
```
spoenemann, line 275: Proposal: `'memory://${grammar.name}.langium'`

packages/langium/src/utils/grammar-util.ts
```diff
@@ -270,28 +271,33 @@ function findNameAssignmentInternal(type: ast.AbstractType, cache: Map<ast.Abstr
...
+    astNode.$document = document;
```
spoenemann, line 276: I think this assignment should be done in `LangiumDocumentFactory.fromModel`

spoenemann (commented):
This is a great step forward!

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -113,7 +130,12 @@ export class DefaultJsonSerializer implements JsonSerializer {
                 }
             }
         };
-        internalRevive(object);
-        return object;
+        internalRevive(root);
```
spoenemann, line 133: This helper function is quite hard to understand. Here's an approach for simplifying it (more like pseudo-code, needs to be adapted):
```typescript
function linkNode(node: AstNode, root: AstNode, container?: AstNode, containerProperty?: string, containerIndex?: number): void {
    node.$container = container;
    node.$containerProperty = containerProperty;
    node.$containerIndex = containerIndex;
    for (const [propertyName, item] of Object.entries(node)) {
        if (Array.isArray(item)) {
            for (let index = 0; index < item.length; index++) {
                const element = item[index];
                if (isReference(element)) {
                    element.ref = getAstNode(root, element.$ref);
                } else if (isAstNode(element)) {
                    linkNode(element, root, node, propertyName, index);
                }
            }
        } else if (isReference(item)) {
            item.ref = getAstNode(root, item.$ref);
        } else if (isAstNode(item)) {
            linkNode(item, root, node, propertyName);
        }
    }
}
```

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -17,103 +23,84 @@ export interface JsonSerializer {
...
-    protected decycle(object: AstNode, ignore: string[]): unknown {
-        const objects = new Set<unknown>(); // Keep references to each unique object
+    protected replacer(key: string, value: unknown, refText?: boolean): unknown {
+        if (this.ignoreProperties.has(key)) {
+            return undefined;
+        } else if (isReference(value)) {
+            const refValue = value.ref;
+            return {
+                $refText: refText ? value.$refText : undefined,
+                $ref: '#' + (refValue && this.astNodeLocator.getAstNodePath(refValue))
```
spoenemann, line 68: How do we handle unresolved references? The `$ref` property should probably be undefined in this case. Shall we store some information from the `error` property in this case (e.g. into `$error`)?

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -17,103 +23,84 @@ export interface JsonSerializer {
...
-                            newItem[name] = replace(itemValue);
-                        }
+    protected linkNode(node: GenericAstNode, root: AstNode, container?: AstNode, containerProperty?: string, containerIndex?: number) {
+        for (const [propertyName, item] of Object.entries(node)) {
+            if (Array.isArray(item)) {
+                for (let index = 0; index < item.length; index++) {
+                    const element = item[index];
+                    if (isIntermediateReference(element)) {
+                        item[index] = {
+                            $refText: element.$refText ?? '',
```
spoenemann, line 81: We could use the NameProvider when `$refText` is not given.

spoenemann (approved):
Let's ship it! I have two more remarks though:

Issue #788, opened 2022-11-28
Documents are ignored during initial indexing when already opened in an editor during VSCode startup


sailingKieler:
<img width="1407" alt="Bildschirmfoto 2022-11-28 um 12 10 04" src="https://user-images.githubusercontent.com/8513883/204267994-e5173254-e8b3-4f30-8164-545a94472fe3.png">

## Explanation:
* LS `initialized` event causes the workspace manager beginning its initially scan for files, the subsequent start of the initially document build is guarded by https://github.com/langium/langium/blob/5ab46cf82543c75c26750a38cca4b00323fd7b2a/packages/langium/src/workspace/workspace-manager.ts#L79, which here causes a cancelation because of an immediately following
* `onDidOpenTextDocument` event for `trafficlight.statemachine`, which (after canceling previously scheduled build activities) causes the deletion of the previously created langiumDocument and scheduling of a subsequent (still initial) build of that document via https://github.com/langium/langium/blob/5ab46cf82543c75c26750a38cca4b00323fd7b2a/packages/langium/src/workspace/document-builder.ts#L86-L92 that in turn gets canceled at the cancelation check, as a second
* `onDidOpenTextDocument` event for `trafficlightExtended.statemachine` comes in that  (again after canceling previously scheduled build activities) triggers the deletion of the previously created langiumDocument and scheduling of a subsequent (still initial) build of that document. That build will eventually be executed with the document list of `[ trafficlightExtended.statemachine ]`.
As a consequence, `trafficlight.statemachine` has never been built, especially never indexed.
Thus, error markers on that document in the IDE are not up-to-date, and other documents suffer linking errors.

PR #801, opened 2022-12-01
Keep documents in mind after invalidation, re-parse them during build or on access

sailingKieler:
Solves  #788

### DIFF
--- a/packages/langium/src/workspace/document-builder.ts
+++ b/packages/langium/src/workspace/document-builder.ts
@@ -12,7 +12,7 @@ import { AstNode } from '../syntax-tree';
 import { MultiMap } from '../utils/collections';
 import { interruptAndCheck, MaybePromise } from '../utils/promise-util';
 import { IndexManager } from '../workspace/index-manager';
-import { DocumentState, LangiumDocument, LangiumDocuments } from './documents';
+import { DocumentState, LangiumDocument, LangiumDocuments, LangiumDocumentFactory } from './documents';
 
 export interface BuildOptions {
     validationChecks?: 'none' | 'all'
@@ -62,6 +62,7 @@ export type DocumentUpdateListener = (changed: URI[], deleted: URI[]) => void
 export type DocumentBuildListener = (built: LangiumDocument[], cancelToken: CancellationToken) => Promise<void>
 export class DefaultDocumentBuilder implements DocumentBuilder {
     protected readonly langiumDocuments: LangiumDocuments;
+    protected readonly langiumDocumentFactory: LangiumDocumentFactory;
     protected readonly indexManager: IndexManager;
     protected readonly serviceRegistry: ServiceRegistry;
     protected readonly updateListeners: DocumentUpdateListener[] = [];
@@ -69,6 +70,7 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
 
     constructor(services: LangiumSharedServices) {
         this.langiumDocuments = services.workspace.LangiumDocuments;
+        this.langiumDocumentFactory = services.workspace.LangiumDocumentFactory;
         this.indexManager = services.workspace.IndexManager;
         this.serviceRegistry = services.ServiceRegistry;
     }
@@ -79,7 +81,7 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
 
     async update(changed: URI[], deleted: URI[], cancelToken = CancellationToken.None): Promise<void> {
         for (const deletedDocument of deleted) {
-            this.langiumDocuments.invalidateDocument(deletedDocument);
+            this.langiumDocuments.deleteDocument(deletedDocument);
         }
         this.indexManager.remove(deleted);
         for (const changedUri of changed) {
@@ -128,6 +130,12 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
     }
 
     protected async buildDocuments(documents: LangiumDocument[], options: BuildOptions, cancelToken: CancellationToken): Promise<void> {
+        // 0. Parse content
+        //  parsing is done initially for each document, but
+        //  re-parsing after changes reported by the client might have been canceled by subsequent changes, so re-parse now
+        await this.runCancelable(documents, DocumentState.Parsed, cancelToken, doc =>
+            this.langiumDocumentFactory.update(doc)
+        );
         // 1. Index content
         await this.runCancelable(documents, DocumentState.IndexedContent, cancelToken, doc =>
             this.indexManager.updateContent(doc, cancelToken)
--- a/packages/langium/src/workspace/documents.ts
+++ b/packages/langium/src/workspace/documents.ts
@@ -135,6 +135,10 @@ export class DefaultTextDocumentFactory implements TextDocumentFactory {
 
 /**
  * Shared service for creating `LangiumDocument` instances.
+ *
+ * Register a custom implementation if special (additional) behavior is required for your language(s).
+ * Note: If you specialize {@link fromString} or {@link fromTextDocument} you probably might want to
+ * specialize {@link update}, too!
  */
 export interface LangiumDocumentFactory {
     /**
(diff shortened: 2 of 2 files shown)


packages/langium/src/workspace/documents.ts
```diff
@@ -327,6 +355,22 @@ export class DefaultLangiumDocuments implements LangiumDocuments {
...
+        if (langiumDoc) {
+            langiumDoc.state = DocumentState.Changed;
+            langiumDoc.parseResult = {
+                lexerErrors: [],
+                parserErrors: [],
+                value: { // dummy AstNode to have the parseResult properly initialized
+                    $type: 'AstNode',
+                    $document: langiumDoc
+                }
+            };
```
msujew, line 369: Suggestion: I wouldn't put a dummy node here. The user will always have the information, that the document is outdated through the `state` property. Putting a node in here which doesn't fit the constraints of the `LangiumDocument<T>` parameter might lead to runtime errors, if the user tries to access those.

packages/langium/src/workspace/documents.ts
```diff
@@ -304,18 +311,39 @@ export class DefaultLangiumDocuments implements LangiumDocuments {
...
+    }
+
+    protected reloadDocumentContent(outdatedDocument: LangiumDocument) {
+        const newDocument = this.loadDocumentContent(outdatedDocument.uri);
+
+        // wipe out the existing document object...
+        Object.keys(outdatedDocument).forEach(key => delete outdatedDocument[key as keyof LangiumDocument]);
+        // ...and take over all the properties of the newly create one,
+        Object.assign(outdatedDocument, newDocument);
+        (outdatedDocument.parseResult.value as Mutable<AstNode>).$document = outdatedDocument;
```
msujew, line 333: I'm not too happy with this solution, but since we need to keep the original document reference around, I guess there's no better solution? Another way would be to break the API of `LangiumDocumentFactory` to accept a document to override.

sailingKieler: Right. I also tried to add an optional function argument for replacing the call of `DefaultLangiumDocumentFactor.createLangiumDocument(...)` but that messed up the existing API considerably, so I skipped that.

packages/langium/src/workspace/documents.ts
sailingKieler: Do we state somewhere that client code should check the state before accessing the content a document?
I didn't find a hint on that, by now.

> might lead to runtime errors, if the user tries to access those
That's true. Do you have another proposal?
Just leaving the stale content in there feels also error-prone to me. With EMF we proxified the AST in that case...

packages/langium/src/workspace/document-builder.ts
```diff
@@ -119,6 +120,7 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
             e.state = Math.min(e.state, DocumentState.ComputedScopes); // need to re-index potentially linked references
         });
         const docSet = new Set([
+            ...invalid,
```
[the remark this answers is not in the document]
sailingKieler: Right I'll check that.

packages/langium/src/workspace/documents.ts
spoenemann: We should just document that cross-references might not be resolved when you access a document with state < Linked. If you want to do something with a properly linked document, you can hook into the DocumentBuilder phases.

packages/langium/src/workspace/documents.ts
```diff
@@ -304,18 +311,39 @@ export class DefaultLangiumDocuments implements LangiumDocuments {
         let langiumDoc = this.documentMap.get(uriString);
         if (langiumDoc) {
             // The document is already present in our map
-            return langiumDoc;
+            if (langiumDoc.state < DocumentState.Parsed) {
+                this.reloadDocumentContent(langiumDoc);
```
spoenemann, line 315: This change adds implicit behavior to this method, which I would prefer to make explicit: Add a protected method to DocumentBuilder that is called on `Changed` documents, invokes the parser and assigns a new `parseResult`.

packages/langium/src/workspace/documents.ts
spoenemann: I agree with @msujew and propose to modify the properties explicitly:
 - Clear `references`, `precomputedScopes` and `diagnostics` in `invalidateDocument`
 - Also clear the TextDocument reference by using a "private" property instead of a local variable in `DefaultLangiumDocumentFactory.createLangiumDocument`; you can define such a property with a subinterface that is not exported
 - The `parseResult` is kept (see other discussion) and reassigned by the document builder

sailingKieler:
2nd attempt is pushed, main change is the extraction of https://github.com/langium/langium/pull/801/files#diff-eb9d95b72e597e87b07fd5d97f6df837d272c1d3041d298ed617801ab670193aR220-R223, which is used in document creation _and_ update

packages/langium/src/workspace/documents.ts
```diff
@@ -177,21 +191,37 @@ export class DefaultLangiumDocumentFactory implements LangiumDocumentFactory {
...
             $template: true,
-            languageId: services.LanguageMetaData.languageId,
+            languageId: servicesReqistry.getServices(uri).LanguageMetaData.languageId,
             uri,
             text
         });
     }
 
+    update<T extends AstNode = AstNode>(document: LangiumDocument<T>): LangiumDocument<T> {
+        const textDocument = this.textDocuments.get(document.uri.toString()) ?? document.textDocument;
```
msujew, line 212: Suggestion: I believe we should never re-use the existing `document.textDocument` here, since it will always be outdated. We should resort to asking the file system for the text instead, if the document is not in the list of text documents.

msujew (commented):
Looks way better in my opinion 👍

packages/langium/src/workspace/documents.ts
sailingKieler: You're right. I just checked the case of a change event being triggered from outside with no editor being opened in VS code. I that case it's definitely outdated.

packages/langium/src/workspace/documents.ts
```diff
@@ -177,19 +194,61 @@ export class DefaultLangiumDocumentFactory implements LangiumDocumentFactory {
...
+            );
+        else {
+            const documentTemplate = this.createTextDocumentTemplate(document.uri, text);
+            const textDocumentGetter = this.getTextDocumentGetter(documentTemplate);
+            Object.defineProperty(
+                document, 'textDocument', {
+                    get: textDocumentGetter
+                }
+            );
+        }
```
spoenemann, line 226: This is good. Proposal: do the same in `createLangiumDocument` (i.e. define a value property if the text document is already present), then `getTextDocumentGetter` would be used only with a template.

packages/langium/src/workspace/documents.ts
```diff
@@ -327,6 +404,17 @@ export class DefaultLangiumDocuments implements LangiumDocuments {
     }
 
     invalidateDocument(uri: URI): void {
```
spoenemann, line 406: Idea: return `LangiumDocument | undefined` so it's easier to customize in subclasses (same for `deleteDocument`).

packages/langium/src/workspace/documents.ts
```diff
@@ -327,6 +404,17 @@ export class DefaultLangiumDocuments implements LangiumDocuments {
...
+        const langiumDoc = this.documentMap.get(uriString);
+        if (langiumDoc) {
+            langiumDoc.state = DocumentState.Changed;
+            langiumDoc.references = [];
+            langiumDoc.precomputedScopes = undefined;
+            langiumDoc.diagnostics = [];
+        }
+    }
+
+    deleteDocument(uri: URI): void {
```
spoenemann, line 417: Thought: shall we keep the document state intact in this new method? Rationale: the method name does not imply that the content has changed.

spoenemann (commented):
Great work! I only have detail remarks now.

packages/langium/src/workspace/documents.ts
sailingKieler: Actually I thought about doing that for consistency.
But in 99.9999% of the executions the 'get' variant will be used, so I skipped it.
Would you still include it, too?

packages/langium/src/workspace/documents.ts
sailingKieler: Good point!

packages/langium/src/workspace/documents.ts
spoenemann: You are right, but when you get into the unlikely case it's good if the data has the same shape as always.

packages/langium/src/workspace/documents.ts
sailingKieler: Changed it.

packages/langium/src/workspace/documents.ts
```diff
@@ -198,33 +211,97 @@ export class DefaultLangiumDocumentFactory implements LangiumDocumentFactory {
...
+        const languageId = this.serviceRegistry.getServices(uri).LanguageMetaData.languageId;
+        return {
+            $template: true,
+            languageId,
+            uri,
+            text
+        };
+    }
+
+    protected getTextDocumentGetter(template: TextDocumentTemplate): () => TextDocument {
```
spoenemann, line 292: Now to simplify this further, you can remove the `TextDocumentTemplate` type and pass the `uri` and `text` arguments directly to this method.

spoenemann (approved):
:+1:

PR #806, opened 2022-12-07
Further reworks of the documents loading…

sailingKieler:
... to concentrate accesses the TextDocuments and FS in the 'LangiumDocumentFactory'

followup on #788, #801

### DIFF
--- a/packages/langium/src/workspace/documents.ts
+++ b/packages/langium/src/workspace/documents.ts
@@ -157,6 +157,12 @@ export interface LangiumDocumentFactory {
     fromModel<T extends AstNode = AstNode>(model: T, uri: URI): LangiumDocument<T>;
 
     /**
+     * Create a Langium document for the given URI. The document shall be fetched from the {@link TextDocuments}
+     * service if present, and loaded via the configured {@link FileSystemProvider} otherwise.
+     */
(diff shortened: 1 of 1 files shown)


spoenemann (approved):
Yes, another improvement!

packages/langium/src/workspace/documents.ts
```diff
@@ -156,6 +156,12 @@ export interface LangiumDocumentFactory {
      */
     fromModel<T extends AstNode = AstNode>(model: T, uri: URI): LangiumDocument<T>;
 
+    /**
+     * Create a Langium document for the given URI. The document shall be fetched from the {@link TextDocuments}
+     * service if present, and loaded via the configured {@link FileSystemProvider} otherwise.
+     */
+    create<T extends AstNode = AstNode>(uri: URI): LangiumDocument<T>
```
sailingKieler, line 163: I removed the additional optional `content` parameter that was in here in my first draft.
IMO this justifies the presence of the 'from...' methods much better, the document builder needs just this, and everything below is just implementation detail of `DefaultLangiumDocumentFactory`.

@spoenemann @msujew  Do you agree?

spoenemann: 👍

PR #808, opened 2022-12-08
Fix type-generator.test on Windows

kaisalmen:
Test  type-generator.test.ts is broken on Windows because of different line endings.
In addition `clean` now removes out folders to clean old garbage,
I also would like to add volta to easily enforce a specific node/npm version, but I don't know if you want this?

### DIFF
--- a/packages/langium-cli/test/generator/types-generator.test.ts
+++ b/packages/langium-cli/test/generator/types-generator.test.ts
@@ -14,12 +14,10 @@ describe('Types generator', () => {
 
     test('should generate types file', async () => {
         const result = (await parseHelper<Grammar>(grammar)(TEST_GRAMMAR)).parseResult;
-        const typesFileContent = generateTypesFile(grammar, [result.value]);
-        let expectedTypes = EXPECTED_TYPES;
-        if (process.platform === 'win32') {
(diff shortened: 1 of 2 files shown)


packages/langium-cli/test/generator/types-generator.test.ts
```diff
@@ -15,7 +15,11 @@ describe('Types generator', () => {
     test('should generate types file', async () => {
         const result = (await parseHelper<Grammar>(grammar)(TEST_GRAMMAR)).parseResult;
         const typesFileContent = generateTypesFile(grammar, [result.value]);
```
msujew, line 17: Suggestion: I believe it'd be better to just always remove `\r` from the generated results. Makes it less verbose (though at the cost of making it less explicit why we perform the replacement.

```suggestion
        const typesFileContent = generateTypesFile(grammar, [result.value]).replace(/\r/g, '');
```

kaisalmen: This is actually the better idea. This way the expectation is not changed. I will add a comment to why this is done.

kaisalmen:
@msujew are you ok with the other changes in the `package,json`

msujew (approved):
LGTM 👍

PR #811, opened 2022-12-09
Enable keywords as properties in declared types

luan-xiaokun:
* Enable langium keywords as properties in declared types
* Add a test for using keywords as properties in declared types
* Add semantic token highlighting support for properties in declared types


spoenemann (approved):
Thanks for the contribution!

PR #812, opened 2022-12-09
To run tests node 16 is required

dhuebner:
Using node 14 a lot of tests are failing

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ Thank you for your interest in the Langium project! The following is a set of gu
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


kaisalmen (approved):
This change is fine. I am just asking myself if we should fix/upgrade the npm/node engines entries in the package.json to reflect that.

dhuebner:
@kaisalmen 
If it's possible, sure. 
But I would love to see the change in the documentation anyway, it cost me time to fix the tests. I tried almost everything else but not changing the node version because the documentation says node 14 is okay.

kaisalmen:
@dhuebner go ahead and merge this. We should update engines, etc. when performing dependency updates.

PR #816, opened 2022-12-09
Parsing errors may lead to invalid CST range data

### LINKED ISSUE #695: Parsing errors may lead to invalid CST range data
In a document with a parsing error (superfluous whitespace within an identifier) I observe range data with `NaN` values in the `end` position object.

Once diagnostics are created on the corresponding `AstNode` those range data are taken and send to the language client, see

Bildschirmfoto 2022-09-22 um 16 01 01

That yields exceptions on the client side:

Bildschirmfoto 2022-09-22 um 16 06 44

Bildschirmfoto 2022-09-22 um 16 08 01

dhuebner:
Should fix  #695 bei filtering out nodes inserted in recovery from the CST


msujew (approved):
Looks good to me 👍

Issue #750, opened 2022-11-02
Declared types as container types


pluralia:
Container types are computed based on inferred types only. We should add declared types as container types too.

Lotes:
Forgive my missing knowledge, what do you mean with containers? How is it done for inferred types and how should it look for declared types?

pluralia:
> what do you mean with containers?

Each `AstNode` has a `$container` property. Its type lists all possible `AstNode`s that can contain that `AstNode`. For example, container type of `NumberLiteral` in [the arithmetic example](https://github.com/langium/langium/blob/main/examples/arithmetics/src/language-server/generated/ast.ts#L106) is `BinaryExpression | Definition | Evaluation | FunctionCall` what means, a parent object of a `NumberLiteral` object, can be `BinaryExpression` or `Definition`, etc.

> How is it done for inferred types

This question has a long answer. I leave a link to [the source code](https://github.com/langium/langium/blob/main/packages/langium/src/grammar/type-system/inferred-types.ts#L484-L532) and to [the issues](https://github.com/langium/langium/issues/150) about problems with current calculation as it has some explanation. Let me know if you still have questions.

> how should it look for declared types?

The same like for inferred types. Basically, to close this issue, it's necessary just add declared types in calculation of container types that it would happen not only on inferred ones.

Lotes:
Thank you. I am convinced that the task becomes easier when we would model the rules and alternatives as types with properties inside of a type-property-graph.
The arithmetic example might be a good starter to talk about the contents of the nodes and edges. I can prepare a visualization next week.

Lotes:
![Arithmetic](https://user-images.githubusercontent.com/1525964/200125914-466a8c66-180e-4fa3-9ba9-f692cdb68acb.png)
Done with SimpleMind :3

Lotes:
![Contains](https://user-images.githubusercontent.com/1525964/200126670-553a43ab-dd2f-4696-9860-534f3b9de9e7.png)
Here are the containment relations minus primitive types.

Lotes:
```
foreach t: type:
  find: (container: type)—[has-property]->(:property)-[has-type]->(t:type)
  containers[t].push(container)
  find: (container: type)—[has-property]->(:property)-[has-type]->(:type)<—[extends*]-(t: type)
  containers[t].push(container)
```
Easy! :D

Issue #744, opened 2022-10-31
Generated type from declared type does not include all of them


luan-xiaokun:
A minimal example:

Using the following type declaration
```
grammar Debug

entry Model:
    (items+=AB)*;

interface A {
    a: string;
}
interface B {
    b: string[];
}

type AB = A | B;

A returns A:
    'A' a=ID;

AB returns AB:
    A ({B} b+=ID)*;

// A infers A:
//     'A' a=ID;

// AB:
//     A ({infer B} b+=ID)*;

hidden terminal WS: /\s+/;
terminal ID: /[_a-zA-Z][\w_]*/;
```
generates
```ts
export type AB = A;

export interface A extends AstNode {
    readonly $container: Model;
    a: string
}

export interface B extends AstNode {
    b: Array<string>
}
```

Switching to `infers` generates
```ts
export type AB = A | B;
```

'A' and 'B' are always subtypes of 'AB' in both ways.

luan-xiaokun:
When using type declaration, changing the `b` in `AB returns AB: A ({B} b+=ID)*;` to other undeclared name such as `c` does not raise error.

luan-xiaokun:
As far as I can tell, the skipping below seems to be the cause of this issue. After removing line 103 to 105 the generated type becomes `export type AB = A | B;`. But frankly, I do not really understand the logic behind the skipping.

https://github.com/langium/langium/blob/9f587815f9d14f433d22831ac11d3b9f11696d9f/packages/langium/src/grammar/type-system/inferred-types.ts#L97-L105

luan-xiaokun:
After looking more closely at the code, I think the cause of the problem may be the `distinct` function on the stream.

https://github.com/langium/langium/blob/9f587815f9d14f433d22831ac11d3b9f11696d9f/packages/langium/src/grammar/type-system/type-collector.ts#L19-L34

In this minimal example:
1. When `infers` is used, everything can be directly inferred by the `collectInferredTypes` function, so the variable `declared` now has empty array properties, and nothing goes wrong.
2. When `returns` is used, the type `AB` obtained by `collectInferredTypes` has the property `union` containing only one element, namely `A`, because type `B` is a declared type and it is ignored. The `union` property of `AB` obtained by `collectDeclaredTypes` contains both `A` and `B` as we declared.
3. The `distinct` function keeps the `AB` computed by `collectInferredTypes`, which is incomplete, and discards the other one, which is supposed to be kept.

[...]

Issue #583, opened 2022-07-06
Declared Types Incorrectly extend Inferred Types


montymxb:
Present in Langium version: 0.4.0

Currently, declared types can extend inferred types. This can lead to situations like the following:
```typescript
// generates inferred type 'A'
A: 'a' a=ID;

// partially-declared type 'B', since it is founded on inferred type 'A', but is still only usable where declared types go (like in returns)
interface B extends A {
    b: string
}

B returns B: 'b' b=ID;
```
In this case, we would expect the parser rule for B to have an error, since we're missing property `a`. However, we get no such error.
...

<img width="188" alt="Screen Shot 2022-07-06 at 19 09 33" src="https://user-images.githubusercontent.com/5070304/177605948-05ba84d5-eff8-4eeb-96d6-2570c0d69bd1.png">

This also generates the following AST types, showing that we're missing a link in the inheritance. This explains, in part, why we're not seeing an error for missing property 'a'.

```typescript
export interface A extends AstNode {
    a: string
}

export interface B extends AstNode {
    b: string
}
```

If we do the same thing, but with declared types, we get an error as we would expect.
...

<img width="516" alt="Screen Shot 2022-07-06 at 19 11 41" src="https://user-images.githubusercontent.com/5070304/177606296-aeb4c5aa-cf13-441b-9818-d73976a7032e.png">

We should either:
- properly validate declared types extending inferred types
- prevent declared types from extending inferred types

spoenemann:
The missing validation error affects an inferred type, but for those types our handling of properties is rather lax anyway (see also https://github.com/langium/langium/issues/562#issuecomment-1169798215). We get to the same situation with only inferred types:

```
A: 'a' a=ID;

B: 'b' b=ID;

C infers A: {infer B} 'c' b=ID;
```

So we could just decide to ignore the fuzziness here.

Another approach: consider this in the type inference, e.g. by making the `a` property optional.

PR #819, opened 2022-12-11
Refactoring of type inference and usage

pluralia:
* Closes
  - #447
  - #750
  - #744
  - #583
* Improves collecting type resources for validation:
  - `LangiumGrammarDocument` extends `LangiumDocument` by adding the `validationResources` property 
  - `LangiumGrammarValidationResourcesCollector` service collects the validation resources for each document during document building
* Extracts type validations in a separate `LangiumGrammarTypesValidator`
* Adds `$type` property (discussed in #769)
* Fixes `$container` types calculation for inferred types
* Makes `types-generator.ts` (#754) using the same methods of converting to string as `ast-generator.ts`
* Reorganizes the `type-system` folder


examples/arithmetics/src/language-server/generated/ast.ts
```diff
@@ -44,7 +44,7 @@ export function isBinaryExpression(item: unknown): item is BinaryExpression {
 }
 
 export interface DeclaredParameter extends AstNode {
-    readonly $container: Definition;
+    readonly $container: Definition | Module;
```
spoenemann, line 47: This seems wrong: a DeclaredParameter can only be contained in a Definition.

examples/arithmetics/src/language-server/generated/ast.ts
```diff
@@ -55,7 +55,7 @@ export function isDeclaredParameter(item: unknown): item is DeclaredParameter {
 }
 
 export interface Definition extends AstNode {
-    readonly $container: Module;
+    readonly $container: Definition | Module;
```
spoenemann, line 58: This seems wrong: a Definition can only be contained in a Module.

packages/langium/src/grammar/workspace/document-builder.ts
```diff
@@ -0,0 +1,32 @@
...
+import { LangiumGrammarServices } from '../langium-grammar-module';
+import { LangiumGrammarDocument } from './documents';
+
+export class LangiumGrammarDocumentBuilder extends DefaultDocumentBuilder {
+    constructor(services: LangiumSharedServices) {
+        super(services);
+        this.addTypeCollectionPhase();
+    }
+
+    private addTypeCollectionPhase() {
```
spoenemann, line 21: I would not do this in a subclass because there's nothing to override in this case. Proposal:
 - Move this code to a function in `langium-grammar-module.ts`
 - Call the function in `createLangiumGrammarServices`

packages/langium/src/test/langium-test.ts
```diff
@@ -16,8 +16,9 @@ import { LangiumDocument } from '../workspace/documents';
 import { findNodeForProperty } from '../utils/grammar-util';
 import { SemanticTokensDecoder } from '../lsp/semantic-token-provider';
 import { TextDocument } from 'vscode-languageserver-textdocument';
+import { LangiumGrammarServices } from '../grammar/langium-grammar-module';
 
-export function parseHelper<T extends AstNode = AstNode>(services: LangiumServices): (input: string) => Promise<LangiumDocument<T>> {
+export function parseHelper<T extends AstNode = AstNode>(services: LangiumGrammarServices): (input: string) => Promise<LangiumDocument<T>> {
```
spoenemann, line 21: These testing utilities are not only meant for our own grammar language, but are published to by used for any Langium-based language. Please revert the changes in this file, even if it means we have to type-cast in our grammar language tests.

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -161,3 +161,114 @@ describe('validate declared types', () => {
 
     });
 });
+
+describe('NEW TESTS', () => {
```
spoenemann, line 165: Please find a better description for this test suite.

packages/langium/src/grammar/generated/ast.ts
```diff
@@ -200,7 +201,7 @@ export function isInferredType(item: unknown): item is InferredType {
 }
 
 export interface Interface extends AstNode {
-    readonly $container: Grammar;
+    readonly $container: Alternatives | Assignment | AtomType | CharacterRange | CrossReference | Grammar | Group | NegatedToken | ParserRule | TerminalAlternatives | TerminalGroup | TerminalRule | UnorderedGroup | UntilToken;
```
dhuebner, line 204: This doesn't seem to be right. 
Can e.g. an assignment be a container of an Interface?  Other `$container` types also doesn't seem to be right to me.

spoenemann (commented):
Thank you @pluralia, great step forward!

sailingKieler:
Good job @pluralia 

I tested it with a real world grammar and ... there's an issue. Extracted it into a test case grammar:
```
grammar DeclaredReferenceTypes

entry Named: '{'
    name = ID
'}';

Expression:
    NamedRef | Literal;

Literal:
    lit = STRING
;

NamedRef:
    ref = [Named]
;

interface QualifiedRef extends NamedRef {
    qualifier: Expression
}

QualifiedRef1 infers Expression:
    NamedRef ({QualifiedRef.qualifier=current} '.' ref=[Named])+    // <-- fails with "The assigned type 'NamedRef' is not compatible with the declared property 'qualifier' of type 'Expression': , 'NamedRef' is not expected."
;

QualifiedRef2 infers Expression:
    {infer NamedRef} (ref=[Named]) ({QualifiedRef.qualifier=current} '.' ref=[Named])+    // <-- same here: "The assigned type 'NamedRef' is not compatible with the declared property 'qualifier' of type 'Expression': , 'NamedRef' is not expected."
;

// the original failing case from the real world project:
interface SpecialNamedRef extends NamedRef {}
interface SpecialQualifiedRef extends QualifiedRef {}

SpecialQualifiedRef infers Expression:
    {SpecialNamedRef} (ref=[Named]) ({SpecialQualifiedRef.qualifier=current} '.' ref=[Named])+.   // <-- fails with: "The assigned type 'SpecialNamedRef' is not compatible with the declared property 'qualifier' of type 'NamedRef': , 'SpecialNamedRef' is not expected."
;
// in the latter case, the validation message suggests that there's something wrong internally, as the declared property 'qualifier' is supposed to be of type 'Expression', not 'NamedRef'

terminal STRING: /"[^"]*"|'[^']*'/;
``` 

You can make it even worse with an additional rule like
```
QualifiedRef3 infers Expression:
    Expression ({QualifiedRef.qualifier=current} '.' ref=[Named])+
;
```

Please play with this example, you can derive more test cases from that.

msujew (commented):
This seems to reintroduce https://github.com/langium/langium/issues/559:

image

The error message also contains to many `,` characters.

There's also another issue which I've noticed about the `$type` property placement. For the following grammar (taken from the statemachine example), the following will be generated:

```
Event:
    name=ID;

interface StateRef {
    state: @State
}

interface Transition extends StateRef {
    event: @Event
}

Transition returns Transition:
    event=[Event] '=>' state=[State];
```

_generates_:

```ts
export interface Event extends AstNode {
    readonly $container: Statemachine;
    name: string
}

export interface StateRef extends AstNode {
    readonly $container: State;
    readonly $type: 'Transition';
    state: Reference<State>
}

export const StateRef = 'StateRef';

export function isStateRef(item: unknown): item is StateRef {
    return reflection.isInstance(item, StateRef);
}

export interface Transition extends StateRef {
    readonly $container: State;
    event: Reference<Event>
}
```

Note that `StateRef` correctly gets the `$type` property, while `Event` doesn't. Event should also receive the `$type: 'Event'` property.

packages/langium/src/grammar/generated/ast.ts
pluralia: True, it's not right by the fact: not all listed types are real `$container`s of `Interface`. However, we have them here since we collect `$container` types per connected component and not per type. We have to do it because one type can inherit some types and having `$container` types per type brings TS errors. Try an example:
```
interface A { $container: D; strA: string }
interface B { $container: E; strB: string }
interface C extends A, B { $container: D | E; strC: string }
interface D { a: A }
interface E { b: B }
```
As the result, our solution is having `$container: D | E` everywhere: for `A` and `B` to. It's not a perfect solution, of course, so, we have an issue about improving `$container` types calculation (#150). Current PR doesn't solve the problem, but adds `$container` types for declared types (#750).

msujew: But `Interface` is completely isolated in it's inheritance structure. Why did it change anyway?

dhuebner: Well okay, when the issue is already addressed

examples/arithmetics/src/language-server/generated/ast.ts
pluralia: This is because `Definition` is a container type of `DeclaredParameter` and `Definition`and `DeclaredParameter` are in the same connected component.

packages/langium/src/grammar/generated/ast.ts
pluralia: @msujew , `Interface` is not isolated: we have `type AbstractType = Interface | Type | Action | ParserRule;` that connects it with `ParserRule` that is connected with `AbstractElement` -- what gives so many container types for each participant of this hierarchy.

pluralia:
@sailingKieler, thank you for the wonderful example :)
The validation error that you found has already existed for parser rules before this PR (I opened an issue based on your example: #823). The reason why you didn't see this error for actions before this PR is that we had no validation for actions. This PR adds it and since it reuses code for parser rules validation, you see this error for actions, too.

packages/langium/test/grammar/type-system/inferred-types.test.ts
```diff
@@ -889,10 +892,12 @@ describe('types of `$container` property are correct', () => {
         expect(interfacesString).toBe(s`
             export interface A extends AstNode {
                 readonly $container: D | E;
+                readonly $type: 'C';
```
msujew, line 895: This doesn't seem to be correct. When parsing rule `A` in this grammar, we would expect an element with a `$type` with value `A` not `C`. The `$type` therefore needs to be `'A' | 'C'`.

Also, interface `D` should have `$type: 'D'` and `E` should have `$type: 'E'`.

pluralia:
@spoenemann, this PR doesn't change container types calculation, but only adds it for declared types. So, if we don't add these changes:
1. We just get erasing `$container` property for all types that are generated from declared ones
2. We get exactly the same `$container` types calculation, like now, if we replace declared types with inferred ones. For example, if you replace a declared type `AbstractDefinition` with a parser rule, you get exactly the same container types.

Overall, upcoming changes for container types calculation are useful and helpful for the current way of calculation since it makes it more correct. I'd propose accept them or remove container types calculation at all, but not leave the current state for the release.

> because changing this back to a more constrained set of container types would be a breaking change

I don't think merging the current container types will bring us breaking changes in the future. The current way is really more generic and new way (that will be closer to real container types) is more constrained. However, extra container types that we add now, will just never be met in real life: we know that they are not real container types. So, the new way just cut what was never used.

spoenemann:
Ok if we see the new (too large) set of container types as an issue that needs to be resolved in a 1.x version, we could merge this and regard the fix to more constrained container types as a "mildly" breaking change, meaning that it causes compile errors only in some specific situations.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -5,31 +5,30 @@
...
     constructor(services: LangiumGrammarServices) {
         super(services);
         const validator = services.validation.LangiumGrammarValidator;
+        const typesValidator = services.validation.LangiumGrammarTypesValidator;
         const checks: ValidationChecks<ast.LangiumGrammarAstType> = {
             Action: [
-                validator.checkActionTypeUnions,
-                validator.checkAssignmentReservedName
+                validator.checkAssignmentReservedName,
+                typesValidator.checkActionIsNotUnionType,
```
spoenemann, line 31: It is incorrect to mix checks from different classes in one call to `register` because it accepts only a single argument to be used as `this`. As consequence, if any check from `LangiumGrammarTypesValidator` would try to access a class property, it would fail because it would use `LangiumGrammarValidator` as `this` context. Solution: create a separate `ValidationChecks` object and call `register` a second time.

spoenemann (approved):
Let's ship this! From my perspective you can merge this after resolving today's comments.

packages/langium/src/grammar/validation/validation-resources-collector.ts
```diff
@@ -0,0 +1,90 @@
...
+import { AbstractElement, Action, Grammar, Interface, isAction, isAlternatives, isGroup, isUnorderedGroup, ParserRule, Type } from '../generated/ast';
+import { getActionType, getRuleType } from '../internal-grammar-util';
+import { AstResources, collectTypeResources, TypeResources } from '../type-system/type-collector/all-types';
+import { mergeInterfaces, mergeTypesAndInterfaces } from '../type-system/types-util';
+import { Property } from '../type-system/type-collector/types';
+import { TypeToValidationInfo, ValidationResources } from '../workspace/documents';
+
+export class LangiumGrammarValidationResourcesCollector {
+
+    collectValidationResources(grammar: Grammar, documents: LangiumDocuments): ValidationResources {
```
spoenemann, line 19: The `LangiumDocuments` could be injected in the constructor instead of passing them as parameter.

PR #820, opened 2022-12-12
Prevent keyword/terminal name clashing

msujew:
Closes https://github.com/langium/langium/issues/719
Closes https://github.com/langium/langium/issues/810

Prevents keywords and terminals from sharing the same token name using a validation. The validation looks at:

1. Local terminals clashing with local keywords
2. Local terminals clashing with imported keywords
3. Imported terminals clashing with local keywords
4. Imported terminals clashing with imported keywords

Note that another solution could've been possible, which included internally renaming tokens generated from terminals during the token builder phase. However, that is pretty invasive and non-obvious for adopters, which is why we simply prevent those cases using validations.


spoenemann (approved):
:+1:

PR #821, opened 2022-12-12
Use plain function to register custom validations

spoenemann:
Creating a subclass just to register some checks is overly complicated. Let's do it directly instead, which is possible without changing the API. With this we show a general scheme to register stuff (same approach for build listeners, for example).

### DIFF
--- a/examples/requirements/src/language-server/requirements-and-tests-lang-module.ts
+++ b/examples/requirements/src/language-server/requirements-and-tests-lang-module.ts
@@ -8,9 +8,13 @@ import {
     createDefaultModule, createDefaultSharedModule, DefaultSharedModuleContext, inject,
     LangiumSharedServices
 } from 'langium';
-import { RequirementsAndTestsGeneratedSharedModule, RequirementsGeneratedModule, TestsGeneratedModule } from './generated/module';
+import {
+    RequirementsAndTestsGeneratedSharedModule, RequirementsGeneratedModule, TestsGeneratedModule
+} from './generated/module';
 import { RequirementsLangModule, RequirementsLangServices } from './requirements-lang-module';
+import { registerRequirementsValidationChecks } from './requirements-lang-validator';
 import { TestsLangModule, TestsLangServices } from './tests-lang-module';
+import { registerTestsValidationChecks } from './tests-lang-validator';
 
 /**
  * Create the full set of services required by Langium.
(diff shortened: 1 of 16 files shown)


msujew (approved):
Looks good to me 👍

examples/requirements/src/language-server/requirements-and-tests-lang-module.ts
```diff
@@ -29,24 +33,26 @@ import { TestsLangModule, TestsLangServices } from './tests-lang-module';
  */
 export function createRequirementsAndTestsLangServices(context: DefaultSharedModuleContext): {
     shared: LangiumSharedServices,
-    RequirementsLang: RequirementsLangServices,
-    TestsLang: TestsLangServices
+    requirements: RequirementsLangServices,
+    tests: TestsLangServices
 } {
```
sailingKieler, line 38: Where do these changes come from? Was the former state out-dated? Did you re-generate the whole project with the yeoman generator?

examples/requirements/src/language-server/requirements-lang-validator.ts
```diff
@@ -4,30 +4,22 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { ValidationAcceptor, ValidationChecks, ValidationRegistry } from 'langium';
+import { ValidationAcceptor, ValidationChecks } from 'langium';
 import { RequirementsAndTestsAstType, Requirement, isTestModel } from './generated/ast';
-import { RequirementsLangServices } from './requirements-lang-module';
+import type { RequirementsLangServices } from './requirements-lang-module';
```
sailingKieler, line 9: Why does this `import type` thing show up here? We don't use it anywhere else, do we?

examples/requirements/src/language-server/requirements-and-tests-lang-module.ts
spoenemann: No, manual renaming for consistency with the other example languages.

examples/requirements/src/language-server/requirements-lang-validator.ts
spoenemann: We use it to ensure that there are not runtime dependencies to that file. In this case a runtime dependency would cause a cycle and thus unexpected problems like undefined global definitions.

PR #825, opened 2022-12-14
Extended code generation support

sailingKieler:


packages/langium/src/generator/template-node.ts
```diff
@@ -0,0 +1,290 @@
+/******************************************************************************
+ * Copyright 2021 TypeFox GmbH
```
spoenemann, line 2: Detail: please use the current year in new files.

sailingKieler:
I plan to do some further code polishing and documentation in `template-node.ts`, everything else is ready for review.

examples/statemachine/test/generator.test.ts
```diff
@@ -0,0 +1,212 @@
...
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { EmptyFileSystem, Generated, toString } from 'langium';
+import { parseHelper } from "langium/lib/test";
+import { generateCppContent } from '../src/cli/generator';
+import { Statemachine } from '../src/language-server/generated/ast';
+import { createStatemachineServices } from '../src/language-server/statemachine-module';
+
+describe('Tests the code generator', () => {
```
spoenemann, line 13: The statemachine package.json is missing a `test` script.

packages/langium/src/generator/generator-node.ts
```diff
@@ -4,31 +4,83 @@
...
+            } else if (arg) {
                 this.contents.push(arg);
             }
         }
         return this;
     }
 
-    indent(func?: (indentNode: IndentNode) => void): CompositeGeneratorNode {
-        const node = new IndentNode();
+    appendIf(condition: boolean, args: Generated | ((node: CompositeGeneratorNode) => void)): this {
```
spoenemann, line 54: What is the use case of this method?

packages/langium/src/generator/generator-node.ts
```diff
@@ -4,31 +4,83 @@
...
+        return condition ? this.append(args) : this;
+    }
+
+    newLine(): this {
+        return this.appendNewLine();
+    }
+
+    appendNewLine(): this {
```
spoenemann, line 62: Why two methods that do the same?

packages/langium/src/generator/template-node.ts
```diff
@@ -0,0 +1,290 @@
...
+    }
+}
+
+export interface ForEachOptions<T> {
+    prefix?: (element: T, index: number, isLast: boolean) => Generated|undefined;
+    suffix?: (element: T, index: number, isLast: boolean) => Generated|undefined;
+    appendNewLineIfNotEmpty?: true;
+}
+
+export function forEach<T>(
```
spoenemann, line 265: This name is ambiguous in the global package namespace.

A possible alternative would be to wrap some generator-related functions in a `Generator` namespace in the main `index.ts`.

packages/langium/src/index.ts
```diff
@@ -7,8 +7,8 @@
 export * from './default-module';
 export * from './dependency-injection';
 export * from './generator/generator-node';
-export * from './generator/node-processor';
-export * from './generator/template-string';
+export { expandToNode, forEach } from './generator/template-node';
+export { expandToString, expandToStringWithNL } from './generator/template-string';
```
spoenemann, line 11: Why not export the whole file contents? Do they contain stuff that we use only internally?

spoenemann (commented):
Great work! The statemachine example is superb to show off this new feature.

examples/statemachine/test/generator.test.ts
sailingKieler: Right, I'll add one.

packages/langium/src/generator/generator-node.ts
sailingKieler: It's quiet beneficial, enables conditional appending in chained method calls, like

image

packages/langium/src/generator/generator-node.ts
sailingKieler: It's syntactic sugar. I wasn't sure which syntax to propose, I actually prefer the longer one, but that's probably my personal taste.
Use cases are
 * `node.append(member.name).append(': {').appendNewLine()...`
 * ```
   return toNode`
        #include <iostream>
        #include <map>
        #include <string>

        class ${ctx.statemachine.name};

        ${generateStateClass(ctx)}

        ${generateStatemachineClass(ctx)}

        ${forEachNL(ctx.statemachine.states, state => generateStateDeclaration(ctx, state))}
        ${forEachNL(ctx.statemachine.states, state => generateStateDefinition(ctx, state))}

        typedef void (${ctx.statemachine.name}::*Event)();

        ${generateMain(ctx)}
   `.appendNewline();        // one way of enforcing a terminating line break.
   ```

packages/langium/src/index.ts
sailingKieler: Changed it export everything from `template-node.ts` (restriction was not necessary there).

`template-string.ts` contains definitions that are used internally in other modules, but shall not be published globally, at least not now.

packages/langium/src/generator/generator-node.ts
sailingKieler: Removed duplicate methods, will add js doc now.

examples/statemachine/test/generator.test.ts
sailingKieler: Added!

packages/langium/src/generator/generator-node.ts
```diff
@@ -4,31 +4,75 @@
...
+}
+
+export function toString(object: unknown): string {
+    if (isGeneratorNode(object))
+        return processGeneratorNode(object);
+    else
+        return String(object);
+}
 
 export class CompositeGeneratorNode {
```
spoenemann, line 31: Please add comments to the methods here to explain their use cases.

packages/langium/src/generator/generator-node.ts
```diff
@@ -4,31 +4,75 @@
...
+
+    appendNewLineIf(condition: boolean): this {
+        return condition ? this.append(NL) : this;
+    }
+
+    appendNewLineIfNotEmpty(): this {
+        return this.append(NLEmpty);
+    }
+
+    appendNewLineIfNotEmptyIf(condition: boolean): this {
```
spoenemann, line 70: Maybe `appendNewLineIfNotEmptyAnd`?

spoenemann (approved):
👍

packages/langium/src/generator/generator-node.ts
sailingKieler: Done!

Issue #747, opened 2022-11-01
Alternative including cross reference and string is recognized as reference to string


luan-xiaokun:
An alternative including a cross reference to `X` and strings sometimes can be recognized as a reference to `X | string`

A minimal example is as follows.
```
grammar Debug

entry Model: (items+=(A | B))*;

interface B {
    b: @A | string;
}

A infers A: 'A' a=ID;

B returns B: 'B' b=([A:ID] | 'C');

hidden terminal WS: /\s+/;
terminal ID: /[_a-zA-Z][\w_]*/;
``` 
Langium-vscode gives a diagnostic.
```
The assigned type 'Reference<'C' | A>' is not compatible with the declared property 'b' of type 'Reference<A> | string'.
''C' | A' is not expected.
```
The type of `[A:ID]|'C'` is recognized as `Reference<'C'|A>` instead of `Reference<A>|'C'`. 

Adding an action to the rule `B` resolves this error: `B returns B: {B} 'B' b=([A:ID] | 'C');`.

If the rule `B` is written using `infers` instead of `returns`, there is no diagnostic, but in the generated ast, type of `b` is still incorrect. Also, adding an action in such case does not resolve the error.

spoenemann:
I propose to add a validation to forbid such a property type (alternative of a reference and something else). I'd simply use two separate properties in this case.

luan-xiaokun:
Hey @spoenemann, using two separate properties looks like a very reasonable solution. I agree that such kind of property type should not be used. But current diagnostic message indicates that the type of `[A:ID] | 'C'` is not correctly inferred without using an action, so I think that there may be something not working as we intended.

msujew:
@luan-xiaokun yes, we are planning to add a validation to this. We do have some validations already planned, see [here](https://github.com/langium/langium/labels/validation).

dhuebner:
@spoenemann 
> a validation to forbid such a property type (alternative of a reference and something else)

You mean validate a property inside a defined `interface`?  Or also property assignments like ` b=([A:ID] | 'C')` ?

msujew:
@dhuebner I believe both. This kind of union type (i.e. reference | other type) might break quite a few things in Langium.

dhuebner:
Implementation is in `https://github.com/langium/langium/tree/dh/validateAssigmentType` branch and is on top of changes from the Types refactoring PR: https://github.com/langium/langium/pull/819

dhuebner:
Fixed with  #826

[...]

PR #826, opened 2022-12-14
Adds validation to ensure cross-ref and other types are not mixed

dhuebner:
Prevents problems described in #747

### DIFF
--- a/packages/langium/src/grammar/validation/types-validator.ts
+++ b/packages/langium/src/grammar/validation/types-validator.ts
@@ -4,14 +4,14 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import * as ast from '../generated/ast';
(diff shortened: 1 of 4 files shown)


spoenemann (approved):
Looks good! 👍

dhuebner:
@spoenemann 
I've mixed `AssignableAlternatives` with `AssignableTerminal` inside the `ParenthesizedAssignableElement` rule, that why I thought it will be flat. 
Now changed `streamContents` to `streamAllContents` and  added an additional test case for nested elements: `prop =  (INT | (STRING | [Foo:ID]))`

giacomociti:
I think the [docs ](https://langium.org/docs/reference/semantic-model/#assignments) should be amended:

> ```Y: crossValue=[X:ID] alt=(INT | X | [X:ID]);```

PR #830, opened 2022-12-14
Move custom domainmodel naming to new service

### LINKED ISSUE #236: Domainmodel example: refactor qualified name provider
The current implementation of the domainmodel example extends the `NameProvider` with an additional method to handle qualified names, which leads to type casting of the service in other classes (see https://github.com/langium/langium/pull/229#discussion_r712983533). A simple solution would be to move that method to a new custom service `QualifiedNameProvider`. But we should consider whether generalizing this case makes sense. Xtext does this with a `QualifiedName` type; I'm not sure whether we want to go down that path.

spoenemann:
Closes #236.


msujew (approved):
LGTM 👍

PR #833, opened 2022-12-14
Sort diagnostics found in grammar

spoenemann:
When multiple diagnostics are found while processing a grammar file, we should print them with ascending line number.


msujew (approved):
LGTM 👍

PR #834, opened 2022-12-15
Fixed some small issues

dhuebner:
- Validation throws an error with a broken model: validator.ts
- Report undefined node when about to stream it contents: ast-utils.ts
- Allow recomputing `_ref` when a linkage error was created in a phase before ComputedScopes: linker.ts

The problem in linker.ts occurs when e.g. semanticHighlighting asks for a reference, but the document is still parsing or indexing and can not link to a local reference. Warning is logged: `Attempted reference resolution before document reached ComputedScopes state`


spoenemann (approved):
:+1:

Issue #589, opened 2022-07-08
Misleading error message in empty entry rule


dhuebner:
Example: 
```
grammar Blank

entry Blank:  // <- error at Blank "The entry rule cannot be a data type rule."
    
;
```

I think we should not produce this error in an empty `entry` rule it is misleading and sometimes annoying when writing new grammars. 

Often it is clear that the document is currently broken or incomplete because user is currently typing. I think we should try to not overload the editor with unrelated errors when it's possible.

PR #835, opened 2022-12-15
Don't create empty AstNodes

dhuebner:
Prevents creating AstNodes with an empty Cst content.

Also improves on `isDataType` check. The implementation will not identify a ParserRule with no `definition` or with a unresolved RuleCall inside `definition` as a DataType rule.

### DIFF
--- a/packages/langium/src/grammar/internal-grammar-util.ts
+++ b/packages/langium/src/grammar/internal-grammar-util.ts
@@ -40,8 +40,12 @@ function isDataTypeRuleInternal(rule: ast.ParserRule, visited: Set<ast.ParserRul
     }
     visited.add(rule);
     for (const node of streamAllContents(rule)) {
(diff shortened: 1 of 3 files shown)


dhuebner:
This change will also fix #589

packages/langium/src/grammar/internal-grammar-util.ts
```diff
@@ -50,7 +54,7 @@ function isDataTypeRuleInternal(rule: ast.ParserRule, visited: Set<ast.ParserRul
             return false;
         }
     }
-    return true;
+    return !!rule.definition;
```
spoenemann, line 57: In general I prefer writing `Boolean(rule.definition` over double negation because it expresses more directly what it actually does.

dhuebner: Changed

Issue #700, opened 2022-09-27
Properties of interface definitions never being assigned are ignored during AST generation


sailingKieler:
The generation of interface properties apparently wants to be smart and ignores those properties that are never assigned within a parser rule.
In rare cases it's required to add such property definitions explicitly to have the property accessible from a super type defined as a union type definition.

Thus, please drop the filtering.

PR #837, opened 2022-12-15
Respect inheritance for actions and unassigned rulecalls

msujew:
Fixes an issue I noticed while investigating https://github.com/langium/langium/issues/775. When calling an unassigned action, we assume that the previous existing type becomes the super type of the new type. This wasn't always the case. This change addresses this.


pluralia:
Cool, it looks like you close #700!

pluralia (approved):
I approve changes: maximum call stack size comes anyway, but problem with unassigned rule calls and actions is fixed.

PR #840, opened 2022-12-15
Add type validation with respect to the hierarchy

### LINKED ISSUE #823: Property types validation doesn't pay attention to types hierarchy
Langium version: 0.5.0
Package name: langium

## Steps To Reproduce
```
Named: name = ID;
Expression: NamedRef;
NamedRef: ref=[Named];

interface QualifiedRef extends NamedRef {
    qualifier: Expression
}

QualifiedRef returns QualifiedRef:
    qualifier=NamedRef ref=[Named];

QualifiedRefWithAction infers Expression:
    NamedRef {QualifiedRef.qualifier=current} '.' ref=[Named];
```
 
## The current behavior
The error `The assigned type 'NamedRef' is not compatible with the declared property 'qualifier' of type 'Expression': 'NamedRef' is not expected.` rises for `QualifiedRef` and `QualifiedRefWithAction`.

## The expected behavior
No errors since `NamedRef` is `Expression`.

pluralia:
Closes #823.
Adds `typeToAlias` that maps a type name to its hierarchy root -- a type that doesn't inherit anything. Validation uses this structure to map all types to their roots and compare roots instead of initial types.

### DIFF
--- a/packages/langium/test/grammar/type-system/type-validator.test.ts
+++ b/packages/langium/test/grammar/type-system/type-validator.test.ts
@@ -450,4 +450,34 @@ describe('Property types validation takes in account types hierarchy', () => {
         expect(validation.diagnostics).toStrictEqual([]);
     });
 
+    test('Handling type aliases shouldn\'t disable other validations.', async () => {
+        const validation = await validate(`
+            interface Y {
+                y: Z1
+            }
+            
+            interface Z {
+                name: string
+            }
+            
+            interface Z1 extends Z {
+                z: number
+            }
+            
+            interface Z2 extends Z {
+                a: string
+            }
+            
+            Y returns Y: y=Z2;
+            
+            Z1 returns Z1: z=NUMBER name=ID;
+            Z2 returns Z2: a=ID name=ID;
+
+            terminal ID: /[_a-zA-Z][\\w_]*/;
+            terminal NUMBER returns number: /[0-9]+(\\.[0-9]*)?/;
+        `);
+
+        expect(validation.diagnostics.filter(d => d.severity === DiagnosticSeverity.Error)).toHaveLength(1);
+    });
+
 });
--- a/packages/langium/test/grammar/type-system/type-validator.test.ts
+++ b/packages/langium/test/grammar/type-system/type-validator.test.ts
@@ -394,20 +394,60 @@ describe('Property types validation takes in account types hierarchy', () => {
         expect(validation.diagnostics).toStrictEqual([]);
     });
 
-    // todo: a type can have 2 and more parents
     // here `X` can be `string` or `XY` and `Y` cab be `number` or `XY
-    // test('Usage of child type with some parents is validated correctly.', async () => {
-    //     const validation = await validate(`
-    //         X returns string: 'X';
-    //         Y returns number: NUMBER;
-    //         QualifiedRef: name=NUMBER;
-    //         XY returns XY: X | Y | QualifiedRef;
-    //         terminal NUMBER returns number: /[0-9]+(\\.[0-9]+)?/;
+    test('Usage of child type with some parents is validated correctly.', async () => {
+        const validation = await validate(`
+            X returns string: 'X';
+            Y returns number: NUMBER;
+            QualifiedRef: name=NUMBER;
+            XY returns XY: X | Y | QualifiedRef;
+            terminal NUMBER returns number: /[0-9]+(\\.[0-9]+)?/;
 
-    //         type XY = string | number | QualifiedRef;
-    //     `);
+            type XY = string | number | QualifiedRef;
+        `);
 
-    //     expect(validation.diagnostics).toStrictEqual([]);
-    // });
+        expect(validation.diagnostics).toStrictEqual([]);
+    });
+
+    test('Keywords are subtypes of strings.', async () => {
+        const validation = await validate(`
+            interface BinaryExpression {
+                left: Expression
+                right: Expression
+                operator: string
+            }
+            
+            Expression:
+                PrimaryExpression ({BinaryExpression.left=current} operator=('+' | '-') right=PrimaryExpression)*;
+            
+            PrimaryExpression infers Expression:
+                {infer NumberLiteral} value=NUMBER;
+
+            terminal NUMBER returns number: /[0-9]+(\\.[0-9]*)?/;
+        `);
+
+        expect(validation.diagnostics).toStrictEqual([]);
+    });
+
+    test('Type aliases replace types correctly.', async () => {
+        const validation = await validate(`
+            X: name=ID;
+            AliasX: X;
+            
+            interface YType {
+                prop: AliasX
+            }
+            Y returns YType: prop=X;
+            
+            interface ZType {
+                prop: X
+            }
+            Z returns ZType: prop=AliasX;
+
+            terminal ID: /[_a-zA-Z][\\w_]*/;
+        `);
+
+        expect(validation.diagnostics).toStrictEqual([]);
+    });
 
 });
--- a/packages/langium/test/grammar/type-system/type-validator.test.ts
+++ b/packages/langium/test/grammar/type-system/type-validator.test.ts
@@ -394,4 +394,20 @@ describe('Property types validation takes in account types hierarchy', () => {
         expect(validation.diagnostics).toStrictEqual([]);
     });
 
+    // todo: a type can have 2 and more parents
+    // here `X` can be `string` or `XY` and `Y` cab be `number` or `XY
+    // test('Usage of child type with some parents is validated correctly.', async () => {
+    //     const validation = await validate(`
+    //         X returns string: 'X';
+    //         Y returns number: NUMBER;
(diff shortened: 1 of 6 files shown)


pluralia:
The current algorithm doesn't pay attention to the fact that a type can have 2+ roots -- this situation may not work perfectly (I added a commented test to demonstrate this). I'd suggest putting this problem in a new issue.

msujew (commented):
I noticed that the approach to look for a common supertype between the assigned and expected type doesn't work as expected. In the following case, the validation will not yield an error, while the types are actually not assignable:

```
interface Y {
    y: Z1
}

interface Z {
    name: string
}

interface Z1 extends Z {
    z: number
}

interface Z2 extends Z {
    a: string
}

Y returns Y: y=Z2;

Z1 returns Z1: z=INT name=ID;
Z2 returns Z2: a=ID name=ID;
```

Instead, the validation should only look whether the expected type is either the actual type or one of its super types.

pluralia:
> Instead, the validation should only look whether the expected type is either the actual type or one of its super types.

Sounds reasonable, but in general case expected and action type is a types alternative and not just one type, and we don't know how types of these 2 type alternatives are coupled. Example:
```
type X = A | B;
RuleX returns X: AA | BB;

AA returns AA: aa=ID;
BB returns BB: bb=ID;

interface A {}
interface B {}
interface AA extends A {
    aa: string
}
interface BB extends B {
    bb: string
}
```
Here `RuleX` uses a declared type `X` correctly: both type alternatives are children of type alternatives that `X` declares. However, when we check consistency of `RuleX` and `X`, we don't know how we should compare alternatives: `A` wtih `AA` or `A` with `BB`?

The presented solution solves this problem: type alternatives of and inferred, and declared type will be normalized. As a result, and `A`, and `AA` will be transformed to `A`, and comparison becomes trivial. However, you showed a problem of this solution.

I have another idea. We should collect all super types and aliases per type and produce all possible type alternatives to find the match. We can attach them to the declared type. For the example above to the `X`'s type alternative `A | B`, `A | BB`, `AA | B`, `AA | BB` would be attached. Attaching will allow as reuse calculation of all possible type alternatives for validation of many parser rules.

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -277,83 +277,207 @@ describe('validate properties duplication in types hierarchy', () => {
...
+            Y returns Y: y=Z2;
+            
+            Z1 returns Z1: z=NUMBER name=ID;
+            Z2 returns Z2: a=ID name=ID;
+
+            terminal ID: /[_a-zA-Z][\\w_]*/;
+            terminal NUMBER returns number: /[0-9]+(\\.[0-9]*)?/;
+        `);
+
+        expect(validation.diagnostics.filter(d => d.severity === DiagnosticSeverity.Error)).toHaveLength(1);
```
msujew, line 480:
```suggestion
        const assignment = streamAllContents(validation.document.parseResult.value).filter(isAssignment).toArray()[0];
        expectError(validation, "The assigned type 'Z2' is not compatible with the declared property 'y' of type 'Z1':  'Z2' is not expected.", {
            node: assignment,
            property: {
                name: 'feature'
            }
        });
```

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -277,83 +277,207 @@ describe('validate properties duplication in types hierarchy', () => {
...
+            }
+            Z returns ZType: prop=AliasX;
+
+            terminal ID: /[_a-zA-Z][\\w_]*/;
+        `);
+
+        expect(validation.diagnostics).toStrictEqual([]);
+    });
+
+    test('Handling type aliases shouldn\'t disable other validations.', async () => {
```
msujew, line 453:
```suggestion
    test('Should create error on assignments with incorrect hierarchy', async () => {
```

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -277,83 +277,207 @@ describe('validate properties duplication in types hierarchy', () => {
...
+            
+            interface ZType {
+                prop: X
+            }
+            Z returns ZType: prop=AliasX;
+
+            terminal ID: /[_a-zA-Z][\\w_]*/;
+        `);
+
+        expect(validation.diagnostics).toStrictEqual([]);
```
msujew, line 450:
```suggestion
        expectNoIssues(validation);
```

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -277,83 +277,207 @@ describe('validate properties duplication in types hierarchy', () => {
...
+            Expression:
+                PrimaryExpression ({BinaryExpression.left=current} operator=('+' | '-') right=PrimaryExpression)*;
+            
+            PrimaryExpression infers Expression:
+                {infer NumberLiteral} value=NUMBER;
+
+            terminal NUMBER returns number: /[0-9]+(\\.[0-9]*)?/;
+        `);
+
+        expect(validation.diagnostics).toStrictEqual([]);
```
msujew, line 429:
```suggestion
        expectNoIssues(validation);
```

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -277,83 +277,207 @@ describe('validate properties duplication in types hierarchy', () => {
...
+            X returns string: 'X';
+            Y returns number: NUMBER;
+            QualifiedRef: name=NUMBER;
+            XY returns XY: X | Y | QualifiedRef;
+            terminal NUMBER returns number: /[0-9]+(\\.[0-9]+)?/;
+
+            type XY = string | number | QualifiedRef;
+        `);
+
+        expect(validation.diagnostics).toStrictEqual([]);
```
msujew, line 409: expectNoIssues(validation);

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -277,83 +277,207 @@ describe('validate properties duplication in types hierarchy', () => {
...
+            interface NamedRef {
+                ref: @Named
+            }
+            
+            interface QualifiedRef extends NamedRef {
+                qualifier: Expression
+            }        
+        `);
+
+        expect(validation.diagnostics).toStrictEqual([]);
```
msujew, line 394: expectNoIssues(validation);

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -277,83 +277,207 @@ describe('validate properties duplication in types hierarchy', () => {
...
+    test('Type aliases are validated correctly.', async () => {
+        const validation = await validate(`
+            interface TypeA {
+                name: string
+            }
+            A returns TypeA: name=QualifiedName;
+            QualifiedName returns string: 'QualifiedName';
+        `);
+
+        expect(validation.diagnostics).toStrictEqual([]);
```
msujew, line 373: expectNoIssues(validation);

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -277,83 +277,207 @@ describe('validate properties duplication in types hierarchy', () => {
...
+            node: attribute!,
+            property: { name: 'typeAlternatives' }
         });
     });
+});
+
+// https://github.com/langium/langium/issues/823
+describe('Property types validation takes in account types hierarchy', () => {
+
+    test('Type aliases are validated correctly.', async () => {
```
msujew, line 364:
```suggestion
    test('Type aliases can be assigned to primitive types.', async () => {
```

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -277,83 +277,207 @@ describe('validate properties duplication in types hierarchy', () => {
...
+                name: string
+            }
+            A returns TypeA: name=QualifiedName;
+            QualifiedName returns string: 'QualifiedName';
+        `);
+
+        expect(validation.diagnostics).toStrictEqual([]);
+    });
+
+    test('Usage of child type is validated correctly.', async () => {
```
msujew, line 376:
```suggestion
    test('Child type can be assigned correctly.', async () => {
```

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -277,83 +277,207 @@ describe('validate properties duplication in types hierarchy', () => {
...
+            PrimaryExpression infers Expression:
+                {infer NumberLiteral} value=NUMBER;
+
+            terminal NUMBER returns number: /[0-9]+(\\.[0-9]*)?/;
+        `);
+
+        expect(validation.diagnostics).toStrictEqual([]);
+    });
+
+    test('Type aliases replace types correctly.', async () => {
```
msujew, line 432:
```suggestion
    test('Type aliases can be assigned correctly for types', async () => {
```

msujew (approved):
Works good for me, just some comments on the newly added test code.

Issue #431, opened 2022-03-01
Ensure that generator-langium and langium:generator are in sync


danieldietrich:
The Langium extension generator `generator-langium` is scaffolding new Langium projects. In particular, it generates the Langium configuration read by the `langium generate` CLI command.

`generator-langium` and `langium` are different packages. As a Langium maintainer, I want to ensure that both packages don't break their common contract (e.g. the Langium configuration format).

A solution could be that we add an additional CI task in the form of a GitHub Action that runs `generator-langium` as an integration test. More specifically, the Action could install Yeoman and `npm link` the package `generator-langium` in order to be able to run `yo langium`. Within the resulting hello world project, we could run the `langium:generate` script and check the generated output.

* A first smoke test could be to check if the `langium:generate` script runs with exit code 0.
* A more sophisticated test could check if generated contents match expected contents (via snapshots?)

spoenemann:
A big challenge here is that the generated npm package pulls langium and langium-cli from the npm registry, but its contents may be already using new API from the next version (main branch). I tested the current state and it fails because it uses the old version 0.2.0. If you change that to use the local version from the repository, everything works fine. So an integration test needs to ensure that the local version is used instead of a published version.

dhuebner:
I think, the first step is to add an automatic test that at least ensures that the generator-langium works with the current codebase.
I will take a look

dhuebner:
@danieldietrich @spoenemann 
I've added a yeoman-generator test that will fail if for example generated source files can not be compiled. See #843

spoenemann:
We have compile tests for generator-langium now.

PR #843, opened 2022-12-16
First yoman generator-langium test #431

dhuebner:
Add tests to make sure generator-langium produces compilable result

### DIFF
--- a/packages/generator-langium/package.json
+++ b/packages/generator-langium/package.json
@@ -27,6 +27,7 @@
     "watch": "tsc --watch",
     "lint": "eslint src --ext .ts",
     "run": "yo langium",
+    "test": "vitest run",
     "debug": "npx --node-arg=--inspect yo langium",
     "publish:next": "npm --no-git-tag-version version \"$(semver $npm_package_version -i minor)-next.$(git rev-parse --short HEAD)\" && npm publish --tag next",
     "publish:latest": "npm publish --tag latest"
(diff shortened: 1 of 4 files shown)


packages/generator-langium/package.json
```diff
@@ -38,7 +38,9 @@
   },
   "devDependencies": {
     "@types/lodash": "^4.14.170",
-    "@types/yeoman-generator": "^5.0.0"
+    "@types/yeoman-generator": "^5.0.0",
+    "@types/yeoman-test": "^4.0.3",
+    "yeoman-test": "^7.0.0"
```
spoenemann, line 43: Please add a `"test": "vitest run"` script so we can run this directly from the package folder

spoenemann (approved):
Wow, yeoman-test has quite heavy dependencies 😅 

It seems this also covers the case where compiling the generated package fails:

> Error: Command failed with exit code 1: npm run build

spoenemann:
@dhuebner I found my old comment again: https://github.com/langium/langium/issues/431#issuecomment-1057772654

Does the test already use the workspace version of langium?

dhuebner:
@spoenemann 
> Wow, yeoman-test has quite heavy dependencies 😅

Yes, I also was very surprised. It's also a long running one compared to other tests :( 

> Does the test already use the workspace version of langium?

I can only confirm that the test failed before your commit here: 
https://github.com/langium/langium/commit/d2760bb3c081c4c5d34132ad38095cda44b64e40
But I will look more deeply into what happens internally. A good start would be to change something locally and see how the test does.

dhuebner:
@spoenemann 
It seems like the test runs against the npm registry. Generator just spawns `npm install` in install phase... 
It might be possible to tweak the test using `yeoman-environment` API so it uses local langium packages, but this needs more investigation and time. I will merge the current state as it tests other important parts of the generator already.

Issue #554, opened 2022-06-20
Collecting AST nodes within the TypeGraph for minimizing duplicated symbol lookup


Lotes:
We could collect all type and property information in a graph TOGETHER WITH the AST nodes which are contributing to that type or property. Types and properties could be nodes in that graph and edges are relations between the nodes, like `node1 IS_SUBTYPE_OF node2` or `node1 HAS_PROPERTY node2`. Both, nodes and edges are also collecting the AST nodes as a list of witnesses for that edge or node.

Why should we do this?

1. removing code duplication: We have code for computing the types and we have code to find all references of a symbol. When one changes, the other can be accidentally be forgotten. The code is not identical like 1:1, but the point is that we traverse the AST at least one time too much, in my eyes... Example: We add super types but forget to traverse in one of 4 lookup codes. Which can be a bug later, because a user assumes the correct behavior.

2. Code becomes shorter. Example: Look at this [PR - Finding references](https://github.com/langium/langium/pull/528/files)... imagine we would have this graph, 200 LOC would be instead something very shorter (see the code sketch here -->).
```typescript
return graph.nodes.getByName(typeName) //every type has a node
           .getIncomingEdges()  //with a bunch of incoming or defining edges
           .filter(e => e.type == 'typedef') //filter out the right edges, if needed
           .flatMap(e => e.astNodes) //get the ast nodes that define the type
```

I do not know what edges types and node types really make sense here. But I also heard we already have some kind of graph.

Issue #534, opened 2022-06-09
Type with a property should become an interface


spoenemann:
The grammar rule
```
X: id=ID ({infer Y} 'a' | {infer Z} 'b');
```
currently leads to the types
```typescript
export type X = Y | Z;

export interface Y extends AstNode {
    id: string
}

export interface Z extends AstNode {
    id: string
}
```
but this ignores the fact that `id` was defined on the level of type `X`, not its subtypes. I think we should change this to
```typescript
export interface X extends AstNode {
    id: string
}

export interface Y extends X {
}

export interface Z extends X {
}
```

See also discussion in #512.

PR #845, opened 2022-12-19
Internal type split

msujew:
This change splits the type system into a "plain" (string-based) type system and a concrete (reference-based) type system. Splitting the types allows for easier handling of purely generated information (i.e. container types, union/interface hierarchies and type names). It also has an impact on how we deal with property types and union types in general, which are now way more flexible.

Aims to resolve multiple issues:
* Closes https://github.com/langium/langium/issues/511 (The grammar has been slightly modified)
* Closes https://github.com/langium/langium/issues/476
* Closes https://github.com/langium/langium/issues/752
* Closes https://github.com/langium/langium/issues/534
* Closes https://github.com/langium/langium/issues/868

### DIFF
--- a/packages/langium/src/grammar/type-system/ast-collector.ts
+++ b/packages/langium/src/grammar/type-system/ast-collector.ts
@@ -6,52 +6,80 @@
 
 import { Grammar } from '../generated/ast';
 import { LangiumDocuments } from '../../workspace/documents';
-import { addSubTypes, sortInterfacesTopologically } from './types-util';
-import { AstTypes, InterfaceType, isInterfaceType, isUnionType, TypeOption, UnionType } from './type-collector/types';
-import { collectTypeResources } from './type-collector/all-types';
-import { isPrimitiveType } from '../internal-grammar-util';
+import { collectTypeHierarchy, sortInterfacesTopologically } from './types-util';
+import { AstTypes, isArrayType, isInterfaceType, isPrimitiveType, isPropertyUnion, isStringType, isUnionType, isValueType, PropertyType, TypeOption, UnionType } from './type-collector/types';
+import { collectTypeResources, ValidationAstTypes } from './type-collector/all-types';
+import { PlainAstTypes, PlainInterface, plainToTypes, PlainUnion } from './type-collector/plain-types';
 
 /**
  * Collects all types for the generated AST. The types collector entry point.
- * @param documents Documents to resolve imports that were used in the given grammars.
- * @param grammars Grammars for which it's necessary to build an AST.
+ *
+ * @param grammars All grammars involved in the type generation process
+ * @param documents Additional documents so that imports can be resolved as necessary
  */
 export function collectAst(grammars: Grammar | Grammar[], documents?: LangiumDocuments): AstTypes {
     const { inferred, declared } = collectTypeResources(grammars, documents);
-
-    const astTypes = {
-        interfaces: sortInterfacesTopologically(mergeAndRemoveDuplicates<InterfaceType>(inferred.interfaces, declared.interfaces)),
-        unions: mergeAndRemoveDuplicates<UnionType>(inferred.unions, declared.unions),
-    };
-
-    specifyAstNodeProperties(astTypes);
-    return astTypes;
+    return createAstTypes(inferred, declared);
 }
 
-function mergeAndRemoveDuplicates<T extends { name: string }>(inferred: T[], declared: T[]): T[] {
-    return Array.from(inferred.concat(declared)
+/**
+ * Collects all types used during the validation process.
+ * The validation process requires us to compare our inferred types with our declared types.
+ *
+ * @param grammars All grammars involved in the validation process
+ * @param documents Additional documents so that imports can be resolved as necessary
+ */
+export function collectValidationAst(grammars: Grammar | Grammar[], documents?: LangiumDocuments): ValidationAstTypes {
+    const { inferred, declared, astResources } = collectTypeResources(grammars, documents);
+
+    return {
+        astResources,
+        inferred: createAstTypes(declared, inferred),
+        declared: createAstTypes(inferred, declared)
+    };
+}
+
+export function createAstTypes(first: PlainAstTypes, second?: PlainAstTypes): AstTypes {
+    const astTypes: PlainAstTypes = {
+        interfaces: sortInterfacesTopologically(mergeAndRemoveDuplicates<PlainInterface>(...first.interfaces, ...second?.interfaces ?? [])),
+        unions: mergeAndRemoveDuplicates<PlainUnion>(...first.unions, ...second?.unions ?? []),
+    };
+
+    const finalTypes = plainToTypes(astTypes);
+    specifyAstNodeProperties(finalTypes);
+    return finalTypes;
+}
+
+/**
+ * Merges the lists of given elements into a single list and removes duplicates. Elements later in the lists get precedence over earlier elements.
+ *
+ * The distinction is performed over the `name` property of the element. The result is a name-sorted list of elements.
+ */
+function mergeAndRemoveDuplicates<T extends { name: string }>(...elements: T[]): T[] {
+    return Array.from(elements
         .reduce((acc, type) => { acc.set(type.name, type); return acc; }, new Map<string, T>())
         .values()).sort((a, b) => a.name.localeCompare(b.name));
 }
 
 export function specifyAstNodeProperties(astTypes: AstTypes) {
     const nameToType = filterInterfaceLikeTypes(astTypes);
-    addSubTypes(nameToType);
-    buildContainerTypes(nameToType);
-    buildTypeTypes(nameToType);
+    const array = Array.from(nameToType.values());
+    addSubTypes(array);
+    buildContainerTypes(array);
+    buildTypeNames(nameToType);
 }
 
-function buildTypeTypes(nameToType: Map<string, TypeOption>) {
+function buildTypeNames(nameToType: Map<string, TypeOption>) {
     const queue = Array.from(nameToType.values()).filter(e => e.subTypes.size === 0);
     const visited = new Set<TypeOption>();
     for (const type of queue) {
         visited.add(type);
-        type.typeTypes.add(type.name);
-        const superTypes = Array.from(type.realSuperTypes)
-            .map(superType => nameToType.get(superType))
+        type.typeNames.add(type.name);
+        const superTypes = Array.from(type.superTypes)
+            .map(superType => nameToType.get(superType.name))
             .filter(e => e !== undefined) as TypeOption[];
         for (const superType of superTypes) {
-            type.typeTypes.forEach(e => superType.typeTypes.add(e));
+            type.typeNames.forEach(e => superType.typeNames.add(e));
         }
         queue.push(...superTypes.filter(e => !visited.has(e)));
     }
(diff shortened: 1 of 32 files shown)


spoenemann:
The result looks mostly really good! But it doesn't properly solve #534: with the rule
```
entry X: id=ID ({infer Y} 'a' | {infer Z} 'b');
```
 I get the types
```
export type X = Y | Z;

export interface Y extends AstNode {
    readonly $type: 'Y';
}

export interface Z extends AstNode {
    readonly $type: 'Z';
}
```
so the `id` property is forgotten. Could you add a test for this?

msujew:
@spoenemann Thanks for the hint 👍 It should as expected now.

spoenemann:
How does this relate to #554?

pluralia:
The PR closes also #868.

msujew:
> How does this relate to https://github.com/langium/langium/issues/554?

It's closely related, but not 100% there yet. My code can be used as a basis to build the graph mentioned in #554.

msujew:
> After generating the code, the empty line at the end of these files is deleted

Unrelated to the changes in my PR. It might be due to some recent changes in the generator API.

spoenemann:
>> After generating the code, the empty line at the end of these files is deleted
>
> Unrelated to the changes in my PR. It might be due to some recent changes in the generator API.

It's my fault: I had to manually fix the generated files (https://github.com/langium/langium/pull/888#issuecomment-1408359809), then @msujew fixed the generator, but we didn't regenerate the files. The file ending is different because the text editor automatically changes it on save.

packages/langium/src/grammar/validation/types-validator.ts
```diff
@@ -233,120 +227,63 @@ function validateDeclaredAndInferredConsistency(typeInfo: InferredInfo & Declare
...
-        } else if (expectedPropertyType.array !== foundPropertyType.array || expectedPropertyType.reference !== foundPropertyType.reference) {
-            errorsInfo.push({ typeAsString, errorMessage: arrayReferenceError(foundPropertyType, expectedPropertyType) });
-        }
+function validateAlternativesConsistency(
+    inferred: PropertyType,
+    declared: PropertyType,
+    applyErrorToInferredTypes: (errorMessage: string) => void
+) {
+    if (!isTypeAssignable(inferred, declared)) {
+        applyErrorToInferredTypes(`Cannot assign type '${propertyTypeToString(inferred, 'DeclaredType')}' to '${propertyTypeToString(declared, 'DeclaredType')}'`);
```
montymxb, line 236: Shouldn't the first one be 'InferredType' instead of 'DeclaredType'?

packages/langium/test/grammar/type-system/inferred-types.test.ts
```diff
@@ -5,694 +5,331 @@
...
-        beforeAll(async () => {
-            const newTypes = await getTypes(grammar);
-            types.interfaces = newTypes.interfaces;
-            types.unions = newTypes.unions;
-        });
-        description.call(undefined, types);
-    });
-}
-
-describeTypes('inferred types of simple grammars', `
```
montymxb, line 31: I'm not sure about these testing rewrites here. Doing validation on types by checking formatted strings feels very brittle to me going forward. Can we still not get the types that are produced by the collector, before printing, and verify on those instead?

If we could have the existing test structure, verifying our internal representation, plus this extra checking on the generated output, that would also work well I think.

montymxb: To add as well, it's hard to follow if the integrity of the original tests is maintained with such a large refactor. I didn't get to look at it closely enough, but I would recommend a double check here by another colleague as well.

packages/langium/test/parser/langium-parser-builder.test.ts
```diff
@@ -437,25 +437,25 @@ describe('Parser calls value converter', () => {
...
     test('Should parse bool correctly', () => {
         expectValue('b true', true);
-        // this is the current 'boolean' behavior when a prop type can't be resolved to just a boolean
-        // either true/undefined, no false in this case
-        expectValue('b false', undefined);
-        // ...then no distinguishing between the bad parse case when the type is unclear
-        expectValue('b asdfg', undefined);
+        expectValue('b false', false);
+        // Any value that cannot be parsed correctly is automatically false
+        expectValue('b asdfg', false);
```
montymxb, line 442: This feels like a breaking change, especially if there's any upstream code that strictly relies on `undefined` to be produced.

packages/langium/test/parser/langium-parser-builder.test.ts
```diff
@@ -437,25 +437,25 @@ describe('Parser calls value converter', () => {
...
     });
 
     test('Should parse BigInt correctly', () => {
         expectValue('big 9007199254740991n', BigInt('9007199254740991'));
-        expectValue('big 9007199254740991', undefined);
-        expectValue('big 1.1', undefined);
-        expectValue('big -19458438592374', undefined);
+        // Any value that cannot be parsed correctly is automatically false
+        expectValue('big 9007199254740991', false);
```
montymxb, line 448: I would expect a numeric value that cannot be parsed to be `undefined`, rather than `false`. Is this fixed now by the design of the new implementation? If not, I would support reverting this.

montymxb (commented):
General suggestions & comments. In particular I'm curious if we can still retain some (or all) of the previous inferred type tests structure. and then there's the point about the subtle change from `undefined` to `false` for values when parsing fails.

packages/langium/test/parser/langium-parser-builder.test.ts
msujew: That doesn't make a lot of sense, since a type that contains `boolean` is guaranteed to be `false` as a fallback (even if it's not parsed). Langium never generates a property as `value?: boolean`. This is actually a bug fix rather than a new feature.

packages/langium/test/parser/langium-parser-builder.test.ts
msujew: It's a bug fix, see the explanation above.

packages/langium/test/parser/langium-parser-builder.test.ts
montymxb: I'm aware it was a bug from before, but it's possible it will break something. I should clarify, in the next release notes we should make this change clear; but beyond that the basis for booleans not having an undefined value -- where it should be false -- is clear.

packages/langium/src/grammar/type-system/ast-collector.ts
```diff
@@ -18,40 +18,58 @@ import { isPrimitiveType } from '../internal-grammar-util';
...
 
-    specifyAstNodeProperties(astTypes);
-    return astTypes;
+export function createAstTypes(first: PlainAstTypes, second?: PlainAstTypes): AstTypes {
+    const astTypes: PlainAstTypes = {
+        interfaces: mergeAndRemoveDuplicates<PlainInterface>(...first.interfaces, ...second?.interfaces ?? []),
+        unions: mergeAndRemoveDuplicates<PlainUnion>(...first.unions, ...second?.unions ?? []),
+    };
+
+    sortInterfacesTopologically(astTypes.interfaces);
```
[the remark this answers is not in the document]
msujew: Right, thanks for the headsup! It's called correctly on `main` but I broke something during my latest rebase.

packages/langium/src/grammar/type-system/ast-collector.ts
```diff
@@ -96,19 +96,44 @@ function filterInterfaceLikeTypes({ interfaces, unions }: AstTypes): Map<string,
     return nameToType;
 }
 
+function isDataType(property: PropertyType, visited: Set<PropertyType>): boolean {
```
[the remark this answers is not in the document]
msujew: I personally like to keep the argument mandatory, since I usually forget to pass the arguments in the recursive call :)

packages/langium/test/parser/langium-parser-builder.test.ts
montymxb: The boolean case makes sense above, but here less so. It would be a topic I would like to discuss this in a later dev meeting perhaps, and we can move forward with this as is. Just want to be sure that we don't mislead or confuse end-users with how `false` is now a possible value for these property, even when the type is not a boolean.

packages/langium/src/grammar/type-system/ast-collector.ts
```diff
@@ -96,19 +96,44 @@ function filterInterfaceLikeTypes({ interfaces, unions }: AstTypes): Map<string,
...
+            return isDataType(value.type, visited);
+        } else {
+            return false;
+        }
+    } else {
+        return isPrimitiveType(property) || isStringType(property);
+    }
+}
+
+function addSubTypes(types: TypeOption[]) {
```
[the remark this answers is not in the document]
msujew: There is no further transformation happening at this point.

packages/langium/test/parser/langium-parser-builder.test.ts
msujew: Well, the type is a boolean. The generated type is `value: boolean | number | bigint | string | Date`.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -659,14 +661,14 @@ export class LangiumGrammarValidator {
...
-    checkInterfacePropertyTypes(interfaceDecl: ast.Interface, accept: ValidationAcceptor): void {
-        interfaceDecl.attributes.filter(attr => attr.typeAlternatives.length > 1).forEach(attr => {
-            const typeKind = (type: ast.AtomType) => type.isRef ? 'ref' : 'other';
-            const firstKind = typeKind(attr.typeAlternatives[0]);
-            if (attr.typeAlternatives.slice(1).some(type => typeKind(type) !== firstKind)) {
-                accept('error', this.createMixedTypeError(attr.name), { node: attr, property: 'typeAlternatives' });
-            }
-        });
+    checkInterfacePropertyTypes(): void {
+        // interfaceDecl.attributes.filter(attr => attr.typeAlternatives.length > 1).forEach(attr => {
```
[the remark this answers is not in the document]
msujew: I forgot to implement it after temporarily removing the validation 👍

packages/langium/src/grammar/validation/types-validator.ts
msujew: This is only for display. The other option would be `AstNode` which is for generation purposes only.

packages/langium/test/grammar/type-system/inferred-types.test.ts
msujew: The existing test structure was very tightly coupled with the exact layout of our old types. To prevent another test refactoring, I recommend that we only perform testing of types on the generated strings, as they contain all information that we get from the object model anyway.

packages/langium/src/grammar/validation/types-validator.ts
```diff
@@ -199,29 +190,32 @@ function validateDeclaredAndInferredConsistency(typeInfo: InferredInfo & Declare
             accept('error', errorMessage, { node, property: ast.isAssignment(node) || ast.isAction(node) ? 'feature' : 'name' })
         );
 
+    // todo add actions
```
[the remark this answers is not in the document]
msujew: I didn't touch any of that code. So I assume so. Maybe @pluralia knows a bit more.

packages/langium/test/parser/langium-parser-builder.test.ts
montymxb: Ahhh Right, that's a good point, in this example booleans are part of the prop, and so it would fallback...

So, in cases where `boolean` is not part of the type we should still get `undefined` then as before? I was looking over our cases and don't see any specific tests for checking that we get such a value when the parser fails (but no boolean type for the value). If that's how it works than I'm all good 👍 .

msujew: Right, I'll add a test for that :)

packages/langium/test/grammar/type-system/inferred-types.test.ts
montymxb: Okay. I'll take a closer look at the tests to verify things look good in the restructure.

montymxb (approved):
Tests look good after a double check. There was one that was slightly changed (inferred types using chained actions), where G, H, and I are changed around. But this looks okay based on testing locally. Also checked this against some other Langium projects as well, and nothing appears out of the ordinary from what I can tell.

PR #846, opened 2022-12-20
Add webworker to arithmetics and domainmodel example

emilkrebs:
- Added a `main-browser.ts` to the arithmetics and domain model example
- Added Monarch generation for arithmetic and domain model


spoenemann (approved):
:+1:

PR #847, opened 2022-12-20
Attempt to test against local langium packages

dhuebner:
Should run the tests against local langium packages instead of pulling them from registry

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ node_modules/
 *.vsix
 *.tsbuildinfo
 examples/*/example/generated
+examples/hello-world
--- a/packages/generator-langium/src/index.ts
+++ b/packages/generator-langium/src/index.ts
@@ -135,7 +135,9 @@ class LangiumGenerator extends Generator {
         const extensionPath = this._extensionPath();
 
         const opts = { cwd: extensionPath };
-        this.spawnCommandSync('npm', ['install'], opts);
+        if(!this.args.includes('skip-install')) {
+            this.spawnCommandSync('npm', ['install'], opts);
+        }
         this.spawnCommandSync('npm', ['run', 'langium:generate'], opts);
         this.spawnCommandSync('npm', ['run', 'build'], opts);
     }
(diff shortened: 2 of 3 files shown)


msujew (commented):
Instead of skipping the install, why not just link to the local `langium` and `langium-cli` package?

kaisalmen:
One observation I made recently with npm workspaces (using a current npm 8 version). If you specify a local package (from the same workspace) with a version that is not released it has the same effect as linking it (=provide relative path). This is one strong argument for incrementing the versions after release (next or dev), because you get resolution for free after `npm i`. We should give this a try on a branch.

dhuebner:
@msujew 
>  why not just link to the local langium and langium-cli package?

You mean create a `node_modules` folder after the generator created everything else and then, before `install()` is called, create links? This sounds less hacky but still tricky. Nevertheless I would also prefer to generate in some temporary folder and not in examples. AFAIK there is no hook between `writing()` and `install()` in the API, but one can "prepare" the target folder.

@kaisalmen 
> If you specify a local package (from the same workspace) with a version that is not released it

This would mean to change the generated `hello-world/package.json`, right? 
Right now `npm install` just looks for installed packages in parent folder where `langium` and `langium-cli` already linked to the compiled local state.

msujew:
@dhuebner No, way simpler, you just run `npm link` inside of `langium` and `langium-cli` that instructs npm to always prioritize the local link instead of loading it from the registry.

dhuebner:
@msujew 
Would it not change the local workspace state after one run the test?

msujew:
@dhuebner I wouldn't run the `npm link` in the test but only in the CI.

kaisalmen:
@dhuebner I was referring to all packages referencing each other in the workspace. I will work on a POC next week. Will see if I told you rubbish above. 🙂

kaisalmen:
Here is the POC/Draft PR: https://github.com/langium/langium/pull/851

dhuebner:
@msujew 
> I wouldn't run the npm link in the test but only in the CI (i.e. within a github action).

And how to run the test locally? Test should run the same way locally, otherwise they are hard to reproduce. 
Overall is seems like Miro's trick with parent folder look-up works better as other suggestions.

spoenemann (approved):
This approach exploits the module resolution mechanism of Node.js. Although there are certainly cleaner approaches, this one has the advantage that it's very simple and is reproducible both locally and in CI. I propose to take this and change it in case it causes problems.

PR #850, opened 2022-12-21
Correctly sort types topologically

msujew:
Closes https://github.com/langium/langium/issues/849


msujew:
@spoenemann We should get this bugfix into `1.0.1` as well.

packages/langium/src/grammar/type-system/ast-collector.ts
```diff
@@ -20,21 +20,18 @@ export function collectAst(grammars: Grammar | Grammar[], documents?: LangiumDoc
...
-
     specifyAstNodeProperties(astTypes);
     return astTypes;
 }
 
 function mergeAndRemoveDuplicates<T extends { name: string }>(inferred: T[], declared: T[]): T[] {
     return Array.from(inferred.concat(declared)
         .reduce((acc, type) => { acc.set(type.name, type); return acc; }, new Map<string, T>())
-        .values());
+        .values()).sort((a, b) => a.name.localeCompare(b.name));
```
pluralia, line 34: Interfaces do not need this sorting -- it's only for unions. I'd suggest applying it directly to unions like `sortInterfacesTopologically`.

pluralia (approved):

packages/langium/src/grammar/type-system/ast-collector.ts
msujew: Interfaces also benefit from this, since it keeps interfaces stable in case the grammar changes. The topological sorting function is stable and so we get less changes in our output files.

PR #851, opened 2022-12-21
POC: Use same development version in complete workspace

kaisalmen:
This is a proof of concept that allows to reconfigure a new version throughout the workspace.

All packages have and reference the same version. Supplied node js script uses shell js to perform the changes

If this version specified is not available from a registry, then it is resolved locally **without any special adjustments** (like npm link). 
Resolution between examples and langium package works. When `npm run watch` is executed cli execution works: Tested with naive console.log in `default-module.ts` and arithmetics example.

Idea is to move to an "incremented" version immediately after release and then use this in development. WDYT?

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -24,5 +24,6 @@ jobs:
       shell: bash
       run: |
         npm ci
+        npm run build
         npm test
         npm run lint
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -140,6 +140,14 @@
             "args": ["run", "${relativeFile}"],
             "smartStep": true,
             "console": "integratedTerminal",
-        }
+        },
+        {
+            "name": "Version Upgrade",
+            "type": "node",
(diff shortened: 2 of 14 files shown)


msujew (commented):
> If this version specified is not available from a registry, then it is resolved locally without any special adjustments (like npm link).

I believe the yeoman generator will place the newly generated project outside of the langium repository structure. Will this change still link correctly across repos in this case? I would think that `npm link` is still necessary.

kaisalmen:
> Will this change still link correctly across repos in this case? I would think that npm link is still necessary.

The new project is independent of the workspace and it will therefore not work. You have to manually link `langium` and `langium-cli`.

Having a local version independent of the release or at least the capability to generate one by invoking a command makes sense / is good practices.

kaisalmen:
I picked a bad branch name and will open a new PR once it is ready (rebased, files at proper location, documentation, potential GH Action intergration, etc.)

Issue #778, opened 2022-11-24
Malformed input breaks leaf node computation


msujew:
Langium version: 0.5
Package name: langium

## Steps To Reproduce

Follow the reproduction steps in https://github.com/langium/langium/discussions/777

## The current behavior

No completion appears, as the `node` cannot be deduced from malformed input

## The expected behavior

Completion should show `BOOL` and `BYTE`.

PR #854, opened 2022-12-28
Bugfix for completion for malformed input

insafuhrmann:
Bugfix for completion break due to malformed/incomplete input. Concerns #778.

### DIFF
--- a/packages/langium/src/utils/cst-util.ts
+++ b/packages/langium/src/utils/cst-util.ts
@@ -135,6 +135,9 @@ export function findLeafNodeAtOffset(node: CstNode, offset: number): LeafCstNode
                 return findLeafNodeAtOffset(n, offset);
             }
         }
+        if (firstChild === lastChild) {
+            return findLeafNodeAtOffset(node.children[firstChild], offset);
+        }
     }
     return undefined;
 }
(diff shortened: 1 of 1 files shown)


packages/langium/src/utils/cst-util.ts
```diff
@@ -125,6 +125,7 @@ export function findLeafNodeAtOffset(node: CstNode, offset: number): LeafCstNode
         let firstChild = 0;
         let lastChild = node.children.length - 1;
         while (firstChild <= lastChild) {
+            if (firstChild === lastChild) return findLeafNodeAtOffset(node.children[firstChild], offset);
```
spoenemann, line 128: Suggestion: change the loop condition to `firstChild <= lastChild` and move this case to after the loop. That would separate the cases more clearly.

I'd also suggest to wrap the `if` body in a block with curly braces for the sake of consistency. I know we often used to write with a more compact style with Xtend, but I see consistency as more important than brevity nowadays.

insafuhrmann: Thanks for the suggestion! I have implemented that (I assumend from context you meant "change the loop condition to `firstChild < lastChild`").

insafuhrmann:
@spoenemann
Thanks for your suggestions, they are implemented now. I assumed that you meant that the loop condition should be changed to "less than", as otherwise there would be no suggested change and no separated cases, `<=` was the previous condition.

spoenemann (approved):
Oh yes, thanks!

PR #857, opened 2022-12-29
Fix generator support on Windows

msujew:
PR https://github.com/langium/langium/pull/825 relies on the fact that the input is correctly normalized to the current OS, since it uses the OS EOL constant to split lines. This isn't always the case, which leads the generator to print unexpectedly indented output.

This PR fixes this issue and addresses the downstream testing issues which appear on Windows. Also adds a windows runtime to our test suite so that contributors are aware that their tests do not work on Windows.


.github/workflows/actions.yml
```diff
@@ -25,4 +29,20 @@ jobs:
...
+      uses: actions/checkout@v3
+    - name: Use Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: '16'
+    - name: Build
+      shell: bash
+      run: |
+        npm ci
         npm run lint
```
spoenemann, line 47: The build is probably faster if we keep the linting step in the matrix job?

msujew: For some reason, running `npm run lint` on Windows GitHub Actions runs into some sort of freeze. The CI times out after 20 minutes, see [here](https://github.com/langium/langium/actions/runs/3812281182/jobs/6485406094). That's why I split them. It's not necessary to run the lint twice.

The build is actually faster if we don't, since these actions run in parallel not sequentially.

msujew: Also, it might make sense to split tests and linting anyway, we have done the same in Theia recently. Contributors don't want to see every pipeline fail just because of an extra whitespace somewhere. Just failing the lint pipeline is good enough.

kampf0keks:
tested it on Windows with the statemachine example. Identations of the generated cpp code look perfectly fine

spoenemann (approved):

PR #866, opened 2023-01-09
Add JSDoc support

msujew:
Adds utilities to parse and render JSDoc comments. Also:

* Extends the `AstNodeDescription` for range information
* Modifies the default hover implement to render `{@link}` inline JSDoc tags.
* Improves the langium language configuration to better deal with JSDoc.


packages/langium/src/grammar/references/grammar-scope.ts
```diff
@@ -141,14 +124,21 @@ export class LangiumGrammarScopeComputation extends DefaultScopeComputation {
...
             }
         }
     }
+
+    protected createInterfaceDescription(node: AstNode, name: string, document: LangiumDocument = getDocument(node)): AstNodeDescription {
+        const nameNode = this.nameProvider.getNameNode(node) ?? node.$cstNode;
+        return {
+            node,
+            name,
+            segment: nameNode && toDocumentSegment(nameNode),
```
spoenemann, line 137: This pattern is used quite frequently. Maybe change  `toDocumentSegment` to accept `undefined`? It can be kept backwards-compatible with multiple function signatures (one for defined, one for undefined).

packages/langium/src/lsp/hover-provider.ts
```diff
@@ -79,15 +96,43 @@ export class MultilineCommentHoverProvider extends AstNodeHoverProvider {
...
-        const split = commentText.split('\n').map(e => {
-            e = e.trim();
-            if (e.startsWith('*')) {
-                e = e.substring(1).trim();
+    protected getAdditionalHoverContent(_node: AstNode): string | undefined {
+        return undefined;
+    }
+
+    protected hoverLinkRenderer(node: AstNode, name: string, display: string): string | undefined {
+        const description = this.findNameInPrecomputedScopes(node, name) ?? this.findNameInGlobalScope(node, name);
```
spoenemann, line 104: I fear that this would force some users who have heavily customized the scope provider to apply similar customizations here as well. Could you try reusing the scope provider service here? The `property` can be set to an exported constant like `$jsdoc`, so language implementers can add special case handling in their scope provider if they want.

packages/langium/src/syntax-tree.ts
```diff
@@ -75,6 +75,14 @@ export function isReference(obj: unknown): obj is Reference {
 export interface AstNodeDescription {
     /** The target node; should be present only for local references (linking to the same document). */
     node?: AstNode;
+    /**
+     * The document segment that represents the range of the name of the AST node.
+     */
+    segment?: DocumentSegment;
```
spoenemann, line 81: Name this `nameSegment` to be clear?

packages/langium/src/workspace/ast-descriptions.ts
```diff
@@ -37,15 +38,20 @@ export interface AstNodeDescriptionProvider {
...
 
     protected readonly astNodeLocator: AstNodeLocator;
+    protected readonly nameProvider: NameProvider;
 
     constructor(services: LangiumServices) {
         this.astNodeLocator = services.workspace.AstNodeLocator;
+        this.nameProvider = services.references.NameProvider;
     }
 
     createDescription(node: AstNode, name: string, document: LangiumDocument = getDocument(node)): AstNodeDescription {
```
spoenemann, line 48: Now that we have the name provider here, we could make `name` optional as well.

Lotes:
I was expecting some new service for the Langium module. Something like this:

```ts
interface DocumentationCommentHandler {
  // extracts the comment string from the hidden CST in front of the given node
  getDocumentation(node: AstNode): string|undefined;
}
```

Which can be used by JSDOC parser…

packages/langium/src/lsp/hover-provider.ts
msujew: Not sure how this would work. The scoping algorithm always expects that we have some kind of reference type. With the current scoping implementation, the reflection will just always throw.

I'm sure there's a clever solution for this somewhere, but I believe it's out of scope for this PR. The hover implementation is just a showcase on how this could be used. We can improve it later on.

msujew:
@Lotes I've tried to keep the JSDoc implementation to be as independent of any services as possible. It's just a function, not a whole service.

packages/langium/src/documentation/index.ts
```diff
@@ -0,0 +1,8 @@
+/******************************************************************************
+ * Copyright 2023 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+export * from './documentation-provider';
+export * from './jsdoc';
```
spoenemann, line 8: We don't have such index files for all subfolders. Do we want that in general?

packages/langium/src/utils/cst-util.ts
```diff
@@ -59,7 +59,12 @@ export function tokenToRange(token: IToken): Range {
     };
 }
 
-export function toDocumentSegment(node: CstNode): DocumentSegment {
+export function toDocumentSegment(node: CstNode): DocumentSegment;
+export function toDocumentSegment(node?: CstNode): DocumentSegment | undefined;
```
spoenemann, line 63:
```suggestion
export function toDocumentSegment(node: undefined): undefined;
```

packages/langium/src/workspace/ast-descriptions.ts
```diff
@@ -25,30 +26,40 @@ export interface AstNodeDescriptionProvider {
...
     constructor(services: LangiumServices) {
         this.astNodeLocator = services.workspace.AstNodeLocator;
+        this.nameProvider = services.references.NameProvider;
     }
 
-    createDescription(node: AstNode, name: string, document: LangiumDocument = getDocument(node)): AstNodeDescription {
+    createDescription(node: AstNode, name: string | undefined, document: LangiumDocument = getDocument(node)): AstNodeDescription {
+        name ??= this.nameProvider.getName(node);
+        const path = this.astNodeLocator.getAstNodePath(node);
+        if (name === undefined) {
```
spoenemann, line 51: I think it's ok to check for `!name` here because an empty string is not a name.

spoenemann (approved):
Could you create two follow-up issues (both low priority):
 - Extend the default go-to-definition handling to support `@link` tags
 - Extend the default completion provider to support `@link` tags

packages/langium/src/utils/cst-util.ts
msujew: It doesn't work that way. The input type will still be `CstNode | undefined`, whereas this function would only allow either type `CstNode` or `undefined` not `CstNode | undefined`.

packages/langium/src/documentation/index.ts
msujew: I believe it might make sense, yeah.

packages/langium/src/utils/cst-util.ts
spoenemann: Ok, strange. To me it looks like that should be compatible

msujew: I've tested it, because I wasn't completely sure myself. For overloads, TypeScript never builds a union over the overloads, but only ever picks one. Since the input type is a "larger" union than either of the possible input types of the overloads, the assignability check doesn't pass.

PR #873, opened 2023-01-16
Switch to specific vitest imports

kaisalmen:
This is triggered by #872 and removing any globals for Jest or Vitest. It includes these changes as well. Rebasing this PR on the main when the other one is merged is easily possible.

Instead of relying on global imports (like jest did), this PR switches to specific imports in every single test. All test folders in all packages and examples are now added to `npm run lint`

Additionally: 
- `tsconfig.test.json` are added in all packages and examples with noEmit. These are used by main `tsconfig.build.json` to compile all code
- GitHub Actions update: Performs build, test on both Ubuntu and Windows and lint only on Ubuntu. Tests are executed even if build fails as Vitest does not rely on that, correct?!
- vscode settings are updated
- Recommended extensions for Vitest and editorconfig are added
- eslint ecmaVersion has been aligned with tsconfig (2017)
- I removed the top-level `prepare` as you can't run `npm install` with a full rebuild. IMO it is better to have control of that: `prepare` is renamed to `build:clean` and the build pipeline is updated accordingly

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,10 +6,10 @@
     "npm": ">= 7.7.0"
   },
   "scripts": {
-    "prepare": "npm run clean && npm run build && npm run build --workspace=langium-vscode",
-    "clean": "shx rm -rf packages/**/lib packages/**/out packages/**/tsconfig.tsbuildinfo",
+    "clean": "shx rm -rf packages/**/lib packages/**/out packages/**/*.tsbuildinfo",
     "build": "tsc -b tsconfig.build.json",
     "watch": "tsc -b tsconfig.build.json -w",
+    "build:clean": "npm run clean && npm run build && npm run build --workspace=langium-vscode",
     "lint": "npm run lint --workspaces",
     "test": "vitest",
     "test-ui": "vitest --ui",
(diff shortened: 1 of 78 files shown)


msujew:
@kaisalmen Please rebase on top of the latest changes

kaisalmen:
Done.

kaisalmen:
@msujew I rebased again due to conflicts. This is what is new:
- I added volta instructions to all package.json now. My default is now node 18 and when this the case it may sneaks in unexpectedly on package level. This prevents it for everybody using volta
- I removed the outdated version of eslint from **langium-vscode**. vite(st) brings in a newer version and the directly declared one was overruled because it it is a workspace level dependency? This even lead to an execution error, but may have been Windows related
- I updated the release guide in CONTRIBUTING.md. It did not yet reflect the `npm run build:clean` change

kaisalmen:
⬆️ Changes: 
- `@types/vscode` and `@types/node` have been moved to top-level of the workspace other references have been removed
- Dependency versions have been updated
- `~` is now the default dependency prefix

msujew (approved):
Another approval after re-review. Everything works as expected on CI/locally.

package.json
```diff
@@ -6,10 +6,10 @@
     "npm": ">= 7.7.0"
   },
   "scripts": {
-    "prepare": "npm run clean && npm run build && npm run build --workspace=langium-vscode",
```
spoenemann, line 9: Why don't we use the `prepare` script anymore?

This broke the Gitpod setup. @kaisalmen could you add the build step to gitpod.yml?

kaisalmen: > I removed the top-level prepare as you can't run npm install without a full rebuild. IMO it is better to have control of that: prepare is renamed to build:clean and the build pipeline is updated accordingly

That is the reason. I adjusted GH Actions, but forgot GP and noticed that yesterday when testing #910

Issue #864, opened 2023-01-05
langium extract-types acts unexpectedly for multi-languages


l-ehlers:
When using `langium extract-types` (introduced in #754) on a grammar that is spread across different files, the type extraction does not work correctly. In fact, if no explicit types have been declared for the specified grammar file, only the lastly imported grammar file is taken into account by the langium-cli tool.

**Note: I know the tool is not published/documented yet, and I have no need for a fix of the tool, but I wanted to share my findings with you anyway**

Probably, the described behavior is caused by this line: https://github.com/langium/langium/blob/main/packages/langium-cli/src/generate.ts#L305

Langium version: v1.0.1
Package name: langium-cli

## Steps To Reproduce

1. Create the multi-language depicted below
2. Run `langium extract-types A.langium`
3. Swap `A`'s imports to `B` and `C`
4. Run `langium extract-types A.langium` again

### A.langium
```
grammar A

import "B"
import "C"

entry A: b=B c=C;
```

### B.langium
```
import "terminals"

B: valB=ID;
```

### C.langium
```
import "terminals"

C: valC=ID;
```

### terminals.langium
```
terminal ID: /[_a-zA-Z][\w_]*/;
```

### First generation
```
interface C {
    valC: string
}
```

### Second generation (with swapped imports in A)
```
interface B {
    valB: string
}
```

## The expected behavior

Either a correct typing file such as 
```
interface A {
    b: B
    c: C
}

interface B {
    valB: string
}

interface C {
    valC: string
}
```

or one typing file for each grammar file (maybe configurable)

or just an error "multi-languages are not supported by `extract-types`"

(Interestingly, the type generation seems to work just fine for multi-languages with types that have already been declared)

dhuebner:
Fixed on master

PR #874, opened 2023-01-17
Start type generation from given grammar file #864

dhuebner:
Fixes bogus behavior in multi grammar projects described in #864

I'm not sure how we handle versions changes in langium-cli. I could imagine to update langium-cli dependencies in examples to ~1.0.1, but it is not really necessary.

### DIFF
--- a/packages/langium-cli/src/generate.ts
+++ b/packages/langium-cli/src/generate.ts
@@ -295,21 +295,27 @@ export async function generateTypes(options: ExtractTypesOptions): Promise<void>
         }
     }
 
-    const grammarFileUri = URI.file(grammarPath);
-    const document = documents.getOrCreateDocument(grammarFileUri);
-    const allUris = eagerLoad(document);
+    const grammarDoc = await doLoadAndUpdate(documents.getOrCreateDocument(URI.file(grammarPath)));
+    const genTypes =  generateTypesFile(grammarServices, [grammarDoc.parseResult.value as Grammar]);
+    await writeWithFail(typesFilePath, genTypes, { watch: false });
+    log('log', { watch: false }, `Generated type definitions to: ${chalk.white.bold(typesFilePath)}`);
+    return;
+}
+
+/**
+ * Builds the given grammar document and all imported grammars.
+ */
+async function doLoadAndUpdate(grammarDoc: LangiumDocument): Promise<LangiumDocument> {
+    const allUris = eagerLoad(grammarDoc);
     await sharedServices.workspace.DocumentBuilder.update(allUris, []);
-    let grammarDoc;
     for (const doc of documents.all) {
         await sharedServices.workspace.DocumentBuilder.build([doc]);
-        grammarDoc = doc;
+        if(doc.uri === grammarDoc.uri) {
+            // update grammar doc after rebuild
+            grammarDoc = doc;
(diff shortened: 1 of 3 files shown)


packages/langium-cli/src/generate.ts
```diff
@@ -295,21 +295,27 @@ export async function generateTypes(options: ExtractTypesOptions): Promise<void>
...
+    log('log', { watch: false }, `Generated type definitions to: ${chalk.white.bold(typesFilePath)}`);
+    return;
+}
+
+/**
+ * Builds the given grammar document and all imported grammars.
+ */
+async function doLoadAndUpdate(grammarDoc: LangiumDocument): Promise<LangiumDocument> {
+    const allUris = eagerLoad(grammarDoc);
     await sharedServices.workspace.DocumentBuilder.update(allUris, []);
```
sailingKieler, line 310: does this call of `update` make sense? ... also wrt. the loop subsequently

dhuebner: I think the `update` call does make sense.

In my case I do not have the  `LangiumConfig`. And yes, `doLoadAndUpdate ` was inspired by `buildAll` func :-)

sailingKieler: > In my case I do not have the `LangiumConfig`. And yes, `doLoadAndUpdate ` was inspired by `buildAll` func :-)

Right. But maybe you could put this https://github.com/langium/langium/blob/2e6acf43c3281c26e1c6788cc62f4ae7ee0884b3/packages/langium-cli/src/generate.ts#L164-L172 of  `buildAll` into a dedicated method taking a varargs parameter of URI or so.

`buildAll` might than call this method with `config.languages.map(lang => URI.file(path.resolve(relPath, languageConfig.grammar)))`.

🙏

packages/langium-cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "langium-cli",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
spoenemann, line 3: Please don't increase the version number now. We'll do it during the publishing process.

packages/langium-cli/src/generate.ts
spoenemann: > What do @msujew and @spoenemann think?

Consolidating is fine if it doesn't increase the code complexity too much.

dhuebner: @spoenemann 
I thought about re-using existing implementation, but decided to leave it as is, especially because I need a return value in my case. 
I'm also just about to fix a bug in the current implementation and don't want to do much changes on existing code.

@sailingKieler 
I could provide an other optimization PR which address your concerns about the code style. Doing so we can better discuss the outcome in sense of "complexity vs. duplicate code". WDYT?

spoenemann: 👍  reusing code can sometimes increase complexity instead of reducing it.

This is internal code, not public API, so we can always refactor it as we see fit.

packages/langium-cli/package.json
dhuebner: @spoenemann 
I reverted this change

spoenemann (approved):

PR #876, opened 2023-01-19
generation support: some bugfixes and improvements

sailingKieler:
... and additional tests

* introduces a 'separator' argument to `joinToNode`
* improves the generation of IndentNodes by `expandToNode`
* refines/fixes the application of (non-immediate) indentation handling in the generator node processor

### DIFF
--- a/packages/langium/src/generator/generator-node.ts
+++ b/packages/langium/src/generator/generator-node.ts
@@ -5,6 +5,7 @@
  ******************************************************************************/
 
 import { processGeneratorNode } from './node-processor';
+import { expandToNode } from './template-node';
 
 export const EOL = (typeof process === 'undefined') ? '\n' : (process.platform === 'win32') ? '\r\n' : '\n';
 
@@ -16,6 +17,13 @@ export const EOL = (typeof process === 'undefined') ? '\n' : (process.platform =
  */
 export type Generated = GeneratorNode | string | undefined;
 export type GeneratorNode = CompositeGeneratorNode | IndentNode | NewLineNode;
+
+export interface IndentConfig {
+    indentedChildren?: Generated[] | ((indented: IndentNode) => void);
+    indentation?: string|number;
+    indentEmptyLines?: boolean;
+    indentImmediately?: boolean;
+}
 
 export function isGeneratorNode(node: unknown): node is CompositeGeneratorNode | IndentNode | NewLineNode {
     return node instanceof CompositeGeneratorNode
@@ -85,7 +93,7 @@ export class CompositeGeneratorNode {
      *      'The end!'
      *   );
      */
-    append(...content: Array<Generated | ((node: CompositeGeneratorNode) => void)>): this {
+    append(...content: Array<Generated | ((node: this) => void)>): this {
         for (const arg of content) {
             if (typeof arg === 'function') {
                 arg(this);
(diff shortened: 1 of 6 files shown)


packages/langium/src/generator/generator-node.ts
```diff
@@ -214,8 +214,8 @@ export class CompositeGeneratorNode {
      *       '}'
      *   );
      */
-    indent(configurator?: (indentingNode: IndentNode) => void, indentation?: string|number): this {
-        const node = new IndentNode(indentation, false);
+    indent(configurator?: (indentingNode: IndentNode) => void, indentation?: string|number, indentImmediately = true): this {
```
spoenemann, line 217: What's the use case of the configurator function? Could the `indentImmediately` property be configured with that?

I like to avoid Boolean function arguments. Alternative signature (backwards-compatible):
```
indent(configurator?: (indentingNode: IndentNode) => void, options?: string | number | { indentation?: string | number,  indentImmediately: boolean })
```

sailingKieler: The `configurator` is a callback allowing to use `indent` in a fluent api style, see the example of the jsdoc:
https://github.com/langium/langium/blob/d53d019cd3f2842e5596f78fa1733e0050f4b7a7/packages/langium/src/generator/generator-node.ts#L205-L213

What's the problem with boolean args? Readability?

I expect `indentImmediately` to be omitted by client code in 99% of the use cases, and I don't expect any further things to be added there. In the initial draft I hoped I could avoid that parameter, it need to be `false` when creating indent nodes via `expandToNode`. That's why I hard-coded it to `false`. However, that breaks the above example, as the first line would not be indented as expected.

spoenemann: > What's the problem with boolean args? Readability?

Yes, calling `myFunction(false)` is more obscure than `myFunction({ indentImmediately: false })`

Let's leave that method as you suggested, we can still add an object-based alternative  signature later.

spoenemann: > I expect indentImmediately to be omitted by client code in 99% of the use cases

Rethinking about this, an argument that is expected to be rarely used is even more suited to be wrapped in an object _(named argument)_.

packages/langium/src/generator/generator-node.ts
```diff
@@ -17,6 +17,13 @@ export const EOL = (typeof process === 'undefined') ? '\n' : (process.platform =
 export type Generated = GeneratorNode | string | undefined;
 export type GeneratorNode = CompositeGeneratorNode | IndentNode | NewLineNode;
 
+export interface IndentConfig {
+    indentation?: string|number;
+    indentedChildren?: Generated[] | ((indented: IndentNode) => void);
```
spoenemann, line 22: Now that this is wrapped in an object, you could as well separate these variants in two properties `indentedChildren` and `addChildren` or similar.

spoenemann (commented):
The changes look good, just one comment

packages/langium/src/generator/generator-node.ts
sailingKieler: Right. On the other hand I like the similarity to the non-object-based definition, but that's maybe personal taste.

I'm just adding some missing documentation to `joinToNode` and a few additional test cases, than it's done I hope...

spoenemann (approved):
Ok

PR #877, opened 2023-01-19
arithmetics example: moved util method to dedicated module to avoid importing 'fs' via 'main-browser'

sailingKieler:
* added explicit package exports

### DIFF
--- a/examples/arithmetics/package.json
+++ b/examples/arithmetics/package.json
@@ -47,6 +47,17 @@
         "arithmetics-cli": "./bin/cli"
     },
     "main": "./out/extension.js",
+    "types": "./out/extension.d.ts",
+    "exports": {
+        ".": {
+            "types": "./out/extension.d.ts",
+            "default": "./out/extension.js"
+        },
+        "./api": {
+            "types": "./out/language-server/generated/ast.d.ts",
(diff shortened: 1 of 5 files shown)


examples/arithmetics/package.json
```diff
@@ -47,6 +47,17 @@
         "arithmetics-cli": "./bin/cli"
     },
     "main": "./out/extension.js",
+    "types": "./out/extension.d.ts",
+    "exports": {
```
msujew, line 51: Are there any special reasons to add the exports?

sailingKieler: I need to access the AST types in my code gen example, and I didn't want copy it over, and lengthy imports like `langium-arithmetics-dsl/lib/language-server/generated/ast.ts` also feel odd.

But I don't insist on the export, esp. since the visible types and methods are limited to the exported ones with this `export` declaration.

sailingKieler: Right, I just added that for completeness as https://nodejs.org/api/packages.html#package-entry-points states that `export` might override `main`; better safe than sorry, added `types` also just for consistency.

msujew: I see, alright.

msujew (approved):
LGTM 👍

PR #886, opened 2023-01-27
generation support: tracing API and implementation

sailingKieler:
This PR contributes:
* Definition of the API for gathering tracing information within code generators
* Implementation of the processing of tracing data to pairs of text regions
* Extension of the JsonSerializer to optionally add `$textRegion` properties to the AstNodes while serializing them

### DIFF
--- a/packages/langium/src/generator/generator-node.ts
+++ b/packages/langium/src/generator/generator-node.ts
@@ -4,8 +4,8 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { AstNode, Properties } from '../syntax-tree';
-import type { TraceRegion, TraceSourceSpec } from './generator-tracing';
+import { AstNode, isAstNode, Properties } from '../syntax-tree';
+import type { SourceRegion, TraceRegion, TraceSourceSpec } from './generator-tracing';
 import { processGeneratorNode } from './node-processor';
 import { expandToNode, expandTracedToNode } from './template-node';
 
@@ -76,7 +76,7 @@ export class CompositeGeneratorNode {
 
     readonly contents: Array<(GeneratorNode | string)> = [];
 
-    tracedSource?: TraceSourceSpec;
+    tracedSource?: TraceSourceSpec | SourceRegion | SourceRegion[];
 
     /**
      * Constructor.
@@ -98,7 +98,8 @@ export class CompositeGeneratorNode {
     }
 
     /**
-     * Adds tracing information to `this` generator node. Overwrites existing trace data, if set previously.
+     * Adds tracing information in form of `{astNode, property?, index?}` to `this` generator node.
+     * Overwrites existing trace data, if set previously.
      *
      * The given data are kept as they are, the actual resolution of text positions within the DSL text
      * is done at the final processing of `this` node as part of {@link toStringAndTrace()}.
@@ -114,10 +115,42 @@ export class CompositeGeneratorNode {
      *
      * @returns `this` {@link CompositeGeneratorNode} for convenience.
      */
-    trace<T extends AstNode>(astNode: T, property?: Properties<T>, index?: number): this {
-        this.tracedSource = <TraceSourceSpec>{ astNode, property, index };
-        if (this.tracedSource.property === undefined && this.tracedSource.index !== undefined && this.tracedSource.index > -1) {
-            throw new Error("Generation support: 'property' argument must not be 'undefined' if a non-negative value is assigned to 'index' in 'CompositeGeneratorNode.trace(...)'.");
+    trace<T extends AstNode>(astNode: T, property?: Properties<T>, index?: number): this;
+
+    /**
+     * Adds tracing information in form of concrete coordinates to `this` generator node. Complete coordinates
+     * are provided by the {@link AstNode AstNodes}' corresponding {@link AstNode.$cstNode AstNode.$cstNodes}.
+     * Overwrites existing trace data, if set previously.
+     *
+     * @param sourceRegion a text region within some file in form of concrete coordinates,
+     *  if `undefined` no tracing will happen
+     *
+     * @returns `this` {@link CompositeGeneratorNode} for convenience.
+     */
+    trace(sourceRegion: SourceRegion | undefined): this;
+
+    /**
+     * Adds tracing information in form of a list of concrete coordinates to `this` generator node.
+     * Overwrites existing trace data, if set previously.
+     *
+     * The list of regions in `sourceRegion` is reduced to the smallest encompassing region
+     *  of all the contained source regions.
+     *
+     * @param sourceRegion a list of text regions within some file in form of concrete coordinates,
+     *  if empty no tracing will happen
+     *
+     * @returns `this` {@link CompositeGeneratorNode} for convenience.
+     */
+    trace(sourceRegion: SourceRegion[]): this;
+
+    trace<T extends AstNode>(source: T | undefined | SourceRegion | SourceRegion[], property?: Properties<T>, index?: number): this {
+        if (isAstNode(source)) {
+            this.tracedSource = <TraceSourceSpec>{ astNode: source, property, index };
+            if (this.tracedSource.property === undefined && this.tracedSource.index !== undefined && this.tracedSource.index > -1) {
+                throw new Error("Generation support: 'property' argument must not be 'undefined' if a non-negative value is assigned to 'index' in 'CompositeGeneratorNode.trace(...)'.");
+            }
+        } else {
+            this.tracedSource = source;
         }
         return this;
     }
@@ -280,9 +313,7 @@ export class CompositeGeneratorNode {
      *   .appendNewLineIfNotEmpty()
      */
     appendTemplateIf(condition: boolean): (staticParts: TemplateStringsArray, ...substitutions: unknown[]) => this {
-        return (staticParts, ...substitutions) => {
-            return condition ? this.appendTemplate(staticParts, ...substitutions) : this;
-        };
+        return condition ? (staticParts, ...substitutions) => this.appendTemplate(staticParts, ...substitutions) : () => this;
     }
 
     /**
@@ -334,7 +365,8 @@ export class CompositeGeneratorNode {
     }
 
     /**
-     * Convenience method for appending content to `this` generator node including tracing information.
+     * Convenience method for appending content to `this` generator node including tracing information
+     * in form of `{astNode, property?, index?}`.
      *
      * This method returns a helper function that takes the desired `content` and does the processing.
      * The returned function delegates to {@link append}, with the provided `content` being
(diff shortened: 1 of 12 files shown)


packages/langium/src/serializer/json-serializer.ts
```diff
@@ -137,3 +154,27 @@ export class DefaultJsonSerializer implements JsonSerializer {
         return this.astNodeLocator.getAstNode(root, path.substring(1))!;
     }
 }
+
+function addAstNodeRegionWithAssignmentsTo(root: AstNodeWithTextRegions) {
```
msujew, line 158: Suggestion: This function should be part of the serializer or exported so that users that override the service can use the function as well.

packages/langium/src/generator/generator-node.ts
```diff
@@ -78,6 +89,14 @@ export class CompositeGeneratorNode {
         return this.contents.length === 0;
     }
 
+    trace<T extends AstNode>(astNode: T, property?: keyof T, index: number = -1): this {
```
msujew, line 92: Suggestion: We have a special type for the `property` parameter here, that prevents users from accidentally using reserved property names such as `$type` or `$container`:
```suggestion
    trace<T extends AstNode>(astNode: T, property?: Properties<T>, index: number = -1): this {
```

packages/langium/src/generator/generator-tracing.ts
```diff
@@ -0,0 +1,122 @@
...
+import type { AstNodeRegionWithAssignments, AstNodeWithTextRegions } from '../serializer/json-serializer';
+import type { AstNode, GenericAstNode } from '../syntax-tree';
+import { findNodesForProperty } from '../utils/grammar-util';
+import { TreeStreamImpl } from '../utils/stream';
+import type { DocumentSegment } from '../workspace/documents';
+
+export interface TraceSourceSpec {
+    astNode: AstNode,
+    property?: string,
+    index?: number,
```
msujew, line 16: Minor:
```suggestion
    astNode: AstNode;
    property?: string;
    index?: number;
```

packages/langium/src/generator/node-processor.ts
```diff
@@ -67,10 +94,30 @@ class Context {
...
+    context.traceData.push(InternalTraceRegion.create(undefined, 0));
+
     processNodeInternal(node, context);
-    return context.content;
+
+    const trace = context.traceData.pop()! as InternalTraceRegion;
+    trace.complete && trace.complete(context.content.length);
+
+    const singleChild = trace.children && trace.children.length === 1 ? trace.children[0] : undefined;
+    const singleChildTargetRegion = singleChild && singleChild.targetRegion;
```
msujew, line 107: Minor:
```suggestion
    const singleChildTargetRegion = singleChild?.targetRegion;
```

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -137,3 +154,27 @@ export class DefaultJsonSerializer implements JsonSerializer {
...
+        offset: cstNode.offset,
+        end:    cstNode.end,
+        length: cstNode.length,
+        range:  cstNode.range,
+    };
+
+    streamAst(root).forEach((node: AstNodeWithTextRegions) => {
+        if (node.$cstNode) {
+            const textRegion = node.$textRegion = createDocumentSegment(node.$cstNode);
+            const assignments: { [key: string]: DocumentSegment[] } = textRegion.assignments = {};
```
msujew, line 169: Minor:
```suggestion
            const assignments: Record<string, DocumentSegment[]> = textRegion.assignments = {};
```

packages/langium/src/generator/generator-node.ts
```diff
@@ -59,6 +68,8 @@ export class CompositeGeneratorNode {
 
     readonly contents: Array<(GeneratorNode | string)> = [];
 
+    tracedSource?: TraceSourceSpec;
```
msujew, line 71: Suggestion: We should make this readonly as well:
```suggestion
    get tracedSource(): TraceSourceSpec | undefined {
        return this._tracedSource;
    }
    
    private _tracedSource?: TraceSourceSpec;
```

packages/langium/src/generator/generator-node.ts
sailingKieler: Good to know, thanks 👍

packages/langium/src/generator/generator-tracing.ts
sailingKieler: 👍

packages/langium/src/generator/node-processor.ts
sailingKieler: 👍

packages/langium/src/serializer/json-serializer.ts
sailingKieler: 👍

packages/langium/src/generator/generator-node.ts
sailingKieler: I don't think so, as the content of `contents` can also be manipulated externally.
Besides the current `trace` method doesn't allow to reset this field, so I wouldn't be pedantic at this point.

If somebody wants to break something: feel free!

packages/langium/src/generator/generator-node.ts
```diff
@@ -78,6 +97,31 @@ export class CompositeGeneratorNode {
...
+     * @param property the value property name (string) corresponding to `this` node's content,
+     *  e.g. if this node's content corresponds to some `string` or `number` property; is optional
+     *
+     * @param index the index of the value within a list property corresponding to `this` node's content,
+     * if the property contains a list of elements, is ignored otherwise,
+     * must not be given if no `property` is given ; is optinal
+     *
+     * @returns `this` {@link CompositeGeneratorNode} for convenience.
+     */
+    trace<T extends AstNode>(astNode: T, property?: Properties<T>, index?: number): this {
```
spoenemann, line 117: It would be cool to be able to trace to a CST node as alternative (use two different method signatures).

Same for the other trace methods.

packages/langium/src/generator/generator-node.ts
```diff
@@ -41,14 +43,29 @@ export function isNewLineNode(node: unknown): node is NewLineNode {
...
+ * Converts instances of {@link GeneratorNode} into `text` accompanied by a corresponding `trace`.
+ *
+ * @param defaultIndentation the indentation to be applied if no explicit indentation is configured
+ *  for particular {@link IndentNode IndentNodes}, either a `string` or a `number` of repeated single spaces,
+ *  defaults to 4 single spaces, see {@link processGeneratorNode} -> `Context`.
+ *
+ * @returns an object of type `{ text: string, trace: TraceRegion }` containing the desired `text` and `trace` data
+ */
+export function toStringAndTrace(input: GeneratorNode, defaultIndentation?: string | number): { text: string, trace: TraceRegion } {
+    return processGeneratorNode(input, defaultIndentation);
```
spoenemann, line 66: Why offer the same functionality with two names? Can't we export `processGeneratorNode` to users, or rename it?

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -6,13 +6,25 @@
...
     space?: string | number
     refText?: boolean
+    textRegions?: boolean
+}
+
+export interface AstNodeWithTextRegions extends AstNode {
+    $textRegion?: AstNodeRegionWithAssignments;
+}
+
+export interface AstNodeRegionWithAssignments extends DocumentSegment {
```
spoenemann, line 25: What do these assignments and document URI mean?

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -79,10 +91,39 @@ export class DefaultJsonSerializer implements JsonSerializer {
...
+            range:  cstNode.range,
+        };
+
+        if (node.$cstNode) {
+            const textRegion = node.$textRegion = createDocumentSegment(node.$cstNode);
+            const assignments: Record<string, DocumentSegment[]> = textRegion.assignments = {};
+
+            Object.keys(node).filter(key => !key.startsWith('$')).forEach( key => {
+                const propertyAssignments = findNodesForProperty(node.$cstNode, key).map( createDocumentSegment );
+                if (propertyAssignments.length !== 0)
```
spoenemann, line 118: I'd add curly braces even to single lines for consistency.

packages/langium/src/generator/generator-node.ts
sailingKieler: Maybe yes. I don't want to expose too many parts as API and wanted to leave us freedom of migrating the implementation. For that reason `processGeneratorNode` is not exposed via some `index.ts`.

packages/langium/src/serializer/json-serializer.ts
sailingKieler: There's documentation missing, right.

packages/langium/src/serializer/json-serializer.ts
sailingKieler: 👍

packages/langium/src/generator/generator-node.ts
sailingKieler: What do you have in mind? Being able to provide a basically arbitrary text region?

sailingKieler: > Trace things that are not covered by a property. Examples: comments, unassigned keywords.

Indeed. I'm just thinking whether we can add this later without breaking anything.

packages/langium/src/generator/generator-node.ts
spoenemann: > putting all the stuff in one file makes also little sense

It depends. Microsoft is doing that for parts of the VS Code and TypeScript code bases.

packages/langium/src/generator/generator-node.ts
spoenemann: You'd have to change the `TraceSourceSpec` type, which might be breaking.

I'm mentioning this because we have seen quite extensive use of the CST (Node Model) with Xtext in earlier projects, including large-scale generators. Of course we can argue that if you need the information in a generator, you should assign it to an AST property. So yes, we can wait for a concrete case where we need it, even if it partly breaks API.

sailingKieler: >I'm mentioning this because we have seen quite extensive use of the CST (Node Model) with Xtext in earlier projects

It's fine, I've added that as you're totally right; implementation is almost done ... will push another update soon ;-)

msujew: I agree with @spoenemann here. Basically all APIs that we have that are somehow related to pointing to a certain range usually accept properties, cst nodes and ranges. [The semantic token provider](https://github.com/langium/langium/blob/c7e15ee9663d37dab676dea4ffce479e3af0fea3/packages/langium/src/lsp/semantic-token-provider.ts#L79-L98) shows this quite well.

sailingKieler: This test case illustrates one example of what I can imagine to be relevant in practice: https://github.com/langium/langium/blob/bb5d1800fc850013ba5a023035d2564f0ec5d851/packages/langium/test/generator/generation-tracing.test.ts#L147-L170

sailingKieler:
I declare this PR to be complete now. 👊 😉

packages/langium/src/generator/generator-tracing.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    if (prev.fileURI || curr.fileURI) {
+        const prevURI = prev.fileURI;
+        const currURI = curr.fileURI;
+        const fileURI = prevURI && currURI && prevURI !== currURI ? `<unmergable text regions of ${prevURI}, ${currURI}>` : prevURI ?? currURI;
+        result.fileURI = fileURI;
+    }
+    return result;
+}
```
msujew, line 196: Nitpick: Missing newline at end of file.

packages/langium/src/generator/generator-node.ts
```diff
@@ -78,6 +97,64 @@ export class CompositeGeneratorNode {
...
+     * @param property the value property name (string) corresponding to `this` node's content,
+     *  e.g. if this node's content corresponds to some `string` or `number` property; is optional
+     *
+     * @param index the index of the value within a list property corresponding to `this` node's content,
+     * if the property contains a list of elements, is ignored otherwise,
+     * must not be given if no `property` is given ; is optinal
+     *
+     * @returns `this` {@link CompositeGeneratorNode} for convenience.
+     */
+    trace<T extends AstNode>(astNode: T, property?: Properties<T>, index?: number): this;
```
msujew, line 118: Minor: We should use multiple overloads for this, so that users cannot enter `property: undefined` with `index` set. We already validate against that anyway, so it would be good to let the overload reflect this.
```suggestion
    trace<T extends AstNode>(astNode: T): this;
    trace<T extends AstNode>(astNode: T, property: Properties<T>, index?: number): this;
```

packages/langium/test/generator/generation-tracing.test.ts
```diff
@@ -0,0 +1,505 @@
+/******************************************************************************
+ * Copyright 2022 TypeFox GmbH
```
msujew, line 2: Minor:
```suggestion
 * Copyright 2023 TypeFox GmbH
```

msujew (approved):
Looks good imo 👍

I have some minor comments, but afterwards this is good to go.

packages/langium/src/generator/generator-tracing.ts
sailingKieler: 👍

packages/langium/test/generator/generation-tracing.test.ts
sailingKieler: 👍

packages/langium/src/generator/generator-node.ts
msujew: It's not too important. Just wanted to have a quick-type-safety-win if possible. If it leads to other issues, we can keep it as it is for now.

sailingKieler: It definitely makes sense, I didn't think into this direction. Thanks for this hint!

sailingKieler: Just pushed an additional commit splitting the previous `(astNode: T, property?: Properties<T>, index?: number)` signatures into pairs of `(astNode: T, property?: Properties<T>)` and `(astNode: T, property: Properties<T>, index: number | undefined)`.

Let's see in action how well e.g. the TS content assist deals with them.

msujew: Yep, I think that's good :)

spoenemann (approved):
Great work!

dhuebner:
@sailingKieler @spoenemann 
Tracing API feels good to use. All the common cases are covered, for corner cases there are low-level "back-doors" open. With the naming here and there I have my own flavor, but this is okay :)

dhuebner (approved):

insafuhrmann (approved):
Looks good to me, API seems nice to me, fairly easy to slip also into existing code, very helpful comment documentation and tests.

PR #888, opened 2023-01-27
Support string escaping

### LINKED ISSUE #885: Support string escaping
We should support string escaping in the grammar language:

 - Add `\` escaping to the `STRING` rule in langium-grammar.langium
 - Process escaped characters in the `convertString` method of DefaultValueConverter
 - Adapt the syntax highlighting config of the grammar language to support escaped characters

### Discussed in https://github.com/langium/langium/discussions/884

<div type='discussions-op-text'>

<sup>Originally posted by **ballcoach12** January 26, 2023</sup>
The following rule generates an error in the grammar editor:

```
terminal S_BYTE_CHAR_STR returns string:
 	'\'' S_BYTE_CHAR_VALUE* '\''
 ;
```

The second backslash causes the following error: `Unexpected character value: ->\<-`.   Do I need to escape this character somehow?</div>

spoenemann:
Closes #885:
 - Add `\` escaping to the `STRING` rule in langium-grammar.langium
 - Process escaped characters in the `convertString` method of DefaultValueConverter
 - Adapt the syntax highlighting config of the grammar language to support escaped characters

### DIFF
--- a/packages/langium-vscode/data/langium.tmLanguage.json
+++ b/packages/langium-vscode/data/langium.tmLanguage.json
@@ -26,12 +26,22 @@
         {
             "name": "string.quoted.double.langium",
             "begin": "\"",
-            "end": "\""
+            "end": "\"",
+            "patterns": [
+				{
+					"include": "#string-character-escape"
+				}
+			]
         },
         {
             "name": "string.quoted.single.langium",
             "begin": "'",
-            "end": "'"
+            "end": "'",
+            "patterns": [
+				{
+					"include": "#string-character-escape"
+				}
+			]
         }
     ],
     "repository": {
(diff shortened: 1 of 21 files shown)


packages/langium-vscode/data/langium.tmLanguage.json
```diff
@@ -25,13 +25,11 @@
         },
         {
             "name": "string.quoted.double.langium",
-            "begin": "\"",
-            "end": "\""
+            "match": "\"(\\\\.|[^\"\\\\])*\""
```
msujew, line 28: Suggestion: We should try to keep the `begin` and `end` properties, since the `match` properties only highlights full matches, not partial matches (i.e. lines which have a `"` in them but no end delimiter).

Add the following rule to the `repository` object:

```json
"string-character-escape": {
    "name": "constant.character.escape.ts",
    "match": "\\\\(x[0-9A-Fa-f]{2}|u[0-9A-Fa-f]{4}|u\\{[0-9A-Fa-f]+\\}|[0-2][0-7]{0,2}|3[0-6][0-7]?|37[0-7]?|[4-7][0-7]?|.|$)"
},
```

```suggestion
            "begin": "\"",
            "end": "\"",
            "patterns": [
				{
					"include": "#string-character-escape"
				}
			]
```

msujew (commented):
Should we apply the changes also to the examples and the `generator-langium` template?

spoenemann:
The TextMate generator doesn't handle the new string regexp correctly. The function `getTerminalParts` creates a too large end value. I'm not sure how to solve that (probably in `CommentRegexVisitor`) – @msujew could you look into that?

spoenemann:
I forgot to fix the tests, I'll look into those.

msujew (approved):
Looks good to me 👍

PR #891, opened 2023-01-31
JsonSerializer: Added option to provide a custom serialization 'replacer' 

sailingKieler:
During my work on the generation tracing stuff, I temporarily used such a hook to modify the AST serialization.

I would love to have it in the code base, as it gives flexibility for use case specific serializations without the need to subclass the JsonSerializer. What do you think?

### DIFF
--- a/packages/langium/src/serializer/json-serializer.ts
+++ b/packages/langium/src/serializer/json-serializer.ts
@@ -13,6 +13,7 @@ import { AstNodeLocator } from '../workspace/ast-node-locator';
 export interface JsonSerializeOptions {
     space?: string | number
     refText?: boolean
+    replacer?: (key: string, value: unknown, defaultReplacer: (key: string, value: unknown) => unknown) => unknown
 }
 
 /**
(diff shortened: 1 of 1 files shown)


spoenemann:
Isn't it sufficient to override the `replacer` method of JsonSerializer?

sailingKieler:
> Isn't it sufficient to override the replacer method of JsonSerializer?

You would have to create an additional class und register that.
With this hook one can just call the default serializer like
```
const json = jsonSerializer.serialize(document.parseResult.value, {
    replacer: (key, value: unknown, defaultReplacer) => {
        if (key === 'foo') {
            return value.toString();
        } else {
            return defaultReplacer(key, value);
        }
    }
);
```
That was super helpful for me.

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -53,7 +54,10 @@ export class DefaultJsonSerializer implements JsonSerializer {
     }
 
     serialize(node: AstNode, options?: JsonSerializeOptions): string {
-        return JSON.stringify(node, (key, value) => this.replacer(key, value, options?.refText), options?.space);
+        const defaultReplacer = (key: string, value: unknown) => this.replacer(key, value, options?.refText);
```
spoenemann, line 57: While we're at it, could you change the third parameter of the `replacer` method to the JsonSerializeOptions object so we can extend it more easily in the future?

spoenemann (commented):
Ok the change is not obtrusive, 👍 for shipping it

packages/langium/src/serializer/json-serializer.ts
sailingKieler: I had that in mind, too, need to sync with #886 (generator) as there will be a conflict.

msujew (approved):
Looks good to me 👍

PR #898, opened 2023-02-06
Fix array assignment in fragment rules

msujew:
Closes https://github.com/langium/langium/issues/896

Previously, only the full property was assigned for unassigned subrule calls (both fragment/non-fragment). This change addresses this by correctly merging the array values.


montymxb (approved):
:octocat: approved! Nice clean fix, and the test case is excellent.

PR #903, opened 2023-02-07
Remove error results for missing LSP providers

msujew:
Inspired by [this discussion comment](https://github.com/langium/langium/discussions/899#discussioncomment-4897363). When we have multiple languages in our language server, we might send capabilities back to the language client, which might not be supported by every language.

In this case, the user receives an error message, about a missing service implementation. I believe it makes more sense to just keep the default behavior (e.g. the one that is seen when no provider is registered for any language) for language which don't have a special implementation of a service. That effectively means returning `[]`, `undefined` or `null` where fitting.


spoenemann (approved):
👍

PR #906, opened 2023-02-08
Add support for inlay hint providers

msujew:
Also adds optional options to our AST streaming methods that allow to specify a range. Only AST nodes within range will be returned in the resulting streams.


packages/langium/src/lsp/semantic-token-provider.ts
```diff
@@ -196,28 +198,28 @@ export abstract class AbstractSemanticTokenProvider implements SemanticTokenProv
         this.clientCapabilities = clientCapabilities;
     }
 
-    semanticHighlight(document: LangiumDocument, _params: SemanticTokensParams, cancelToken = CancellationToken.None): SemanticTokens {
+    async semanticHighlight(document: LangiumDocument, _params: SemanticTokensParams, cancelToken = CancellationToken.None): Promise<SemanticTokens> {
```
spoenemann, line 201: These changes are breaking. Are they "mild" enough to be included in 1.1 nonetheless?

packages/langium/src/services.ts
```diff
@@ -85,7 +86,7 @@ export type LangiumLspServices = {
     SemanticTokenProvider?: SemanticTokenProvider
     RenameProvider?: RenameProvider
     Formatter?: Formatter
-    SignatureHelp?: SignatureHelpProvider
+    SignatureHelpProvider?: SignatureHelpProvider
```
spoenemann, line 89: This is definitely breaking, I'm not sure how to handle it in the 1.x version stream.

packages/langium/src/services.ts
```diff
@@ -152,6 +153,7 @@ export type LangiumDefaultSharedServices = {
         CodeLensProvider?: CodeLensProvider
         DocumentLinkProvider?: DocumentLinkProvider
         ExecuteCommandHandler?: ExecuteCommandHandler
+        InlayHintProvider?: InlayHintProvider
```
spoenemann, line 156: Why are this and `CodeLensProvider`, `DocumentLinkProvider` shared service and not language-specific?

packages/langium/src/utils/cst-util.ts
```diff
@@ -74,6 +74,36 @@ export function toDocumentSegment(node?: CstNode): DocumentSegment | undefined {
     };
 }
 
+export enum RangeComparison {
+    Before = 0,
+    OverlapFront = 1,
+    Inside = 2,
+    OverlapBack = 3,
+    After = 4
```
spoenemann, line 82: If you order these differently, you could use `<` or `>` comparisons to check whether there's an overlap.

packages/langium/src/services.ts
msujew: All of these language services offer a way to "resolve" their respective results. For example, you can return an inlay hint just with it's text and position, and calculate all the other information once requested (for example the tooltip property if the user hovers over it). It's the same for the code lens and document link provider

For these special resolve requests, we don't get the URI of the original document, only the requested code lens, inlay hint or document link. That's the reason why these services are shared.

In theory we could remove the option to resolve these items, that would allow us to move them to the language specific services.

packages/langium/src/services.ts
msujew: I've reverted the change.

packages/langium/src/lsp/semantic-token-provider.ts
msujew: There is very little reason for developers to override these methods. I think it's mild enough for a 1.1.

packages/langium/src/services.ts
spoenemann: > In theory we could remove the option to resolve these items, that would allow us to move them to the language specific services.

packages/langium/src/utils/cst-util.ts
```diff
@@ -74,6 +74,36 @@ export function toDocumentSegment(node?: CstNode): DocumentSegment | undefined {
     };
 }
 
+export enum RangeComparison {
+    Before = 0,
+    After = 1,
+    OverlapFront = 2,
+    Inside = 3,
```
spoenemann, line 81: I'd put this to the last position because a partial overlap "is less than" a full overlap.

spoenemann (approved):
Thanks!

packages/langium/src/services.ts
msujew: I've created https://github.com/langium/langium/issues/925.

PR #907, opened 2023-02-08
Simplify property check for validation testing

msujew:
Currently, users always have to wrap their property name in an object, even when there's no index. This change makes the API a bit more flexible and allows users to just use a string.


spoenemann (approved):
:+1:

PR #910, opened 2023-02-09
Improve root build & watch scripts

msujew:
The new `build` and `watch` scripts automatically build the TypeScript code as well as the langium-vscode extension. Running both watcher processes in parallel required two terminals previously. Now it runs in the same terminal.


kaisalmen (approved):
Tested with Gitpod. Looks good! LGTM

kaisalmen:
Gitpod changes work fine! Thank you for fixing it. 🙂 LGTM 🎉

packages/langium-vscode/esbuild.js
```diff
@@ -20,7 +20,7 @@ require('esbuild').build({
     loader: { '.ts': 'ts' },
     external: ['vscode'], // the vscode-module is created on-the-fly and must be excluded.
     platform: 'node', // VSCode extensions run in a node process
-    sourcemap: true,
+    sourcemap: !minify,
```
spoenemann, line 23: Should we use different settings for dev and production?

msujew: Since `dev` is effectively without minification, I don't really see a reason right now.

PR #911, opened 2023-02-09
Simplify yeoman generator

msujew:
Removes a lot of complexity from the yeoman generator by moving a lot of it's functionality inside of methods/out of the loops themselves.

Also adds an end option that allows to directly open the project in vscode if installed.


spoenemann (approved):
:+1:

PR #916, opened 2023-02-10
Fix incorrect yeoman vscode-engines version

msujew:
Using `~` prevents vscode from picking up the extension compared to `^`.


kaisalmen (approved):
LGTM 🎉

PR #917, opened 2023-02-10
Support terminal lookahead

msujew:
In some cases (mostly for complex terminal definitions) it might make sense to create regexes with positive/negative lookahead support. While users could always use the direct regex syntax to do this, they lose the option to then use the normal terminal syntax. This change adds a `?=` (positive) and `?!` (negative) lookahead syntax to our grammar to improve on this.


spoenemann (approved):
👍

PR #927, opened 2023-02-13
Fix CST nodes pointing to stale AST nodes when parsed from actions

daumantas-kavolis-sensmetry:
This fixes CST node `element` pointing to stale AST nodes when parsed from an action. This is important for CST based services such as a formatter which can fail because keywords or properties are incorrectly assumed to be owned by another AST node. Removing the `element` assignment fails the test.


msujew:
Hey @daumantas-kavolis-sensmetry, thanks for contributing. Any specific reason why you closed this PR?

daumantas-kavolis-sensmetry:
Found some issues, the element should only be updated for action CSTs. As it is now, it breaks the rest of formatting.

Issue #926, opened 2023-02-13
Generate tasks.json to ensure grammar is generated before stating extension


dhuebner:
Sometime users start debugging the extension (F5) without generating/re-generating the grammar. 
This might lead to bad user experience especially by newcomers #913.

The solution could be to create a new 'Generate <language-name>' task with yo generator and start it as a `preLaunchTask` in  `Run Extension` launch configuration.

dhuebner:
Merged in master

PR #928, opened 2023-02-13
yeoman - Create tasks.json to generate grammar files

dhuebner:
Yeoman now generates a `tasks.json` file that contains a task to run `langium:generate` script. This task is defined as a `preLaunchTask`  in `Run extension` launch config.

See #926

### DIFF
--- a/packages/generator-langium/langium-template/.vscode/tasks.json
+++ b/packages/generator-langium/langium-template/.vscode/tasks.json
@@ -1,0 +1,21 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Build <%= language-id %>",
+            "command": "npm run langium:generate && npm run build",
+            "type": "shell",
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            },
+            "detail": "Langium: Generate grammar and build the <%= language-id %> language",
+            "icon": {
+                "color": "terminal.ansiGreen",
+                "id": "server-process"
+            }
+        }
+    ]
+}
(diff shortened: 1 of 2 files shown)


packages/generator-langium/langium-template/.vscode/tasks.json
```diff
@@ -0,0 +1,17 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Generate <%= language-id %>",
```
spoenemann, line 7: This is more than just generating, it also involves the TypeScript compiler. Rename to "Build"?

spoenemann (commented):
I'm not sure this is the best approach. If a user is already running the watchers for langium-cli and tsc, there's no need to run that task before every launch, and it would considerably increase the turnaround time for larger projects.

Note that there's the concept of a build task, which can be invoked by pressing cmd+shift+B:
https://code.visualstudio.com/docs/editor/tasks#_typescript-hello-world

I propose to configure the task as build task so users can decide themselves when to launch it.

dhuebner:
@spoenemann 
Thank you for your feedback!

> I'm not sure this is the best approach. If a user is already running the watchers for langium-cli and tsc, there's no need to run that task before every launch, and it would considerably increase the turnaround time for larger projects.

I was thinking about users that are starting a new project using yoeman and not about experts that have a long running large project. Maybe that could be a yoeman option?

> Note that there's the concept of a build task, which can be invoked by pressing cmd+shift+B: https://code.visualstudio.com/docs/editor/tasks#_typescript-hello-world
> 
> I propose to configure the task as build task so users can decide themselves when to launch it.

Then we need to describe it better in the documentation for people who just use F5 to start working after they changed the grammar.

I will remove the `preLaunchTask` entry and turn the task to a build task (default = true?).

I think, however, that the first contact with Langium should run as smoothly as possible to have the best first impression about the project.

packages/generator-langium/langium-template/.vscode/tasks.json
dhuebner: Done

spoenemann (approved):
:+1:

PR #932, opened 2023-02-14
Fix CST nodes pointing to stale AST nodes when parsed from subrules

daumantas-kavolis-sensmetry:
This fixes CST node `element` pointing to stale AST nodes when parsed from a subrule. This is important for CST based services such as a formatter which can fail because keywords or properties are incorrectly assumed to be owned by another AST node. Also tested on https://gitlab.com/sensmetry/public/sysml-2ls/-/tree/feat/formatting to make sure it didn't break anything else.

Opening this as draft for now since I could not find a generic way to apply this to any subrule nesting level as shown by the added failing test. I will mark it ready if the test case should be removed and addressed in a separate PR or there is a fix for it.


msujew:
Superseded by https://github.com/eclipse-langium/langium/pull/1547. cc @daumantas-kavolis-sensmetry in case you are interested in testing that PR.

PR #935, opened 2023-02-14
Allow completion providers to override options

msujew:
Developers currently need to override the language server to provide completion options. This change addresses this by allowing to override the options in the `CompletionProvider` itself.

### DIFF
--- a/packages/langium/src/lsp/language-server.ts
+++ b/packages/langium/src/lsp/language-server.ts
@@ -122,10 +122,7 @@ export class DefaultLanguageServer implements LanguageServer {
                     commands: commandNames
                 },
                 textDocumentSync: TextDocumentSyncKind.Incremental,
-                completionProvider: hasCompletionProvider ? {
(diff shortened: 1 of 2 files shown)


packages/langium/src/lsp/language-server.ts
```diff
@@ -120,7 +122,10 @@ export class DefaultLanguageServer implements LanguageServer {
                     commands: commandNames
                 },
                 textDocumentSync: TextDocumentSyncKind.Incremental,
-                completionProvider: hasCompletionProvider ? {} : undefined,
+                completionProvider: hasCompletionProvider ? {
+                    triggerCharacters: completionOptions.triggerCharacters,
+                    allCommitCharacters: completionOptions.allCommitCharacters
+                } : undefined,
```
spoenemann, line 128: Why not just `hasCompletionProvider ? completionOptions  : undefined`?

spoenemann (approved):

PR #942, opened 2023-02-17
Prevent generation on non-AST union types

msujew:
Closes https://github.com/langium/langium/issues/941
Closes https://github.com/langium/langium/issues/957
Closes https://github.com/langium/langium/issues/958

### DIFF
--- a/packages/langium/src/grammar/type-system/types-util.ts
+++ b/packages/langium/src/grammar/type-system/types-util.ts
@@ -195,13 +195,13 @@ export function findReferenceTypes(type: PropertyType): string[] {
     return [];
 }
 
-export function hasAstType(type: PropertyType): boolean {
+export function isAstType(type: PropertyType): boolean {
     if (isPropertyUnion(type)) {
-        return type.types.some(hasAstType);
+        return type.types.every(isAstType);
     } else if (isValueType(type)) {
         const value = type.value;
         if ('type' in value) {
-            return hasAstType(value.type);
+            return isAstType(value.type);
         } else {
             // Is definitely an interface type
             return true;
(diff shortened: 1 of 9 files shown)


sailingKieler (changes requested):
Great, thanks @msujew for the prompt reaction.

However, 2 wishes:
 * add a bit of prose to the definition of `AstTypeList` & `ValiditionChecks` on what it is supposed to describe/denote (daniel and me tried to guess it yesterday)
 * create at least a related unit test

msujew:
We don't really have a testing infrastructure in place for type testing. Since this is a type change (it doesn't actually change any runtime code), there's little need for runtime tests.

packages/langium/src/grammar/type-system/types-util.ts
```diff
@@ -194,3 +194,18 @@ export function findReferenceTypes(type: PropertyType): string[] {
     }
     return [];
 }
+
+export function hasAstType(type: PropertyType): boolean {
+    if (isPropertyUnion(type)) {
+        return type.types.some(hasAstType);
```
spoenemann, line 200: Should we better use `every`? A union that combines AST types with primitives is not usable as AST type.

spoenemann (commented):
I don't see #958 fixed, the invalid switch case in `computeIsSubtype` is still generated.

sailingKieler (approved):
I basically agree, although I would love to see some comments here and there simplifying to understand what's going on and what aim of the particular filter criteria is (e.g. `hasAstType(e.type)`)

Provided a few suggestions for improving the documentation

spoenemann:
AFAICT https://github.com/langium/langium/pull/942#pullrequestreview-1315362762 is still valid.

msujew:
@spoenemann Indeed, I pushed a fix for that. Thanks 👍

spoenemann (approved):
:+1:

PR #947, opened 2023-02-17
Improve type validation for type unions

msujew:
Closes https://github.com/langium/langium/issues/943

Fixes two main issues:

1. Allow to use type unions as return types of data type rules (if they are also a return type)
2. Resolve union types as early as possible during assignability checking. This allows to use type unions as property types of declared interfaces.

### DIFF
--- a/packages/langium/test/grammar/type-system/type-validator.test.ts
+++ b/packages/langium/test/grammar/type-system/type-validator.test.ts
@@ -26,20 +26,15 @@ describe('validate params in types', () => {
         terminal ID: /[a-zA-Z_][\\w_]*/;
         `.trim();
         // verify we only have 1 error, associated with a missing 'name' prop
-        const validation = await validate(prog);
+        const document = await parseDocument(grammarServices, prog);
+        let diagnostics: Diagnostic[] = await grammarServices.validation.DocumentValidator.validateDocument(document);
+        diagnostics = diagnostics.filter(d => d.severity === DiagnosticSeverity.Error);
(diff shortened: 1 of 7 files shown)


packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -26,15 +26,20 @@ describe('validate params in types', () => {
...
+                start: {
+                    line: 4,
+                    character: 8
+                },
+                end: {
+                    line: 4,
+                    character: 10
+                }
+            }
+        });
```
montymxb, line 42: This seems like an unnecessary change. I would roll this back, as it's not related to the PR. This change also alters the test semantics, making it looser than before (several diagnostics can occur, which was expressly forbidden before).

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -51,15 +56,20 @@ describe('validate params in types', () => {
...
+                start: {
+                    line: 5,
+                    character: 8
+                },
+                end: {
+                    line: 5,
+                    character: 34
+                }
+            }
+        });
```
montymxb, line 72: I would roll this back as well, same as above, the meaning has changed.

montymxb (changes requested):
Added some comments on extra test changes that should be reverted. Rest of the new test additions look nice however 👍 .

montymxb (approved):
:octocat: approved! Looks good 👍 .

PR #950, opened 2023-02-20
Mandatory properties are also optional

daumantas-kavolis-sensmetry:
This makes mandatory properties not trigger the missing property validation. The generated grammar and AST are still valid as mandatory properties are automatically created during parsing if they are missing.


msujew (approved):
Thanks, I was just about to fix this issue as well 👍

PR #951, opened 2023-02-21
Fix potentially infinite loop in get(Previous|Next)Node

daumantas-kavolis-sensmetry:
Fixes potentially infinite loop in `get(Previous|Next)Node` if `hidden = false` and the neighbouring CST node is hidden


msujew (approved):
Thanks, that makes a lot more sense than the existing logic 👍

PR #953, opened 2023-02-23
Add gitignore to yeoman generator

### LINKED ISSUE #918: suggest .gitignore for generated projects
It would helpful if a `.gitignore` would be scaffolded with the hello-world project, so that new users will know exactly what should be ignored from commit.

---

So far mine looks like this:

```conf
.vscode/
!.vscode/extensions.json
!.vscode/launch.json
node_modules/
out/
syntaxes/
src/language-server/generated/
```

Is that correct?

gfontorbe:
Closes #918 

Adds a .gitignore file for newly created project via the yeoman generator.

### DIFF
--- a/packages/generator-langium/langium-template/.gitignore
+++ b/packages/generator-langium/langium-template/.gitignore
@@ -1,0 +1,8 @@
+.vscode/
+!.vsode/extensions.json
+!.vscode/launch.json
+!.vscode/tasks.json
+node_modules/
+out/
+syntaxes/
+src/language-server/generated/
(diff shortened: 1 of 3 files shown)


packages/generator-langium/langium-template/.gitignore
```diff
@@ -0,0 +1,8 @@
+.vscode/
+!.vsode/extensions.json
```
msujew, line 2:
```suggestion
!.vscode/extensions.json
```

spoenemann (approved):
👍

PR #954, opened 2023-02-23
Add yeoman generator for web setup

### LINKED ISSUE #760: Add a command for the CLI to create a in-browser worker script of the language server
* [ ] Put a generator into `generator-langium` package
* [ ] Add a question during Langium generation which environment should be included: `vscode` or `web` (checkboxes)
* [ ] Make sure the resulting worker script is compatible with our Monaco React component (WIP)

gfontorbe:
Closes #760 

This adds an option in the yeoman generator for configuring a new project to run in a web browser using a Monaco editor.

I followed the tutorial on [Langium's documentation](https://langium.org/tutorials/langium_and_monaco/) for the structure of the project.

All the build steps are done by the generator so users only have to run `npm run serve` to have a functional project running on an express server.


packages/generator-langium/langium-template-web/src/static/setup.js
```diff
@@ -0,0 +1,58 @@
...
+
+buildWorkerDefinition('./monaco-editor-workers/workers', new URL('', window.location.href).href, false);
+
+MonacoEditorLanguageClientWrapper.addMonacoStyles('monaco-editor-styles');
+
+const client = new MonacoEditorLanguageClientWrapper();
+const editorConfig = client.getEditorConfig();
+editorConfig.setMainLanguageId('<%= language-id %>');
+
+editorConfig.setMonarchTokensProvider({
```
cdietrich, line 12: i wonder if we should properly build and or load the original generated (when option is set) monaco syntax here.
i did this here:
https://github.com/cdietrich/langium-1x-experiments/blob/029baeb4baf8103e8e3f924bcaaa13ad4e1c0078/src/static/setup.js#L13
https://github.com/cdietrich/langium-1x-experiments/blob/main/tsconfig-monarch.json
https://github.com/cdietrich/langium-1x-experiments/blob/029baeb4baf8103e8e3f924bcaaa13ad4e1c0078/package.json#L54
but as i am a node noob, no idea if there is a better idea.

gfontorbe: @cdietrich  Generating a `.js` file from the Monarch syntax and moving it to the public folder should be possible, I will look into that.

packages/generator-langium/langium-template-web/package.json
```diff
@@ -0,0 +1,73 @@
...
+    },
+    "dependencies": {
+        "chalk": "~4.1.2",
+        "chevrotain": "~10.4.2",
+        "commander": "~10.0.0",
+        "express": "~4.18.2",
+        "langium": "~1.1.0",
+        "monaco-editor-workers": "~0.34.2",
+        "monaco-editor-wrapper": "~1.6.0",
+        "vscode-languageclient": "~8.0.2",
```
kaisalmen, line 59: I was just wondering if we shouldn't just rely on the transitive dependencies `langium` provides. These versions will only get out of sync (forgotten to update) and increase maintenance.

packages/generator-langium/langium-template-web/src/static/setup.js
kaisalmen: Good idea, this will make this file shorter and more focused.

packages/generator-langium/src/index.ts
```diff
@@ -129,15 +141,35 @@ class LangiumGenerator extends Generator {
                 }
             );
         }
+
+        if (this.answers.targetEnvironment === 'web') {
```
kaisalmen, line 145: The other regular target is always created. So, web is always optional or do web want an exclusive or (XOR))?

packages/generator-langium/src/index.ts
```diff
@@ -129,15 +141,35 @@ class LangiumGenerator extends Generator {
...
-    install(): void {
+    async install(): Promise<void> {
         const extensionPath = this._extensionPath();
 
         const opts = { cwd: extensionPath };
         this.spawnCommandSync('npm', ['install'], opts);
         this.spawnCommandSync('npm', ['run', 'langium:generate'], opts);
         this.spawnCommandSync('npm', ['run', 'build'], opts);
+
+        if (this.answers.targetEnvironment === 'web') {
```
kaisalmen, line 170: See optional/XOR question.

packages/generator-langium/test/yeoman-generator.test.ts
```diff
@@ -33,6 +42,22 @@ describe('Check yeoman generator works', () => {
         context.cleanup(); // clean-up
     }, 120_000);
 
+    test('Should produce files for web environment', async () => {
```
kaisalmen, line 45: See optional/XOR question.

kaisalmen (commented):
Looking good already. There are only some remarks.

packages/generator-langium/langium-template-web/package.json
msujew: There is no transitive dependency, since this is the project that implements the language server.

kaisalmen: Sorry, you are right. For the `vscode-languageclient` there is none, but for `vscode-languageserver` and `vscode-uri` we could rely on the ones defined by langium.

packages/generator-langium/src/index.ts
gfontorbe: Most of the files generated are shared between `vscode` and `web` environment. 
I didn't want to duplicate the vscode files (and increase maintenance cost) for the web configuration. Instead I just overwrite what is needed.

kaisalmen: Ok, makes sense. Thanks for explaining.

packages/generator-langium/langium-template-web/package.json
gfontorbe: What about `chevrotain`, should it also be removed from this package.json?

packages/generator-langium/src/index.ts
```diff
@@ -11,6 +11,7 @@ import path from 'path';
 import which from 'which';
 
 const TEMPLATE_DIR = '../langium-template';
```
spoenemann, line 13: Let's rename the original folder `langium-template-vscode`?

spoenemann (commented):
Trying it in Gitpod, the result was the union of both templates:
Screenshot 2023-02-27 at 12 31 16

packages/generator-langium/langium-template-web/package.json
kaisalmen: Makes sense to me, too. @msujew WDYT?

msujew: I guess so, yes.

spoenemann (commented):
Did I understand correctly that the final package.json is the union of the `.package.json` files of the selected templates? Then I'd propose to remove all properties that are already contained in the core template to reduce maintenance effort, especially duplicate dependencies.

Further proposals:
 - Rename `src/language-server` to `src/language` (makes more sense in a CLI or library context)
 - Rename `src/cli/index.ts` to `src/cli/main.ts` (update in `bin/cli`)
 - Move `src/extension.ts` to `src/extension/main.ts` (update in package.json)
 - Move tasks.json to the core template

packages/generator-langium/langium-template-web/src/static/styles.css
```diff
@@ -0,0 +1,29 @@
+html,body {
+    background: rgb(33,33,33);
+    font-family: 'Lucida Sans', 'Lucida Sans Regular', 'Lucida Grande', 'Lucida Sans Unicode', Geneva, Verdana, sans-serif;
+    color: white;
+    /* for monaco */
+    margin: 8px auto;
+    width: 80%;
+    height: 80%;
```
msujew, line 8: Suggestion: This doesn't work well for smaller screen sizes (or non-maximized browser windows). The editor should never exceed the window:

<details>
<summary>Image</summary>

image

</details>

msujew (commented):
Looks quite good, though some changes require a bit more work.

packages/generator-langium/langium-template-web/src/static/styles.css
gfontorbe: Seems like the monaco editor is getting a fixed size in its inline styles.
@kaisalmen any idea where this is set?

kaisalmen: The wrapper uses `automaticLayout` as default (it wasn't always like that) and it should no longer require any fixes size overrides/outer definitions.

spoenemann (approved):
Looks really good! Last comments from me

packages/generator-langium/langium-template-vscode/.package.json
```diff
@@ -1,4 +1,7 @@
 {
+    "name": "<%= extension-name %>",
```
spoenemann, line 2: I wasn't specific enough here: `name` should be in the core template because it's used by npm as well.

spoenemann (approved):
Last detail: in .gitignore change `src/language-sever/generated` to `src/language/generated`

PR #955, opened 2023-02-24
Don't Format Documents with Lexer/Parser Errors

### LINKED ISSUE #912: grammar formatter deletes code
Langium version: 1.0.1
Package name:

## Steps To Reproduce

1. Put the following in a `.langium` file. Note the missing semi colon for the Param line:
   ```rb
   grammar Code
   
   Param: name=ID
   
   hidden terminal WS: /\s+/;
   ```
3. Run Lagium formatter in VS Code. It removes content, resulting in:
   ```rb
   grammar Code
   
   Param: name=ID
   
   hidden terminal WS
   ```

Link to code example:

https://langium.org/playground?grammar=OYJwhgthYgBAwgewCYFMBQ6AKNIC5YA7SVAXgEkARTACwEtk1DYAXVECO4gG1gHUAygQD0AHQDOAamEBudGw5cwvKiIDaAfTABaAF4BBbQC0AumtEB3DSYBUsoA&content=Q

## The current behavior

Strips content.

## The expected behavior

If there's a syntax error, it should avoid formatting, or (more difficult) should be forgiving and not delete content.

montymxb:
Closes #912 by producing no formatting when a given document has either lexer or parser errors. Just to add there's no checking of error diagnostics, since these can still be produced for a syntactically valid document.

The result is that trying to format the previously given example produces no changes, at least not until the document has been corrected. Gif below for reference.

We should also make sure to publish a patched version onto the playground as well.

Feb-24-2023 10-11-08

### DIFF
--- a/packages/langium/src/lsp/formatter.ts
+++ b/packages/langium/src/lsp/formatter.ts
@@ -81,7 +81,7 @@ export abstract class AbstractFormatter implements Formatter {
      * @param range Formatting range to check for safety
      * @returns Whether the given formatting range does not overlap with or follow any regions with an error
      */
-    private isFormatRangeErrorFree(document: LangiumDocument, range: Range): boolean {
+    protected isFormatRangeErrorFree(document: LangiumDocument, range: Range): boolean {
         const pr = document.parseResult;
         if (pr.lexerErrors.length || pr.parserErrors.length) {
             // collect the earliest error line from either
--- a/packages/langium/src/lsp/formatter.ts
+++ b/packages/langium/src/lsp/formatter.ts
@@ -74,9 +74,15 @@ export abstract class AbstractFormatter implements Formatter {
         }
     }
 
-    formatDocumentRange(document: LangiumDocument, params: DocumentRangeFormattingParams): MaybePromise<TextEdit[]> {
+    /**
+     * Returns whether a range for a given document is error free, i.e. safe to format
+     *
+     * @param document Document to inspect for lexer & parser errors that may produce an unsafe range
+     * @param range Formatting range to check for safety
+     * @returns Whether the given formatting range does not overlap with or follow any regions with an error
+     */
+    private isFormatRangeErrorFree(document: LangiumDocument, range: Range): boolean {
         const pr = document.parseResult;
-
         if (pr.lexerErrors.length || pr.parserErrors.length) {
             // collect the earliest error line from either
             const earliestErrLine = Math.min(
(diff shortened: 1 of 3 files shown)


packages/langium/test/grammar/formatting.test.ts
```diff
@@ -0,0 +1,121 @@
+/******************************************************************************
+ * Copyright 2023 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { beforeAll, describe, test, expect } from 'vitest';
+import { createLangiumGrammarServices, EmptyFileSystem } from 'langium';
+import { expectFormatting, expectFunction } from 'langium/test';
```
msujew, line 9: For some reason does this style of importing break the `expectFunction` behavior. The following suggestion should fix it though.

```suggestion
import { createLangiumGrammarServices, EmptyFileSystem } from '../../src';
import { expectFormatting } from '../../src/test';
```

packages/langium/src/lsp/formatter.ts
```diff
@@ -64,10 +64,36 @@ export abstract class AbstractFormatter implements Formatter {
...
+            // safe to format
+            return this.doDocumentFormat(document, params.options);
+        } else {
+            // don't format a potentially broken document, return no edits
+            return [];
+        }
     }
 
     formatDocumentRange(document: LangiumDocument, params: DocumentRangeFormattingParams): MaybePromise<TextEdit[]> {
+        const pr = document.parseResult;
```
msujew, line 78: Suggestion: I would like to see the same logic applied to the `formatDocumentOnType` method.

packages/langium/src/lsp/formatter.ts
```diff
@@ -64,10 +64,36 @@ export abstract class AbstractFormatter implements Formatter {
...
+                earliestErrLine = Math.min(lexerError.line ?? Number.MAX_VALUE, earliestErrLine);
+            }
+            for (const parserError of pr.parserErrors) {
+                earliestErrLine = Math.min(parserError.token.startLine ?? Number.MAX_VALUE, earliestErrLine);
+            }
+
+            // if the earliest error line occurs before or at the end line of the range, then don't format
+            if (Math.min(earliestErrLine, params.range.end.line) === earliestErrLine) {
+                return [];
+            }
```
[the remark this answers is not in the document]
montymxb: That's perfect! Yeah let's roll with this.

packages/langium/test/grammar/formatting.test.ts
montymxb: The import of `expectFunction` is needed to override how `expectFormatting` behaves below. Without it the tests fail on my end. This is interesting if you're seeing the exact opposite however, as there may be something up with the implementation based on platform then.

montymxb: Also going with `../../src/test` causes the `expectFormatting` usage to break on my end.

montymxb: Nvm, got it, helps rectify the issue with `expectFunction` needing to used in the first place

packages/langium/src/lsp/formatter.ts
```diff
@@ -74,33 +74,52 @@ export abstract class AbstractFormatter implements Formatter {
...
 
-    formatDocumentRange(document: LangiumDocument, params: DocumentRangeFormattingParams): MaybePromise<TextEdit[]> {
+    /**
+     * Returns whether a range for a given document is error free, i.e. safe to format
+     *
+     * @param document Document to inspect for lexer & parser errors that may produce an unsafe range
+     * @param range Formatting range to check for safety
+     * @returns Whether the given formatting range does not overlap with or follow any regions with an error
+     */
+    private isFormatRangeErrorFree(document: LangiumDocument, range: Range): boolean {
```
msujew, line 84: We should never declare a method as `private`, as that makes overriding it unnecessarily complicated. Please use `protected`.

montymxb: That's fair, one moment...

msujew (approved):
Looks good to me, thanks 👍 Don't forget to squash before merging!

PR #968, opened 2023-02-27
Added a validator to check for duplicate names

emilkrebs:
The current version of statemachine allows you to create duplicate events or states with the same name. 
This throws some problems when writing generators or previews for the statemachine. 
So I wrote a validator to check for duplicates: [link to the file](https://github.com/langium/langium/blob/emilkrebs/statemachineFixDuplicates/examples/statemachine/src/language-server/statemachine-validator.ts).

Here is a code example to replicate the bug:

```
statemachine TrafficLight

events 
    switchCapacity
    next

initialState PowerOff

state PowerOff
    switchCapacity => RedLight
end

state RedLight
    switchCapacity => PowerOff
    next => GreenLight
end

state YellowLight
    switchCapacity => PowerOff
    next => RedLight
end

state GreenLight
    switchCapacity => PowerOff
    next => YellowLight
end

// this would work
state GreenLight 
    switchCapacity => PowerOff
    next => YellowLight
end

// this would work either
state switchCapacity
    switchCapacity => PowerOff
    next => YellowLight
end
```

### DIFF
--- a/examples/statemachine/src/language-server/statemachine-validator.ts
+++ b/examples/statemachine/src/language-server/statemachine-validator.ts
@@ -5,20 +5,25 @@
  ******************************************************************************/
 
 import { ValidationAcceptor, ValidationChecks } from 'langium';
-import { State, StatemachineAstType } from './generated/ast';
+import { State, Statemachine, StatemachineAstType } from './generated/ast';
 import type { StatemachineServices } from './statemachine-module';
 
 export function registerValidationChecks(services: StatemachineServices) {
     const registry = services.validation.ValidationRegistry;
     const validator = services.validation.StatemachineValidator;
     const checks: ValidationChecks<StatemachineAstType> = {
-        State: validator.checkStateNameStartsWithCapital
+        State: validator.checkStateNameStartsWithCapital,
+        Statemachine: validator.checkUniqueStatesAndEvents
     };
     registry.register(checks, validator);
 }
 
 export class StatemachineValidator {
-
+    /**
+     * Checks if the state name starts with a capital letter.
+     * @param state the state to check
+     * @param accept the acceptor to report errors
+     */
     checkStateNameStartsWithCapital(state: State, accept: ValidationAcceptor): void {
         if (state.name) {
             const firstChar = state.name.substring(0, 1);
@@ -26,6 +31,30 @@ export class StatemachineValidator {
                 accept('warning', 'State name should start with a capital letter.', { node: state, property: 'name' });
             }
         }
+    }
+    /**
+     * Checks if there are duplicate state and event names.
+     * @param statemachine the statemachine to check
+     * @param accept the acceptor to report errors
+     */
+    checkUniqueStatesAndEvents(statemachine: Statemachine, accept: ValidationAcceptor): void {
+        const variables: string[] = [];
+
+        // let's check for duplicate state names
+        statemachine.states.forEach(state => {
+            if (variables.find(name => name === state.name)) {
+                accept('error',  `Name '${state.name}' does already exist.`, {node: state, property: 'name'});
+            }
+            variables.push(state.name);
+        });
+
+        // let's check for duplicate event names
+        statemachine.events.forEach(event => {
+            if (variables.find(name => name === event.name)) {
+                accept('error',  `Name '${event.name}' does already exist.`, {node: event, property: 'name'});
+            }
+            variables.push(event.name);
+        });
     }
 
 }


examples/statemachine/src/language-server/statemachine-validator.ts
```diff
@@ -27,5 +32,29 @@ export class StatemachineValidator {
             }
         }
     }
+    /**
+     * Checks if there are duplicate state and event names.
+     * @param statemachine the statemachine to check
+     * @param accept the acceptor to report errors
+     */
+    checkUniqueStatesAndEvents(statemachine: Statemachine, accept: ValidationAcceptor): void {
```
emilkrebs, line 40: Here is the validator which checks for duplicates

examples/statemachine/src/language-server/statemachine-validator.ts
```diff
@@ -5,20 +5,25 @@
...
+import { State, Statemachine, StatemachineAstType } from './generated/ast';
 import type { StatemachineServices } from './statemachine-module';
 
 export function registerValidationChecks(services: StatemachineServices) {
     const registry = services.validation.ValidationRegistry;
     const validator = services.validation.StatemachineValidator;
     const checks: ValidationChecks<StatemachineAstType> = {
-        State: validator.checkStateNameStartsWithCapital
+        State: validator.checkStateNameStartsWithCapital,
+        Statemachine: validator.checkUniqueStatesAndEvents
```
emilkrebs, line 16: I added the validator to the checks

examples/statemachine/src/language-server/statemachine-validator.ts
```diff
@@ -27,5 +32,29 @@ export class StatemachineValidator {
             }
         }
     }
+    /**
+     * Checks if there are duplicate state and event names.
+     * @param statemachine the statemachine to check
+     * @param accept the acceptor to report errors
+     */
+    checkUniqueStatesAndEvents(statemachine: Statemachine, accept: ValidationAcceptor): void {
+        const variables: string[] = [];
```
msujew, line 41: There are two issues with this approach of checking duplicated elements. One is semantic in nature, the other is more technical.

1. We usually want to show both occurrences of duplicated elements. Right now the error is only raised on second element, never the first.
2. Using an array and `find` leads to [quadratic runtime complexity](https://jarednielsen.com/big-o-quadratic-time-complexity/). While it generally works, it's not a good showcase for users who are looking how to implement such a validation.

We have an optimized of this algorithm version available [here](https://github.com/langium/langium/blob/712bd01d58dc3c683ccbb27f813e1c0e7fc941f6/packages/langium/src/grammar/validation/validator.ts#L165-L189) which runs in linear time and addresses the first issue as well.

examples/statemachine/src/language-server/statemachine-validator.ts
```diff
@@ -27,5 +32,35 @@ export class StatemachineValidator {
...
     }
-
+    /**
+     * Checks if there are duplicate state and event names.
+     * @param statemachine the statemachine to check
+     * @param accept the acceptor to report errors
+     */
+    checkUniqueStatesAndEvents(statemachine: Statemachine, accept: ValidationAcceptor): void {
+        // check for duplicate state and event names and add them to the map
+        const names = new Map<string, State | Event | undefined>();
```
msujew, line 42: Minor: The return type of the `get` function of `Map` is already `T | undefined`. We don't need it explicitly here.
```suggestion
        const names = new Map<string, State | Event>();
```

examples/statemachine/src/language-server/statemachine-validator.ts
```diff
@@ -27,5 +32,35 @@ export class StatemachineValidator {
...
+        for (const state of statemachine.states) {
```
msujew, line 43: This new approach is better, but it still suffers from a few issues:

1. The first element is littered with duplicate error messages, one for each duplicated element. However, one element should only ever raise one error.
image
2. Differentiating between duplicated states and events leads to unexpected error messages, see the image above. All elements should be checked in the same loop and receive the same error message.

examples/statemachine/src/language-server/statemachine-validator.ts
```diff
@@ -28,4 +33,27 @@ export class StatemachineValidator {
...
+                const duplicate = names.get(symbol.name);
+                if (duplicate && !duplicates.includes(symbol.name)) {
+                    duplicates.push(symbol.name);
+                    accept('error', `Duplicate identifier name: ${symbol.name}`, { node: duplicate, property: 'name' });
+                }
+                accept('error', `Duplicate identifier name: ${symbol.name}`, { node: symbol, property: 'name' });
+            } else {
+                names.set(symbol.name, symbol);
+            }
+        }
```
msujew, line 57: We're getting there! With `includes` we suffer again from quadratic complexity, since we need to look through the whole array of `duplicates` to confirm that an item doesn't exist in the array yet. This would be an optimized version:
```suggestion
        const names = new MultiMap<string, State | Event>();
        const allSymbols = [...statemachine.states, ...statemachine.events];
        const duplicates: string[] = [];
        for (const symbol of allSymbols) {
            names.add(symbol.name, symbol);
        }
        for (const [name, symbols] of names.entriesGroupedByKey()) {
            if (symbols.length > 1) {
                for (const symbol of symbols) {
                    accept('error', `Duplicate identifier name: ${name}`, { node: symbol, property: 'name' });
                }
            }
        }
```

emilkrebs: Thanks! I added this approach to arithmetics and statemachine!

msujew (approved):
Thanks, looks good to me 👍

PR #972, opened 2023-02-28
Add utilities for resolving cross-references

Lotes:
I added two generic types that helped me a lot during the resolution of the cross-references. It assigns all possible types to the scope provider variables, so you can handle one by one until you reach NEVER, which is guarded by `assertUnreachable(....);`.

The big advantage is that you immediately get compile time feedback if you left out some case. Isn't this great?!

You can now add case by case and be sure that every scenario is handled :)...

# Example

```typescript
    override getScope(context: ReferenceInfo): Scope {
        const container = context.container as AstNodeTypesWithCrossReferences<SqlAstType>;
        if(isTableDefinition(container)) {
            const property = context.property as CrossReferencesOfAstNodeType<TableDefinition>;
            switch(property) {
                case 'schemaName': {
                    return this.getSchemasFromGlobalScope(context);
                }
                default:
                    assertUnreachable(property); //will complain only if you add a new crossers to TableDefinition without adding a case in this switch
            }
        } else if(isPrimaryKeyDefinition(container)) {
            const property = context.property as CrossReferencesOfAstNodeType<PrimaryKeyDefinition>;
            switch(property) {
                case 'primaryKeys':
                    const tableDef = getContainerOfType(container, isTableDefinition)!;
                    return this.streamColumnDefinitions(tableDef.columns.filter(isColumnDefinition));
                default:
                    assertUnreachable(property); //same...
            }
        } else {
            assertUnreachable(container); //will complain because there are more than 2 AST nodes with cross-references
        }

        return super.getScope(context);
    }            
```

### DIFF
--- a/packages/langium-cli/src/generator/ast-generator.ts
+++ b/packages/langium-cli/src/generator/ast-generator.ts
@@ -44,7 +44,7 @@ function generateAstReflection(config: LangiumConfig, astTypes: AstTypes): Gener
     const crossReferenceTypes = buildCrossReferenceTypes(astTypes);
     const reflectionNode = new CompositeGeneratorNode();
 
-    reflectionNode.append(`export interface ${config.projectName}AstType {`, NL);
+    reflectionNode.append(`export type ${config.projectName}AstType {`, NL);
     reflectionNode.indent(astTypeBody => {
         for (const type of typeNames) {
             astTypeBody.append(type, ': ', type, NL);
--- a/packages/langium-cli/src/generator/ast-generator.ts
+++ b/packages/langium-cli/src/generator/ast-generator.ts
@@ -44,7 +44,7 @@ function generateAstReflection(config: LangiumConfig, astTypes: AstTypes): Gener
     const crossReferenceTypes = buildCrossReferenceTypes(astTypes);
     const reflectionNode = new CompositeGeneratorNode();
 
-    reflectionNode.append(`export type ${config.projectName}AstType {`, NL);
+    reflectionNode.append(`export type ${config.projectName}AstType = {`, NL);
     reflectionNode.indent(astTypeBody => {
         for (const type of typeNames) {
             astTypeBody.append(type, ': ', type, NL);
--- a/packages/langium/src/syntax-tree.ts
+++ b/packages/langium/src/syntax-tree.ts
@@ -263,6 +263,6 @@ export type AstTypeList<T> = Record<keyof T, AstNode>;
  * Returns all types that contain cross-references, A is meant to be the interface `XXXAstType` fromm your generated `ast.ts` file.
  * Meant to be used during cross-reference resolution in combination with `assertUnreachable(context.container)`.
  */
-export type AstNodeTypesWithCrossReferences<E, A extends AstTypeList<E>> = {
+export type AstNodeTypesWithCrossReferences<A extends AstTypeList<A>> = {
     [T in keyof A]: CrossReferencesOfAstNodeType<A[T]> extends never ? never : A[T]
 }[keyof A];
--- a/packages/langium/src/syntax-tree.ts
+++ b/packages/langium/src/syntax-tree.ts
@@ -257,6 +257,6 @@ export type CrossReferencesOfAstNodeType<N extends AstNode> = (
  * Returns all types that contain cross-references, A is meant to be the interface `XXXAstType` fromm your generated `ast.ts` file.
  * Meant to be used during cross-reference resolution in combination with `assertUnreachable(context.container)`.
  */
-export type AstNodeTypesWithCrossReferences<A> = {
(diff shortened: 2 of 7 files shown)


Lotes:
Wait I forgot to test the refactoring of these types. One change will come for sure!

Lotes:
Now ready for review!

packages/langium/src/syntax-tree.ts
```diff
@@ -240,3 +240,23 @@ export interface RootCstNode extends CompositeCstNode {
...
+type ExtractKeysOfValueType<T, K> = { [I in keyof T]: T[I] extends K ? I : never }[keyof T];
+
+/**
+ * Returns the property names (!) of an AstNode that are cross-references.
+ * Meant to be used during cross-reference resolution in combination with `assertUnreachable(context.property)`.
+ */
+export type CrossReferencesOfAstNodeType<N extends AstNode> = (
+    ExtractKeysOfValueType<N, Reference|undefined>
+    | ExtractKeysOfValueType<N, Array<Reference|undefined>|undefined>
+) & Record<string, never>;
```
spoenemann, line 254: When I apply this to the `ParserRule` type of the grammar language, I get

```
type Bar = ("hiddenTokens" & Record<string, never>) | ("returnType" & Record<string, never>)
```

That looks strange.

packages/langium/src/syntax-tree.ts
```diff
@@ -240,3 +240,23 @@ export interface RootCstNode extends CompositeCstNode {
...
+
+/**
+ * Returns all types that contain cross-references, A is meant to be the interface `XXXAstType` fromm your generated `ast.ts` file.
```
spoenemann, line 257: We have the type `AstTypeList` defined in the same file for this. Proposal: move that type right before these new definitions and add a JSDoc comment to it.

spoenemann (commented):
Thanks! While you're at it, could you change the generated `XXXAstType` from `interface` to `type` in langium-cli?

packages/langium/src/syntax-tree.ts
Lotes: The reason is the following listing error:
Bildschirmfoto 2023-03-01 um 09 00 36
But I will use `{}` again and suppress the rule, then it works the best.

packages/langium/src/syntax-tree.ts
Lotes: Nice! Done

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -44,7 +44,7 @@ function generateAstReflection(config: LangiumConfig, astTypes: AstTypes): Gener
     const crossReferenceTypes = buildCrossReferenceTypes(astTypes);
     const reflectionNode = new CompositeGeneratorNode();
 
-    reflectionNode.append(`export interface ${config.projectName}AstType {`, NL);
+    reflectionNode.append(`export type ${config.projectName}AstType {`, NL);
```
spoenemann, line 47:
```suggestion
    reflectionNode.append(`export type ${config.projectName}AstType = {`, NL);
```

packages/langium/src/syntax-tree.ts
```diff
@@ -240,3 +238,31 @@ export interface RootCstNode extends CompositeCstNode {
...
+/**
+ * Represents the enumeration-like type, that lists all AstNode types of your grammar.
+ */
+export type AstTypeList<T> = Record<keyof T, AstNode>;
+
+/**
+ * Returns all types that contain cross-references, A is meant to be the interface `XXXAstType` fromm your generated `ast.ts` file.
+ * Meant to be used during cross-reference resolution in combination with `assertUnreachable(context.container)`.
+ */
+export type AstNodeTypesWithCrossReferences<E, A extends AstTypeList<E>> = {
```
spoenemann, line 266: Why are there two type parameters now? Can't it be `A extends AstTypeList<A>`?

spoenemann:
Please regenerate all grammars after the CLI change (`npm run langium:generate` in the root)

packages/langium-cli/src/generator/ast-generator.ts
Lotes: Sorry missed that

packages/langium/src/syntax-tree.ts
Lotes: Done, thx. Brilliant!

Lotes:
> Please regenerate all grammars after the CLI change (`npm run langium:generate` in the root)

Weird. It doesn't change anything in the staging area...

Lotes:
Added the new ASTs

Lotes:
I think it would be neat to also have the type of the crossref... so you can ensure that the right stuff is processed...

spoenemann (approved):
I think this is already good

PR #973, opened 2023-03-02
Fixes for deep and declared hierarchies

### LINKED ISSUE #949: Type inference issues
Langium version: 1.1

There are still some issues with type inference which results in missing `$type` values and false positive property compatibility checks. Even a simple type hierarchy can result in invalid generated AST. Not sure where the issue stems from.

## Steps To Reproduce

1. Run `pnpm run langium:generate` in https://gitlab.com/sensmetry/public/sysml-2ls/-/tree/f4959ffa5cf29e3182c60c11c4b68e59d9a70771 with #946 (or comment out diagnostic filtering and `toAstTypesString` patch in `scripts/generate.ts` and run `pnpm run grammar:generate`)

Link to code example:

https://gitlab.com/sensmetry/public/sysml-2ls/-/tree/f4959ffa5cf29e3182c60c11c4b68e59d9a70771

Some explicit interfaces:

```ts
interface ElementReference {
    chain: @Element[];
}
interface TypeReference extends ElementReference {}
interface ClassifierReference extends TypeReference {}
interface FeatureReference extends TypeReference {}
interface MetaclassReference extends TypeReference {}
interface ConjugatedPortReference extends ClassifierReference {}

...

interface MetadataUsage extends MetadataFeature, ItemUsage {}
```

## The current behavior

```ts
export interface ElementReference extends AstNode {
    readonly $type: 'ElementReference' | 'FeatureReference' | 'MetaclassReference' | 'TypeReference';
}

export interface TypeReference extends ElementReference {
    readonly $type: 'ClassifierReference' | 'ConjugatedPortReference' | 'FeatureReference' | 'MetaclassReference' | 'TypeReference';
}
```

```
src/language-server/grammar/KerML.langium:342:21 - The assigned type 'ClassifierReference[]' is not compatible with the declared property 'specializes' of type 'TypeReference[]'.
...
src/language-server/grammar/SysML.langium:287:4 - The assigned type 'MetadataUsage[]' is not compatible with the declared property 'metadata' of type 'MetadataFeature[]'.
```

## The expected behavior

```ts
export interface ElementReference extends AstNode {
    readonly $type: 'ClassifierReference' | 'ConjugatedPortReference' | 'ElementReference' | 'FeatureReference' | 'MetaclassReference' | 'TypeReference';
}

export interface TypeReference extends ElementReference {
    readonly $type: 'ClassifierReference' | 'ConjugatedPortReference' | 'FeatureReference' | 'MetaclassReference' | 'TypeReference';
}
```

daumantas-kavolis-sensmetry:
Fixes #949:

- `$type` missing some subtype names occasionally - names are collected recursively
- assignment validation using inferred hierarchy which is very likely to be incomplete with declared types - replaced the inferred property type with a matching declared type for assignment checks only

This does not fix `$container` being non-optional on grammar entry types and its supertypes.

### DIFF
--- a/packages/langium/test/grammar/type-system/inferred-types.test.ts
+++ b/packages/langium/test/grammar/type-system/inferred-types.test.ts
@@ -543,6 +543,44 @@ describe('types of `$container` and `$type` are correct', () => {
             }
         `);
     });
+
+    test('types of `$type` for declared deep hierarchies', async () => {
+        await expectTypes(`
+            interface A {}
+            interface B extends A {}
+            interface C extends B {}
+            interface D extends C {}
+            interface E extends C {}
+            interface F extends D {}
+            interface G extends D {}
+            interface H extends A {}
+        `, expandToString`
+            export interface A extends AstNode {
+                readonly $type: 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G' | 'H';
+            }
+            export interface B extends A {
+                readonly $type: 'B' | 'C' | 'D' | 'E' | 'F' | 'G';
+            }
+            export interface H extends A {
+                readonly $type: 'H';
+            }
+            export interface C extends B {
+                readonly $type: 'C' | 'D' | 'E' | 'F' | 'G';
+            }
+            export interface D extends C {
+                readonly $type: 'D' | 'F' | 'G';
+            }
+            export interface E extends C {
+                readonly $type: 'E';
+            }
+            export interface F extends D {
+                readonly $type: 'F';
+            }
+            export interface G extends D {
+                readonly $type: 'G';
+            }
+        `);
+    });
 });
 
 // https://github.com/langium/langium/issues/744
--- a/packages/langium/test/grammar/type-system/type-validator.test.ts
+++ b/packages/langium/test/grammar/type-system/type-validator.test.ts
@@ -497,4 +497,28 @@ describe('Property types validation takes in account types hierarchy', () => {
         });
     });
 
+    test('No false positive on declared type assignment', async () => {
+        const validation = await validate(`
+        interface A {}
+        interface B extends A {}
+        interface C extends B {}
+        interface D extends C {}
+        interface E extends C {}
+        interface F extends D {}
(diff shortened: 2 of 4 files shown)


packages/langium/test/grammar/type-system/inferred-types.test.ts
```diff
@@ -543,6 +543,44 @@ describe('types of `$container` and `$type` are correct', () => {
             }
         `);
     });
+
+    test('types of `$type` for declared deep hierarchies', async () => {
```
msujew, line 547: Suggestion: It might make sense to split this test into multiple tests that test for the following scenarios:

1. Linear (`C extends B, B extends A`)
2. Deviating (`B extends A, C extends A`)
3. Joining (`C extends A and B`)
4. A mix of all of the above.

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -497,4 +497,27 @@ describe('Property types validation takes in account types hierarchy', () => {
...
+    test('No false positive on declared type assignment', async () => {
+        const validation = await validate(`
+        interface A {}
+        interface B extends A {}
+        interface C extends B {}
+        interface D extends C {}
+        interface E extends C {}
+        interface F extends D {}
+        interface G extends D {}
+        interface H extends A {}
```
msujew, line 509: Suggestion: I feel like there are a few interfaces in here which only add noise. We can probably just trim  of them down to `A, B, C` and accomplish the same effect.

msujew (commented):
Thanks, that's very helpful! I have a few improvements on the testing code, but otherwise this looks already pretty good 👍

daumantas-kavolis-sensmetry:
I have split the `$type` tests into smaller ones but only the last one failed without the fix, and removed as much as possible from the assignment test such that it failed without a fix.

daumantas-kavolis-sensmetry:
Wait, the `$type` test doesn't fail on `main`, I'll look into it.

msujew (approved):
Looks good to me 👍

PR #995, opened 2023-03-21
Add prism syntax highlighting generator

msujew:
Closes https://github.com/langium/langium/issues/993
Also addresses the discussion https://github.com/langium/langium/discussions/785

See this generator in use in [this file](https://github.com/langium/langium-website/blob/main/hugo/static/prism/langium-prism.js) for the corresponding website PR https://github.com/langium/langium-website/pull/140. As outlined by the comment that is generated into the file, this is a best guess attempt at syntax highlighting and needs to be extended for the users/devs needs. How this can be done can be seen in the linked file.


montymxb (approved):
:octocat: approved! Looks good, short and sweet 😎

billiegoose:
Oooh! I can't wait to try this out!

cdietrich:
@msujew was there a reasoning behind using `Prism.languages.camelCaseGrammar = ...`
and not
`Prism.languages["language-id"]= ...`

msujew:
@cdietrich No specific one, no. I'm fine with changing it, though I assumed that adopters would modify the generated file anyway, given that the generated file is doing very little beside keyword highlighting.

cdietrich:
https://github.com/eclipse-langium/langium/pull/1617

PR #1000, opened 2023-03-27
Fix document highlighting for non-local references

### LINKED ISSUE #998: Highlighter LSP service yields incorrect results when declaration is in another file
Langium version: 1.1.0
Package name: Langium

## Steps To Reproduce

1. Reference an element from another file
2. Click on the reference text. It doesn't highlight as expected.

Link to code example: `requirements` example language

image

## The current behavior

The node is either not highlighted at all, or highlighting appears at unexpected locations.

## The expected behavior

Highlighting is correctly performed on the selected reference.

msujew:
Closes #998

When designing the last refactoring of the `References` services, we chose to implement local and global references separately. This was a bad idea after all. We now just implement a filtering for selected documents.

Also adds testing methods for the `DocumentHighlightProvider`.


packages/langium/src/references/references.ts
```diff
@@ -46,7 +47,13 @@ export interface References {
 }
 
 export interface FindReferencesOptions {
-    onlyLocal?: boolean;
```
spoenemann, line 49: Shall we keep this and mark it deprecated?

packages/langium/src/references/references.ts
```diff
@@ -101,50 +108,18 @@ export class DefaultReferences implements References {
...
-                        sourcePath: this.nodeLocator.getAstNodePath(cstNode.element),
-                        targetUri: getDocument(targetNode).uri,
-                        targetPath: this.nodeLocator.getAstNodePath(targetNode),
-                        segment: toDocumentSegment(cstNode),
-                        local: equalURI(getDocument(cstNode.element).uri, getDocument(targetNode).uri)
-                    });
-                }
-            });
-        });
+        refs.push(...indexReferences);
```
spoenemann, line 122: The protected method removals need to be documented as breaking changes.

packages/langium/src/test/langium-test.ts
```diff
@@ -17,17 +17,23 @@ import { findNodeForProperty } from '../utils/grammar-util';
 import { SemanticTokensDecoder } from '../lsp/semantic-token-provider';
 import { TextDocument } from 'vscode-languageserver-textdocument';
 import { expect } from 'vitest';
-import { BuildOptions } from '../workspace/document-builder';
 
-export function parseHelper<T extends AstNode = AstNode>(services: LangiumServices): (input: string, buildOptions?: BuildOptions) => Promise<LangiumDocument<T>> {
+export interface ParseHelperOptions {
+    validationChecks?: boolean;
+    documentUri?: string;
+}
```
spoenemann, line 24: Why not extend the `BuildOptions` to avoid a breaking change and allow full flexibility (we might extend the BuildOptions in the future)?

spoenemann (commented):
In the following grammar snippet, the reference finder hits the first assignment to `environments`, but not the second one.

```
interface Requirement {
    name: string
    text: string
    environments: @Environment[]
}

Requirement returns Requirement:
    'req' name=ID text=STRING
    ('applicable' 'for' environments+=[Environment] (',' environments+=[Environment])*)?;
```

packages/langium/src/test/langium-test.ts
msujew: Note that this wouldn't have been a breaking change, since the `BuildOptions` argument has been introduced in a recent PR that hasn't been released yet. I have adopted your suggestion anyway, to improve flexibility.

montymxb (approved):
:octocat: approved! Looked over this a couple of times, checked locally, but can't find anything that stands out except for that one function. Rest looks good.

PR #1001, opened 2023-03-27
Return back to global `expect` imports

msujew:
The current way of imports require to use `vitest` as the testing dependency for any user. Using globals also allows `jest` and correctly throws an error if no test function is provided.


sailingKieler:
Please spread the word about this change, I assume it will be quite confusing for the langium devs...

spoenemann:
Isn't `packages/langium/src/test` the only critical part for users of Langium? Could we keep the explicit imports for our framework tests and examples?

PR #1002, opened 2023-03-27
Improvements to `expectSymbols`, `expectFoldings`, and `expectCompletion` test functions

sailingKieler:
* Improved & extended function `expectSymbols`, `expectFoldings`, `expectCompletion` to optionally apply a custom expectation checker
  * changed behavior in case of different cardinalities to avoid (initial) failure messages like
  `AssertionError: Expected 1 but received 4 completion items: expected 4 to deeply equal 1` that are too obscure IMO
* export all helper functions to allow to copy & modify the `expect...` functions without any need to copy the required helpers, too

### DIFF
--- a/packages/langium/src/test/langium-test.ts
+++ b/packages/langium/src/test/langium-test.ts
@@ -129,7 +129,7 @@ export function expectFoldings(services: LangiumServices): (input: ExpectedFoldi
     };
 }
 
-function textDocumentParams(document: LangiumDocument): { textDocument: TextDocumentIdentifier } {
+export function textDocumentParams(document: LangiumDocument): { textDocument: TextDocumentIdentifier } {
     return { textDocument: { uri: document.textDocument.uri } };
 }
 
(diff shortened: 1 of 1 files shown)


packages/langium/src/test/langium-test.ts
```diff
@@ -101,33 +101,43 @@ export function expectFoldings(services: LangiumServices): (input: ExpectedBase)
...
+    assertCompletionList: (completions: CompletionList) => void;
+}
+
 export interface ExpectedCompletion extends ExpectedBase {
     index: number
     expectedItems: Array<string | CompletionItem>
 }
 
-export function expectCompletion(services: LangiumServices): (completion: ExpectedCompletion) => Promise<void> {
+export function expectCompletion(services: LangiumServices): (completion: AssertCompletion|ExpectedCompletion) => Promise<void> {
```
msujew, line 118: Instead of accepting different kinds of assertion objects, these types should be structured more clearly:

```ts
interface ExpectedItemCompletion {
    index: number
    expectedItems: Array<string | CompletionItem>
}
interface ExpectedCallbackCompletion {
    index: number;
    assert: (completions: CompletionList) => void;
}
type ExpectedCompletion = ExpectedItemCompletion | ExpectedCallbackCompletion;
```

msujew (commented):
I am a bit opposed to export our internal utility testing functions (such as `textDocumentParams`) to the outside. Is there a specific reason to do that? The only reason I can imagine is that we're missing some LSP testing functions. If so, we should add them to the framework instead of exposing our utilities.

sailingKieler:
@msujew  more changes here.
Didn't take the optional pojo-based comparison into account initially, should be better now. Applied the pattern to `expectSymbols` and partly to `expectFoldings`.
Let's see whether the `assert` option makes sense for other test utilities, too

msujew (approved):
I like the changes, thanks!

PR #1003, opened 2023-03-28
Generate `isType` function for primitive datatype rules

### LINKED ISSUE #990: Generate checker functions for data type rules
It's easy to generate checker functions for rules that consist of an alternative of string literals. We might even consider extending this to terminal rules by using their corresponding regular expression, but it might be hard to check arbitrary data type rules (which can be composed of multiple lexer tokens).

_Originally posted by @goto40 in https://github.com/langium/langium/discussions/389#discussioncomment-3428725_

> Is there a function generated by the langium code generator to check the validity of a string concerning a given union type from the grammar? E.g., `isState` in the example above.

gfontorbe:
closes #990 

The idea is to generate checker functions for datatype rules through langium-cli.
This only applies on datatype rules that are only composed of strings or terminals. I also had to slightly touch the type inference to provide the regex for terminals.

Example:
```
A returns string:
    'a';

AB returns string:
    A | 'b';

ABC returns string:
    AB | ID;
```

will generate the following functions in the generated `ast.ts` file:
```
export function isA(item: string): item is A {
    return item === 'a';
}

export function isAB(item: string): item is AB {
    return isA(item) || item === 'b';
}

export function isABC(item: string): item is ABC {
    return isAB(item) || /[_a-zA-Z][\w_]*/.test(item);
}
```


packages/langium-cli/test/generator/ast-generator.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    AB | ID;
+
+hidden terminal WS: /\\s+/;
+terminal ID: /[_a-zA-Z][\\w_]*/;
+
+hidden terminal ML_COMMENT: /\\/\\*[\\s\\S]*?\\*\\//;
+hidden terminal SL_COMMENT: /\\/\\/[^\\n\\r]*/;
+`;
+
+const EXPECTED_AST_FILE = normalizeEOL(`/******************************************************************************
```
sailingKieler, line 47: You can use `expandToString` from `template-string.ts` here to properly indent and format the expectation text, like
```
const EXPECTED_AST_FILE = expandToString`
    /*******
    ....
`
```

(not sure whether `normalizeEOL` would still be required for the test to pass on windows).

spoenemann:
Could you run `npm run langium:generate` in the root and check in the resulting changes?

packages/langium/src/grammar/type-system/type-collector/types.ts
```diff
@@ -376,6 +388,114 @@ function pushReflectionInfo(node: CompositeGeneratorNode, name: string) {
...
+    const regexes = collectRegexesFromDataType(union);
+    const returnString = createDataTypeCheckerFunctionReturnString(subTypes, strings, regexes);
+    node.append(`export function is${union.name}(item: string): item is ${union.name} {`, NL);
+    node.indent(body => {
+        body.append(returnString + ';', NL);
+    });
+    node.append('}', NL);
+}
+
+function createDataTypeCheckerFunctionReturnString(subTypes: string[], strings: string[], regexes: string[]): string {
```
msujew, line 403: Can we also add a check for `number` (`typeof item === 'number')`), `bigint` (`typeof item === 'bigint'`) and `Date` (`item instanceof Date`)? That would make the checks complete.

packages/langium/src/grammar/type-system/type-collector/types.ts
```diff
@@ -376,6 +388,114 @@ function pushReflectionInfo(node: CompositeGeneratorNode, name: string) {
...
+    }
+    if (regexes.length > 0) {
+        for (let i = 0; i < regexes.length; i++) {
+            returnString += `/${regexes[i]}/.test(item)`;
+            if (i < regexes.length - 1) {
+                returnString +=' || ';
+            }
+        }
+    }
+    return returnString;
```
msujew, line 435: This code seems rather lengthy. We should just collect the check/test code snippets once and then just join them together with `snippets.join(' || ')`.

msujew (commented):
I like it, thanks!

Can we also include generated checks for `number`, `bigint` and `Date` types? I think that would complete the feature on this part.

packages/langium-cli/test/generator/ast-generator.test.ts
```diff
@@ -0,0 +1,836 @@
...
+            A returns string:
+                'a' | 'b' | 'c';
+
+            hidden terminal WS: /\\s+/;
+            terminal ID: /[_a-zA-Z][\\w_]*/;
+            hidden terminal ML_COMMENT: /\\/\\*[\\s\\S]*?\\*\\//;
+            hidden terminal SL_COMMENT: /\\/\\/[^\\n\\r]*/;
+        `;
+        const expectedASTFile = normalizeEOL(expandToString`
+        /******************************************************************************
```
msujew, line 107: Suggestion: These type of tests should be performed as isolated as possible, so that unrelated changes don't affect the test. You should cut the unrelated parts off using something like this:

```ts
const relevantPart = generatedCode.substring(
  generatedCode.indexOf('export') // from first exported type...
  generatedCode.indexOf('export class') // ... until exported reflection class
);
```

That way, we only test the generated types declarations

packages/langium/src/grammar/type-system/type-collector/types.ts
```diff
@@ -376,6 +385,156 @@ function pushReflectionInfo(node: CompositeGeneratorNode, name: string) {
...
+                regexes.push(type.regex);
+            } else if (isPropertyUnion(type)) {
+                regexes.push(...collectRegexesFromDataType(type));
+            }
+        }
+    }
+    return regexes;
+}
+
+function pushDataTypeReflectionInfoForNumber(node: CompositeGeneratorNode, union: UnionType) {
```
msujew, line 518: I believe we can add another layer of abstraction in there, right now there's a lot of repetition in the `pushDataTypeReflectionFor...` functions.

PR #1004, opened 2023-03-29
Improve completion behavior for cross references

msujew:
Fixes two minor issues:

* In some grammars, the completion might be slightly different if we use the following AST node (i.e. the AST node right at the cursor position). In most languages, this is the correct behavior. We now create two separate contexts if the node after the cursor and the one before are different and create completion for both.
* Sometimes, we evaluate the same keyword/cross reference multiple times over the course of the completion provider. Users would see the same proposals multiple times in their list. This change implements a deduplication algorithm, which removes duplicated entries.

### DIFF
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -138,7 +138,7 @@ export class DefaultCompletionProvider implements CompletionProvider {
         if (!astNode) {
             const parserRule = getEntryRule(this.grammar)!;
             await this.completionForRule(context, parserRule, acceptor);
-            return CompletionList.create(items, true);
+            return CompletionList.create(this.deduplicateItems(items), true);
         }
 
         const contexts: CompletionContext[] = [context];
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -123,20 +123,41 @@ export class DefaultCompletionProvider implements CompletionProvider {
             }
         };
 
-        const node = findLeafNodeAtOffset(cst, this.backtrackToAnyToken(text, offset));
+        const lastTokenOffset = this.backtrackToAnyToken(text, offset);
+
+        const astNode = findLeafNodeAtOffset(cst, lastTokenOffset)?.element;
 
         const context: CompletionContext = {
             document,
             textDocument,
-            node: node?.element,
+            node: astNode,
             offset,
             position: params.position
         };
 
-        if (!node) {
+        if (!astNode) {
             const parserRule = getEntryRule(this.grammar)!;
             await this.completionForRule(context, parserRule, acceptor);
             return CompletionList.create(items, true);
+        }
+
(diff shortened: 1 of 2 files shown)


packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -158,19 +179,29 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
                 stream(afterFeatures)
                     .exclude(beforeFeatures, distinctionFunction)
                     .distinct(distinctionFunction)
-                    .map(e => this.completionFor(context, e, acceptor))
+                    .map(e => this.completionForContexts(contexts, e, acceptor))
             );
         }
 
-        return CompletionList.create(items, true);
+        return CompletionList.create(this.deduplicateItems(items), true);
```
spoenemann, line 194: Is there an explanation for why exactly duplicate items can appear? Can we prevent that at the root cause instead of post-processing?

msujew: There are a lot of edge cases in which the scope provider might return an element multiple times during the course of the completion. For example:

```
Rule: ref1=[Definition:ID] 'x' | ref2=[Definition:ID] 'y';
```

Or

```
Rule: ref1=[Super:ID] 'x' | ref2=[Sub:ID] 'y';

interface Sub extends Super { }
```

In both cases we have different cross references which yield some of the same elements. We do try to prevent duplicates already beforehand (see [here](https://github.com/langium/langium/blob/c2797380edc01ed170d0f9ada926551209b01fcf/packages/langium/src/lsp/completion/completion-provider.ts#L160) and [here](https://github.com/langium/langium/blob/c2797380edc01ed170d0f9ada926551209b01fcf/packages/langium/src/lsp/completion/completion-provider.ts#L274)), but some of these are not preventable through these measures.

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -123,22 +123,43 @@ export class DefaultCompletionProvider implements CompletionProvider {
             }
         };
```
Lotes, line 125: This method feels a bit complex (or `unübersichtlich`), because of the inserted `if` statements. You could split it into phases, like `collectCompletionContexts`, `completeOnContext` (already there), `mergeCompletionResults`... maybe you find better names.
Code reads better with less branches.

msujew: I disagree (at least to a certain degree). The nesting is two levels deep at most and most and fairly clear in my opinion. The reason why I'm quite hesitant to split this method is that every part of it requires a lot of offset+ast+document information. We also have a few early exit conditions (such as the one where we perform completion at the start of the document), which become very difficult to handle once we split this into different methods.

dhuebner (approved):
Unfortunately no tests, otherwise looks good to me.

msujew:
@dhuebner I added a test for the deduplication. A test for the other minor bug is unfortunately rather difficult to build, since it requires a lot of scoping setup.

Issue #956, opened 2023-02-24
CLI launcher fails on linux "/usr/bin/env: ‘node\r’: No such file or directory"


rbbl-dev:
**Situation**: freshly created langium hello world project.

**Input**: `./bin/cli --help`

**Error**:  
> /usr/bin/env: ‘node\r’: No such file or directory
> /usr/bin/env: use -[v]S to pass options in shebang lines

tested on fedora 37 & ubuntu 20.04

**Fix**: Saving it with LF line endings.

![image](https://user-images.githubusercontent.com/78143613/221142510-4a90bbc5-472d-4254-b3a5-61e636af8660.png)

PR #1005, opened 2023-03-29
Normalize line endings for yeoman generator

msujew:
Closes https://github.com/langium/langium/issues/956

When uploading the npm package from windows, the generated files contain windows line endings. We simply replace those with the line endings of the selected file system.

### DIFF
--- a/packages/generator-langium/src/index.ts
+++ b/packages/generator-langium/src/index.ts
@@ -9,6 +9,7 @@ import _ from 'lodash';
 import chalk from 'chalk';
 import path from 'path';
 import which from 'which';
+import { EOL } from 'os';
 
 const TEMPLATE_CORE_DIR = '../langium-template-core';
 const TEMPLATE_VSCODE_DIR = '../langium-template-vscode';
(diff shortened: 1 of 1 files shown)


sailingKieler (commented):
So you idea is to allow mixed line breaks in the templates and consolidate them while expanding the templates?

I'm not sure whether `import { EOL } from 'os';` is a good idea.
Would `\n` work on windows?
That's mainly relevant for the scripts section in the `package.json`, isn't it?

sailingKieler:
> They aren't mixed. They are in the format of the source system. In my case that's windows, which yields Windows style line endings on all generated files.

What I meant is: If you contribute to the templates and I do, too, it's likely that the templates contain both line ending versions arbitrarily.

The problem described in #956 is about such a peace of code:
image

Hence, if you create a langium project on your windows system and publish scripts like the above with windows line endings, the issue will still be there on every linux system, right? And since `langium:generate` doesn't touch it, users have to fix it manually.
Thus, I'm in favor of generating linux line endings, at least for the cli binary part.

msujew:
> If you contribute to the templates and I do, too, it's likely that the templates contain both line ending versions arbitrarily.

That's a non-issue, since Git will always checkout files using the line endings of the target system. If you commit something with Linux line endings, and I'll check it out on Windows, it will automatically use Windows line endings.

PR #1007, opened 2023-03-29
Fix json config schema validation

msujew:
Closes https://github.com/langium/langium/issues/1006

There were a few issues which prevented the json validation from working as expected. This change should address all of them.

### DIFF
--- a/packages/langium-cli/src/package.ts
+++ b/packages/langium-cli/src/package.ts
@@ -68,7 +68,7 @@ export function getFilePath(absPath: string, config: LangiumConfig): string {
     return path.relative(base, absPath);
 }
 
-export async function loadConfigs(options: GenerateOptions): Promise<LangiumConfig[]> {
+export async function loadConfig(options: GenerateOptions): Promise<LangiumConfig> {
     let filePath: string;
     if (options.file) {
         filePath = path.normalize(options.file);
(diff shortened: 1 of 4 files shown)


packages/langium-cli/src/package.ts
```diff
@@ -83,26 +83,13 @@ export async function loadConfigs(options: GenerateOptions): Promise<LangiumConf
...
-            return setProjectNames(config);
-        } else {
-            config[RelativePath] = relativePath;
-        }
-        return setProjectNames([config]);
+        const config: LangiumConfig = path.basename(filePath) === 'package.json' ? obj.langium : obj;
+        config.projectName = _.camelCase(config.projectName);
+        config.projectName = config.projectName.charAt(0).toUpperCase() + config.projectName.slice(1);
+        config[RelativePath] = relativePath;
+        return config;
```
spoenemann, line 90: This is a breaking change – shall we defer it to 2.0?

msujew: I'm not sure whether anyone actually used multiple langium configs in one file (i.e. the ability to generate multiple Langium based projects using a single invocation of the `langium-cli`). It was something I added for testing purposes which we never ended up needing, due to being able to interpret Langium grammars on the fly.

While it was working in theory, the feature was completely undocumented and I'm quite certain that no one used it. I would keep it in this PR for 1.2.0.

spoenemann: Ok, but please document its removal in CHANGELOG of langium-cli

spoenemann (approved):

PR #1008, opened 2023-03-29
Use `assert` for test utilities

msujew:
Effectively reverts the changes from https://github.com/langium/langium/pull/1001 by utilizing `assert` from the node.js runtime instead of trying to use `jest` or `vitest` `expect` functions.

### DIFF
--- a/packages/langium/src/test/langium-test.ts
+++ b/packages/langium/src/test/langium-test.ts
@@ -17,6 +17,7 @@ import { findNodeForProperty } from '../utils/grammar-util';
 import { SemanticTokensDecoder } from '../lsp/semantic-token-provider';
 import { TextDocument } from 'vscode-languageserver-textdocument';
 import { BuildOptions } from '../workspace/document-builder';
+import assert from 'assert';
 
 export function parseHelper<T extends AstNode = AstNode>(services: LangiumServices): (input: string, buildOptions?: BuildOptions) => Promise<LangiumDocument<T>> {
     const metaData = services.LanguageMetaData;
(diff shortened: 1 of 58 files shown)


kaisalmen:
@msujew shouldn't then
```
"types": [
     "vitest/globals"
]
```
be only added to the langium tsconfig.json? Is it really clever to have this mixed bag now?

msujew:
@kaisalmen I've updated the PR according to our previous discussion about using `assert`.

packages/langium/src/test/langium-test.ts
```diff
@@ -34,20 +35,13 @@ export function parseHelper<T extends AstNode = AstNode>(services: LangiumServic
...
-            // With `jest-expect-message`
-            expect(actual, message).toEqual(expected);
-        } else {
-            expect(actual).toEqual(expected);
-        }
-    } else {
-        throw new Error('No expect function provided. Use the `expectFunction` function to supply a custom expect function or install `vitest` or `jest`.');
-    }
+    assert.deepStrictEqual(actual, expected, message);
 };
```
sailingKieler, line 39: Why not keeping this? Would leave us flexibility!

packages/langium/src/test/langium-test.ts
```diff
@@ -34,20 +35,13 @@ export function parseHelper<T extends AstNode = AstNode>(services: LangiumServic
...
 
 /**
- * Overrides the assertion function used by tests. Tries to use Jest by default.
+ * Overrides the assertion function used by tests. Uses `assert.deepStrictEqual` by default
+ *
+ * @deprecated Since 1.2.0. Do not override the assertion functionality.
```
sailingKieler, line 44: Why not keeping this (non-deprecated)? Would leave us flexibility!

msujew: I'm not sure how much this flexibility is actually an advantage. `assert.deepStrictEqual` should already cover all of our bases, especially since it's only used by the test functions that we export.

PR #1013, opened 2023-04-05
Improved Arithmetics Language

emilkrebs:
**New**
- Modolo operator (%)
- Potential (^)
- Validation to check if the given parameters match the function
- Validation to check if the parameter names are unique

**Changes**
- Changed the arithmetic language server to evaluate and send the expressions needed for this PR: https://github.com/langium/langium-website/pull/143 **(!!)**


examples/arithmetics/src/language-server/arithmetics-util.ts
```diff
@@ -4,9 +4,11 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-export function applyOp(op: '+' | '-' | '*' | '/'): (x: number, y: number) => number {
+export function applyOp(op: '+' | '-' | '*' | '/' | '^' | '%'): (x: number, y: number) => number {
```
emilkrebs, line 7: added the new operations here

examples/arithmetics/src/language-server/arithmetics.langium
```diff
@@ -26,7 +26,14 @@ Addition infers Expression:
     Multiplication ({infer BinaryExpression.left=current} operator=('+' | '-') right=Multiplication)*;
 
 Multiplication infers Expression:
-    PrimaryExpression ({infer BinaryExpression.left=current} operator=('*' | '/') right=PrimaryExpression)*;
+    Potentiation ({infer BinaryExpression.left=current} operator=('*' | '/') right=Potentiation)*;
+
+Potentiation infers Expression:
```
msujew, line 31: Suggestion: You likely mean `Exponentiation`:
```suggestion
Exponentiation infers Expression:
```

examples/arithmetics/src/language-server/arithmetics.langium
```diff
@@ -26,7 +26,14 @@ Addition infers Expression:
...
+    Modulo ({infer BinaryExpression.left=current} operator='^' right=Modulo)*;
+
+Modulo infers Expression:
```
msujew, line 34: Suggestion: This is usually called the `Remainder` operation. However, please group it with the `Multiplication` parser rule, since it has the same associativity (i.e. operator precedence).

examples/arithmetics/src/language-server/main-browser.ts
```diff
@@ -7,6 +7,8 @@
 import { startLanguageServer, EmptyFileSystem, DocumentState } from 'langium';
 import { BrowserMessageReader, BrowserMessageWriter, createConnection, Diagnostic, NotificationType } from 'vscode-languageserver/browser';
 import { createArithmeticsServices } from './arithmetics-module';
+import { interpretEvaluations } from '../cli/evaluator';
```
msujew, line 10: While an import from `/cli/evaluator` is not harmful by itself, it is usually assumed that a `CLI` would use `node` dependencies. Since this is the browser module, we should make clear that the code imported here does is not explicitly part of a CLI. Please move the `evaluator` to a common area that is supposed to be accessed by every consumer of the language.

examples/arithmetics/src/cli/evaluator.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        const localContext = new Map<string, number | Definition>(ctx.context);
+        for (let i = 0; i < valueOrDef.args.length; i += 1) {
+            localContext.set(valueOrDef.args[i].name, evalExpression(ctx, expr.args[i]));
+        }
+        return evalExpression({module: ctx.module, context: localContext, result: ctx.result}, valueOrDef.expr);
+    }
+
+    console.error('Impossible type of Expression.');
+    process.exit(1);
```
msujew, line 77: Suggestion: `process.exit` is only available in node and not in the browser. It will yield unexpected runtime errors in the browser case. You should just throw an `Error` here instead.

examples/arithmetics/src/language-server/arithmetics.langium
Lotes: Wait! Isn't the ^ operator not right-associative?!

msujew: My comment is about the `%` operator, not `^`.

Lotes: Sorry! I totally ignored that :-(. I will provide another comment with a solution

examples/arithmetics/src/language-server/arithmetics.langium
Lotes, line 31: The potentiation operator is right-associative. Your rule will make it left-associative. It is just a good old pattern for compiler creators.
You can see the right-associative pattern in this article [here](https://www.typefox.io/blog/parsing-expressions-with-xtext).
Should be something like:
```
Potentiation infers Expression:
 PrimaryExpression ({BinaryExpression.left=current} '^' right=Potentiation)?;
```

The syntax tree will then grow to the right and not the left. You can draw the syntax tree for practicing. 4^3^2 will have two parse trees, but only one is the commonly used one.

luan-xiaokun:
Since now we have the remainder operator %, it would be nice if we could also check that the modulus is not zero when checking that the divisor is not zero, such as changing the condition to `(binExpr.operator === '/' || binExpr.operator === '%') && ...`.
https://github.com/langium/langium/blob/82613e6618123f62bec9433b63234b425a05a9a3/examples/arithmetics/src/language-server/arithmetics-validator.ts#L24-L28

examples/arithmetics/src/language-server/main-browser.ts
```diff
@@ -26,14 +28,20 @@ const documentChangeNotification = new NotificationType<DocumentChange>('browser
...
+                const cstNode = evaluation.expression.$cstNode;
+                if (cstNode) {
+                    json.push({range: evaluation.expression.$cstNode?.range, text: evaluation.expression.$cstNode?.text, value: value});
+                }
+            }
+        }
         connection.sendNotification(documentChangeNotification, {
             uri: document.uri.toString(),
-            content: json,
+            content: JSON.stringify({ ast: jsonSerializer.serialize(module), evaluations: json }),
```
sailingKieler, line 43: That's a tricky change, as it affects my envisioned code generation demo (which you can't be aware of 😉).
Please keep the AST serialization as it was above including the instruction of
```javascript

{
    sourceText: true,
    textRegions: true,
}
```
Question: Could do the interpretation also on client side?
With the serialized AST including the original source text and the text regions of the particular AST elements that shouldn't be difficult. Let's chat about this if you like.

sailingKieler (changes requested):

examples/arithmetics/src/language-server/arithmetics.langium
sailingKieler, line 1: In order to express e.g. the fibonacci function the easiest solution would be to include a case distinction.
I encourage you to also replace
```
Expression:
    Addition;

```
by
```

Expression:
    ConditionalExpression;

ConditionalExpression:
    Comparison ({infer ConditionalExpression.left=current} '?' then=Expression ':' else=Expression);

Comparison:
    Addition ({infer BinaryExpression.left=current} operator=('==' | '!=' | '<' | '<=' | '>=' | '>') right=Addition)*;
```

Of course the interpreter needs to be supplemented to support `ConditionalExpression`, which shouldn't be much effort.

examples/arithmetics/src/language-server/arithmetics-util.ts
sailingKieler, line 7:
```suggestion
export function applyOp(op: BinaryExpression['operator']): (x: number, y: number) => number {
```

examples/arithmetics/src/language-server/arithmetics-util.ts
```diff
@@ -4,9 +4,11 @@
...
     if (op === '+') return (x, y) => x + y;
     if (op === '-') return (x, y) => x - y;
     if (op === '*') return (x, y) => x * y;
+    if (op === '^') return (x, y) => Math.pow(x, y);
+    if (op === '%') return (x, y) => x % y;
     return (x, y) => y === 0 ? x : x / y;
```
sailingKieler, line 13: Instead of the `if`s you might use a switch statement, like
```suggestion
    switch(op) {
        case '+':  return (x, y) => x + y;
        case '-':  return (x, y) => x - y;
        case '*':  return (x, y) => x * y;
        case '/':  return (x, y) => x / y;
        case '*':  return (x, y) => x * y;
        case '/':  return (x, y) => x / y;
        case '^':  return (x, y) => Math.pow(x, y);
        case '%':  return (x, y) => x % y;
        case '==': return (x, y) => x == y ? 1 : 0;
        case '!=': return (x, y) => x != y ? 1 : 0;
        case '<':  return (x, y) => x  < y ? 1 : 0;
        case '<=': return (x, y) => x <= y ? 1 : 0;
        case '>':  return (x, y) => x  > y ? 1 : 0;
        case '>=': return (x, y) => x >= y ? 1 : 0;
    }
```

examples/arithmetics/src/language-server/main-browser.ts
emilkrebs: Thanks, I added the JSON part back! 
I actually talked about this and we concluded that sending the results would be more efficient for the preview. 
Also for people looking at the example, it's helpful to know how to process the last inside of the language server

examples/arithmetics/src/language-server/arithmetics.langium
emilkrebs: Great idea, conditional expressions would be very cool to have in an arithmetic programming language! 
However, this would make parts of the validation obsolete and very complex. Maybe we should push this feature to a different branch.

examples/arithmetics/src/language-server/arithmetics-util.ts
emilkrebs: Same thoughts apply here

examples/arithmetics/src/language-server/arithmetics-util.ts
emilkrebs: Yeah, switch-case statements would be great to simplify this case, I added them!

examples/arithmetics/src/language-server/arithmetics-util.ts
```diff
@@ -4,9 +4,13 @@
...
-    return (x, y) => y === 0 ? x : x / y;
+export function applyOp(op: '+' | '-' | '*' | '/' | '^' | '%'): (x: number, y: number) => number {
+    switch (op) {
+        case '+': return (x, y) => x + y;
+        case '-': return (x, y) => x - y;
+        case '*': return (x, y) => x * y;
+        case '^': return (x, y) => Math.pow(x, y);
+        case '%': return (x, y) => x % y;
+        default: return (x, y) => y === 0 ? x : x / y;
+    }
```
sailingKieler, line 15: I suggest to explicitly mention the `/` instead of `default`, the type system will then add an error here once further operators are added to the grammar if they are missing here. Applying a division by `default` is quite unconventional 😉.

examples/arithmetics/src/language-server/main-browser.ts
sailingKieler, line 43: @emilkrebs I have the impression you added the above mentioned config objects in the `domainmodel` example (which might also be helpful someday), but I need it here in line 43

examples/arithmetics/src/language-server/arithmetics-util.ts
```diff
@@ -4,9 +4,13 @@
...
-    if (op === '*') return (x, y) => x * y;
-    return (x, y) => y === 0 ? x : x / y;
+export function applyOp(op: '+' | '-' | '*' | '/' | '^' | '%'): (x: number, y: number) => number {
+    switch (op) {
+        case '+': return (x, y) => x + y;
+        case '-': return (x, y) => x - y;
+        case '*': return (x, y) => x * y;
+        case '^': return (x, y) => Math.pow(x, y);
+        case '%': return (x, y) => x % y;
+        default: return (x, y) => y === 0 ? x : x / y;
```
msujew, line 14: In addition to what @sailingKieler already said about this: It's fine to throw an error here when we try to divide by 0. I know it's old code, but we should catch it appropriately. Simply returning `x` is very much unexpected in this instance.

sailingKieler (changes requested):
Hi @emilkrebs, I still have a few remarks, please have a look.

examples/arithmetics/src/language-server/arithmetics-util.ts
```diff
@@ -11,7 +11,7 @@ export function applyOp(op: '+' | '-' | '*' | '/' | '^' | '%'): (x: number, y: n
         case '*': return (x, y) => x * y;
         case '^': return (x, y) => Math.pow(x, y);
         case '%': return (x, y) => x % y;
-        default: return (x, y) => {
+        case '/': return (x, y) => {
```
montymxb, line 14: Generally speaking, it's always good to have a default case in a switch. Having a concrete case for / is good, but we should add one more default that throws when an unknown operator is passed. If we ever extend the language, that error will tell us where we need to add handling for it.

examples/arithmetics/src/language-server/arithmetics-util.ts
montymxb: This would be quite nice to add actually. No need to re-declare the binary ops.

examples/arithmetics/src/language-server/arithmetics-util.ts
montymxb: This got added in sometime earlier this week.

examples/arithmetics/src/language-server/arithmetics.langium
montymxb: I'd vote to add conditionals in a follow-up PR.

montymxb (commented):
Nice! This is looking really good. I have a few more points I wanted to have addressed, which are outlined in the review. In terms of functionality that we wanted to add, I think we're at a good point where we can tie off this last aspects and merge the PR in. As for new functionality, such as conditionals and the like, those would be nice to have but I think that would be more appropriate to introduce in a separate PR; especially to keep this one from growing beyond its original scope.

examples/arithmetics/src/language-server/main-browser.ts
```diff
@@ -26,14 +29,28 @@ const documentChangeNotification = new NotificationType<DocumentChange>('browser
...
+                }
+            }
+        }
+
+        // add the evaluations to the ast object
+        (module as unknown as { evaluations: Array<{
+            range: Range | undefined;
+            text: string | undefined;
+            value: number;
+        }> }).evaluations = json;
```
sailingKieler, line 49: One remark on this: I encourage you to change the field `evaluations` to `$evaluations` to emphasize that those data do not belong to the AST.

Otherwise I'm fine. 😀

sailingKieler (approved):

sailingKieler:
@emilkrebs would you please `Squash and merge` the changes? 40 commits are quite a lot.

examples/arithmetics/src/language-server/main-browser.ts
emilkrebs: Aah, yes, because it's metadata

PR #1014, opened 2023-04-06
Replicated signature overloads to 'StreamImpl'

sailingKieler:
this way they're also applied to vars of the ...Impl types
(TypeScript apparently doesn't apply interface overloads to implementing classes)


msujew (approved):
Looks good to me 👍

PR #1017, opened 2023-04-11
Add validation for cross-reference to union type

### LINKED ISSUE #966: Missing validation for cross-reference targets
When the target type of a cross-reference is a type union, we should validate that all alternatives are AST nodes (not primitives).

gfontorbe:
Closes #966 

Adds validation for cross-reference to union type that are not composed only of AST nodes.

I'm pretty sure I am missing some test cases but couldn't think of other ways this validation would be triggered. If you think of some other cases please let me know and I will add tests for them.


msujew (approved):
Looks good to me 👍

PR #1018, opened 2023-04-11
Import types instead of import and prepare update to TS5.0

### LINKED ISSUE #960: Langium TypeScript output should use `type` imports for non-value imports
Currently if a user updates their tsconfig to have the option `"importsNotUsedAsValues": "error"`, they will get a type error in generated code:

Screenshot 2023-02-25 at 12 29 57 PM

So what Langium can do is output type-only imports like the following, for anything that is type-only:

```ts
import type { LangiumGeneratedServices, LangiumGeneratedSharedServices, LangiumSharedServices, LangiumServices, LanguageMetaData, Module } from 'langium';
```

This is nice because, when bundling a lib for release, we want to import only files that are actually used. This setting tells TypeScript that it can safely remove the import statement from the output because no values are actually used, it is only for type checking.

If someone *does* need a side-effect import (or the generated code does), while they have this option enabled, the way to do that would be like so:

```ts
import type {SomeType} from 'something-that-has-side-effects'
import 'something-that-has-side-effects' // This will be kept
```

gfontorbe:
Closes #960 

* Prepare upgrade to TypeScript 5.0.x
* Adds new linter rule to use `import type` instead of `import` for imports that are only used for types and not values
* Fixed linter issues for the new rule
* Adapted langium-cli and generators to reflect those changes
* Some styling here and there

### DIFF
--- a/packages/langium/test/grammar/langium-grammar-validator.test.ts
+++ b/packages/langium/test/grammar/langium-grammar-validator.test.ts
@@ -5,7 +5,7 @@
  ******************************************************************************/
 
 import type { AstNode, Properties } from '../../src';
-import type { Assignment, ParserRule, UnionType } from '../../src/grammar/generated/ast';
+import type { Assignment, CrossReference, ParserRule, UnionType } from '../../src/grammar/generated/ast';
 import type { ValidationResult } from '../../src/test';
 import { afterEach, beforeAll, describe, expect, test } from 'vitest';
 import { DiagnosticSeverity } from 'vscode-languageserver';


package-lock.json
```diff
@@ -71,6 +71,28 @@
         "vscode": "^1.67.0"
       }
     },
+    "examples/hello-world": {
```
dhuebner, line 74: This should not happen. Failed the CLI test, or is the clean-up after test disabled? 
Afaik I set it up to clean up even on test failure.

dhuebner (changes requested):
Please remove the test project from the package-lock. Also check that clean-up step is inside a `finally` block

gfontorbe:
> Please remove the test project from the package-lock. Also check that clean-up step is inside a `finally` block

The clean-up step is already in a finally block, not sure why this was added to the package-lock...

dhuebner:
> > Please remove the test project from the package-lock. Also check that clean-up step is inside a `finally` block
> 
> The clean-up step is already in a finally block, not sure why this was added to the package-lock...

Do you have this folder locally? Maybe it is there since you was adding new CLI tests?

dhuebner (approved):

dhuebner:
@gfontorbe 
It is really difficult to track all the changes.
Just hope we have enough tests and enough time to find potential bugs :) 
Would be nice if someone else could also look through.

msujew (changes requested):
It isn't noticeable since the `ast.ts` files get generated with an `eslint-disable` instruction, but these files aren't using the `import type` statements correctly:

```ts
import type { AstNode, Reference, ReferenceInfo, TypeMetaData } from 'langium';
import { AbstractAstReflection } from 'langium';
```

We currently only generate
```ts
import { AstNode, AbstractAstReflection, Reference, ReferenceInfo, TypeMetaData } from 'langium';
```

msujew (approved):
Looks good, please rebase on top of the latest `main` and merge at the end of the week as discussed 👍

PR #1024, opened 2023-04-14
Add IPA for 'Langium'

msujew:
Since it's not 100% clear how Langium is supposed to be pronounced, I've added the IPA pronounciation for it to the readme.


sailingKieler (approved):
Like it! 👌

PR #1031, opened 2023-04-20
Automate release process

msujew:
Simplifies and mostly automates the release process:

1. Updating tags only requires two commands to run now.
2. Publishing will be performed from within GitHub actions whenever a new release is created.

Since the `npm version` script does not automatically update dependency versions between the packages, we need to update them via a script. This is what `version:dependencies` does.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,54 @@
+name: Publish
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  publish:
+    name: Langium Publish
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Use Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: '16'
+        registry-url: 'https://registry.npmjs.org'
+    - name: Build
+      shell: bash
+      run: |
+        npm ci
+        npm run build:clean
+    - name: Test
+      if: success() || failure()
+      shell: bash
+      run: |
+        npm run test
+    - name: Publish NPM Packages
+      shell: bash
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      # Update the following list when a new package is added
+      run: |
+        npm run publish:latest --workspace=langium
+        npm run publish:latest --workspace=langium-cli
+        npm run publish:latest --workspace=langium-sprotty
+        npm run publish:latest --workspace=generator-langium
+        npm run publish:latest --workspace=langium-arithmetics-dsl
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -110,43 +110,19 @@ const debugOptions = { execArgv: ['--nolazy', '--inspect-brk=6009'] };
...
- 10. Create a GitHub release from the new tag
- 11. Close the corresponding GitHub milestone
+The release process is mostly automated and requires running only a few commands.
+After commiting, pushing, tagging and releasing the changes, a GitHub Action will publish all artifacts (npm packages and vscode extensions).
+
+1. Pull the latest changes
+2. Uplift the package versions
+    * Run `npm version major|minor|patch --no-git-tag-version --workspaces`
+3. Update the dependency versions
+    * Run `npm run version:dependencies`
```
kaisalmen, line 120: langium-vscode does not have the "~" at the langium version and therefore the script does not work there.

scripts/update-version.js
```diff
@@ -0,0 +1,33 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+async function runUpdate() {
+    const langiumPath = getPath('langium', true);
+    const langiumPackage = await fs.readJson(langiumPath);
+    const version = langiumPackage.version;
+    await Promise.all([
+        replaceAll('langium-cli', true, version),
```
kaisalmen, line 9: You have to add `langium` here as well otherwise the `langium-cli` devDependency is unchanged and then polluting the lock file with old versions.

kaisalmen (commented):
Nice work. Tested it locally. After performing the suggested changes it looks good.

kaisalmen (approved):
Looks good. LGTM 🎉

.github/workflows/publish.yml
```diff
@@ -0,0 +1,54 @@
...
+        npm run publish:latest --workspace=langium-sprotty
+        npm run publish:latest --workspace=generator-langium
+        npm run publish:latest --workspace=langium-arithmetics-dsl
+        npm run publish:latest --workspace=langium-domainmodel-dsl
+        npm run publish:latest --workspace=langium-statemachine-dsl
+    - name: Publish VSCode Extension
+      shell: bash
+      env:
+        VSCE_TOKEN: ${{ secrets.VSCE_TOKEN }}
+        OVSX_TOKEN: ${{ secrets.OVSX_TOKEN }}
```
spoenemann, line 47: These variables seem unused.

spoenemann (changes requested):
Please add the `--provenance` flag to make use of the new supply chain security badge:

https://github.blog/2023-04-19-introducing-npm-package-provenance/

spoenemann (approved):
Great improvement, thank you!

PR #1038, opened 2023-05-02
Fix missing files in generator

msujew:
Closes https://github.com/langium/langium/issues/1037

We missed the new files in the `package.json#files` entry. I've simplified it a bit so we only need a single `templates` entry, even if we add a new template.

I would push an update to the generator afterwards, which is why I updated the version to `1.2.1` already.


montymxb (approved):
Looks good to me :octocat:

PR #1049, opened 2023-05-16
[Draft] Extract playground from Langium website

Lotes:


spoenemann:
Seeing all the additional dependencies, I'm thinking whether a separate repo might be the better option. Let's discuss it next week with @msujew.

Lotes:
I setuped the package with create React App. That was not the best option and deps exploded. I will try another starter script that has only a few deps

Lotes:
Closed because I will use a dedicated repository to keep things simple.

PR #1054, opened 2023-05-18
Make parser rule reachable when used in crossref

hpopov:
This removes redundant hint "This rule is declared but never referenced" during Langium validation for Parser Rules that are only referenced from cross reference in the grammar.

An example of cases when this hint is shown incorrectly: [ImportableElement presents a type union, and is only referenced by crossref](https://github.com/Crystal-VPL/crystal-core-lang/blob/f9f93963e20f4834c002ad015dc5fedd3b548071/src/language-server/behavior/behavior.langium).

Please, let me know, if I should raise a corresponding bug ticket.


msujew:
I believe the validation is working as intended. The actual parser rule is never being used, so it is unreachable by the parser. If you want to fix the hint correctly, you need to replace the parser rule with a type:

```
type ImportableElement = Classifier | BoundaryObject;
```

PR #1055, opened 2023-05-21
Fix container types calculation

### LINKED ISSUE #150: Improve container types calculation
An `AstNode` instance has an optional field `$container`, which presents its container -- a parental `AstNode`. The feature makes casting an `AstNode` to its container pretty comfortable, however it works nicely exclusively, when there is the only one possible container for the `AstNode`. Current questions:

- How the case of multiple container types will be used?
- How to improve multiple container types calculation?

**The problem of multiple container types calculation**
There are types `A`, `B` and `C`, and `C` extends `A` and `B`.  If now we add a type `D`, containing a field with a type `A`, `A` will get a container `D`. Also we add a type `E`, containing a field with a type `B`, and `B` gets a container `E`. These manipulations make `C` invalid -- what is the `C`'s container type?

**The current solution**
Share container types of parents. For the example above: the container type of `A` will become `D | E`, and the container type of `B` will become `D | E`, so the container type of `C` will be determined: `D | E`.

**Problems of the solution**
1. Number of container types can grow very fast. Suppose, we add a type `F`, which extends our type `B` and one more type `G`. To have consistent `F`, we have to share container types of `G` with `A` and `B` and add `D | E` to `G`'s container types.
2. Container types become senseless. Container types of `G` can be senseless for types `A` and `B`.

pluralia:
Closes #150.
* drops union types from the calculation of container types
* drops sharing of container types over connected components: now the container types are lifted only lifted to the parents
* if one of the children has no container types, the parent also loses container types

Examples:
```
interface A extends AstNode { $container: E | F | G; strA: string }
interface B extends AstNode { $container: E | F | H; strB: string }
interface C extends A, B { $container: E; strC: string }
interface D extends A, B { $container: F; strC: string }
interface E extends AstNode{ c: C }
interface F extends AstNode{ d: D }
interface G extends AstNode{ a: A }
interface H extends AstNode{ b: B }
```

`X` has no container types -- `A` and `B`, too
```
interface A extends AstNode { strA: string }
interface B extends AstNode { strB: string }
interface C extends A, B { $container: E; strC: string }
interface D extends A, B { $container: F; strC: string }
interface X extends A, B { strC: string }
interface E extends AstNode{ c: C }
interface F extends AstNode{ d: D }
interface G extends AstNode{ a: A }
interface H extends AstNode{ b: B }
```

### DIFF
--- a/packages/langium/src/grammar/generated/ast.ts
+++ b/packages/langium/src/grammar/generated/ast.ts
@@ -52,7 +52,6 @@ export function isTypeDefinition(item: unknown): item is TypeDefinition {
 }
 
 export interface AbstractElement extends AstNode {
-    readonly $container: Alternatives | Assignment | CharacterRange | CrossReference | Grammar | Group | NegatedToken | ParserRule | TerminalAlternatives | TerminalGroup | TerminalRule | UnorderedGroup | UntilToken;
     readonly $type: 'AbstractElement' | 'Action' | 'Alternatives' | 'Assignment' | 'CharacterRange' | 'CrossReference' | 'Group' | 'Keyword' | 'NegatedToken' | 'RegexToken' | 'RuleCall' | 'TerminalAlternatives' | 'TerminalGroup' | 'TerminalRuleCall' | 'UnorderedGroup' | 'UntilToken' | 'Wildcard';
     cardinality?: '*' | '+' | '?'
     lookahead?: '?!' | '?='
@@ -146,7 +145,7 @@ export function isInferredType(item: unknown): item is InferredType {
 }
 
 export interface Interface extends AstNode {
-    readonly $container: Alternatives | Assignment | CharacterRange | CrossReference | Grammar | Group | NegatedToken | ParserRule | TerminalAlternatives | TerminalGroup | TerminalRule | UnorderedGroup | UntilToken;
+    readonly $container: Grammar;
     readonly $type: 'Interface';
     attributes: Array<TypeAttribute>
     name: string
@@ -222,7 +221,7 @@ export function isParameterReference(item: unknown): item is ParameterReference 
 }
 
 export interface ParserRule extends AstNode {
-    readonly $container: Alternatives | Assignment | CharacterRange | CrossReference | Grammar | Group | NegatedToken | ParserRule | TerminalAlternatives | TerminalGroup | TerminalRule | UnorderedGroup | UntilToken;
+    readonly $container: Grammar;
     readonly $type: 'ParserRule';
     dataType?: PrimitiveType
     definesHiddenTokens: boolean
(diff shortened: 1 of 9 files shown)


packages/langium/src/grammar/generated/ast.ts
```diff
@@ -418,7 +411,7 @@ export function isGroup(item: unknown): item is Group {
 }
 
 export interface Keyword extends AbstractElement {
-    readonly $container: Alternatives | Assignment | CharacterRange | CrossReference | Grammar | Group | NegatedToken | ParserRule | TerminalAlternatives | TerminalGroup | TerminalRule | UnorderedGroup | UntilToken;
+    readonly $container: CharacterRange;
```
spoenemann, line 414: This looks wrong: CharacterRange is not the only possible container of Keyword.

spoenemann (commented):
Thanks, it looks mostly good (found only one wrongly computed container in the grammar).

packages/langium/src/grammar/generated/ast.ts
pluralia: @spoenemann, it's correctly for the current calculation:
* `Keyword` is used in the assignment in the only one place -- in `CharacterRange`
* One more usage is the rule `PredicatedKeyword` that infers `Keyword` -- 1) the interface `PredicatedKeyword` will not be generated by the grammar definition, so no its income to container types; 2) if we would have the interface `PredicatedKeyword`, it would be a child of `Keyword` -- no income to container type, too, because we don't lift children container types in parents anymore
* Other usages are in type alternatives and are ignored by the new algorithm: `AssignableTerminal`, `CrossRefTerminal`, `AbstractTerminal`

spoenemann: There's a rather complex chain of unassigned rule calls here:

Alternatives > ConditionalBranch > UnorderedGroup > Group > AbstractToken > AbstractTokenWithCardinality > AbstractTerminal > Keyword

Alternatives has two possible containers: Group and ParserRule. Keyword should inherit those containers.

I've seen that in simpler examples of unassigned rule calls, the container types are fine. So maybe this is just a special case for which we can create an issue, but we ignore it for now?

pluralia: By my understanding it's not an issue, but an expected behaviour for the current algorithm.

The moment is the algorithm is working with inferred / declared types (from the `ast.ts`, basically) and not with parser rules content. That means, not all information that you can see in the `.langium` is available for the algorithm. As a result:
1. Most of types with the names from the chain doesn't exist
2. `Group` and `ParserRule` contain only `AbstractElement` and not `Alternatives`

spoenemann (approved):
Thanks!

packages/langium/src/grammar/generated/ast.ts
spoenemann: Ok let's ignore the issue for now and get back to it in case it turns out to be a real problem. I created #1063 so we don't forget.

PR #1070, opened 2023-06-02
Add support for regex flags

msujew:
Closes https://github.com/langium/langium/issues/127

This change adds support for the following regex flags:

* `i` - case insensitivity
* `s` - allow to match newline characters with `.`
* `u` - enable unicode support

Other flags (`g`, `y`, `m`) are not applicable in the context of Langium/Chevrotain and return a warning if used.

Right now, only regex flags defined on regular expressions themselves are valid for this feature. We cannot specify regex flags on composite terminals themselves. That would require some dedicated syntax, but we can add that in a separate PR.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 *.js
+*.monarch.ts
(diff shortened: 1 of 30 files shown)


spoenemann (approved):
Great! Works as expected.

PR #1072, opened 2023-06-05
feat: Add `jsExtension` to `LangiumConfig`

sidharthv96:
When downstream projects are using `nodenext` module resolution, file extensions are mandatory.
So if `jsExtension` is true, a `.js` extension will be added to the ast and grammar imports.

Context: https://github.com/mermaid-js/mermaid/pull/4450#issuecomment-1575549546


msujew (approved):
@sidharthv96 Thanks for the contribution. I took the freedom to update the `jsExtension` to a more general `importExtension` field, in case someone wants to use `.cjs` or `.mjs` style imports.

Is this change blocking for https://github.com/mermaid-js/mermaid/pull/4450? I can create a bugfix release for this for the `langium-cli` package once this is merged, if you need it.

sidharthv96:
Yup, the value being configurable is better, thank you.

> I can create a bugfix release for this for the langium-cli package once this is merged, if you need it.

That would be wonderful @msujew, I guess it'll make @Yokozuna59's life a lot easier. 😇

Issue #585, opened 2022-07-08
Validate cross references to global but not imported types


dhuebner:
Currently any objects from the global scope can be referenced in the Grammar.
This will fail in CLI, as in CLI we index or lookup only Langium documents that are directly or transitively imported in the root 
Grammar.

In some places we are already validating references to a not imported object, like e.g. for rules, but interfaces and types can still be used without explicitly importing them.

Example:
grammar.langium
```js
grammar Test_file_ref

// import "types" <-- no import

entry Root returns IRoot: // IRoot can still be used. Should be marked as error
    name=ID;
```

types.langium
```js
interface IRoot {

}
```

We need a validation for that.
Alternatively we can decide to shrink the default Langium scope, but have an extended scoping implementation for the content assist.

msujew:
@hpopov Thanks, you're right, it's been fixed by https://github.com/langium/langium/pull/763 👍

PR #1073, opened 2023-06-05
Add test cases for cross-references resolution in Langium grammar

hpopov:
This verifies the issue https://github.com/langium/langium/issues/585 is already implemented.

### DIFF
--- a/packages/langium/test/grammar/langium-grammar-validator.test.ts
+++ b/packages/langium/test/grammar/langium-grammar-validator.test.ts
@@ -7,7 +7,7 @@
 import type { AstNode, Properties } from '../../src';
 import type { Assignment, CrossReference, ParserRule, UnionType } from '../../src/grammar/generated/ast';
 import type { ValidationResult } from '../../src/test';
-import { afterEach, beforeAll, describe, expect, test } from 'vitest';
+import { afterEach, beforeAll, beforeEach, describe, expect, test } from 'vitest';
 import { DiagnosticSeverity } from 'vscode-languageserver';
 import { createLangiumGrammarServices, EmptyFileSystem, GrammarAST, streamAllContents, streamContents } from '../../src';
 import { isAssignment } from '../../src/grammar/generated/ast';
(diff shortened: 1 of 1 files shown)


hpopov:
Dear @spoenemann, while investigating the implementation, that resolved the issue #585, I got concerned about hardcoded checks `des.type === Interface || des.type === Type` for references to `AbstractType`. ([see on gitHub](https://github.com/langium/langium/blob/1de4e459c644c4d679e20c2e73d5fb9ac4f64ad8/packages/langium/src/grammar/references/grammar-scope.ts#LL64C1-L66C10)):
```ts
        if (referenceType === AbstractType) {
            importedElements = importedElements.filter(des => des.type === Interface || des.type === Type);
        }
```
These basically means, that wherever in the grammar a cross-reference is specified to AbstractType, it will in fact only be resolved for Interface or Type. As far as a I was able to get it, it is expected, because whenever there is a reference to AbstractType in Langium Grammar, it in fact means a reference to either `Type` or `Interface`. Is it an expected behavior? If so, shouldn't rather the grammar be refactored? I.e., by introducing new union type `Interface | Type`, and using it as a substitute for references to `AbstractType`.

hpopov:
Thank you for elaboration!
After revieweing scope computation for grammar, I see what you mean. There seems to be more weird parts in the grammar, than I assumed. I have several concerns, like:

- `type AbstractType` has `Action` as one of its subtypes. However, in places where AbstractType:ID is referenced, it is never `Action`, since `Action` has no `ID` attribute.
- `type AbstractType` has also `ParserRule` in it. However, since `LangiumGrammarScopeComputation` exports (and adds to local scope too) either the ParserRule itself, or its inferredType as an Interface, does it make sense to try to resolve the reference against abstract types other than Type || Interface? I reckon not, and it seems, that `LangiumGrammarScopeProvider` doing exactly the thing: it ignores `AbstractTypes` other than `Type` or `Interface`.

I would suggest:
- to remove both `Action` and `ParserRule` from the `AbstractType` union;
- to rename `InferredType` into `InferredTypeDeclaration`, since it is not a type, but its declaration is processed as a part of either `ParserRule` or `Action`
- To use references to `AbstractType` without additional filtering. Optionally, a syntatic `interface InferredType` can be introduced in the grammar, added to the `AbstractType` union, and used instead of 'Interface' for synnthetically inferred types.

Yet let me check the code more thoroughly to see, will it break some other logic (e.g., code generation).

hpopov:
Otherwise it can be made plainly as `DeclaredType`

spoenemann:
The complexity you are seeing is due to the "magic" of combining explicitly declared types and implicitly generated types. Luckily the weird details that result from this are not so relevant for most users of Langium (it should "just work").

PR #1075, opened 2023-06-05
Generate syntax railroad diagrams

msujew:
Closes https://github.com/langium/langium/issues/827
Closes https://github.com/langium/langium/issues/1062

This change introduces a new package, `langium-railroad` that is able to generate railroad syntax diagrams from Langium grammars. The package is reused in `langium-cli` and `langium-vscode`:

1. To generate a railroad html file for each entry grammar (if specified in the langium-config.json).
2. Adds a preview command to the current langium editor that shows a webview with the railroad syntax diagram for the currently selected langium editor.

The vscode webview follows the current editor and updates on changes to the grammar file.

The new `langium-railroad` package can also be used in the playground for an inline syntax diagram view of the current grammar.

### DIFF
--- a/packages/langium-railroad/src/grammar-railroad.ts
+++ b/packages/langium-railroad/src/grammar-railroad.ts
@@ -5,53 +5,93 @@
  ******************************************************************************/
 
 import { GrammarAST, findNameAssignment } from 'langium';
-import * as railroad from 'railroad-diagrams';
+import type { FakeSVG } from 'railroad-diagrams';
+import * as rr from 'railroad-diagrams';
+
+export const defaultCss = `
+svg.railroad-diagram {
+    background-color: hsl(30,20%,95%);
+}
+svg.railroad-diagram path {
+    stroke-width: 3;
+    stroke: black;
+    fill: rgba(0,0,0,0);
+}
+svg.railroad-diagram text {
+    font: bold 14px monospace;
+    text-anchor: middle;
+}
+svg.railroad-diagram text.label {
+    text-anchor: start;
+}
+svg.railroad-diagram text.comment {
+    font: italic 12px monospace;
+}
+svg.railroad-diagram rect {
+    stroke-width: 3;
+    stroke: black;
+    fill: hsl(120,100%,90%);
+}
+`.trim();
 
 export function createGrammarDiagramHtml(grammar: GrammarAST.Grammar): string {
-    return createGrammarDiagram(grammar);
+    let text = `<!DOCTYPE HTML>
+<html>
+<head>
+<style>
+${defaultCss}
+</style>
+</head>
+<body>
+`;
+    text += createGrammarDiagram(grammar);
+    text += `</body>
+</html>`;
+
+    return text;
 }
 
 export function createGrammarDiagram(grammar: GrammarAST.Grammar): string {
     const nonTerminals = grammar.rules.filter(GrammarAST.isParserRule);
     let text = '';
     for (const nonTerminal of nonTerminals) {
-        text += '<h2>' + nonTerminal.name + '</h2>';
+        text += '<h2>' + nonTerminal.name + '</h2>\n';
         text += createRuleDiagram(nonTerminal);
     }
     return text;
 }
 
 function createRuleDiagram(rule: GrammarAST.ParserRule): string {
-    const diagram = new railroad.Diagram(toRailroad(rule.definition));
+    const diagram = new rr.Diagram(toRailroad(rule.definition));
     return diagram.toString();
 }
 
-function toRailroad(element: GrammarAST.AbstractElement): railroad.FakeSVG[] {
+function toRailroad(element: GrammarAST.AbstractElement): FakeSVG[] {
     if (GrammarAST.isAssignment(element)) {
         return wrapCardinality(element.cardinality, toRailroad(element.terminal));
     } else if (GrammarAST.isAlternatives(element)) {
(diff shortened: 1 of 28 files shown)


packages/langium-vscode/package.json
```diff
@@ -19,6 +19,23 @@
     "vscode": "^1.67.0"
   },
   "contributes": {
+    "commands": [
+      {
+        "command": "langium.showRailroadDiagram",
+        "title": "Show railroad syntax diagram",
```
spoenemann, line 25: Use Title Case for command titles.

packages/langium-vscode/src/railroad-webview.ts
```diff
@@ -0,0 +1,118 @@
...
+export class RailroadDiagramPanel implements vscode.Disposable {
+
+    uri: string;
+
+    private panel: vscode.WebviewPanel;
+    private client: LanguageClient;
+    private disposables: vscode.Disposable[] = [];
+
+    static current?: RailroadDiagramPanel;
+    static readonly viewType = 'railroadDiagram';
```
spoenemann, line 56: Maybe call it `LangiumRailroadDiagram` or so to ensure it's unique?

spoenemann (approved):
Great new feature! I also like how it's realized with relatively little code.

spoenemann:
@msujew could you also add a description of the new webview feature to the extension README (and update it while you're at it)?

montymxb (commented):
This is cool! I tried it out on a few of our example languages, and it works quite nicely. Just one suggestion below regarding terminals.

packages/langium-railroad/src/grammar-railroad.ts
```diff
@@ -0,0 +1,125 @@
...
+    text += createGrammarDiagram(grammar);
+    text += `</body>
+</html>`;
+
+    return text;
+}
+
+export function createGrammarDiagram(grammar: GrammarAST.Grammar): string {
+    const nonTerminals = grammar.rules.filter(GrammarAST.isParserRule);
+    let text = '';
```
Lotes, line 68: Are rule fragments a problem? Are they inlined or normal rules?

packages/langium-railroad/src/grammar-railroad.ts
```diff
@@ -0,0 +1,125 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { GrammarAST, findNameAssignment } from 'langium';
+import type { FakeSVG } from 'railroad-diagrams';
+import * as rr from 'railroad-diagrams';
+
+export const defaultCss = `
+svg.railroad-diagram {
```
Lotes, line 12: Can we also just export the SVG?

Lotes: When thinking about the size of grammar, this might be not so a great idea ^^*

packages/langium-railroad/src/grammar-railroad.ts
```diff
@@ -0,0 +1,125 @@
...
+`;
+    text += createGrammarDiagram(grammar);
+    text += `</body>
+</html>`;
+
+    return text;
+}
+
+export function createGrammarDiagram(grammar: GrammarAST.Grammar): string {
+    const nonTerminals = grammar.rules.filter(GrammarAST.isParserRule);
```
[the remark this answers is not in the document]
msujew: Our terminals are all regex based (internally at least). I don't think that the railroad diagrams package can transform them into a DFA style representation by itself.

packages/langium-railroad/src/grammar-railroad.ts
msujew: They are treated as being normal rules. I'm wondering whether it makes sense to inline them.

Lotes: I think it is fine to treat them as normal rules.

packages/langium-vscode/src/railroad-webview.ts
```diff
@@ -0,0 +1,120 @@
...
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import * as vscode from 'vscode';
+import type { LanguageClient } from 'vscode-languageclient/node';
+import { DOCUMENTS_VALIDATED_NOTIFICATION, RAILROAD_DIAGRAM_REQUEST } from './language-server/messages';
+
+export function registerRailroadWebview(client: LanguageClient): void {
+    vscode.commands.registerCommand('langium.showRailroadDiagram', (uri?: vscode.Uri) => {
+        if (uri instanceof vscode.Uri) {
```
Lotes, line 13: Can we not use the active document instead?
This case is coming up when one uses the Command list. So there should be a reaction, also when it is not a grammar, like: *Error* can only process *.langium files.

Lotes:
I think it would be a visual improvement if would choose the same background color as we use for the diagrams. Or make the diagrams transparent, but that would conflict with dark/light mode (lines are black and will not be visible in dark mode).

msujew:
> I think it would be a visual improvement if would choose the same background color as we use for the diagrams. Or make the diagrams transparent, but that would conflict with dark/light mode (lines are black and will not be visible in dark mode).

I think that we can leave those for later :) This PR does what it should, any additional changes can be performed on top.

Lotes:
The decision might be already made, but my first try to use the railroads was through the Langium CLI. I was a bit surprised that there is no `--help` that lists me the option to have diagrams.
I see that we have these flags in a Langium config file. Maybe one place too less. `langium init` command or a `langium generate --railroad` maybe?

PR #1077, opened 2023-06-07
Add support for langium-cli mode

msujew:
To improve bundle size when building a Langium based language server/parser, generating more succinct code is beneficial. This change adds a `--mode` argument for generating a compressed version of the grammar json.

### DIFF
--- a/packages/langium-cli/src/generator/grammar-serializer.ts
+++ b/packages/langium-cli/src/generator/grammar-serializer.ts
@@ -7,8 +7,9 @@ import type { Grammar, LangiumServices } from 'langium';
 import type { LangiumConfig } from '../package';
 import { CompositeGeneratorNode, NL, normalizeEOL, toString } from 'langium';
 import { generatedHeader } from './util';
+import type { GenerateOptions } from '../generate';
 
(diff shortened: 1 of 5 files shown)


packages/langium-cli/src/generator/grammar-serializer.ts
```diff
@@ -26,12 +28,16 @@ export function serializeGrammar(services: LangiumServices, grammars: Grammar[],
     for (let i = 0; i < grammars.length; i++) {
         const grammar = grammars[i];
         if (grammar.name) {
+            const production = (options.mode ?? config.mode) === 'production';
```
spoenemann, line 31: Suggestion: overwrite `config.mode` with the value from `options.mode`, if present, somewhere at the beginning of the execution, to ensure that we use a consistent setting everywhere (we might use the new option in other places later).

spoenemann (approved):
Good feature!

Lotes:
The term mode is very broad. I would associate this with a deployment environment or bundling behavior.

Mode is like type or kind. Very unspecific and confusing. You can get some inspiration from other build tools I hope.

spoenemann:
@Lotes here's the inspiration: https://webpack.js.org/configuration/mode/

PR #1079, opened 2023-06-09
fix(generator): prevent copied `syntaxes` from being  `syntaxes` child dir

Yokozuna59:
When executing the `copy:monarch-syntax` script, it performs the following:

```
public/
├── ...
├── syntaxes/                       # old from HelloWorld example
│   ├── language.monarch.js         # old from HelloWorld example
│   ├── language.monarch.js.map     # old from HelloWorld example
│   │
│   └── syntaxes/                   # new from actual grammar
│       ├── language.monarch.js     # new from actual grammar
│       └── language.monarch.js.map # new from actual grammar
└── ...
```

Whereas is should be:

```
public/
├── ...
├── syntaxes/                   # new from actual grammar
│   ├── language.monarch.js     # new from actual grammar
│   └── language.monarch.js.map # new from actual grammar
└── ...
```

### DIFF
--- a/packages/generator-langium/templates/web/.package.json
+++ b/packages/generator-langium/templates/web/.package.json
@@ -4,7 +4,7 @@
         "prepare:public": "shx mkdir -p ./public && shx cp -fr ./src/static/* ./public/",
         "copy:monaco-editor-wrapper": "shx cp -fr ./node_modules/monaco-editor-wrapper/bundle ./public/monaco-editor-wrapper",
         "copy:monaco-workers": "shx cp -fr ./node_modules/monaco-editor-workers/dist/ ./public/monaco-editor-workers",
-        "copy:monarch-syntax": "shx cp -fr ./out/syntaxes/ ./public/syntaxes",
+        "copy:monarch-syntax": "shx cp -fr ./out/syntaxes/ ./public/",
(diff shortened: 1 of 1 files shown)


packages/generator-langium/templates/web/.package.json
```diff
@@ -4,7 +4,7 @@
         "prepare:public": "shx mkdir -p ./public && shx cp -fr ./src/static/* ./public/",
         "copy:monaco-editor-wrapper": "shx cp -fr ./node_modules/monaco-editor-wrapper/bundle ./public/monaco-editor-wrapper",
         "copy:monaco-workers": "shx cp -fr ./node_modules/monaco-editor-workers/dist/ ./public/monaco-editor-workers",
-        "copy:monarch-syntax": "shx cp -fr ./out/syntaxes/ ./public/syntaxes",
+        "copy:monarch-syntax": "shx cp -fr ./out/syntaxes/ ./public/",
```
msujew, line 7: It seems like ending the source of `cp` with a `/` is [bad practice](https://stackoverflow.com/questions/18158248/should-i-put-trailing-slash-after-source-and-destination-when-copy-folders), We should use something like this instead, which explains better that the target is `public/syntaxes`: 
```suggestion
        "copy:monarch-syntax": "shx cp -fr ./out/syntaxes ./public/syntaxes",
```

Yokozuna59: Done.

msujew (approved):
Looks good to me, thanks 👍

PR #1090, opened 2023-06-21
Use `RootCstNode` for root node property

msujew:
We already assume that the `root` property is a root node [here](https://github.com/langium/langium/blob/0cd59e78a0b8e1dd681a379df2f1c06debee514a/packages/langium/src/parser/cst-node-builder.ts#L109). This fact wasn't reflected in the type yet.


spoenemann (approved):
:+1:

PR #1091, opened 2023-06-21
Perform name segment resolving lazily

msujew:
The current AST descriptions all contain the `nameSegment`, which is rarely used, but fairly costly to compute. This change makes the value lazily evaluated.


spoenemann (commented):
Looks good, thanks. Shall we also offer a utility function for making an AstNodeDescription serializable, i.e. replacing all getters with respective values? That can be useful when you want to optimize workspace loading by saving the index to a file.

dhuebner (approved):
LGTM. Thanks!

msujew:
> Shall we also offer a utility function for making an AstNodeDescription serializable, i.e. replacing all getters with respective values?

It is serializable, even when using getters. `get` properties get treated as normal object fields by serializers such as `JSON.stringify`.

spoenemann (approved):
Oh cool, good to know :+1:

PR #1094, opened 2023-06-26
Refactored document builder and affected analysis

spoenemann:
This PR fixes two issues that have not been noticed until recently:
 - When you open a DSL file, the whole workspace is validated even if the initial build skips validation. This is because `DefaultDocumentBuilder.collectDocuments` includes all documents for which `state < DocumentState.Validated`.
 - Even if you fix that, the whole workspace is validated in most cases:
     - The language extension is often activated when a DSL file is opened for the first time.
     - After starting and initializing the language server, the client immediately sends a `didOpen` notification, which triggers an update of the respective `TextDocument` (that's the standard behavior of `TextDocuments` provided by Microsoft).
     - The update cancels the initial build if it's still running (if the workspace is sufficiently large, it's very likely that it's still running).
     - The documents sent to the `DocumentBuilder` for the initial build are now continued with the new `BuildOptions`, which enable validation.

The issues are fixed with the following changes:
 - `DocumentBuilder` tracks the state of building (completed or not), including the last used build options.
 - This new information is now used to collect unfinished documents after a previous build was canceled.
 - Every document has its individual options during build; this is currently used to control validation.

Additional improvements:
 - API changes to simplify `IndexManager` and improve performance with large workspaces.
 - Added a public `initialBuildOptions` property to `WorkspaceManager` to control validation for the initial build.
 - Added a public `updateBuildOptions` property to `DocumentBuilder` to control validation for updates (document changes).
 - Added validation categories `'fast'` (default) and `'slow'`. You can set the category when you register a record of validation checks. The category can be selected in the build options. If no category is specified, all checks are executed.
 - Lexer, parser and linker errors are collected if the `'built-in'` validation category is enabled.
 - Added more validation options to enable reporting only partial results, e.g. don't do custom validations if there are parsing errors.
 - Made document builder phases more consistent: the `state` of the document is always set by the document builder

### DIFF
--- a/packages/langium/src/workspace/document-builder.ts
+++ b/packages/langium/src/workspace/document-builder.ts
@@ -10,7 +10,6 @@ import type { LangiumSharedServices } from '../services';
 import type { AstNode } from '../syntax-tree';
 import type { MaybePromise } from '../utils/promise-util';
 import type { ValidationOptions } from '../validation/document-validator';
-import type { ValidationCategory } from '../validation/validation-registry';
 import type { IndexManager } from '../workspace/index-manager';
 import type { LangiumDocument, LangiumDocuments, LangiumDocumentFactory } from './documents';
 import { CancellationToken, Disposable } from 'vscode-languageserver';
@@ -20,7 +19,7 @@ import { stream } from '../utils/stream';
 import { DocumentState } from './documents';
 
 export interface BuildOptions {
-    validationChecks?: 'none' | 'all' | ValidationCategory
+    validation?: boolean | ValidationOptions
 }
 
 export interface DocumentBuildState {
@@ -75,7 +74,9 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
 
     updateBuildOptions: BuildOptions = {
         // Default: run only the validation checks in the _fast_ category (includes those without category)
-        validationChecks: 'fast'
+        validation: {
+            category: 'fast'
+        }
     };
 
     protected readonly langiumDocuments: LangiumDocuments;
@@ -110,6 +111,7 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
             if (!invalidated) {
                 this.langiumDocuments.getOrCreateDocument(changedUri);
             }
+            this.buildState.delete(changedUri.toString());
         }
         // Set the state of all documents that should be relinked to `ComputedScopes` (if not already lower)
         const allChangedUris = stream(changed).concat(deleted).map(uri => uri.toString()).toSet();
@@ -259,11 +261,10 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
 
     /**
      * Determine whether the given document should be validated during a build. The default
-     * implementation checks the `validationChecks` property of the build options.
+     * implementation checks the `validation` property of the build options.
      */
     protected shouldValidate(document: LangiumDocument): boolean {
-        const options = this.getBuildOptions(document);
-        return options.validationChecks !== undefined && options.validationChecks !== 'none';
+        return Boolean(this.getBuildOptions(document).validation);
     }
 
     /**
@@ -271,11 +272,9 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
      */
     protected async validate(document: LangiumDocument, cancelToken: CancellationToken): Promise<void> {
         const validator = this.serviceRegistry.getServices(document.uri).validation.DocumentValidator;
-        const buildOptions = this.getBuildOptions(document);
-        const validationOptions: ValidationOptions = {
-            category: buildOptions.validationChecks === 'all' ? undefined : buildOptions.validationChecks as ValidationCategory
-        };
-        const diagnostics = await validator.validateDocument(document, validationOptions, cancelToken);
+        const validationSetting = this.getBuildOptions(document).validation;
+        const options = typeof validationSetting === 'object' ? validationSetting : undefined;
+        const diagnostics = await validator.validateDocument(document, options, cancelToken);
         document.diagnostics = diagnostics;
         document.state = DocumentState.Validated;
     }
--- a/packages/langium/src/workspace/document-builder.ts
+++ b/packages/langium/src/workspace/document-builder.ts
@@ -19,11 +19,19 @@ import { stream } from '../utils/stream';
 import { DocumentState } from './documents';
 
 export interface BuildOptions {
+    /**
+     * Control the validation phase with this option:
+     *  - `true` enables all validation checks
+     *  - An object with the `category` property enables a subset of validation checks
+     *  - `false` or `undefined` disables all validation checks
+     */
     validation?: boolean | ValidationOptions
 }
 
 export interface DocumentBuildState {
+    /** Whether a document has completed its last build process. */
     completed: boolean
+    /** The options used for the last build process. */
     options: BuildOptions
 }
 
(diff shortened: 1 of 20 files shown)


msujew (commented):
This feels very heavy-handed and workaround-y. I don't think that the document should know about ephemeral build information (aside from its own state). I can also imagine that this property being set only once per completed build might lead to inconsistencies when multiple builds with different options are started. Maybe you want `none` to override `all`.

I would much rather like to see a change in `BuildOptions.validationChecks` to include a `only-open` value that only performs validation on files that are in the `TextDocuments` service. And use that as the default for the `textDocument/open` request we receive from the language client.

spoenemann:
I used this opportunity to add tests for ValidationRegistry, DocumentValidator and DocumentBuilder (there weren't any).

packages/langium/src/workspace/document-builder.ts
```diff
@@ -20,7 +19,7 @@ import { stream } from '../utils/stream';
 import { DocumentState } from './documents';
 
 export interface BuildOptions {
-    validationChecks?: 'none' | 'all' | ValidationCategory
+    validation?: boolean | ValidationOptions
```
sailingKieler, line 22: Please clarify the relation between `true`,  `slow`, and `fast`.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -63,12 +71,21 @@ export interface DocumentBuilder {
 export type DocumentUpdateListener = (changed: URI[], deleted: URI[]) => void
 export type DocumentBuildListener = (built: LangiumDocument[], cancelToken: CancellationToken) => void | Promise<void>
 export class DefaultDocumentBuilder implements DocumentBuilder {
+
+    updateBuildOptions: BuildOptions = {
+        // Default: run only the validation checks in the _fast_ category (includes those without category)
```
sailingKieler, line 76: Please add a similar hint to the definition of `ValidationCategory`.

sailingKieler (approved):
I have a few minor suggestions, no major objections

packages/langium/src/workspace/document-builder.ts
```diff
@@ -114,27 +150,16 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
         });
     }
 
-    protected collectDocuments(changed: LangiumDocument[], deleted: URI[]): LangiumDocument[] {
```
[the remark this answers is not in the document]
spoenemann: Yes I doubt that anyone will need to override the now simplified collection logic.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -207,8 +238,8 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
      * Determine whether the given document should be validated during a build. The default
      * implementation checks the `validationChecks` property of the build options.
      */
-    protected shouldValidate(_document: LangiumDocument, options: BuildOptions): boolean {
-        return options.validationChecks === 'all';
+    protected shouldValidate(document: LangiumDocument): boolean {
+        return document.build?.options.validationChecks === 'all';
```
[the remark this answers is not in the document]
spoenemann: This diff is outdated. The current version is
```typescript
return Boolean(this.getBuildOptions(document).validation);
```

msujew (approved):
After rethinking my initial restraint to approve this, I'm happy with the new design. I'm sure there's still some space to iterate on the details, but this works well for now 👍

spoenemann:
I changed where the `state` of a document is updated in the build process because that was quite inconsistent.

The new options merging logic in the `build` method of DocumentBuilder has become rather complex, @msujew @sailingKieler please have a look.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -104,7 +105,38 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+                        this.buildState.set(key, {
+                            completed: false,
+                            options: {
+                                validation: {
+                                    ...options.validation,
+                                    categories
+                                }
+                            }
+                        });
+                        document.state = DocumentState.IndexedReferences;
```
msujew, line 133: This method of identifying missing categories lacks a real delta. See the following test:

```ts

await builder.build([document1], { validation: { categories: ['built-in', 'fast'] } });
expect(document1.state).toBe(DocumentState.Validated);
expect(document1.diagnostics?.map(d => d.message)).toEqual([
    'Value is too large: 11'
]);

await builder.build([document1], { validation: { categories: ['slow'] } });
expect(document1.state).toBe(DocumentState.Validated);
expect(document1.diagnostics?.map(d => d.message)).toEqual([
    'Value is too large: 11',
    'Bar is too long: AnotherStrangeBar'
]);

await builder.build([document1], { validation: { categories: ['fast'] } });
expect(document1.state).toBe(DocumentState.Validated);
expect(document1.diagnostics?.map(d => d.message)).toEqual([
    'Value is too large: 11',
    'Bar is too long: AnotherStrangeBar'
]);
```

During the last step (running `fast` again), we actually get 3 diagnostics (we get `'Value is too large: 11'` twice), since the document builder "forgot" that we have run `fast` already, and only remembers that we ran `slow`.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -114,58 +213,73 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+            return linker.link(doc, cancelToken);
+        });
         // 4. Index references
         await this.runCancelable(documents, DocumentState.IndexedReferences, cancelToken, doc =>
             this.indexManager.updateReferences(doc, cancelToken)
         );
         // 5. Validation
-        const validateDocs = documents.filter(doc => this.shouldValidate(doc, options));
-        await this.runCancelable(validateDocs, DocumentState.Validated, cancelToken, doc =>
+        const toBeValidated = documents.filter(doc => this.shouldValidate(doc));
```
dhuebner, line 245: If `shouldValidate` returns `false` the document will never reach the Validated state. This state is checked in Line 113. Does that mean that the custom implementation inside the `shouldValidate` should take care about setting the build state? If so we should add this information to the documentation.

spoenemann: No, the build state is set in `runCancelable` (line 282). Custom implementations of `shouldValidate` don't need to check or modify that property.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -114,58 +213,73 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
-        return Array.from(docSet);
-    }
-
+    /**
+     * Build the given documents by stepping through all build phases. If a document's state indicates
+     * that a certain build phase is already done, the phase is skipped for that document.
+     */
     protected async buildDocuments(documents: LangiumDocument[], options: BuildOptions, cancelToken: CancellationToken): Promise<void> {
+        this.prepareBuild(documents, options);
         // 0. Parse content
```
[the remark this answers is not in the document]
spoenemann: IMO the current numbering makes sense because it expresses that document of that state will be included in each phase. Example: documents with state Changed = 0 are included in the "Parse content" phase.

dhuebner (approved):

packages/langium/src/workspace/document-builder.ts
dhuebner: `toBeValidated` will not contain the document for that `shouldValidate ` returns false. For this document `runCancelable`  will not be called. Or?

spoenemann: Right, so the state won't be set to Validated, but that's fine. The check you mentioned in line 113 is only for documents where validation is already partially done.

PR #1095, opened 2023-06-26
Add comment provider

Lotes:
- serializes to `$comment` property
- comes with default multiline comments
- replaces consumption of comments in documentation provider by a comment provider call

- one thing that bothers me is that comments can get duplicated (I guess)

```bnf
A ::= B c
B ::= a b
```

For file:

```
/** comment */
a b c
```

- because then `$comment(A) == $comment(B) == '/** comment */'`
- @msujew what do you think? Can we live with this or should we filter out the comment for upper nodes when a lower node already has the comment?

### DIFF
--- a/packages/langium/src/documentation/comment-provider.ts
+++ b/packages/langium/src/documentation/comment-provider.ts
@@ -5,6 +5,7 @@
  ******************************************************************************/
 
 import type { GrammarConfig } from '../grammar';
+import type { AstNodeWithComment } from '../serializer';
 import type { LangiumServices } from '../services';
 import type { AstNode } from '../syntax-tree';
 import { findCommentNode } from '../utils/cst-util';
@@ -13,13 +14,6 @@ import { findCommentNode } from '../utils/cst-util';
  * Provides comments for AST nodes.
  */
 export interface CommentProvider {
-    /**
-     * Defines the token types that are considered as comments.
-     *
-     * The default implementation `DefaultCommentProvider` will return the `multilineCommentRules` from the grammar config.
-     */
-    getCommentTokenTypes(): string[];
-
     /**
      * Returns the comment associated with the specified AST node.
      * @param node The AST node to get the comment for.
@@ -33,10 +27,10 @@ export class DefaultCommentProvider implements CommentProvider {
     constructor(services: LangiumServices) {
         this.grammarConfig = () => services.parser.GrammarConfig;
     }
-    getCommentTokenTypes(): string[] {
-        return this.grammarConfig().multilineCommentRules;
-    }
     getComment(node: AstNode): string | undefined {
-        return findCommentNode(node.$cstNode, this.getCommentTokenTypes())?.text;
+        if('$comment' in node) {
+            return (node as AstNodeWithComment).$comment;
+        }
+        return findCommentNode(node.$cstNode, this.grammarConfig().multilineCommentRules)?.text;
     }
 }
--- a/packages/langium/src/documentation/comment-provider.ts
+++ b/packages/langium/src/documentation/comment-provider.ts
@@ -1,0 +1,22 @@
+/******************************************************************************
+ * Copyright 2023 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import type { AstNode } from '../syntax-tree';
+import { findCommentNode } from '../utils/cst-util';
+
+export interface CommentProvider {
+    getCommentTokenTypes(): string[];
+    getComment(node: AstNode): string | undefined;
+}
+
+export class DefaultCommentProvider implements CommentProvider {
+    getCommentTokenTypes(): string[] {
+        return ['SL_COMMENT', 'ML_COMMENT'];
+    }
+    getComment(node: AstNode): string | undefined {
+        return findCommentNode(node.$cstNode, this.getCommentTokenTypes())?.text;
+    }
+}
--- a/packages/langium/src/documentation/comment-provider.ts
+++ b/packages/langium/src/documentation/comment-provider.ts
@@ -15,12 +15,12 @@ export interface CommentProvider {
 }
 
 export class DefaultCommentProvider implements CommentProvider {
-    protected readonly grammarConfig: GrammarConfig;
+    protected readonly grammarConfig: () => GrammarConfig;
     constructor(services: LangiumServices) {
-        this.grammarConfig = services.parser.GrammarConfig;
+        this.grammarConfig = () => services.parser.GrammarConfig;
     }
     getCommentTokenTypes(): string[] {
-        return this.grammarConfig.multilineCommentRules;
+        return this.grammarConfig().multilineCommentRules;
     }
     getComment(node: AstNode): string | undefined {
         return findCommentNode(node.$cstNode, this.getCommentTokenTypes())?.text;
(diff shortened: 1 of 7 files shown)


packages/langium/src/serializer/json-serializer.ts
```diff
@@ -130,6 +141,12 @@ export class DefaultJsonSerializer implements JsonSerializer {
                 astNode ??= { ...value };
                 astNode.$sourceText = value.$cstNode?.text;
             }
+            if (comments && isAstNode(value)) {
+                astNode ??= {
+                    ...value,
+                    $comment: this.commentProvider.getComment(value)
+                } as AstNodeWithComment;
```
msujew, line 148: This will fail in case either `textRegions` or `sourceText` is true. The right side of `??=` is only evaluated when the left side is `undefined`. Therefore, if it's not undefined, `$comment` will not be set.

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -30,6 +32,13 @@ export interface AstNodeWithTextRegion extends AstNode {
     $textRegion?: AstNodeRegionWithAssignments;
 }
 
+/**
+ * {@link AstNode}s that may carry a comment CST node in front of itself.
+ */
```
msujew, line 37: Suggestion: Given that the `CommentProvider` might return any kind of different comment (only the default is "right in front of the node"):

```suggestion
/**
 * {@link AstNode}s that may carry a semantically relevant comment.
 */
```

packages/langium/src/documentation/comment-provider.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Provides comments for AST nodes.
+ */
+export interface CommentProvider {
+    /**
+     * Defines the token types that are considered as comments.
+     *
+     * The default implementation `DefaultCommentProvider` will return the `multilineCommentRules` from the grammar config.
+     */
+    getCommentTokenTypes(): string[];
```
msujew, line 21: Suggestion: I thought about to use this to replace `GrammarConfig#multilineCommentRules`, but they actually serve different purposes. I'm paddling back on this and propose to remove this interface method.

packages/langium/src/documentation/comment-provider.ts
```diff
@@ -0,0 +1,42 @@
...
+export class DefaultCommentProvider implements CommentProvider {
+    protected readonly grammarConfig: () => GrammarConfig;
+    constructor(services: LangiumServices) {
+        this.grammarConfig = () => services.parser.GrammarConfig;
+    }
+    getCommentTokenTypes(): string[] {
+        return this.grammarConfig().multilineCommentRules;
+    }
+    getComment(node: AstNode): string | undefined {
+        return findCommentNode(node.$cstNode, this.getCommentTokenTypes())?.text;
```
msujew, line 40: Suggestion: The default implementation should also look for `$comment` in the `AstNode` and return it if it exists. Otherwise, currently no service makes use of the `$comment` property.

packages/langium/src/documentation/comment-provider.ts
Lotes: So I guess I have to use `GrammarConfig#multilineCommentRules` in-place

packages/langium/src/serializer/json-serializer.ts
Lotes: Something is wrong with this snippet:
```ts
else {
  let astNode: AstNodeWithTextRegion | undefined = undefined;
  if (textRegions && isAstNode(value)) {
      astNode = this.addAstNodeRegionWithAssignmentsTo({ ...value });
      if ((!key || value.$document) && astNode?.$textRegion) {
          try {
              astNode.$textRegion.documentURI = getDocument(value).uri.toString();
          } catch (e) { /* do nothing */ }
      }
  }
  if (sourceText && !key && isAstNode(value)) {
      astNode ??= { ...value };
      astNode.$sourceText = value.$cstNode?.text;
  }
  if (comments && isAstNode(value)) {
      astNode ??= {
          ...value,
          $comment: this.commentProvider.getComment(value)
      } as AstNodeWithComment;
  }
  return astNode ?? value;
}
```

When `textRegions` and `sourceText` are true, only `astNode.$sourceText` will be set. And I did some similar mistake by copy-pasting.

Lotes: Ah! No sorry, now I see it. Very weird way of writing this

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -30,6 +32,13 @@ export interface AstNodeWithTextRegion extends AstNode {
     $textRegion?: AstNodeRegionWithAssignments;
 }
 
+/**
+ * {@link AstNode}s that may carry a semantically relevant comment.
+ */
+export interface AstNodeWithComment extends AstNode {
```
spoenemann, line 38: Why is this defined in this file? Why not an optional property of `AstNode`?

spoenemann (commented):
@Lotes could you describe the main use cases for this, besides the documentation provider?

packages/langium/src/serializer/json-serializer.ts
msujew: For the same reason that the `$sourceText` and `$textRegion` properties are defined in this file as well: They are only relevant on serialized AST nodes, and will not be set any AST node during normal operation of Langium.

msujew:
@spoenemann There are two relevant design changes in here:

1. Adopters of Langium can choose to override the `getComment` method to change how other services derive the comment from a given AST node. Right now, all services use `findCommentNode`, which isn't overridable.
2. Since we lose the CST during serialization, we also lose all semantically relevant comments. You could argue that comments should never be semantically relevant, but this is not the case for Langium. For example, what if I want to generate interface definitions [similar to TSDoc](https://tsdoc.org/) from a serialized version of my AST? We need the AST comment for that, which is exactly what the serializer change in this PR accomplishes.

msujew:
> what do you think? Can we live with this or should we filter out the comment for upper nodes when a lower node already has the comment?

We can live with that, since filtering it out might be potentially destructive.

packages/langium/src/documentation/comment-provider.ts
```diff
@@ -0,0 +1,36 @@
...
+    getComment(node: AstNode): string | undefined;
+}
+
+export class DefaultCommentProvider implements CommentProvider {
+    protected readonly grammarConfig: () => GrammarConfig;
+    constructor(services: LangiumServices) {
+        this.grammarConfig = () => services.parser.GrammarConfig;
+    }
+    getComment(node: AstNode): string | undefined {
+        if('$comment' in node) {
```
spoenemann, line 31: Please add a function `isAstNodeWithComment` to check whether the property is present.

I would use `typeof (node as AstNodeWithComment).$comment === 'string'` instead of the `in` operator, as that's safer for type checking.

spoenemann (commented):
Ok, makes sense.

packages/langium/src/documentation/comment-provider.ts
Lotes: Done

packages/langium/src/serializer/json-serializer.ts
```diff
@@ -130,6 +145,10 @@ export class DefaultJsonSerializer implements JsonSerializer {
                 astNode ??= { ...value };
                 astNode.$sourceText = value.$cstNode?.text;
             }
+            if (comments && isAstNode(value)) {
+                astNode ??= { ...value };
+                (astNode as AstNodeWithComment).$comment =  this.commentProvider.getComment(value);
```
msujew, line 150: Minor: Formatting
```suggestion
                (astNode as AstNodeWithComment).$comment = this.commentProvider.getComment(value);
```

msujew (approved):
Looks good to me, thanks 👍 

One minor comment, see below.

PR #1100, opened 2023-06-29
Support workspace symbols

msujew:
Closes https://github.com/langium/langium/issues/1099


packages/langium/src/lsp/workspace-symbol-provider.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    protected readonly serviceRegistry: ServiceRegistry;
+    protected readonly documents: LangiumDocuments;
+
+    constructor(services: LangiumSharedServices) {
+        this.serviceRegistry = services.ServiceRegistry;
+        this.documents = services.workspace.LangiumDocuments;
+    }
+
+    async getSymbols(params: WorkspaceSymbolParams, cancelToken = CancellationToken.None): Promise<WorkspaceSymbol[]> {
```
spoenemann, line 29: The [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_symbol) says:

> The workspace symbol request is sent from the client to the server to list project-wide symbols

Wouldn't it be more accurate to use `IndexManager.allElements` to find the workspace symbols?

packages/langium/src/services.ts
```diff
@@ -154,6 +155,7 @@ export type LangiumDefaultSharedServices = {
         DocumentLinkProvider?: DocumentLinkProvider
         ExecuteCommandHandler?: ExecuteCommandHandler
         InlayHintProvider?: InlayHintProvider
+        WorkspaceSymbolProvider?: WorkspaceSymbolProvider
```
spoenemann, line 158: We don't have to make this optional if we have a good default?

msujew: We currently make every LSP service optional, so adopters can just remove them if they want to.

packages/langium/src/lsp/workspace-symbol-provider.ts
msujew: Yeah, that probably makes more sense 👍

packages/langium/src/lsp/document-symbol-provider.ts
```diff
@@ -68,8 +71,10 @@ export class DefaultDocumentSymbolProvider implements DocumentSymbolProvider {
         return undefined;
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    protected getSymbolKind(type: string): SymbolKind {
+    /**
+     * @deprecated Use the `NodeKindProvider` service instead.
```
spoenemann, line 75: I think we can remove this unless you want to add the change to v1.3.0. Note that users won't easily notice the deprecation because the TypeScript editor doesn't indicate that you're overriding a deprecated method.

packages/langium/src/lsp/node-kind-provider.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright 2023 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import type { AstNode, AstNodeDescription } from '../syntax-tree';
+import { CompletionItemKind, SymbolKind } from 'vscode-languageserver';
+
+/**
+ * This interface consolidates the logic for gathering LSP kind information based on AST nodes or their descriptions.
```
spoenemann, line 11:
```suggestion
 * This service consolidates the logic for gathering LSP kind information based on AST nodes or their descriptions.
```

packages/langium/src/lsp/node-kind-provider.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export interface NodeKindProvider {
+    /**
+     * Returns a `SymbolKind` as used by `WorkspaceSymbolProvider` or `DocumentLinkSymbolProvider`.
+     */
+    getSymbolKind(node: AstNode | AstNodeDescription): SymbolKind;
+    /**
+     * Returns a `CompletionItemKind` as used by the `CompletionProvider`.
+     */
+    getCompletionKind(node: AstNode | AstNodeDescription): CompletionItemKind;
```
spoenemann, line 21: Name this method `getCompletionItemKind` for consistency?

packages/langium/src/lsp/workspace-symbol-provider.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { IndexManager } from '../workspace';
+import type { MaybePromise} from '../utils/promise-util';
+import type { AstNodeDescription } from '../syntax-tree';
+import type { NodeKindProvider } from './node-kind-provider';
+import { CancellationToken } from 'vscode-languageserver';
+import { interruptAndCheck } from '../utils/promise-util';
+
+export interface WorkspaceSymbolProvider {
+    getSymbols(params: WorkspaceSymbolParams, cancelToken?: CancellationToken): MaybePromise<WorkspaceSymbol[]>;
+    resolveSymbol?(symbol: WorkspaceSymbol, cancelToken?: CancellationToken): MaybePromise<WorkspaceSymbol>;
```
spoenemann, line 18: Please add the standard comments as for other LSP services.

packages/langium/src/lsp/workspace-symbol-provider.ts
```diff
@@ -0,0 +1,61 @@
...
+        this.indexManager = services.workspace.IndexManager;
+        this.nodeKindProvider = services.lsp.NodeKindProvider;
+    }
+
+    async getSymbols(params: WorkspaceSymbolParams, cancelToken = CancellationToken.None): Promise<WorkspaceSymbol[]> {
+        const workspaceSymbols: WorkspaceSymbol[] = [];
+        const query = params.query.toLowerCase();
+        for (const description of this.indexManager.allElements()) {
+            await interruptAndCheck(cancelToken);
+            if (description.name.toLowerCase().includes(query)) {
```
spoenemann, line 36: Shall we use the same matching logic as you made for completion?

spoenemann (commented):
Thanks, this looks good!

packages/langium/src/lsp/workspace-symbol-provider.ts
msujew: I've extracted the logic into a `FuzzyMatcher` service.

spoenemann (approved):
Very good, just one details seems to be odd with the fuzzy matching: I would expect any substring of `Statemachine` to match that symbol, but only a prefix does it.

Screenshot 2023-07-03 at 08 41 26

msujew:
> just one details seems to be odd with the fuzzy matching: I would expect any substring of Statemachine to match that symbol, but only a prefix does it.

The fuzzy matching done by VSCode only matches characters included in a word, if the word start/boundary is matching. That's working as expected and the implementation of the fuzzy matching is a 1:1 translation from the TypeScript source code.

spoenemann:
Ok but why does `ma` match `Multiplication`?

BTW TypeScript completion behaves differently in this regard. Do we need two different versions of FuzzyMatch?
Screenshot 2023-07-03 at 18 06 59

msujew:
`ma` matches `Multiplication` because of **M**ultiplic**a**tion. Fuzzy matching is case-insensitive. As far as I could tell, the TypeScript completion behaves exactly like ours. Do you have a specific example that I can use to create a unit test?

spoenemann:
I observe different behavior between TypeScript and our symbol matching.

TypeScript, searching for `DocumentBuilder`:
 - `documentbuilder`, `DB`, `DocBui`, `docBui` match
 - `docbui`, `DcBi`, `ocuuil` don't match

FuzzyMatcher, searching for `BinaryExpression`:
 - `binaryexpression`, `BE`, `BinExp`, `binExp`, `binexp`, `BnEp` match
 - `inxp` doesn't match

PR #1105, opened 2023-07-03
Fix generated directory paths in cli

msujew:
The old logic really didn't make a lot of sense:

```ts
const parentDir = path.dirname(grammarPath).split(path.sep).pop();
parentDir && await mkdirWithFail(parentDir, options);
```

This effectively created a directory in the `cwd` with the name of the last path segment before the `grammarPath` file name. This wasn't want this feature was supposed to do. It should ensure that the directory for any file that we write actually exists. This change fixes the behavior.


spoenemann (approved):
👍

PR #1106, opened 2023-07-03
Support completion for grammar imports

msujew:
Closes https://github.com/langium/langium/issues/350

Fixes a few minor issues in the completion provider related to completion in strings. Provides a bit more information to consumers of the API about the current token position. This allows to replace the current token more easily.

### DIFF
--- a/packages/langium/src/grammar/lsp/grammar-completion-provider.ts
+++ b/packages/langium/src/grammar/lsp/grammar-completion-provider.ts
@@ -1,0 +1,118 @@
+/******************************************************************************
+ * Copyright 2023 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import type { Range } from 'vscode-languageserver-types';
+import { CompletionItemKind } from 'vscode-languageserver-types';
+import type { NextFeature } from '../../lsp';
+import type { CompletionAcceptor, CompletionContext } from '../../lsp/completion/completion-provider';
+import { DefaultCompletionProvider } from '../../lsp/completion/completion-provider';
+import type { LangiumServices } from '../../services';
+import type { MaybePromise} from '../../utils';
+import { getContainerOfType} from '../../utils';
+import { equalURI } from '../../utils';
+import type { LangiumDocument, LangiumDocuments } from '../../workspace';
+import type { AbstractElement } from '../generated/ast';
+import { isAssignment } from '../generated/ast';
+import { Utils } from 'vscode-uri';
+
+export class LangiumGrammarCompletionProvider extends DefaultCompletionProvider {
+
+    private readonly documents: LangiumDocuments;
+
+    constructor(services: LangiumServices) {
+        super(services);
+        this.documents = services.shared.workspace.LangiumDocuments;
+    }
+
+    protected override completionFor(context: CompletionContext, next: NextFeature<AbstractElement>, acceptor: CompletionAcceptor): MaybePromise<void> {
+        const feature = getContainerOfType(next.feature, isAssignment);
+        if (feature?.feature === 'path') {
+            this.completeImportPath(context, acceptor);
+        } else {
+            return super.completionFor(context, next, acceptor);
+        }
+    }
+
+    private completeImportPath(context: CompletionContext, acceptor: CompletionAcceptor): void {
+        const text = context.textDocument.getText();
+        const existingText = text.substring(context.tokenOffset, context.offset);
+        let allPaths = this.getAllFiles(context.document);
+        let range: Range = {
+            start: context.position,
+            end: context.position
+        };
+        if (existingText.length > 0) {
+            const existingPath = existingText.substring(1);
+            allPaths = allPaths.filter(path => path.startsWith(existingPath));
+            // Completely replace the current token
+            const start = context.textDocument.positionAt(context.tokenOffset + 1);
+            const end = context.textDocument.positionAt(context.tokenEndOffset);
+            range = {
+                start,
+                end
+            };
+        }
+        for (const path of allPaths) {
+            // Only insert quotes if there is no `path` token yet.
+            const delimiter = existingText.length > 0 ? '' : '"';
+            const completionValue = `${delimiter}${path}${delimiter}`;
+            acceptor({
+                label: path,
+                textEdit: {
+                    newText: completionValue,
+                    range
+                },
+                kind: CompletionItemKind.File,
+                sortText: '0'
+            });
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


packages/langium/src/grammar/lsp/grammar-completion-provider.ts
```diff
@@ -0,0 +1,118 @@
...
+                const docUri = doc.uri.toString();
+                const uriWithoutExt = docUri.substring(0, docUri.length - Utils.extname(doc.uri).length);
+                const relativePath = this.createRelativePath(dirname, uriWithoutExt);
+                paths.push(relativePath);
+            }
+        }
+        return paths;
+    }
+
+    private createRelativePath(fromPath: string, toPath: string): string {
```
spoenemann, line 90: We already have the utility function `relativeURI`, can we somehow consolidate this?

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -40,6 +40,8 @@ export interface CompletionContext {
     node?: AstNode
     document: LangiumDocument
     textDocument: TextDocument
+    tokenOffset: number
+    tokenEndOffset: number
     offset: number
     position: Position
```
spoenemann, line 46: Please add comments to these properties so it's clear how to use them in custom completion code.

spoenemann (commented):
Thanks! Observation: I don't get any completion if the cursor is at the end of the import string and the string is not empty. For example, it works with `"|"` and `"./|asdf"`, but not with `"./|"` (`|` means cursor position).

msujew:
> I don't get any completion if the cursor is at the end of the import string and the string is not empty. For example, it works with "|" and "./|asdf", but not with "./|" (| means cursor position).

This led me down a pretty deep completion related rabbit hole. However, I believe the new approach to completion context computation is more robust (and more accurate). It even works inside of strings.

packages/langium/src/grammar/lsp/grammar-completion-provider.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    private readonly documents: () => LangiumDocuments;
+
+    constructor(services: LangiumServices) {
+        super(services);
+        this.documents = () => services.shared.workspace.LangiumDocuments;
+    }
+
+    protected override completionFor(context: CompletionContext, next: NextFeature<AbstractElement>, acceptor: CompletionAcceptor): MaybePromise<void> {
+        const feature = getContainerOfType(next.feature, isAssignment);
```
dhuebner, line 31: It is a bit confusing. Should it be better named 'assignment'?

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -184,19 +156,20 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
 
-        await Promise.all(
-            stream(beforeFeatures)
-                .distinct(distinctionFunction)
-                .map(e => this.completionForContexts(contexts, e, acceptor))
-        );
-
-        if (reparse) {
+        const completedFeatures: NextFeature[] = [];
+        for (const context of contexts) {
```
dhuebner, line 160: It would be nice to have a way to stop processing other contextes inside a called complitionFor(). Or at least move this for loop to a separate protected method

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -109,73 +132,22 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
-        if (!cst) {
-            return undefined;
-        }
         const items: CompletionItem[] = [];
-        const textDocument = document.textDocument;
-        const text = textDocument.getText();
-        const offset = textDocument.offsetAt(params.position);
-        const acceptor: CompletionAcceptor = value => {
-            const completionItem = this.fillCompletionItem(textDocument, offset, value);
+        const contexts = this.buildContexts(document, params.position);
```
dhuebner, line 142: Can we turn `items` to a Stream?

msujew: What for? We need to constantly append to it, and once we're done, we need to convert it to an array for the LSP response.

packages/langium/src/grammar/lsp/grammar-completion-provider.ts
```diff
@@ -0,0 +1,93 @@
...
+            // Only insert quotes if there is no `path` token yet.
+            const delimiter = existingText.length > 0 ? '' : '"';
+            const completionValue = `${delimiter}${path}${delimiter}`;
+            acceptor(context, {
+                label: path,
+                textEdit: {
+                    newText: completionValue,
+                    range
+                },
+                kind: CompletionItemKind.File,
```
dhuebner, line 68: In case we have already a path: `import <|>"../common"`  one could remove `"../common"` using a textEdit

dhuebner (approved):
Works really good, nice feature! 
I left some comments and improvement suggestions.

packages/langium/src/lsp/completion/completion-provider.ts
msujew: It didn't really make sense as a stream, but I've changed `buildContexts` into a generator function. That way, if we end early, we don't need to compute the other contexts.

packages/langium/src/grammar/lsp/grammar-completion-provider.ts
```diff
@@ -0,0 +1,93 @@
...
+            const end = context.textDocument.positionAt(context.tokenEndOffset - 1);
+            range = {
+                start,
+                end
+            };
+        }
+        for (const path of allPaths) {
+            // Only insert quotes if there is no `path` token yet.
+            const delimiter = existingText.length > 0 ? '' : '"';
+            const completionValue = `${delimiter}${path}${delimiter}`;
```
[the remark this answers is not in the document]
msujew: We don't have reverse value conversion yet (i.e. converted value -> text value). I don't think implementing this for this PR is necessary.

packages/langium/src/grammar/lsp/grammar-completion-provider.ts
msujew: I don't think we should do that:

1. The completion context is generally unaware of any token that starts at or after the cursor position. This is the expected behavior for completion, as it should be independent of any following tokens.
2. It works the same way in TypeScript, where text is replaced if you're inside of a string constant, whereas text is inserted if you're at the start of a string constant.
image
image

packages/langium/src/lsp/completion/completion-provider.ts
msujew: I've added a `continueCompletion` method, that's supposed to return whether to continue with the next context.

msujew:
@spoenemann Do you want to take another look?

PR #1107, opened 2023-07-05
Move shared LSP services to language services

msujew:
Closes https://github.com/langium/langium/issues/925

Removes the `resolve` requests, which aren't usually necessary in Langium. Note that consumers off Langium can still opt to manually edit the language server initialization return value to support these requests manually.


spoenemann (approved):
👍

PR #1108, opened 2023-07-05
Add parser error message provider service

msujew:
Closes https://github.com/langium/langium/issues/1058

Adds a small abstraction layer (`AbstractParserErrorMessageProvider`) so that adopters can change specific messages more easily.


packages/langium/src/parser/langium-parser.ts
```diff
@@ -350,27 +353,22 @@ export interface IParserDefinitionError {
     ruleName?: string
 }
 
-export class LangiumParserErrorMessageProvider implements IParserErrorMessageProvider {
+export class AbstractParserErrorMessageProvider implements IParserErrorMessageProvider {
```
spoenemann, line 356: The class is not abstract; shall we call it `Default...`?

spoenemann (approved):

PR #1109, opened 2023-07-05
Improve regular expression generator

msujew:
Closes https://github.com/langium/langium/issues/740

This was initially fixed back in https://github.com/langium/langium/pull/1070. However, the fix to escape `/` was just a workaround. While escaping `/` works as expected, it does not need to be escaped. The only reason it has to be escaped in JS is that `/` is the regex literal delimiter.

Using `new RegExp().toString()` automatically escapes `/` as needed.


spoenemann (commented):
There are test failures, maybe you need to update expectations.

spoenemann (approved):
:+1:

PR #1113, opened 2023-07-06
Resolve transitive imports in grammar

### LINKED ISSUE #924: Support transitive imports
I have several XText grammars with dependencies as follows:
```
grammar com.epistimis.face.Face with com.epistimis.uddl.Uddl
grammar com.epistimis.uddl.Uddl with org.eclipse.xtext.common.Terminals
```
I have used XText2Langium, v. 0.4.0 to generate the equivalent Langium grammars.

When I generate from the Uddl grammar, I get the following in `Uddl.langium`:
```
grammar Uddl
import 'Terminals'
```

When I generate from the Face grammar, I get the following in `Face.langium`:
```
grammar Face
import 'Uddl'
```

When I open the `Face.langium` grammar, I get a lot of errors `Could not resolve reference to AbstractRule rule named 'ID'`

I can resolve this by adding `Import Terminals` at the top of the `Face.langium` file. But I shouldn't have to. Since `Uddl.langium` imports `'Terminals'`, importing `'Uddl'` should transitively import everything it imports. This is how C++ includes work. 

If you decide that Langium imports should not work like C++ includes, then this is a bug for XText2Langium - it should recurse into all grammar dependencies and import those dependencies.

spoenemann:
Fixes #924.

### DIFF
--- a/packages/langium/src/grammar/references/grammar-scope.ts
+++ b/packages/langium/src/grammar/references/grammar-scope.ts
@@ -9,20 +9,23 @@ import type { LangiumServices } from '../../services';
 import type { AstNode, AstNodeDescription, ReferenceInfo } from '../../syntax-tree';
 import type { Stream } from '../../utils/stream';
 import type { AstNodeLocator } from '../../workspace/ast-node-locator';
-import type { DocumentSegment, LangiumDocument, PrecomputedScopes } from '../../workspace/documents';
+import type { DocumentSegment, LangiumDocument, LangiumDocuments, PrecomputedScopes } from '../../workspace/documents';
+import type { Grammar } from '../generated/ast';
 import { DefaultScopeComputation } from '../../references/scope-computation';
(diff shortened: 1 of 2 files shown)


msujew (approved):
Looks good to me. I believe there's some room for optimization (we currently recompute all imports on every scope computation), but that's something for another PR.

dhuebner (approved):
Works really good! 
We could also change the requirements grammar to reflect the new feature by removing 
https://github.com/langium/langium/blob/main/examples/requirements/src/language-server/tests.langium#L2

dhuebner:
I was trying to test the requirements example with a proposed change, but getting an error `Error: Undefined rule type: RuleCall at 4:13` when running `npm run langium:generate` . 
Maybe I misconfigured my setup... Could somebody check that it works using cli?

dhuebner:
Works now, thanks @msujew.

PR #1116, opened 2023-07-11
Infer optional properties for guarded groups

msujew:
Closes https://github.com/langium/langium/issues/1047

While collecting the AST types, this chanages how optionality is computed for guarded groups. If a group is guarded, all properties within are potentially optional.

### DIFF
--- a/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
@@ -329,7 +329,9 @@ function newTypePart(element?: ParserRule | Action | string): TypePart {
  * @param element The given AST element, from which it's necessary to extract the type.
  */
 function collectElement(graph: TypeGraph, current: TypePart, element: AbstractElement): TypePart {
-    const optional = isOptionalCardinality(element.cardinality);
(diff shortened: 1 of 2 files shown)


packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
```diff
@@ -329,7 +329,9 @@ function newTypePart(element?: ParserRule | Action | string): TypePart {
  * @param element The given AST element, from which it's necessary to extract the type.
  */
 function collectElement(graph: TypeGraph, current: TypePart, element: AbstractElement): TypePart {
-    const optional = isOptionalCardinality(element.cardinality);
+    const optional = isOptionalCardinality(element.cardinality)
```
Lotes, line 332: Lack of knowledge from my side :) : can `|| (isGroup(element) && element.guardCondition)` be moved into `isOptionalCardinality`? Because you are also using the function in the completion provider.
If you need this condition in the completion provider as well the condition should be moved into the function.

Lotes: Nevermind, saw the parameters xD it is not using element ha ha ha

spoenemann (approved):
Looks good! 👍

PR #1120, opened 2023-07-13
generator-util: fixed minor bug in 'node-processor.ts', added corresponding tests

sailingKieler:


msujew (commented):
Is this really a bug? I'd say this is pretty much intended behavior. I've changed the test to the following:

```ts
        const node = n`
                {
                    
                }
        `;
```

and 

```ts
        const node = n`
                {
                    ${''}
                }
        `;
```

In both times I receive a result that includes the newline in there. However, adding `undefined` seems to remove the line instead. I find it to be quite counter-intuitive that adding an `undefined` removes the line it is placed on. My expectation would be that it is handled the same as an empty string.

spoenemann:
The most intuitive might be to regard everything _falsy_ as empty, at least it would be in line with JS semantics.

sailingKieler:
> Is this really a bug? I'd say this is pretty much intended behavior.

The bug (without the above change) is the following: 
```
const node = n`
    not-indented
        {
            ${''}
        }
    not-indented
`;
```
results in
```
not-indented
    {
        
}
not-indented
```

> In both times I receive a result that includes the newline in there. However, adding undefined seems to remove the line instead. I find it to be quite counter-intuitive that adding an undefined removes the line it is placed on. My expectation would be that it is handled the same as an empty string.

The answer is here: https://www.typefox.io/blog/code-generation-for-langium-based-dsls-2

spoenemann:
That makes sense. Is the behavior mentioned in the JSDoc comments?

msujew (approved):
I see, I wasn't aware of the erroneous behavior. Ok, with the blog post in mind, I'm alright with the changes. Please document the behavior though if it isn't done yet.

sailingKieler:
Added documentation on the _magic line break handling_ in https://github.com/langium/langium/blob/8b535a16f53568b1079759050e6083096e080218/packages/langium/src/generator/template-node.ts#L55-L63

PR #1121, opened 2023-07-15
Improve global scoping mechanism

msujew:
Improves a few performance critical areas in the index manager and scoping mechanism. Introduces a map based scope.

### DIFF
--- a/examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
+++ b/examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
@@ -17,7 +17,7 @@ export class ArithmeticsScopeProvider extends DefaultScopeProvider {
     }
 
     protected override getGlobalScope(referenceType: string): Scope {
-        return new StreamScope(this.indexManager.allElements(referenceType), undefined, { caseInsensitive: true });
+        return this.indexManager.globalScope(referenceType, { caseInsensitive: true });
     }
 
 }
--- a/examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
+++ b/examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
@@ -18,7 +18,12 @@ export class ArithmeticsScopeProvider extends DefaultScopeProvider {
     }
 
     protected override getGlobalScope(referenceType: string): Scope {
-        return new MapScope(this.indexManager.allElements(referenceType), undefined, { caseInsensitive: true });
+        let scope = this.globalScopeCache.get(referenceType);
+        if (!scope) {
+            scope = new MapScope(this.indexManager.allElements(referenceType), undefined, { caseInsensitive: true });
+            this.globalScopeCache.set(referenceType, scope);
+        }
+        return scope;
     }
 
 }
--- a/examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
+++ b/examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
@@ -5,6 +5,7 @@
  ******************************************************************************/
 
 import type { AstNodeDescription, Scope, ScopeOptions } from 'langium';
+import { MapScope } from 'langium';
 import { DefaultScopeProvider, stream, StreamScope } from 'langium';
 
 /**
@@ -17,7 +18,7 @@ export class ArithmeticsScopeProvider extends DefaultScopeProvider {
     }
 
     protected override getGlobalScope(referenceType: string): Scope {
-        return this.indexManager.globalScope(referenceType, { caseInsensitive: true });
+        return new MapScope(this.indexManager.allElements(referenceType), undefined, { caseInsensitive: true });
     }
 
 }
--- a/examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
+++ b/examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
@@ -18,12 +18,7 @@ export class ArithmeticsScopeProvider extends DefaultScopeProvider {
     }
 
     protected override getGlobalScope(referenceType: string): Scope {
-        let scope = this.globalScopeCache.get(referenceType);
-        if (!scope) {
-            scope = new MapScope(this.indexManager.allElements(referenceType), undefined, { caseInsensitive: true });
-            this.globalScopeCache.set(referenceType, scope);
-        }
-        return scope;
+        return this.globalScopeCache.get(referenceType, () => new MapScope(this.indexManager.allElements(referenceType), undefined, { caseInsensitive: true }));
     }
 
 }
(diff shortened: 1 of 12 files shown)


packages/langium/src/references/scope.ts
```diff
@@ -0,0 +1,113 @@
...
+                ? element.name.toLowerCase()
+                : element.name;
+            this.elements.set(name, element);
+        }
+        this.outerScope = outerScope;
+    }
+
+    getElement(name: string): AstNodeDescription | undefined {
+        name = this.caseInsensitive ? name.toLowerCase() : name;
+        return this.elements.get(name);
```
spoenemann, line 94: If the element is not found locally, fall back to the outer scope.

packages/langium/src/workspace/index-manager.ts
```diff
@@ -60,12 +60,12 @@ export interface IndexManager {
     isAffected(document: LangiumDocument, changedUris: Set<string>): boolean;
 
     /**
-     * Compute a global scope, optionally filtered using a type identifier.
+     * Compute a list of all exported elements, optionally filtered using a type identifier.
      *
      * @param nodeType The type to filter with, or `undefined` to return descriptions of all types.
-     * @returns a `Stream` of existing `AstNodeDescription`s filtered by their type
+     * @returns a `Stream` containing all globally visible nodes (of a given type).
```
spoenemann, line 66: Add `@param` documentation for `uris`.

packages/langium/src/workspace/index-manager.ts
```diff
@@ -106,38 +108,53 @@ export class DefaultIndexManager implements IndexManager {
...
+        }
+        let map = this.simpleTypeIndex.get(uri);
+        if (!map) {
+            map = new Map();
+            this.simpleTypeIndex.set(uri, map);
+        }
+        let descriptions = map.get(nodeType);
+        if (!descriptions) {
+            const allFileDescriptions = this.simpleIndex.get(uri) ?? [];
+            descriptions = allFileDescriptions.filter(e => this.astReflection.isSubtype(e.type, nodeType));
```
spoenemann, line 134: The result type of `allElements` is a Stream and in most cases the stream is traversed only once, so I think it would make sense to avoid creating intermediate arrays here.
```suggestion
            descriptions = stream(allFileDescriptions).filter(e => this.astReflection.isSubtype(e.type, nodeType));
```

packages/langium/src/workspace/index-manager.ts
```diff
@@ -106,38 +108,53 @@ export class DefaultIndexManager implements IndexManager {
...
+    }
 
-        const cached = this.globalScopeCache.get(nodeType);
-        if (cached) {
-            return stream(cached);
-        } else {
-            const elements = this.globalScopeCache.get('')!.filter(e => this.astReflection.isSubtype(e.type, nodeType));
-            this.globalScopeCache.set(nodeType, elements);
-            return stream(elements);
+    protected getFileDescriptions(uri: string, nodeType?: string): AstNodeDescription[] {
```
spoenemann, line 122:
```suggestion
    protected getFileDescriptions(uri: string, nodeType?: string): Stream<AstNodeDescription> {
```

packages/langium/src/workspace/index-manager.ts
```diff
@@ -106,38 +108,53 @@ export class DefaultIndexManager implements IndexManager {
...
-            this.globalScopeCache.set('', Array.from(this.simpleIndex.values()).flat());
+    allElements(nodeType?: string, uris?: Set<string>): Stream<AstNodeDescription> {
+        const allUris = this.documents.all.map(doc => doc.uri.toString());
+        const arrays: AstNodeDescription[][] = [];
+        for (const uri of allUris) {
+            if (!uris || uris.has(uri)) {
+                arrays.push(this.getFileDescriptions(uri, nodeType));
+            }
         }
+        return stream(arrays).flat();
```
spoenemann, line 119: I would use the key set of simpleIndex instead of the LangiumDocuments service. And avoid intermediate arrays.
```typescript
const allUris = this.simpleIndex.keys();
return stream(allUris)
    .filter(uri => !uris || uris.has(uri))
    .map(uri => this.getFileDescriptions(uri, nodeType))
    .flat();
```

packages/langium/src/references/scope-provider.ts
```diff
@@ -165,7 +98,12 @@ export class DefaultScopeProvider implements ScopeProvider {
      * Create a global scope filtered for the given reference type.
      */
     protected getGlobalScope(referenceType: string, _context: ReferenceInfo): Scope {
-        return new StreamScope(this.indexManager.allElements(referenceType));
+        let scope = this.globalScopeCache.get(referenceType);
+        if (!scope) {
+            scope = new MapScope(this.indexManager.allElements(referenceType));
+            this.globalScopeCache.set(referenceType, scope);
+        }
+        return scope;
```
spoenemann, line 106: We should extract this cache access, either to a utility class
```typescript
return this.globalScopeCache.get(referenceType, () => new MapScope(this.indexManager.allElements(referenceType)));
```
or to another protected method
```typescript
return this.fromGlobalScopeCache(referenceType, () => new MapScope(this.indexManager.allElements(referenceType)));
```

examples/arithmetics/src/language-server/arithmetics-scope-provider.ts
```diff
@@ -17,7 +18,7 @@ export class ArithmeticsScopeProvider extends DefaultScopeProvider {
     }
 
     protected override getGlobalScope(referenceType: string): Scope {
-        return new StreamScope(this.indexManager.allElements(referenceType), undefined, { caseInsensitive: true });
+        return new MapScope(this.indexManager.allElements(referenceType), undefined, { caseInsensitive: true });
```
spoenemann, line 21: A MapScope does not bring any performance improvement when used without caching because our standard usage of scopes traverses the scope only once.

packages/langium/src/workspace/index-manager.ts
msujew: > in most cases the stream is traversed only once

That's not correct. The decision to make this an array was made deliberately, as the `descriptions` array is usually only computed once per document change. I.e. if a document changes and requests the global scope during linking, it can reuse the cached `descriptions` arrays of the unaffected documents. Making it into a stream would require to recompute the filter step, which is quite expensive.

packages/langium/src/references/scope-provider.ts
spoenemann: This will become easier once we switch to the CachingService (#1123).

msujew:
I've added a benchmark for the domainmodel language in the PR that can be extended as necessary. For current `main` (@ dd09c7c) I've measured:

```
Benchmark duration for '1000 small files': 167.192ms
Benchmark duration for 'single large file': 64.587ms
Benchmark duration for '50 medium sized files': 304.724ms
```

The latest commit in this PR shows:

```
Benchmark duration for '1000 small files': 131.314ms
Benchmark duration for 'single large file': 65.473ms
Benchmark duration for '50 medium sized files': 150.806ms
```

As expected, in the case where we only require the local scope (`single large file`), nothing has changed. We see meaningful improvements on a lot of small files and especially large gains on the most common usecase, a large amount of medium sized files. I'm sure that the benchmark method can be improved, but this already shows the efficacy of this PR.

spoenemann (approved):

PR #1122, opened 2023-07-16
Use exact project name for code generation

jindong-zhannng:
The default case type conversion rule don't always make sense for abbreviated project names. For example:

```json
{
  "projectName": "FormML",
  "languages": [
    ...
  ],
}
```

Then generated codes will be like `FormMlAstType`. But the expectation may be `FormMLAstType` to keep abbreviations in uppercase.

So I add an option `exact` to indicate that don't do any change on project name in generation.

### DIFF
--- a/examples/requirements/langium-config.json
+++ b/examples/requirements/langium-config.json
@@ -1,5 +1,5 @@
 {
-    "projectName": "Requirements and Tests",
+    "projectName": "RequirementsAndTests",
     "languages": [{
         "id": "requirements-lang",
         "grammar": "src/language-server/requirements.langium",
--- a/packages/langium-cli/langium-config-schema.json
+++ b/packages/langium-cli/langium-config-schema.json
@@ -129,7 +129,8 @@
     "properties": {
         "projectName": {
             "description": "The name of your Langium project",
-            "type": "string"
+            "type": "string",
+            "pattern": "^[a-zA-Z_$][0-9a-zA-Z_$]*$"
         },
         "languages": {
             "description": "Your language configurations",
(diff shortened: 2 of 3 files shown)


spoenemann (commented):
I propose changing this to _always_ use the exact project name, and terminate with an error during initial validation if the project name is not a valid identifier.

Actually we're already making this conversion in the Yeoman generator:

https://github.com/jindong-zhannng/langium/blob/fc1601c1e2321d3814fd5224e25412de4289d172/packages/generator-langium/src/index.ts#L150-L152

jindong-zhannng:
> I propose changing this to always use the exact project name, and terminate with an error during initial validation if the project name is not a valid identifier.

👍. You mean an error from yeoman generator or langium cli?

```js
  const languageName = _.upperFirst(
      _.camelCase(this.answers.rawLanguageName)
  );
  const languageId = _.kebabCase(this.answers.rawLanguageName);
```

I propose to never do case conversions like these, as it makes users lose control of the final result. We can prompt user to request names on different cases (maybe with suggested value) in project generation.

It's obviously a breaking change, maybe can be shipped with v2.0.0?

msujew:
@jindong-zhannng What Miro meant is that the CLI should accept every name accept for those that are not valid identifiers in TypeScript. For example, a Projectname `42` would lead to a non-valid TypeScript identifier. Everything else should be accepted in the CLI. You don't have to change anything in the yeoman generator.

msujew:
Interesting, that didn't happen to me for a while. Sometimes GitHub has troubles with rebases and simply deletes the branch. Please create a new PR 👍

jindong-zhannng:
@msujew That because I did some change 😂.

Now I remove the case conversion behavior and add a regex to validate project name.

spoenemann:
@jindong-zhannng that's perfect! But your change revealed that the `requirements` example language doesn't meet this criterion. Could you fix the config of that language so `npm run langium:generate` works without errors?

spoenemann (approved):
Thank you!

PR #1123, opened 2023-07-17
Implement workspace and document cache

msujew:
Implements a feature similar to the Xtext `ResourceScopeCache`.


packages/langium/src/utils/caching.ts
```diff
@@ -0,0 +1,155 @@
...
+    get(documentUri: URI, key: K): V | undefined;
+    delete(documentUri: URI, key: K): boolean;
+    clear(): void;
+    clear(documentUri: URI): void;
+}
+
+export interface WorkspaceCache<K, V> extends Disposable {
+    has(key: K): boolean;
+    set(key: K, value: V): void;
+    get(key: K): V | undefined;
```
spoenemann, line 29: The most obvious pattern of a cache is
```typescript
if (cache.has(key)) {
    return cache.get(key);
} else {
    const value = computeValue();
    cache.set(key, value);
    return value;
}
```
IMO every cache implementation should support this pattern, e.g. like this:
```suggestion
    get(key: K, valueProvider: () => V): V;
    get(key: K): V | undefined;
```
or move this to another method name like `access`.

packages/langium/src/services.ts
```diff
@@ -158,6 +159,7 @@ export type LangiumGeneratedSharedServices = {
  */
 export type LangiumDefaultSharedServices = {
     ServiceRegistry: ServiceRegistry
+    Caching: CachingService
```
spoenemann, line 162:
```suggestion
    CachingService: CachingService
```

spoenemann (commented):
Thanks @msujew! Could you add a proposal about how to use the new service to the PR? We should have a clear picture of where such Map-based caches could help to improve performance.

AFAICT #1121 could make use of a WorkspaceCache in the scope provider, right?

msujew:
> AFAICT https://github.com/langium/langium/pull/1121 could make use of a WorkspaceCache in the scope provider, right?

Yes, exactly :)

spoenemann:
I think it would make sense to integrate these changes into #1121 then, to see the new utility in action and to clean up the changes made there.

spoenemann (commented):
I propose a different class structure to make the underlying cache implementation more reusable and reduce the necessary API surface:
 - `SimpleCache<K, V>` as a simple wrapper around `Map`
 - `ContextCache<C, K, V>` with a two-stage map
 - `WorkspaceCache<K, V>` (without interface) extends `SimpleCache<K, V>`, creates and disposes the DocumentBuilder listener internally (receives LangiumSharedServices in constructor)
 - `DocumentCache<K, V>` (without interface) extends `ContextCache<string, K, V>`, creates and disposes the DocumentBuilder listener internally (receives LangiumSharedServices in constructor)
 - Dispose the `DisposableCache` superclass
 - `CachingService` is no longer necessary because you can easily create the caches in your service constuctor:
```typescript
class MyService {
    private myCache: DocumentCache<A, B>;
    constructor(services: MyLanguageServices) {
        this.myCache = new DocumentCache(services.shared);
    }
}
```
- If you need a short-living cache, create a factory in your constructor:
```typescript
class MyService {
    private myCacheFactory: () => DocumentCache<A, B>;
    constructor(services: MyLanguageServices) {
        this.myCacheFactory = () => new DocumentCache(services.shared);
    }
}
```

packages/langium/src/utils/caching.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    dispose(): void {
+        this.throwIfDisposed();
+        this.clear();
+        this.isDisposed = true;
+        this.toDispose.dispose();
+    }
+
+    throwIfDisposed(): void {
```
spoenemann, line 27: This should be protected.

packages/langium/src/utils/caching.ts
```diff
@@ -0,0 +1,163 @@
...
+import type { Disposable } from 'vscode-languageserver';
+import type { URI } from 'vscode-uri';
+import type { LangiumSharedServices } from '../services';
+
+export abstract class DisposableCache implements Disposable {
+
+    protected toDispose: Disposable;
+    protected isDisposed = false;
+
+    constructor(disposable: Disposable) {
```
spoenemann, line 16: Proposal: add a method `onDispose` that adds a disposable to an array to separate this from construction.

packages/langium/src/utils/caching.ts
```diff
@@ -0,0 +1,163 @@
...
+        this.cache.clear();
+    }
+}
+
+export class ContextCache<I, C, K, V> extends DisposableCache {
+
+    private cache = new Map<C, Map<K, V>>();
+    private converter: (input: I) => C;
+
+    constructor(disposable: Disposable, converter: (input: I) => C) {
```
spoenemann, line 80: `converter` could be optional for `I` = `C`?

packages/langium/src/utils/caching.ts
```diff
@@ -0,0 +1,163 @@
...
+    clear(contextKey?: I): void {
+        this.throwIfDisposed();
+        if (contextKey) {
+            this.getDocumentCache(contextKey).clear();
+        } else {
+            this.cache.clear();
+        }
+    }
+
+    protected getDocumentCache(contextKey: I): Map<K, V> {
```
spoenemann, line 127:
```suggestion
    protected getCacheForContext(contextKey: I): Map<K, V> {
```

spoenemann (commented):
This looks really good! More detailed comments:

PR #1125, opened 2023-07-17
Update all packages to esm

msujew:
Closes https://github.com/langium/langium/issues/905

Main reasons to move to ESM:

* Able to use the newest and cutting edge dependencies (in our case explicitly `chevrotain@11.0.0`).
* Decreased bundle size (roughly ~10% less)
* Better runtime support outside of node.js

Concrete changes in this PR:

* Update our `package.json` and `tsconfig.json` files to use ESM style module resolution. This includes:
    * `"type": "module"` in all `package.json` so that node.js knows that our projects are using ESM style module imports.
    * Add `"exports"` fields for all of our packages.
    * Update all imports to include the `.js` file extension. This isn't strictly required when bundling code, but is good practice when used in Node.js without a bundler.
* Replace all straight imports from node such as `fs` and `path` with `node:fs` and `node:path` respectively. This makes Langium easier to consume in other runtimes such as [deno](https://deno.land/) or [bun](https://bun.sh/).
* Replace builtin properties from node.js such as `__dirname` with their ESM equivalent.
* Update used node and npm version.
* Update Chevrotain and Chalk to use their newest ESM versions.
* Bundle all example languages so they continue to run in vscode as expected.

As for tests:

* Ensure that the langium-cli and vscode extensions still work as expected
* Ensure that the yeoman generated project works as expected


packages/langium/src/documentation/jsdoc.ts
```diff
@@ -4,11 +4,11 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { CstNode } from '../syntax-tree';
+import type { CstNode } from '../syntax-tree.js';
 import { Position, Range } from 'vscode-languageserver';
-import { URI } from 'vscode-uri';
-import { NEWLINE_REGEXP } from '../generator/template-string';
-import { escapeRegExp } from '../utils/regex-util';
+import vscodeUri from 'vscode-uri';
```
kaisalmen, line 9: Not only applicable here, but why did you change the import? Importing and using just `URI` is fine, isn't it?

kaisalmen: Or is this a problem only at runtime, because `vscode-uri` is cjs package?

kaisalmen (commented):
Looking good. I still need to build/use/test it in a real world application.

packages/langium/src/documentation/jsdoc.ts
msujew: The problem is at runtime, since `vscode-uri` (even though it is esm) is missing the correct package.json structure. See also https://github.com/microsoft/vscode-uri/pull/25

kaisalmen: @msujew ok, I guessed there must be a reason. :-) It seems something was merged in June. Let's see when a new version becomes available.

packages/generator-langium/templates/vscode/esbuild.mjs
```diff
@@ -0,0 +1,54 @@
...
+    platform: 'node',
+    sourcemap: !minify,
+    minify,
+    plugins
+});
+
+if (watch) {
+    await ctx.watch();
+} else {
+    await ctx.rebuild();
```
Lotes, line 52: Minor thing...

Use a try-finally here, insert the if-then-else without dispose and for the finally block insert the dispose...

One could argue to remove the dispose because the process will be dead so or so.

examples/statemachine/esbuild.mjs
```diff
@@ -0,0 +1,26 @@
...
+    bundle: true,
+    target: "es6",
+    format: 'cjs',
+    loader: { '.ts': 'ts' },
+    external: ['vscode'],
+    platform: 'node',
+    sourcemap: true
+});
+
+if (watch) {
```
Lotes, line 21: Minor: Try finally?

packages/generator-langium/templates/vscode/esbuild.mjs
msujew: I don't think that esbuild throws. It will just callback with the error. And even if it were to throw, then the process would end either way, making it dispose all resources. Using try-finally here would only make things more complicated.

packages/langium-vscode/src/language-server/grammar-workspace-manager.ts
```diff
@@ -36,18 +33,18 @@ export class LangiumGrammarWorkspaceManager extends DefaultWorkspaceManager {
     }
 
     override async initializeWorkspace(folders: WorkspaceFolder[], cancelToken = CancellationToken.None): Promise<void> {
-        const buildConf: WorkspaceManagerConf = await this.configurationProvider.getConfiguration(LangiumGrammarLanguageMetaData.languageId, CONFIG_KEY);
+        const buildConf: WorkspaceManagerConf = await this.configurationProvider.getConfiguration('langium', CONFIG_KEY);
         const ignorePatterns = buildConf.ignorePatterns?.split(',')?.map(pattern => pattern.trim())?.filter(pattern => pattern.length > 0);
-        this.matcher = ignorePatterns ? ignore().add(ignorePatterns) : undefined;
+        this.matcher = ignorePatterns ? ignore.default().add(ignorePatterns) : undefined;
```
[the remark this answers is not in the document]
msujew: The `ignore` package has a weird export structure, that isn't correctly picked up for ESM code (since the `exports` fields are missing) see source code:

```js
const factory = options => new Ignore(options)

// Fixes typescript
factory.default = factory

module.exports = factory
```

examples/arithmetics/esbuild.mjs
```diff
@@ -0,0 +1,26 @@
...
+    loader: { '.ts': 'ts' },
+    external: ['vscode'],
+    platform: 'node',
+    sourcemap: true
+});
+
+if (watch) {
+    await ctx.watch();
+} else {
+    await ctx.rebuild();
```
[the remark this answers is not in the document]
msujew: Errors are printed to the console in case they are encountered. If we were to print every successful build, we would have a lot of cluttering in the console, since we start 5 esbuild processes when running `npm run watch`.

examples/domainmodel/test/cross-refs.test.ts
```diff
@@ -4,12 +4,12 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { AstNode, LangiumDocument, ReferenceDescription } from 'langium';
-import type { Domainmodel } from '../src/language-server/generated/ast';
+import type { AstNode, LangiumDocument, ReferenceDescription } from '../../../packages/langium/lib/index.js';
```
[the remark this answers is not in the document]
msujew: For some reason, inside of `test` folders, the TypeScript resolution does not work to `langium`:

image

examples/arithmetics/src/cli/cli-util.ts
```diff
@@ -5,10 +5,10 @@
  ******************************************************************************/
 
 import type { AstNode, LangiumDocument, LangiumServices } from 'langium';
-import fs from 'fs';
-import path from 'path';
+import vscodeUri from 'vscode-uri';
+import * as fs from 'node:fs';
```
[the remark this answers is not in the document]
msujew: I feel like this is a matter of taste. I always perform `*` imports on node built-ins. I feel like the methods exposed by `fs` and `path` are otherwise pretty non-descriptibe - a `join` function can mean literally everything, whereas `path.join` is quite explicit.

packages/generator-langium/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Yeoman generator for Langium - the language engineering tool",
   "homepage": "https://langium.org",
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=16.0.0"
```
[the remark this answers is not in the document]
msujew: Most node libraries aim to support as much as possible without losing out on too much functionality. We don't gain anything from using node 18, so I think it makes sense to stick with 16.

examples/arithmetics/esbuild.mjs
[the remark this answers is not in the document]
kaisalmen: > we would have a lot of cluttering in the console

Yes, I know. It is about consistency with tsc behaviour. But also not worth discussing further. 🙂 Up to you.

examples/arithmetics/src/cli/cli-util.ts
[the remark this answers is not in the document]
kaisalmen: Alright, good pro argument.

packages/generator-langium/templates/vscode/esbuild.mjs
Lotes: I have a very splitted opinion here. It is actually ignorable, because it really makes no sense due to the end of the lifetime of the process.

For the watch branch we say: It will never throw, so cleanup is when process is killed.
For the rebuild branch we dispose, because there is a no endless loop. So it makes sense to cleanup. But if you step out the process is cleanup nevertheless...

You do not need to act here, it is just a weird situation, you can have four different scenarios (no clean up in both branches, cleanup in both branches, cleanup only in one branch x 2), but they all will lead to the same result... meh

msujew: > For the rebuild branch we dispose, because there is a no endless loop. So it makes sense to cleanup. But if you step out the process is cleanup nevertheless...

Note that we cleanup because `ctx.rebuild()` does something in the background which prevents the process from ending. Calling `dispose` ends that background task and allows the process to end. If the `rebuild` were to throw (which I've never experienced, since an erroneous build is a part of the "happy path"), the process would just crash.

montymxb (commented):
Seems good on my end. Checked the changes (a bit quickly 😅 ) and ran several tests against existing projects and setting up fresh ones, nothing stood out as a problem there.

examples/domainmodel/test/cross-refs.test.ts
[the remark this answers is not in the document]
kaisalmen: vscode relies on `tsconfig.json` to configure itself and therefore it needs have both `src` and `test` with `noEmit=true` in this config. Another `tsconfig.src.json` must be introduced to drive the regular source build. 

Because it is easier to define then to explain, I created the branch: https://github.com/langium/langium/tree/kaisalmen/esm-test-langium-import
It demonstrates this. All import of `src` in tests have been adapted to use `langium` or `langium/test`.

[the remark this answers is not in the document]
msujew: Thanks for the explanation! Sure, can you push your commit to this PR branch?

packages/generator-langium/package.json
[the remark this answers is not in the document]
kaisalmen: Ok.

packages/langium-cli/tsconfig.test.json
```diff
@@ -4,9 +4,10 @@
     "noEmit": true,
     "rootDir": "test"
   },
-  "references": [{
-    "path": "tsconfig.json"
-  }],
+  "references": [
+    { "path": "tsconfig.json" },
```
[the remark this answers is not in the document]
kaisalmen: Nevermind, that is corret!

packages/langium/src/grammar/internal-grammar-util.ts
```diff
@@ -4,14 +4,17 @@
...
-import * as ast from '../grammar/generated/ast';
-import { Utils } from 'vscode-uri';
-import { getDocument, streamAllContents } from '../utils/ast-util';
-import { TypeResolutionError } from './type-system/type-collector/types';
-import { escapeRegExp } from '../utils/regex-util';
+import type { URI } from 'vscode-uri';
+import type { LangiumDocuments } from '../workspace/documents.js';
+import type { AstNode } from '../syntax-tree.js';
+import * as ast from '../grammar/generated/ast.js';
+import vscodeUri from 'vscode-uri';
```
Yokozuna59, line 11: This line is throwing errors when trying to bundle:

```ts
import vscodeUri from 'vscode-uri';
```

```
"default" is not exported by "node_modules/.pnpm/vscode-uri@3.0.7/node_modules/vscode-uri/lib/esm/index.js", imported by "node_modules/.pnpm/langium@2.0.0-next.239179f/node_modules/langium/lib/grammar/internal-grammar-util.js".
file: node_modules/.pnpm/langium@2.0.0-next.239179f/node_modules/langium/lib/grammar/internal-grammar-util.js:7:7
5:  ******************************************************************************/
6: import * as ast from '../grammar/generated/ast.js';
7: import vscodeUri from 'vscode-uri';
          ^
8: import { getDocument, streamAllContents } from '../utils/ast-util.js';
9: import { TypeResolutionError } from './type-system/type-collector/types.js';
```

Yokozuna59 (commented):
I'm not really sure if we were supposed to use [`langium@2.0.0-next.239179f`](https://www.npmjs.com/package/langium/v/2.0.0-next.239179f), but I tried updating it in here https://github.com/mermaid-js/mermaid/pull/4450 and I came to two issues:

1. [This](https://github.com/mermaid-js/mermaid/blob/b0edfab9264411283b2b34097e5ace64b6e630f4/packages/mermaid/src/diagrams/pie/pieParser.ts#L7) and [this](https://github.com/mermaid-js/mermaid/blob/b0edfab9264411283b2b34097e5ace64b6e630f4/packages/mermaid/src/diagrams/pie/pieParser.ts#L10) lines started throwing errors:

```
Type 'import("packages/mermaid/node_modules/mermaid-parser/node_modules/langium/lib/parser/langium-parser", { assert: { "resolution-mode": "import" } }).LangiumParser' is not assignable to type 'import("packages/mermaid/node_modules/langium/lib/parser/langium-parser").LangiumParser'.
  Types have separate declarations of a private property 'linker'.ts(2322)
```

```
Type 'ParseResult<AstNode>' is not assignable to type 'ParseResult<Pie>'.
  Type 'AstNode' is missing the following properties from type 'Pie': sections, showData ts(2322)
```

Although it works fine when removing the `LangiumParser` type from the variable.

packages/langium/src/grammar/internal-grammar-util.ts
msujew: Well, that's just great... Without using the default import, it doesn't bundle in esbuild - but with it, it doesn't seem to bundle in vite...

msujew:
> Type 'import("packages/mermaid/node_modules/mermaid-parser/node_modules/langium/lib/parser/langium-parser", { assert: { "resolution-mode": "import" } }).LangiumParser' is not assignable to type 'import("packages/mermaid/node_modules/langium/lib/parser/langium-parser").LangiumParser'.
  Types have separate declarations of a private property 'linker'.ts(2322)

That's a pretty weird one, no idea. Maybe two different versions of Langium getting pulled here?

> Type 'ParseResult<AstNode>' is not assignable to type 'ParseResult<Pie>'.
  Type 'AstNode' is missing the following properties from type 'Pie': sections, showData ts(2322)

That's correct, you need to write `parser.parse<Pie>(input)` to get the correct result back. The upcast is not valid.

packages/langium/src/grammar/internal-grammar-util.ts
Yokozuna59: Rollup to be more specific :)

msujew:
> I'm not sure, but I removed all node_modules folders so I don't think so.

I'm confused why it's downloading that twice anyway. I'm not too familiar with pnpm, but shouldn't it only hoist the second `langium` dependency in case it's versions are different? Can you look into the two `langium` packages and see which versions were actually downloaded?

Yokozuna59:
@msujew Sorry, I updated langium version in `packages/parser` and forgot to update the one in `packages/mermaid` 😬

Yokozuna59:
One last thing, it still logs that `vscode-jsonrpc`, `vscode-languageserver-protocol` are missing:

```
"vscode-jsonrpc/browser" is imported by "vscode-jsonrpc/browser?commonjs-external", but could not be resolved – treating it as an external dependency.
"vscode-languageserver-protocol/browser" is imported by "vscode-languageserver-protocol/browser?commonjs-external", but could not be resolved – treating it as an external dependency.
```

After updating `chevrotain` to `v11.0.0` it stopped logging that `lodash` is also missing.

msujew:
@Yokozuna59 Yeah, that's what I already assumed [here](https://github.com/mermaid-js/mermaid/pull/4450#issuecomment-1637975419). Rollup probably needs some additional info/configuration to resolve those correctly, not sure what exactly though. This works out of the box in esbuild which we are usually using to bundle Langium for node/web.

Yokozuna59:
@msujew Is there any update on this? I'd like to try it out, even if it's another pre-release. So we could resolve issues that might occur while integrating parsers.

msujew:
@Yokozuna59 I've just published `2.0.0-next.a57102f` versions of all Langium packages.

PR #1128, opened 2023-07-18
Use data instead of code for code actions

### LINKED ISSUE #560: Code actions: use data instead of code to identify diagnostic type
The `code` property of diagnostics is shown in the UI. As we use that only programmatically to identify which code actions we should apply to a diagnostic, we should change to another property, for example `data`.

spoenemann:
Resolves #560.


msujew (approved):
Looks good to me 👍

Issue #839, opened 2022-12-15
Add validation of cyclic inheritance for declared types


pluralia:
Currently, no validation of this code, where type `X` extends `Y`, `Y` extends `Z` and `Z` extends `X`.
```
interface X extends Y {}
interface Y extends Z {}
interface Z extends X {}

**Comment by snarkipus** [comment]
I implemented something similar to this when I was attempting to learn how to use Langium.
<details>
  <summary>Helper Functions</summary>

```typescript
export function classHierarchy(c: SJClass): Set<Reference<SJClass>> {
    let visited = new Set<Reference<SJClass>>();
    let current = c.superClass;
    while (current != undefined && !visited.has(current)) {
        visited.add(current);
        current = current.ref?.superClass;
    }
    return visited;
}

export function classHierarchyMethods(c: SJClass): Map<string, SJMethod> {
        let returnMap = new Map<string, SJMethod>();

        [...classHierarchy(c)]
            .reverse()
            .flatMap(r => r.ref?.members as SJMember[])
            .forEach(m => returnMap.set(m.name, m as SJMethod));

        return returnMap;
}

export function classHierarchyMembers(c: SJClass): Array<SJMember> {
    return [...classHierarchy(c)]
                .map(r => r.ref)
                .map(m => m?.members as SJMember[])
                .flat();
}
```
</details>

<details>
  <summary>Validator</summary>

```typescript
    checkClassHierarchy(c: SJClass, accept: ValidationAcceptor): void {
        const hierarchy = classHierarchy(c);
        hierarchy.forEach(ref => {
            if (ref.ref?.name === c.name) {
                accept(
                    'error',
                    "cycle in hierarchy of class '" + c.name + "'",
                    {
                        node: c,
                        property: 'name',
                        code: IssueCodes.ClassHierarchy
                    }
                );
            }
        })
    }
```
</details>

<details>
  <summary>Test Code</summary>

```typescript
describe('Small Java Validator: Valid Hierarchy', () => {
    const text=`
    class A extends C {}
    class C extends B {}
    class B {}
    `;

    let validationResult: ValidationResult<SJProgram>;

    beforeAll(async () => {
        validationResult = await validate(text);
    });

    it('Should detect no cycle hierarchies', () => {
        expectNoIssues(validationResult);
    });
});

describe('Small Java Validator: Hierarchy Cycles', () => {
    const text=`
    class A extends C {}
    class C extends B {}
    class B extends A {}
    `;

[...]

Issue #499, opened 2022-05-16
Cyclic inheritance of types


pluralia:
The issue has 2 tasks:
- [ ] Fix type inference for parser rules to avoid cyclic inheritance
- [ ] Validate cyclic inheritance in declared types

---

Some grammars produce cyclic inherited types. For example, these 2 grammars
```
A infers TypeA:
    B ({infer TypeC.left=current} name=ID)*;

B infers TypeB:
    '(' A ')' | {infer TypeD} value=NUMBER;
```

```
A infers TypeA:
    B C*;

B infers TypeB:
    '(' A ')' | D;
```
generate these types:
```
export type TypeA = TypeB | TypeC;
export type TypeB = TypeA | TypeD;
```
`TypeA` here inherits `TypeB` and `TypeB` inherits `TypeA`. We should generate types another way for such grammars, for example, as described [here](https://github.com/langium/langium/pull/490#issuecomment-1124713910).

also check the declared types that define the cyclic way as:

We should also validate declared types that define cyclic inheritance:
```
type TypeA = TypeB | TypeC;
type TypeB = TypeA | TypeD;
```

jtenner:
I think I'm experiencing something like this problem myself. Is this issue still open?

msujew:
@jtenner I've assigned you :)

PR #1130, opened 2023-07-19
Added validations for cross-references and cyclic types

### LINKED ISSUE #468: Cross references for non-string types
_Originally brought up in https://github.com/langium/langium/discussions/467_

When using non-string terminals for references, they are never resolved correctly. We should fix this behavior or disallow cross references on non-string terminals/data type rules.

spoenemann:
Resolves #468. On the way to solving that, I noticed several problems with cyclic type definitions, which I hopefully resolved here, too.

### DIFF
--- a/packages/langium/src/grammar/type-system/type-collector/types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/types.ts
@@ -265,45 +265,58 @@ export class TypeResolutionError extends Error {
 }
 
 export function isTypeAssignable(from: PropertyType, to: PropertyType): boolean {
+    return isTypeAssignableInternal(from, to, new Map());
+}
+
+function isTypeAssignableInternal(from: PropertyType, to: PropertyType, visited: Map<string, boolean>): boolean {
+    const key = `${propertyTypeToKeyString(from)}»${propertyTypeToKeyString(to)}`;
+    let result = visited.get(key);
+    if (result !== undefined) {
+        return result;
+    }
+    visited.set(key, false);
+    result = false;
     if (isPropertyUnion(from)) {
-        return from.types.every(fromType => isTypeAssignable(fromType, to));
+        result = from.types.every(fromType => isTypeAssignableInternal(fromType, to, visited));
     } else if (isPropertyUnion(to)) {
-        return to.types.some(toType => isTypeAssignable(from, toType));
+        result = to.types.some(toType => isTypeAssignableInternal(from, toType, visited));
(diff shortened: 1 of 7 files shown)


packages/langium/src/grammar/type-system/type-collector/types.ts
```diff
@@ -316,6 +329,27 @@ function isInterfaceAssignable(from: InterfaceType, to: InterfaceType, visited:
     return false;
 }
 
+function propertyTypeToKeyString(type: PropertyType): string {
+    if (isReferenceType(type)) {
+        return `@(${propertyTypeToKeyString(type.referenceType)})}`;
+    } else if (isArrayType(type)) {
+        return `(${propertyTypeToKeyString(type.elementType)})[]`;
+    } else if (isPropertyUnion(type)) {
+        const union = type.types.map(e => propertyTypeToKeyString(e)).join(' | ');
```
msujew, line 338: These should be sorted so that type `X | Y` and `Y | X` get the same key (as they are the same type).

packages/langium/src/grammar/validation/types-validator.ts
```diff
@@ -67,6 +77,31 @@ export class LangiumGrammarTypesValidator {
...
+    } else if (ast.isReferenceType(type)) {
+        return isCyclicType(type.referenceType, visited);
+    } else if (ast.isArrayType(type)) {
+        return isCyclicType(type.elementType, visited);
+    } else if (ast.isUnionType(type)) {
+        for (const t of type.types) {
+            if (isCyclicType(t, visited)) {
+                return true;
+            }
+        }
```
msujew, line 98: Suggestion: Since we return with `false` anyway after the loop, we can immediately return here.

```suggestion
        return type.types.some(t => isCyclicType(t, visited));
```

msujew (commented):
You're missing a case for cyclic types: cyclic interface inheritance isn't covered by this PR yet:

```ts
interface A extends B {
}

interface B extends A {
}
```

packages/langium/src/grammar/type-system/type-collector/types.ts
spoenemann: Is that relevant in practice? It seems like an optimization for very rare cases, while in most other cases the sorting is an unnecessary step.

spoenemann:
This now covers part of #499 (but not for inferred types) and resolves #839.

msujew (approved):
Looks good to me 👍

PR #1131, opened 2023-07-19
Renamed some properties in the CST

spoenemann:
I noticed that our CST has properties named after the "node model" of Xtext. I propose to rename some to improve clarity:
 - `parent`  → `container`
 - `children` → `content`
 - `feature` → `grammarSource`
 - `element` → `astNode`


msujew (approved):
Looks good to me 👍

PR #1135, opened 2023-07-28
Fix edge case behavior in parser value assignment

msujew:
The author over at https://github.com/mermaid-js/mermaid/pull/4450 has hit some interesting edge case behavior when the lexer output contains `NaN`. In that case, the `cstNode.length` property becomes `NaN` and the assignment isn't performed. While this lexer behavior is generally not recommended, Langium should deal with it gracefully. The current behavior is straight up just confusing.

### DIFF
--- a/packages/langium/src/parser/cst-node-builder.ts
+++ b/packages/langium/src/parser/cst-node-builder.ts
@@ -225,7 +225,7 @@ export class CompositeCstNodeImpl extends AbstractCstNode implements CompositeCs
 
     private get firstNonHiddenNode(): CstNode | undefined {
         for (const child of this.content) {
-            if (!child.hidden) {
+            if (!child.hidden && !isNaN(child.offset)) {
                 return child;
             }
         }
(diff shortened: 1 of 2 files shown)


packages/langium/src/parser/langium-parser.ts
```diff
@@ -213,7 +213,7 @@ export class LangiumParser extends AbstractLangiumParser {
             cstNode = this.nodeBuilder.buildCompositeNode(feature);
         }
         const subruleResult = this.wrapper.wrapSubrule(idx, rule, args) as any;
-        if (!this.isRecording() && cstNode && cstNode.length > 0) {
+        if (!this.isRecording() && cstNode && cstNode.content.length > 0) {
```
dhuebner, line 216: I didn't get the intention behind this change. Instead of checking the containing text length we now check if there is more than one child node, even if they are probably empty? It raises a question what will happen with an empty range when later building a reference with `this.linker.buildReference(...)`

packages/langium/src/parser/cst-node-builder.ts
```diff
@@ -225,7 +225,7 @@ export class CompositeCstNodeImpl extends AbstractCstNode implements CompositeCs
 
     private get firstNonHiddenNode(): CstNode | undefined {
         for (const child of this.content) {
-            if (!child.hidden) {
+            if (!child.hidden && !isNaN(child.offset)) {
```
dhuebner, line 228: Would it be possible to draft a test grammar for this edge case you are mentioned in the PR description?

msujew:
Rethinking this, it probably doesn't make a lot of sense to get this into the framework. You can only get into this position if you've modified the lexer to return invalid tokens. I think the current behavior for this is doing just fine.

PR #1138, opened 2023-07-30
Improve completion results for data type rules

msujew:
See also https://github.com/langium/langium/discussions/1137

Simply look for the start of the current data type element. This is a pretty simple trick, which seems to fix FQN related completion issues completely.

### DIFF
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -10,7 +10,7 @@ import type { LangiumCompletionParser } from '../../parser/langium-parser';
 import type { NameProvider } from '../../references/name-provider';
 import type { ScopeProvider } from '../../references/scope-provider';
 import type { LangiumServices } from '../../services';
-import type { AstNode, AstNodeDescription, Reference, ReferenceInfo } from '../../syntax-tree';
+import type { AstNode, AstNodeDescription, CstNode, Reference, ReferenceInfo } from '../../syntax-tree';
 import type { MaybePromise } from '../../utils/promise-util';
 import type { LangiumDocument } from '../../workspace/documents';
 import type { NextFeature } from './follow-element-computation';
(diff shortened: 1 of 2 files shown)


packages/langium/test/lsp/completion-provider.test.ts
```diff
@@ -202,3 +202,75 @@ describe('Path import completion', () => {
...
+        entry Model:
+            (persons+=Person | greetings+=Greeting)*;
+
+        Person:
+            'person' name=FQN;
+
+        Greeting:
+            'Hello' person=[Person:FQN] '!';
+
+        FQN returns string: ID ('.' ID)*;
```
Lotes, line 221: I am wondering,

1. what kind of string value will be FQN for `abc  . /*comment*/ xyz`
2. whether it would be good to add test cases which handle edge cases like
  * having whitespace and comments in-between
  * also test the case when the cursor is directly(!) in front of or behind of the dot '.'

msujew: 1. `abc  . /*comment*/ xyz` will be turned into `abc.xyz`, as hidden tokens aren't respected while building data type rule values.
2. I've added a test for comments and whitespace within FQN. They should result in no completion items being returned. As for the other one, we already have a test for right in front of the period. Having the cursor after the dot is the same edge case (i.e. the cursor is between two tokens of the FQN).

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -222,13 +225,29 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
-        const { nextTokenStart, nextTokenEnd, previousTokenStart, previousTokenEnd } = this.backtrackToAnyToken(text, offset);
         const partialContext = {
             document,
             textDocument,
             offset,
             position
         };
+        // Data type rules need special handling, as their tokens are irrelevant for completion purposes.
+        // If we encounter a data type rule at the current offset, we jump to the start of the data type rule.
+        const dataTypeRuleOffsets = this.findDataTypeRuleStart(cst, offset);
```
Lotes, line 236: What is happening here? When we have a FQN defined as:

```ts
FQN returns string: ID ('.' ID)*;
```

And then request a completion on `Example.For.Full<|>_Qualified.Name`, we get the entire range `<|Example.For.Full_Qualified.Name|>`... within are features like `ID` or?

msujew: Yes, that correct. The `findDataTypeRuleStart` method searches the CST until it finds the top most node of the data type rule and returns the full range.

Lotes (approved):

PR #1152, opened 2023-08-15
add implicit ID rule call for cross refs to getAllReachableRules

### LINKED ISSUE #1151: getAllReachableRules should return implicit ID rule
getAllReachableRules should return implicit ID rule but currently does not.
this leads to error initializing a parser based on a imported grammar
that only uses rules that call ID implicitely

```
test('ID implicit called should be returned by getAllReachableRules', async () => {
        // [A] is short for [A:ID] thus the ID rule is needed by the parser and getAllReachableRules should return it
        const input = `
            grammar HelloWorld

            entry Model: A|B;
            A: name=STRING;
            B: ref=[A];
            terminal STRING: /"(\\.|[^"\\])*"|'(\\.|[^'\\])*'/;
        `;
        const output = await parse(input);

        // act
        const reachableRules = [...getAllReachableRules(output.parseResult.value, true)].map(r => r.name);

        // assert
        expect(reachableRules).toContain('ID');
    });
```

cdietrich:
add implicit ID rule call for cross refs to getAllReachableRules
Fixes #1151

### DIFF
--- a/packages/langium/src/utils/grammar-util.ts
+++ b/packages/langium/src/utils/grammar-util.ts
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright 2021-2022 TypeFox GmbH
+ * Copyright 2021-2023 TypeFox GmbH
  * This program and the accompanying materials are made available under the
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
@@ -47,33 +47,49 @@ export function getHiddenRules(grammar: ast.Grammar) {
  *      this function returns all rules of the specified grammar.
  */
 export function getAllReachableRules(grammar: ast.Grammar, allTerminals: boolean): Set<ast.AbstractRule> {
-    const ruleNames = new Set<string>();
     const entryRule = getEntryRule(grammar);
     if (!entryRule) {
         return new Set(grammar.rules);
     }
 
     const topMostRules = [entryRule as ast.AbstractRule].concat(getHiddenRules(grammar));
+    const collectedRules = new Set<ast.AbstractRule>();
     for (const rule of topMostRules) {
-        ruleDfs(rule, ruleNames, allTerminals);
+        ruleDfs(rule, collectedRules, allTerminals);
     }
 
     const rules = new Set<ast.AbstractRule>();
     for (const rule of grammar.rules) {
-        if (ruleNames.has(rule.name) || (ast.isTerminalRule(rule) && rule.hidden)) {
+        if (collectedRules.has(rule) || ((ast.isTerminalRule(rule) && rule.hidden))) {
             rules.add(rule);
         }
     }
+    for (const rule of collectedRules) {
+        if (!rules.has(rule)) {
+            rules.add(rule);
+        }
+    }
+
     return rules;
 }
 
-function ruleDfs(rule: ast.AbstractRule, visitedSet: Set<string>, allTerminals: boolean): void {
-    visitedSet.add(rule.name);
+function ruleDfs(rule: ast.AbstractRule, visitedRules: Set<ast.AbstractRule> , allTerminals: boolean): void {
+    visitedRules.add(rule);
     streamAllContents(rule).forEach(node => {
         if (ast.isRuleCall(node) || (allTerminals && ast.isTerminalRuleCall(node))) {
             const refRule = node.rule.ref;
-            if (refRule && !visitedSet.has(refRule.name)) {
-                ruleDfs(refRule, visitedSet, allTerminals);
+            if (refRule && !visitedRules.has(refRule)) {
+                ruleDfs(refRule, visitedRules, allTerminals);
+            }
+        } else if (ast.isCrossReference(node)) {
+            const term = getCrossReferenceTerminal(node);
+            if (term !== undefined) {
+                if (ast.isRuleCall(term) || (allTerminals && ast.isTerminalRuleCall(term))) {
+                    const refRule = term.rule.ref;
+                    if (refRule && !visitedRules.has(refRule)) {
+                        ruleDfs(refRule,  visitedRules, allTerminals);
+                    }
+                }
             }
         }
     });
--- a/packages/langium/test/utils/grammar-util.test.ts
+++ b/packages/langium/test/utils/grammar-util.test.ts
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright 2022 TypeFox GmbH
+ * Copyright 2022-2023 TypeFox GmbH
  * This program and the accompanying materials are made available under the
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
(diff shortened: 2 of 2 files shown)


msujew (commented):
I'm not sure this is correct. Langium never uses magic names for convention reasons (only exception is the `name` property). Instead Langium uses the terminal associated with the name property for cross references. See also [this method](https://github.com/eclipse-langium/langium/blob/13f482bf33169092f17cf92e327b799c21f5b4dc/packages/langium/src/utils/grammar-util.ts#L90).

packages/langium/src/utils/grammar-util.ts
```diff
@@ -54,26 +54,38 @@ export function getAllReachableRules(grammar: ast.Grammar, allTerminals: boolean
...
     for (const rule of topMostRules) {
-        ruleDfs(rule, ruleNames, allTerminals);
+        ruleDfs(rule, ruleNames, rules, allTerminals);
     }
 
-    const rules = new Set<ast.AbstractRule>();
+    
     for (const rule of grammar.rules) {
-        if (ruleNames.has(rule.name) || (ast.isTerminalRule(rule) && rule.hidden)) {
+        if ((ast.isTerminalRule(rule) && rule.hidden)) {
```
cdietrich, line 64: do i miss something now?

packages/langium/src/utils/grammar-util.ts
```diff
@@ -54,26 +54,38 @@ export function getAllReachableRules(grammar: ast.Grammar, allTerminals: boolean
...
             rules.add(rule);
         }
     }
     return rules;
 }
 
-function ruleDfs(rule: ast.AbstractRule, visitedSet: Set<string>, allTerminals: boolean): void {
-    visitedSet.add(rule.name);
+function ruleDfs(rule: ast.AbstractRule, visitedRuleNames: Set<string>, visitedRules: Set<ast.AbstractRule> , allTerminals: boolean): void {
```
cdietrich, line 71: is there a more elegant way? do we need to make breath first search / two loops
to catch overriden rules?

or just collect all rule names in a first run and the actual rules in a second?

packages/langium/src/utils/grammar-util.ts
```diff
@@ -54,26 +54,38 @@ export function getAllReachableRules(grammar: ast.Grammar, allTerminals: boolean
...
     streamAllContents(rule).forEach(node => {
         if (ast.isRuleCall(node) || (allTerminals && ast.isTerminalRuleCall(node))) {
             const refRule = node.rule.ref;
-            if (refRule && !visitedSet.has(refRule.name)) {
-                ruleDfs(refRule, visitedSet, allTerminals);
+            if (refRule && !visitedRuleNames.has(refRule.name)) {
+                ruleDfs(refRule, visitedRuleNames, visitedRules, allTerminals);
+            }
+        } else if (ast.isCrossReference(node)) {
+            const term = getCrossReferenceTerminal(node)
```
cdietrich, line 81: what is the override semantic here: rule used at the name place is overriden/also defined in current grmmar

msujew: Langium prevents adopters from overriding imported rules.

packages/langium/src/utils/grammar-util.ts
msujew: Langium prevents adopters from overriding imported rules.

packages/langium/src/utils/grammar-util.ts
```diff
@@ -54,26 +54,38 @@ export function getAllReachableRules(grammar: ast.Grammar, allTerminals: boolean
...
             rules.add(rule);
         }
     }
     return rules;
 }
 
-function ruleDfs(rule: ast.AbstractRule, visitedSet: Set<string>, allTerminals: boolean): void {
-    visitedSet.add(rule.name);
+function ruleDfs(rule: ast.AbstractRule, visitedRuleNames: Set<string>, visitedRules: Set<ast.AbstractRule> , allTerminals: boolean): void {
+    visitedRuleNames.add(rule.name);
```
msujew, line 72: Given that we don't allow rule overrides, I don't think we need this set?

packages/langium/test/utils/grammar-util.test.ts
```diff
@@ -36,4 +37,47 @@ describe('Grammar Utils', () => {
         expect(reachableRules).toContain('Ws');
     });
 
+    test('ID implicit called should be returned by getAllReachableRules', async () => {
```
msujew, line 40:
```suggestion
    test('getAllReachableRules should return rules referenced in cross references', async () => {
```

packages/langium/test/utils/grammar-util.test.ts
```diff
@@ -36,4 +37,47 @@ describe('Grammar Utils', () => {
...
+        entry B: ref=[A];
+        `);
+        await services.shared.workspace.DocumentBuilder.build([grammar2, grammar1]);
+        // act
+        const reachableRules = [...getAllReachableRules(grammar2.parseResult.value, true)].map(r => r.name);
+        // assert
+        expect(reachableRules).toContain('ID');
+    });
+
+    test('ID not implicit called should not be returned by getAllReachableRules', async () => {
```
msujew, line 60:
```suggestion
    test('getAllReachableRules should not return unused rules', async () => {
```

packages/langium/src/utils/grammar-util.ts
cdietrich: yes

packages/langium/src/utils/grammar-util.ts
cdietrich: ok, need to try that out

packages/langium/test/utils/grammar-util.test.ts
```diff
@@ -8,32 +8,74 @@ import type { Grammar } from '../../src';
...
+    //     // act
+    //     const reachableRules = [...getAllReachableRules(output.parseResult.value, true)].map(r => r.name);
+
+    //     // assert
+    //     expect(reachableRules).toContain('Ws');
+    // });
+
+    test('ID implicit called should be returned by getAllReachableRules', async () => {
+        // [A] is short for [A:ID] thus the ID rule is needed by the parser and getAllReachableRules should return ID
+        const grammar1 = await parse(`
```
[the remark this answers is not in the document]
msujew: Yeah, that works fine 👍

cdietrich:
is the build fail related to my changes?

msujew:
You just need to run `npm test` to run the test that the Actions suite is running.

> where is the rules from imported rules sorted in? at the end?

Yes, local rules have priority over imported rules. They should be at the end.

packages/langium/test/utils/grammar-util.test.ts
```diff
@@ -36,4 +37,53 @@ describe('Grammar Utils', () => {
...
+        grammar G2
+        import './${Utils.basename(grammar1.uri)}'
+        entry B: ref=[A] s=STRING c=Called;
+        terminal STRING: /"(\\.|[^"\\])*"|'(\\.|[^'\\])*'/;
+        `);
+        await services.shared.workspace.DocumentBuilder.build([grammar2, grammar1]);
+        // act
+        const reachableRules = [...getAllReachableRules(grammar2.parseResult.value, true)].map(r => r.name);
+        // assert
+        expect(reachableRules).toEqual(['B', 'STRING', 'ID', 'Called', 'INT' ])
```
cdietrich, line 62: is this sufficient as testcase or do we need deeply nested calls too?

spoenemann:
@cdietrich this is very old and still marked as draft. Shall we close it?

Issue #1146, opened 2023-08-08
`parseHelper` and other test helpers that use it can cause non-deterministic test failures


lars-reimann:
If no URI is provided to the `parseHelper` function, it [generates one](https://github.com/langium/langium/blob/13f482bf33169092f17cf92e327b799c21f5b4dc/packages/langium/src/test/langium-test.ts#L29) using a random number and the first registered file extension of the DSL. With many tests, however, collisions become increasingly noticeable, causing some tests to fail non-deterministically with the following error:

```txt
Error: A document with the URI 'file:///6821462.sdspipe' is already present.
    at DefaultLangiumDocuments.addDocument (...\node_modules\langium\src\workspace\documents.ts:328:19)
    at ...\node_modules\langium\src\test\langium-test.ts:33:52
    at parseDocument (...\node_modules\langium\src\test\langium-test.ts:356:52)
    at ...\node_modules\langium\src\test\langium-test.ts:336:32
    at ...\tests\formatting\testFormatter.test.ts:15:40
    at file://.../node_modules/@vitest/runner/dist/index.js:357:140
    at file://.../node_modules/@vitest/runner/dist/index.js:242:72
    at file:///.../node_modules/@vitest/runner/dist/index.js:41:26
    at runTest (file:///.../node_modules/@vitest/runner/dist/index.js:599:17)
    at runSuite (file:///.../node_modules/@vitest/runner/dist/index.js:708:15)
```

There are $10^7$ possibilities for the random number, so at 100 tests the probability of any collision is 0.05%, at 200 tests 0.2%, at 500 tests 1.24%, and at 1000 tests already 4.87%.

While this can be mitigated by specifying a URI when using `parseHelper` directly, other functions like `validationHelper` or `expectFormatting` call `parseHelper` internally but do not allow specifying the URI manually.

Langium version: 1.2.1
Package name: `langium`

## Steps To Reproduce

Since the bug is non-deterministic, I cannot provide exact steps for reproduction. But I reckon the description above suffices.

## The current behavior

More tests increase the probability of collisions to levels that are no longer tolerable.

## The expected behavior

Even with many tests, the generated URIs should not collide. Maybe UUIDs could be used instead of the random numbers.

lars-reimann:
> Hm, it make might sense to just use a continuous id for the tests (or uuid). In any case, you can also just call `clearDocuments` on the services in between tests to clear the documents list.

That's a nice workaround as well, thanks.

PR #1153, opened 2023-08-16
Prevent file names collisions during tests

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1146

Uses a running ID to prevent name collisions during tests. Also adds the parse options to all test methods.

### DIFF
--- a/packages/langium/src/test/langium-test.ts
+++ b/packages/langium/src/test/langium-test.ts
@@ -21,6 +21,9 @@ import type { AsyncDisposable } from '../utils/disposable.js';
 import { Disposable } from '../utils/disposable.js';
 
 export interface ParseHelperOptions extends BuildOptions {
+    /**
+     * Specifies the URI of the generated document. Will use a counter variable if not specified.
+     */
     documentUri?: string;
 }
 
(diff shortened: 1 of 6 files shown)


Lotes:
Running numbers is a working solution. A different solution could be a cleanup of the files in memory. I normally would add a dispose method to the result object. If there are no performance issues during the testing process, I would not care sooo much.

spoenemann (approved):
👍

msujew:
@spoenemann Requesting a small re-review. I've added an `AsyncDisposable` that allows to dispose the data that is being created during a test. Also adds a `dispose` flag to all test methods, which is currently falsy by default. We could set the default to be `true` instead, which would make cleaning up after a test quite a bit easier. WDYT?

packages/langium/src/test/langium-test.ts
```diff
@@ -51,9 +55,11 @@ export function expectFunction(functions: ExpectFunction): void {
 
 export interface ExpectedBase {
     text: string
+    parseOptions?: ParseHelperOptions
     indexMarker?: string
     rangeStartMarker?: string
     rangeEndMarker?: string
+    dispose?: boolean;
```
spoenemann, line 62: The meaning of the property is not fully clear. `disposeAfterCheck`?

spoenemann (commented):
What would be the consequences of making the property default-true? If that's potentially breaking, we should merge this as is and change the default for v3.

msujew:
> What would be the consequences of making the property default-true? If that's potentially breaking, we should merge this as is and change the default for v3.

It is breaking in case adopters are reusing the documents from their test methods. I would change the behavior in 3.0 and created an issue tracking this: https://github.com/eclipse-langium/langium/issues/1183.

packages/langium/src/test/langium-test.ts
```diff
@@ -51,9 +55,16 @@ export function expectFunction(functions: ExpectFunction): void {
 
 export interface ExpectedBase {
     text: string
+    parseOptions?: ParseHelperOptions
     indexMarker?: string
     rangeStartMarker?: string
     rangeEndMarker?: string
```
spoenemann, line 61: This is a good opportunity to add comments to the other properties, too.

spoenemann (approved):

PR #1155, opened 2023-08-17
Add EOF rule if needed

Lotes:
I added some logic to have EOF (end of file) as a token. The logic is "if there is a terminal rule named 'EOF', checks if the content of the rule is `/\z/` and converts it into a Chevrotain-EOF."

WDYT? Is it good like this? Something too much or too less?

CC @msujew


Lotes:
Looking at this PR, I think I should add a test to see whether EOF is working O.o...

Yokozuna59:
@Lotes Thanks! I that was my question in #1048.

Replying on `if needed part`:

> Sounds like we could add `EOF` as a built-in keyword to the grammar language?
>
> https://github.com/eclipse-langium/langium/discussions/1048#discussioncomment-5912344

msujew (commented):
I don't really like having to specify a specific definition for a terminal. I would just add a `EOF` keyword that can be used inside of parser rules. Note that unlike ANTLR, Chevrotain does not support EOF as part of a token definition. It's a dedicated token that cannot be mixed with other token types.

Also, please add parser tests to ensure that the EOF token actually parses the EOF token.

Issue #1157, opened 2023-08-17
Ship `langium` with `mode: 'production'`?


Yokozuna59:
Hi there! Is it possible to ship Langium with the `mode` option set to `production`? The current unminified langium grammar is about ~100kb, which might look small compared to other dependencies, but we want to have the parser in https://github.com/mermaid-js/mermaid/pull/4727 as small as possible. So, is it possible to do so? Or do you need the unminified grammar for debugging?

PR #1158, opened 2023-08-17
Publish Langium's grammar minified

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1157

Runs `langium generate --mode=production`  on the Langium language before publishing the package. Ensures that adopters can enjoy a smaller bundle size when using Langium.


spoenemann (commented):
What are the benefits of keeping the non-minified version in the repo? The only thing that comes to my mind is better diffs after changes to the grammar. Or are there cases where you need to look at the generated JSON?

Issue #1150, opened 2023-08-14
export `langium-cli` intenals?


Yokozuna59:
We currently have `dev` script (similar to `watch` and `serve`), that watches `.ts` and `.js` files and rebuild on change. We want to also make that script that also watches `.langium` files and regenerate on change without the need of running to watch mode, but when trying to import `langium-cli` it throws error since it doesn't export anything except `bin`, that's why I'm requesting to export internals so we could use them in that script.

I'm aware that there's a workaround to do that by running the generate command on the shell, but I feel doing so in typescript is better :)

msujew:
Will be addressed in https://github.com/eclipse-langium/langium/pull/1125 with 24a9eba.

Yokozuna59:
Sorry to disturb you, but how should we actually use the `generate` function? I tried using all parameters, but it always exited with one exit code with `--help` prompt not showing what's the error:

```ts
// inside .esbuild/build.ts
await generate({ file: '../packages/parser/langium-config.json' })
```

```
Usage: build [options] [command]

Options:
  -V, --version                   output the version number
  -h, --help                      display help for command

Commands:
  generate [options]              Generate code for a Langium grammar
  extract-types [options] <file>
  help [command]                  display help for command
 ELIFECYCLE  Command failed with exit code 1.
```

I tried generate by execute it in shell but it appears it can't read files from previous dir:

```sh
$ pnpm langium generate --file=../packages/parser/langium-config.json
Reading config from ../packages/parser/langium-config.json
Failed to read config file. [Error: ENOENT: no such file or directory, open '../packages/parser/langium-config.json'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '../packages/parser/langium-config.json'
}
```


---

A side from that, I'm not sure why, when I try to import `langium-cli` in the root dir to use it in [`.esbuild/build.ts`](https://github.com/Yokozuna59/mermaid/blob/e9d43591ba0973fcbcf300e7b94b90b7c66c587b/.esbuild/build.ts) it throws errors that it can't be found:

![image](https://github.com/eclipse-langium/langium/assets/87622592/4c256fd2-7d20-4d08-9e02-1d7b34cd50d5)

But it can be easily imported in [`packages/*`](https://github.com/Yokozuna59/mermaid/tree/e9d43591ba0973fcbcf300e7b94b90b7c66c587b/packages):

[...]

PR #1160, opened 2023-08-20
Fix CLI API import

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1150

Moves the functions out of the `commander` file. Importing the file led to immediate execution of the CLI with incorrect arguments.


spoenemann (approved):
👍

PR #1162, opened 2023-08-21
Add EOF token to grammar

Lotes:
Adds the ability to explicitly parse the EOF token from Chevrotain.

### DIFF
--- a/packages/langium/src/parser/parser-builder-base.ts
+++ b/packages/langium/src/parser/parser-builder-base.ts
@@ -9,8 +9,8 @@ import type { AbstractElement, Action, Alternatives, Condition, CrossReference, 
 import type { BaseParser } from './langium-parser.js';
 import type { AstNode } from '../syntax-tree.js';
 import type { Cardinality } from '../grammar/internal-grammar-util.js';
-import { EMPTY_ALT } from 'chevrotain';
-import { isAction, isAlternatives, isAssignment, isConjunction, isCrossReference, isDisjunction, isGroup, isKeyword, isLiteralCondition, isNegation, isParameterReference, isParserRule, isRuleCall, isTerminalRule, isUnorderedGroup } from '../grammar/generated/ast.js';
+import { EMPTY_ALT, EOF } from 'chevrotain';
+import { isAction, isAlternatives, isEndOfFile, isAssignment, isConjunction, isCrossReference, isDisjunction, isGroup, isKeyword, isLiteralCondition, isNegation, isParameterReference, isParserRule, isRuleCall, isTerminalRule, isUnorderedGroup } from '../grammar/generated/ast.js';
 import { assertUnreachable, ErrorWithLocation } from '../utils/errors.js';
 import { stream } from '../utils/stream.js';
 import { getTypeName } from '../grammar/internal-grammar-util.js';
@@ -88,6 +88,8 @@ function buildElement(ctx: RuleContext, element: AbstractElement, ignoreGuard = 
         method = buildUnorderedGroup(ctx, element);
     } else if (isGroup(element)) {
         method = buildGroup(ctx, element);
+    } else if(isEndOfFile(element)) {
+        method = () => ctx.parser.consume(0, EOF, element);
     } else {
         throw new ErrorWithLocation(element.$cstNode, `Unexpected element type: ${element.$type}`);
     }
--- a/packages/langium/src/parser/parser-builder-base.ts
+++ b/packages/langium/src/parser/parser-builder-base.ts
@@ -89,7 +89,8 @@ function buildElement(ctx: RuleContext, element: AbstractElement, ignoreGuard = 
     } else if (isGroup(element)) {
         method = buildGroup(ctx, element);
     } else if(isEndOfFile(element)) {
-        method = () => ctx.parser.consume(0, EOF, element);
+        const idx = ctx.consume++;
+        method = () => ctx.parser.consume(idx, EOF, element);
     } else {
         throw new ErrorWithLocation(element.$cstNode, `Unexpected element type: ${element.$type}`);
     }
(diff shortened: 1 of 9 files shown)


packages/langium/test/parser/token-builder.test.ts
```diff
@@ -5,10 +5,10 @@
  ******************************************************************************/
 
 import type { TokenPattern, TokenType } from '@chevrotain/types';
-import type { Grammar } from 'langium';
+import type { Grammar } from '../../src/index.js';
```
Lotes, line 8: WTF? Will fix this.

Lotes: done

Lotes:
@Yokozuna59 I had to discard the other PR. The current implementation only allows EOF within parser rules - so no terminals or terminal fragments.

I am wondering if it is still helpful. When I was searching for test examples, this EOF feature seemed not to be very powerful.

Yokozuna59:
> The current implementation only allows EOF within parser rules - so no terminals or terminal fragments.
> 
> I am wondering if it is still helpful. When I was searching for test examples, this EOF feature seemed not to be very powerful.

@Lotes I think the main usage in my mind about it is to use it as `EOL` rule as @msujew said:

<https://github.com/mermaid-js/mermaid/blob/6e0f41180fcb384df2188b77c93e8620a38d2e5d/packages/mermaid/src/diagrams/pie/parser/pie.jison#L84-L88> 

So it should be used as fragment rule:

```langium
fragment EOL:
  NEWLINE+ | ';' | EOF
;
```

The main issue we're facing is that each attribute should be in a separate line, so we can't use `\s+` as hidden rule. But the last rule mayn't end with `\n`, that why we override `Lexer` to add one:

https://github.com/mermaid-js/mermaid/blob/205360c1093ad0059989e05d5efecc4f019225b7/packages/parser/src/language/common/commonLexer.ts#L6

packages/langium/src/parser/parser-builder-base.ts
```diff
@@ -88,6 +88,8 @@ function buildElement(ctx: RuleContext, element: AbstractElement, ignoreGuard =
         method = buildUnorderedGroup(ctx, element);
     } else if (isGroup(element)) {
         method = buildGroup(ctx, element);
+    } else if(isEndOfFile(element)) {
+        method = () => ctx.parser.consume(0, EOF, element);
```
msujew, line 92: Suggestion: Don't use index `0` here. Chevrotain will throw in error in case we encounter a rule such as `Test: 'x' EOF | EOF;` use the rule context `consume` property instead.

packages/langium/src/parser/token-builder.ts
```diff
@@ -29,6 +29,9 @@ export class DefaultTokenBuilder implements TokenBuilder {
         const reachableRules = stream(getAllReachableRules(grammar, false));
         const terminalTokens: TokenType[] = this.buildTerminalTokens(reachableRules);
         const tokens: TokenType[] = this.buildKeywordTokens(reachableRules, terminalTokens, options);
+        if(reachableRules.some(r => streamAst(r.definition).some(isEndOfFile))) {
```
msujew, line 32: Minor: spacing after `if`.

packages/langium/src/parser/token-builder.ts
```diff
@@ -29,6 +29,9 @@ export class DefaultTokenBuilder implements TokenBuilder {
...
+            tokens.push(EOF);
+        }
```
msujew, line 34: Suggestion: The `EOF` token type should be pushed to the end of the lexer definition. Otherwise the lexer will always try to match it first. Given that it's a fairly unlikely token to encounter, we should move it to the end of the lexer definition.

packages/langium/test/parser/langium-parser-builder.test.ts
```diff
@@ -675,6 +677,57 @@ describe('ALL(*) parser', () => {
...
 
+describe('EOF', () => {
+    test('Handle EOF', async () => {
+        const grammar = `
+        grammar Test
+        entry Main: greet='Hello!' EOF;
+        `;
+        const services = await createLangiumGrammarServices(EmptyFileSystem);
+        const output = await parseHelper(services.grammar)(grammar, {validation: true});
+        console.log(output.parseResult.parserErrors);
```
msujew, line 688: Leftover from testing?

packages/langium/test/parser/langium-parser-builder.test.ts
```diff
@@ -675,6 +677,57 @@ describe('ALL(*) parser', () => {
     });
 });
 
+describe('EOF', () => {
+    test('Handle EOF', async () => {
```
msujew, line 681: Suggestion: Can you give these tests more descriptive names? While they explain what they do, test names should also give information on what they expect.

packages/langium/src/parser/parser-builder-base.ts
Lotes: TBH that line was done by Copilot XD, I forgot to review it, because the tests were working. I am not sure, what to write there. I would expect something like `consume++; if(consume >= input.length) { ...return EOF token }`

Lotes: DONE

msujew (commented):
Looking pretty good, there's a test case I'm missing (which is the main reason to use `EOF` at all): Can you create a test that builds a line based language and confirm that it parses as expected? I.e. something like this:

```antlr
Lines: lines+=Line*;
Line: text=ID (EOL | EOF);
terminal EOL: /\r?\n/;
```

Lotes:
@msujew New tests were pushed and branch was rebased. `Wingardium Approviosa`? :)

msujew (approved):
Alright looks good. Can you please remove the changes to the `package-lock.json` before merging?

Issue #1154, opened 2023-08-16
Default values for declared properties


msujew:
Related discussion over at https://github.com/eclipse-langium/langium/discussions/352

It might be advantageous to have default values on declared properties. Something like this should do the trick:

```typescript
interface Item {
  value: string;
  version: string = "2.0.0"
}
```

Yokozuna59:
Awesome! But I've a question, since `entry` rules must have an assignment, by adding this feature, does that mean no need to assign any values?

For example:

```langium
grammar Example:

interface Model {
  version: string = '2.0.0';
}

entry Model returns Model:
  'model'
;

hidden terminal WS: /\s+/; 
```

msujew:
@Yokozuna59 No parser rule requires an assignment. You can already do the following:

```
grammar Example

interface Model {
  version?: string;
}

entry Model returns Model:
   {Model} 'model'
;

hidden terminal WS: /\s+/; 
```

By using the `Model` action, it forces the instatiation of an object with the `Model` type.

PR #1165, opened 2023-08-23
Support default values on interface properties

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1154
Closes https://github.com/eclipse-langium/langium/issues/1206

Touches multiple parts of the framework to make this work:
* Support default values on the `TypeAttribute` grammar element
* Support setting default values in the parser (based on the existing mandatory-properties architecture)
* Adds validation so that default values always match the types defined on the property

### DIFF
--- a/packages/langium/src/grammar/langium-grammar.langium
+++ b/packages/langium/src/grammar/langium-grammar.langium
@@ -16,13 +16,25 @@ entry Grammar:
 Interface:
     'interface' name=ID
     ('extends' superTypes+=[AbstractType:ID] (',' superTypes+=[AbstractType:ID])*)?
-    SchemaType;
-
-fragment SchemaType:
-    '{' attributes+=TypeAttribute* '}' ';'?;
+    '{' 
+        attributes+=TypeAttribute* 
+    '}'
+';'?;
 
 TypeAttribute:
-    name=FeatureName (isOptional?='?')? ':' type=TypeDefinition ';'?;
+    name=FeatureName (isOptional?='?')? ':' type=TypeDefinition ('=' defaultValue=ValueLiteral)? ';'?;
+
+ValueLiteral:
+    StringLiteral | NumberLiteral | BooleanLiteral | ArrayLiteral;
+
+StringLiteral:
+    value=STRING;
+NumberLiteral:
+    value=NUMBER;
+BooleanLiteral:
+    true?='true' | 'false';
+ArrayLiteral:
+    '[' (elements+=ValueLiteral (',' elements+=ValueLiteral)*)? ']';
 
 TypeDefinition: UnionType;
 
@@ -116,9 +128,6 @@ NamedArgument:
     ( parameter=[Parameter:ID] calledByName?='=')?
     ( value=Disjunction );
 
-LiteralCondition:
-    true?='true' | 'false';
-
 Disjunction infers Condition:
     Conjunction ({infer Disjunction.left=current} '|' right=Conjunction)*;
 
@@ -129,7 +138,7 @@ Negation infers Condition:
     Atom | {infer Negation} '!' value=Negation;
 
 Atom infers Condition:
-    ParameterReference | ParenthesizedCondition | LiteralCondition;
+    ParameterReference | ParenthesizedCondition | BooleanLiteral;
 
 ParenthesizedCondition infers Condition:
     '(' Disjunction ')';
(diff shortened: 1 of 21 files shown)


packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -204,10 +212,11 @@ CharacterRange infers AbstractElement:
     {infer CharacterRange} left=Keyword ('..' right=Keyword)?;
 
 FeatureName returns string:
-    'current' | 'entry' | 'extends' | 'false' | 'fragment' | 'grammar' | 'hidden' | 'import' | 'interface' | 'returns' | 'terminal' | 'true' | 'type' | 'infer' | 'infers' | 'with' | PrimitiveType | ID;
+    'Infinity' | 'NaN' | 'undefined' | 'current' | 'entry' | 'extends' | 'false' | 'fragment' | 'grammar' | 'hidden' | 'import' | 'interface' | 'returns' | 'terminal' | 'true' | 'type' | 'infer' | 'infers' | 'with' | PrimitiveType | ID;
```
spoenemann, line 215: This doesn't work: the keywords take higher precedence than the `NUMBER` token. And we have a validation forbidding the use of these names anyway because they are reserved names of the JavaScript runtime.

packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -16,13 +16,21 @@ entry Grammar:
...
-fragment SchemaType:
-    '{' attributes+=TypeAttribute* '}' ';'?;
+    '{' 
+        attributes+=TypeAttribute* 
+    '}' 
+';'?;
 
 TypeAttribute:
-    name=FeatureName (isOptional?='?')? ':' type=TypeDefinition ';'?;
+    name=FeatureName (isOptional?='?')? ':' type=TypeDefinition ('=' defaultValue=ValueLiteral)? ';'?;
```
Yokozuna59, line 25: Is it possible to support default for arrays of primitive types? Or do you plan to only support primitive types?

My use case that I want to parse `yaml` style code in `mermaid` and some properties have array default values.

```ts
{
  ...,
  timeline: {
    ...,
    sectionColours: ['#fff']
  }
}
```

Or should I just override the parser and set default values if it's empty array? It would take more manual work and time since we need to figure out each default and update in both schema and setter.

msujew: Sure, that's a good idea :)

I'll add that to the PR when I have time. I'm currently quite busy.

msujew: Added the feature to the PR 👍

packages/langium/src/parser/langium-parser.ts
```diff
@@ -283,15 +283,20 @@ export class LangiumParser extends AbstractLangiumParser {
...
-                obj[mandatoryProperty.name] = false;
+            if (obj[mandatoryProperty.name] === undefined) {
+                obj[mandatoryProperty.name] = this.copyDefaultValue(mandatoryProperty.defaultValue!);
             }
         }
     }
 
+    private copyDefaultValue(propertyType: MandatoryPropertyType): MandatoryPropertyType {
+        if (Array.isArray(propertyType)) {
+            return [...propertyType.map(this.copyDefaultValue)];
```
spoenemann, line 294: The method is used with a function reference here. I suggest moving this to a plain function to avoid bugs due to unbound `this` (otherwise you'd need to write `this.copyDefaultValue.bind(this)`).

packages/langium/src/syntax-tree.ts
```diff
@@ -197,19 +197,24 @@ export abstract class AbstractAstReflection implements AstReflection {
 export interface TypeMetaData {
     /** The name of this meta model type. Corresponds to the `AstNode.$type` value. */
     name: string
-    /** A list of mandatory properties. These are defaults for array and boolean based properties (`[]` and `false` respectively). */
-    mandatory: TypeMandatoryProperty[]
+    /** A list of mandatory properties. They contain default values for their respective property in the AST. */
+    mandatory: MandatoryProperty[]
```
spoenemann, line 201: This is a breaking change. Can we keep the old type and mark it deprecated? Otherwise we should wait for v3.0.0 before merging this.
```suggestion
    mandatory: Array<MandatoryProperty | TypeMandatoryProperty>
```

spoenemann: We decided to move the feature to v3.0.0, so the change is ok.

spoenemann (approved):
👍  This is much cleaner than the previous approach, which was focused on properties with certain types.

Issue #1167, opened 2023-08-23
Export grammar language via dedicated ESM export


msujew:
Currently, we export very selected functions/objects via a `langium/internal` export. This is for internal (i.e. `langium-cli` usage) only. However, we actually export our full grammar code (a large JSON structure + related services, mostly validation, type system, etc.) on `langium` directly. Even though adopters don't need the grammar code during runtime, every bundle includes the code.

PR #1171, opened 2023-08-24
Split grammar and main export

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1167

Bundle size benchmark on the arithmetics example:

|main|PR|main, minified|PR, minified|
|-----|--|---------------|-----------|
|1308kb|1019kb|633kb|457kb|

We can see that we remove roughly ~175kb from a minified language server bundle with this change, reducing small language server bundles by 28%.


spoenemann (commented):
Thanks! A few questions, maybe we can discuss them in today's dev meeting:
 - Why did you move so many functions from `internal-grammar-util` to `grammar-util`?
 - How can we guarantee that we don't accidentally introduce unwanted dependencies to grammar code in the core framework?
    - Many files use `grammar/generated/ast` – maybe we should move that out of the `grammar` folder or reexport them
    - `ast-reflection-interpreter` uses more grammar internal code, but is in the `language` folder
    - `value-converter` still uses a function from `internal-grammar-util`

cdietrich:
@msujew we are currently using `getActionAtElement` too.
can you consider moving it?

packages/langium/src/utils/grammar-util.ts
```diff
@@ -281,59 +269,286 @@ function findNameAssignmentInternal(type: ast.AbstractType, cache: Map<ast.Abstr
...
+    wrap?: boolean
+    lookahead?: string
+}): string {
+    if (options.wrap !== false || options.lookahead) {
+        regex = `(${options.lookahead ?? ''}${regex})`;
+    }
+    if (options.cardinality) {
+        return `${regex}${options.cardinality}`;
+    }
+    return regex;
```
spoenemann, line 553: All these functions are exposed as top-level exports. I'm thinking about wrapping them in a namespace so it's clear what they are about. WDYT?

spoenemann (commented):
I know we've talked about including this change in the 2.x version stream, but now it looks like a too big breaking change to me. One option could be to plan for a 3.0.0 version rather soon so we don't have to worry too much about the changes?

packages/langium/src/utils/grammar-util.ts
spoenemann: Let's do this in a follow-up PR as discussed.

spoenemann (commented):
The changes look good to me. I would only suggest to modify the main package export test so it excludes the grammar code (see #1269).

I also thought about generating the grammar related files into different folders, specifically to move the AST file to `src/language/grammar-ast.ts`. This would make code isolation much easier, but we would need a special case for langium-cli (which we already have with the `langiumInternal` flag). WDYT?

packages/langium-cli/src/generator/grammar-serializer.ts
```diff
@@ -16,7 +16,7 @@ export function serializeGrammar(services: LangiumServices, grammars: Grammar[],
     if (config.langiumInternal) {
         node.append(
             `import type { Grammar } from './ast${config.importExtension}';`, NL,
-            `import { loadGrammarFromJson } from '../../utils/grammar-util${config.importExtension}';`);
+            `import { loadGrammarFromJson } from '../../utils/grammar-loader${config.importExtension}';`);
```
Lotes, line 19: Just saying this because I was in the generator PR before: Why not using the `expandToNode` function?

msujew: Because this code predates the `expandToNode` function by ~1 year.

spoenemann: This is going to change in #1287

packages/langium/src/language/index.ts
sailingKieler, line 1: I think I understood the separation of this from `src/grammar` from the conversations within this PR.
I guess it's because types and functions like `RuleCall` and `isRuleCall` are required in client code to implement custom indexing/scope computation/proposals, right? (without the need to rely on `langium/grammar`)

Could you please leave a comment on this here?
On the first glance it looks quite odd that there is `src/grammar/generated/...` and `src/language/generated/...`.
Maybe `languageS` (with lower case 's') expresses the difference better?

sailingKieler (commented):
Looks very good to me 👍, I have only one remark on the duality of `grammar` vs. `language` about making clear its rationale. IMO the name `language` is too ambiguous here, what about `languages` or `language-api`. Or even `grammar-api` as it describes the API to implement evaluations of grammars.

I assume you have tested that this cut get's us the expected reduction in bundle size 😉

packages/langium/src/language/index.ts
msujew: Yeah, `languages` is a better name. I've renamed it.

sailingKieler (approved):
As I already said, some hints in the index files of `languages` and `grammar` on their purpose would be nice for others to understand their purpose.

Apart from that 🚀

Issue #1176, opened 2023-09-01
Completion provider ignores NodeDescriptions with the same name


dhuebner:
Often a language accept different linking targets with the same name. A good example is method overloading.
Right now there is a `duplicateStore` in [complition-provider](https://github.com/eclipse-langium/langium/blob/7259652a1669e236688b6f1f062f92430dcfbfb6/packages/langium/src/lsp/completion/completion-provider.ts#L403) that prevents suggesting completions for same named NodeDescriptions. 
As we already provide a way to filter out NodeDescriptions from scope using `filterCrossReference(_nodeDescription)` I would like to suggest to remove the `duplicateStore`. In addition we can pass a `Set` to the `filterCrossReference` so the overwrites can use it as state, but I don't think it is required.

dhuebner:
An other idea is to use `documentUri + path` as a set entry. This will filter out duplicate NodeDescriptions but accept descriptions with the same name.

msujew:
Closed in https://github.com/eclipse-langium/langium/pull/1178

PR #1178, opened 2023-09-01
Don't filter out same named NodeDescriptions

dhuebner:
Replaces a name based duplicates filter with URI based one.

See #1176

### DIFF
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -398,11 +398,9 @@ export class DefaultCompletionProvider implements CompletionProvider {
             };
             try {
                 const scope = this.scopeProvider.getScope(refInfo);
-                const duplicateStore = new Set<string>();
                 scope.getAllElements().forEach(e => {
-                    if (!duplicateStore.has(e.name) && this.filterCrossReference(e)) {
+                    if (this.filterCrossReference(e)) {
                         acceptor(context, this.createReferenceCompletionItem(e));
-                        duplicateStore.add(e.name);
                     }
                 });
             } catch (err) {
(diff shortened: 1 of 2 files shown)


packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -400,9 +400,10 @@ export class DefaultCompletionProvider implements CompletionProvider {
                 const scope = this.scopeProvider.getScope(refInfo);
                 const duplicateStore = new Set<string>();
                 scope.getAllElements().forEach(e => {
-                    if (!duplicateStore.has(e.name) && this.filterCrossReference(e)) {
+                    const storeId = e.documentUri + e.path;
```
spoenemann, line 403: `${e.documentUri}#${e.path}`

spoenemann (commented):
Sometimes one element is in the scope with multiple names. Shall we use a combination of URI, path and name?

packages/langium/src/lsp/completion/completion-provider.ts
dhuebner: Fragment path starts with an `/` which is already a good separator as doc URI normally ends with a file extension I thought. That why I left out an additional separator, but yes we can add `#`.

dhuebner:
> Sometimes one element is in the scope with multiple names. Shall we use a combination of URI, path and name?

I think it is really about to filter out the same AstNodeDescriptions (probably performance reasons), means filter out descriptions that points to the same element using URI + Element path. 
I would rather completely remove this `duplicateStore` as we have an additional method for filtering and completion items are filtered again later on.  See linked issue.

msujew:
@dhuebner I agree. The duplicateStore was supposed to reduce the amount of duplicate suggestions. We have a better method of deduplicating elements now, so we should remove the store.

dhuebner:
@spoenemann 
I removed the `duplicateStore`, please re-view again.

spoenemann (approved):
👍

PR #1204, opened 2023-09-27
feat: make keyword filtering in completion provider configurable

cdietrich:


packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -442,6 +441,11 @@ export class DefaultCompletionProvider implements CompletionProvider {
         });
     }
 
+    protected isKeywordWorthyToPropose(context: CompletionContext, keyword: ast.Keyword): boolean {
```
msujew, line 444: Minor: Sounds a bit verbose and it probably makes sense to keep this more in line with the existing `filterCrossReference`.
```suggestion
    protected filterKeyword(context: CompletionContext, keyword: ast.Keyword): boolean {
```

msujew (commented):
While you're at it: I've noticed that the `filterCrossReference` method doesn't accept a `CompletionContext` as a parameter. In it's current state, the method is kind of useless. Do you mind adding that as well?

packages/langium/src/lsp/completion/completion-provider.ts
cdietrich: done

cdietrich:
adapted both methods

msujew (approved):
Alright, thanks 👍

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -442,6 +441,11 @@ export class DefaultCompletionProvider implements CompletionProvider {
         });
     }
 
+    protected filterKeyword(_context: CompletionContext, keyword: ast.Keyword): boolean {
```
spoenemann, line 444: The method name is ambiguous. In the JS context, "filter" is usually understood as include, not "filter out".

cdietrich:
@spoenemann got a mail about the naming, but cannot find the comment

PR #1208, opened 2023-09-28
Railroad diagrams separate files

### LINKED ISSUE #1187: Railroad Diagrams in separate files
As part of an user guide for a language developed with Langium, I want to include the railroad syntax diagrams for illustration. However, `langium-cli` only outputs such diagrams into a single HTML file, which makes their extraction difficult. 

Therefore I propose to extend `langium-railroad` and `langium-cli` to output each syntax diagram into a separate picture file on demand. Probably the most straightforward way would be to take the already generated SVG documents per diagram, insert the diagram styles, and then write them into separate `svg` files.

bjthehun:
Closes #1187

Allows `langium-cli` to output railroad diagrams for a given language configuration in single SVG files, that can be used e.g. for language documentation. To use this feature, the `railroad` part of a configuration now contains the `mode` setting:

1. When `mode === 'svg'`, output each grammar rule named `Rule` into single files with the path `out/Rule.svg`.
2. When `mode` is missing or `mode === 'html'`, output all grammar rules into a single HTML file with the path `out`, as is the current behavior.


packages/langium-cli/src/generate.ts
```diff
@@ -365,10 +365,21 @@ export async function runGenerator(config: LangiumConfig, options: GenerateOptio
...
+                const diagram = createGrammarDiagramHtml(grammar);
+                log('log', options, `Writing railroad syntax diagram to ${chalk.white.bold(diagramPath)}`);
+                await writeWithFail(diagramPath, diagram, options);
+            }
+            // Svg files requested -> make dir and write into it.
+            else {
+                const diagrams = createGrammarDiagramSvg(grammar);
+                log('log', options, `Writing railroad syntax diagrams to ${chalk.white.bold(diagramPath)}`);
+                for (const [name, diagram] of diagrams) {
+                    await writeWithFail(`${diagramPath}/${name}.svg`, diagram, options);
```
msujew, line 380: Suggestion: Please use `path.join` or `path.resolve` to join file system paths. This kind of joining will lead to problems in case the given directory ends in `/`.

packages/langium-railroad/src/grammar-railroad.ts
```diff
@@ -39,16 +39,30 @@ export interface GrammarDiagramOptions {
...
+}
+
+/**
+ * Creates a whole HTML file that contains all railroad diagrams.
+ *
+ * @param grammar - GrammarAST.Grammar
+ * @param options? - GrammarDiagramOptions
+ * Use to add additional styling to all diagrams and additional behavior for the created HTML.
+ *
+ * @returns text - string
```
msujew, line 58: Minor: We usually don't annotate our parameter or return type information, since we already generate `.d.ts` files for all JavaScript files.

packages/langium-railroad/src/grammar-railroad.ts
```diff
@@ -39,16 +39,30 @@ export interface GrammarDiagramOptions {
     javascript?: string;
 }
 
+function styling(options?: GrammarDiagramOptions) {
+    return `<style>
+    ${defaultCss}
+    </style>${options?.css ? `
+    <style>
+    ${options.css}
+    </style>` : ''}`;
```
msujew, line 48: Suggestion: This kind of string generation leads to unexpected indentation on every line after the first. Even though it's just HTML, we try to keep the generated code reasonably formatted. 
```suggestion
    return expandToString`
    <style>
    ${defaultCss}
    </style>${options?.css ? `
    <style>
    ${options.css}
    </style>` : ''}`;
```

packages/langium-cli/src/package.ts
```diff
@@ -62,10 +62,12 @@ export interface LangiumLanguageConfig {
...
     }
-    /** Enable generating a railroad syntax diagram */
+    /** Enable generating railroad syntax diagrams */
     railroad?: {
-        /** Output path to the railroad diagram file */
-        out: string
+        /** Output path for railroad diagrams. */
+        out: string;
+        /** Whether to print diagrams all into a single HTML file or in separate SVG files. */
+        mode?: 'html' | 'svg';
```
msujew, line 70: Suggestion: Please add this change to the schema JSON for this type, so that adopters get accurate IDE information about this property.

msujew (commented):
Hey @bjthehun thank you for your contribution. I have a few comments on the code, please see below.

packages/langium-cli/src/generate.ts
bjthehun: Now using `path.join`.

packages/langium-cli/src/package.ts
bjthehun: Added `mode` to both the schema JSON (`langium-config-schema.json`) and documentation (`README.md`).

packages/langium-railroad/src/grammar-railroad.ts
bjthehun: Now using `expandToString` as suggested.

packages/langium-railroad/src/grammar-railroad.ts
bjthehun: Removed parameter and return type annotations.

msujew (approved):
Thanks, looks good to me 👍

PR #1210, opened 2023-09-28
Adapt links and readme for Eclipse Foundation

msujew:
Most links were outdated, with some like the build pipeline straight up just not working. Also added a paragraph about the ECA that needs to be signed.


spoenemann (approved):
Thanks!

PR #1211, opened 2023-09-28
set diagram server manager fields to protected

Rtulpe:
It does not make sense to have `changedUris` and `outdatedDocuments` fields as private, since if I want to extend the default class and change the behavior of lets say `documentsUpdated()` method, I have to re-implement everything.


spoenemann (approved):
Thanks!

Issue #1174, opened 2023-08-28
include imported used rules in `railroad` result


Yokozuna59:
Hi there, the current railroad result don't include rules are imported from another `.langium` file. I would like them to be included in the result.

![image](https://github.com/eclipse-langium/langium/assets/87622592/4b5edb99-a130-41fe-9b07-21f86c33c49d)

![image](https://github.com/eclipse-langium/langium/assets/87622592/d467b686-6f30-4d3d-bdcb-0e0ce86743c2)

weedySeaDragon:
@Yokozuna59 - It is pulling in the imported 'common/common.langium' file.

The diagram does not include terminals (including hidden ones); terminals are filtered out before the diagram is created.

Do you want terminals to be shown?  Perhaps that could be an option.

msujew:
@weedySeaDragon The issue is that the fragment parser rule `TitleAndAccessibilities` from common.langium is missing. Given that the generated syntax diagram includes imported rules, it only makes sense that the vscode extension should show those as well.

weedySeaDragon:
Ah - now I see. I was totally overlooking that`TitleAndAccessibilities` was missing from the Info grammar diagram. 

I don't use VS code.  I have tweaked the code (langium-railroad) to make it easier for me create .html files and the results I have _do_ include the imported files.  
I haven't compared my code to the original yet to see why. When I do, I'll post info about the differences here, FWIW.

PR #1213, opened 2023-10-03
Show all railroad rules in railroad preview

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1174

Includes referenced parser rules from other referenced grammars in the railroad syntax preview in vscode. Currently, it only shows grammar rules from the opened file without any imports. This change adds imported rules to the bottom of the file.

Also adds the ability to add a path to a css file that gets included into the generated railroad syntax diagram file.


montymxb (approved):
:octocat: Approved! Nice, played around with this locally. Imported rules show up, and custom CSS is applied as expected ✅ .

Issue #1203, opened 2023-09-27
NextFeature.property calculation is broken with a datatype rule


cdietrich:
Adapt the hello world grammar terminals to

```
ID returns string: ID1 | ID2;
terminal ID1: /[_a-zA-Z][\w_-]*/;
terminal ID2: /`[^`]*`/;
```

upon completion on

```
person |
```
the value of `next.property` is `undefined`.
i would expect it to be `name`

msujew:
Interesting finding. I can reproduce the issue and created a proposed fix in [here](https://github.com/eclipse-langium/langium/tree/msujew/fix-next-element-compute). I'll continue working on that tomorrow, since I might have found another potential issues related to the `new` flag of the computed value that I want to fix in there as well.

PR #1215, opened 2023-10-04
Improve next element computation in completion

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1203

Improves the completion algorithm in two ways:
1. Correctly propagates the current `NextFeature#property` variable through alternatives (related to the linked issue)
2. Fixes the way in which we create new AST nodes during cross reference completion. Previously, the `new` property wasn't set correctly, which led to cross reference completion on "stale" elements. This change simplifies (and also improves) the computation by only using the `type` property.

Both changes aren't trivial to (automatically) test and require debugging to accurately pinpoint the actual changes in the completion algorithm.

### DIFF
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -19,7 +19,6 @@ import type { FuzzyMatcher } from '../fuzzy-matcher.js';
 import type { GrammarConfig } from '../../grammar/grammar-config.js';
 import type { Lexer } from '../../parser/lexer.js';
 import type { IToken } from 'chevrotain';
-import type { Stream } from '../../utils/stream.js';
 import { CompletionItemKind, CompletionList, Position } from 'vscode-languageserver';
 import * as ast from '../../grammar/generated/ast.js';
 import { getExplicitRuleType } from '../../grammar/internal-grammar-util.js';
@@ -381,17 +380,23 @@ export class DefaultCompletionProvider implements CompletionProvider {
         } else if (ast.isCrossReference(next.feature) && context.node) {
             return this.completionForCrossReference(context, next as NextFeature<ast.CrossReference>, acceptor);
         }
+        // Don't offer any completion for other elements (i.e. terminals, datatype rules)
+        // We - from a framework level - cannot reasonably assume their contents.
+        // Adopters can just override `completionFor` if they want to do that anyway.
     }
 
-    protected completionForCrossReference(context: CompletionContext, crossRef: NextFeature<ast.CrossReference>, acceptor: CompletionAcceptor): MaybePromise<void> {
-        const assignment = getContainerOfType(crossRef.feature, ast.isAssignment);
+    protected completionForCrossReference(context: CompletionContext, next: NextFeature<ast.CrossReference>, acceptor: CompletionAcceptor): MaybePromise<void> {
+        const assignment = getContainerOfType(next.feature, ast.isAssignment);
         let node = context.node;
         if (assignment && node) {
-            if (crossRef.type) {
+            if (next.type) {
+                // When `type` is set, it indicates that we have just entered a new parser rule.
+                // The cross reference that we're trying to complete is on a new element that doesn't exist yet.
+                // So we create a new synthetic element with the correct type information.
                 node = {
-                    $type: crossRef.type,
+                    $type: next.type,
                     $container: node,
-                    $containerProperty: crossRef.property
+                    $containerProperty: next.property
                 };
                 assignMandatoryAstProperties(this.astReflection, node);
             }
@@ -404,9 +409,11 @@ export class DefaultCompletionProvider implements CompletionProvider {
             };
             try {
                 const scope = this.scopeProvider.getScope(refInfo);
-                for (const description of this.filterCrossReferences(context, scope.getAllElements())) {
-                    acceptor(context, this.createReferenceCompletionItem(description));
-                }
+                scope.getAllElements().forEach(e => {
+                    if (this.filterCrossReference(context, e)) {
+                        acceptor(context, this.createReferenceCompletionItem(e));
+                    }
+                });
             } catch (err) {
                 console.error(err);
             }
@@ -429,8 +436,8 @@ export class DefaultCompletionProvider implements CompletionProvider {
         };
     }
 
-    protected filterCrossReferences(context: CompletionContext, nodeDescriptions: Stream<AstNodeDescription>): Stream<AstNodeDescription> {
-        return nodeDescriptions;
+    protected filterCrossReference(_context: CompletionContext, _nodeDescription: AstNodeDescription): boolean {
+        return true;
     }
 
     protected completionForKeyword(context: CompletionContext, keyword: ast.Keyword, acceptor: CompletionAcceptor): MaybePromise<void> {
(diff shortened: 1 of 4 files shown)


msujew:
@cdietrich Are you interested in testing the change to see whether it works for you?

cdietrich:
@msujew i can have a look but i will need some time as i dont want to mess up my production dev env with npm link

cdietrich:
@msujew have problems to get this running

when i run my extension i am getting

```
Error: Cannot find module 'vscode'
Require stack:
- /Users/dietrich/tmp/viechviechviech/dot-expression-problem/out/language/main.cjs
    at Function._resolveFilename (node:internal/modules/cjs/loader:1129:15)
    at node:internal/modules/cjs/loader:974:27
    at Function._load (node:electron/js2c/asar_bundle:2:13377)
    at Module.require (node:internal/modules/cjs/loader:1195:19)
    at require (node:internal/modules/cjs/helpers:110:18)
    at node_modules/vscode-languageclient/lib/node/main.js (/Users/dietrich/tmp/viechviechviech/dot-expression-problem/out/language/main.cjs:35831:20)
    at __require (/Users/dietrich/tmp/viechviechviech/dot-expression-problem/out/language/main.cjs:26:50)
    at Object.<anonymous> (/Users/dietrich/tmp/viechviechviech/dot-expression-problem/out/language/main.cjs:63806:44)
    at Module._compile (node:internal/modules/cjs/loader:1316:14)
    at Object..js (node:internal/modules/cjs/loader:1371:10) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/dietrich/tmp/viechviechviech/dot-expression-problem/out/language/main.cjs'
  ]
  ```
  any idea?

**Comment by msujew** [comment]
@cdietrich Are you accidentally importing `vscode-languageclient` in your language server code? It imports `vscode` to register the language client handlers.

**Comment by cdietrich** [comment]
yes.
i assume i have to import
import { CompletionItemKind, InsertTextFormat } from "vscode-languageserver"

**Comment by cdietrich** [comment]
next.property is now correctly computed. thx.

File: packages/langium/src/lsp/completion/completion-provider.ts [hunk 1]
```diff
@@ -381,28 +376,26 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
             try {
                 const scope = this.scopeProvider.getScope(refInfo);
-                scope.getAllElements().forEach(e => {
-                    if (this.filterCrossReference(context, e)) {
-                        acceptor(context, this.createReferenceCompletionItem(e));
-                    }
-                });
+                for (const description of this.filterCrossReferences(context, scope.getAllElements())) {
+                    acceptor(context, this.createReferenceCompletionItem(description));
+                }
```
sailingKieler, line 398: This is just a different way of formulating the same, correct? (except for the changed signature of `filterCrossReferences(...)`)

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -381,28 +376,26 @@ export class DefaultCompletionProvider implements CompletionProvider {
         const assignment = getContainerOfType(crossRef.feature, ast.isAssignment);
         let node = context.node;
         if (assignment && node) {
-            if (crossRef.type && (crossRef.new || node.$type !== crossRef.type)) {
+            if (crossRef.type) {
                 node = {
```
sailingKieler, line 380: why is `node` re-initialized although it is already truthy?
Is this branch supposed to cover the case of assigning a (the first) property of an astNode not existing in the AST yet?
Under which circumstances is `crossRef.type` (un-)initialized?
The var name `crossRef` is a bit misleading as it actually denotes a `NextFeature` referring to a crossRef. Esp. `crossRef.type` is confusing, as it denotes `NextFeature.type`, not `CrossReference.type`.

Please add some comments here (and there).

sailingKieler (changes requested):
First of all, although you stated that these changes are difficult to mechanically test, I still miss some unit tests, at least for documenting the expected behavior of the *follow-element-computation* algorithm. I would appreciate to have a few, e.g. for being able to easily debug the implementation in case of unexpected behaviors.

General question/remark:
* Is it intended that for terminal or datatype rules no completion is returned at all, according to `completionFor(...)`? I guess some hint on what can be expected as arguments in `completionFor(...)` would be very helpful.

Technical question:
* Why does `interpretStackToken` prefix the `nextFeatures` with the existing `stack`, and `findNextFeatures` drops that prefix, again? Do I overlook any evaluations of the prefixed `nextFeatures`?

Apart from my complaints you build quite impressive stuff! 🤩

msujew:
> First of all, although you stated that these changes are difficult to mechanically test, I still miss some unit tests, at least for documenting the expected behavior of the follow-element-computation algorithm. I would appreciate to have a few, e.g. for being able to easily debug the implementation in case of unexpected behaviors.

Not really in scope of this PR. The current follow element computation is so heavily intertwined with the workings of the completion provider that I would need to replicate half of what we do in the completion provider to build a test for the follow element computation.

Note that you can debug it using the existing completion provider tests.

> Is it intended that for terminal or datatype rules no completion is returned at all, according to completionFor(...)?

If the terminal or datatype rule is part of a cross reference, then we fall into the expected cross reference case. Otherwise, there's no reasonable way to assume the expected value of a terminal or datatype value, so we don't offer any completion for that.

> Why does interpretStackToken prefix the nextFeatures with the existing stack, and findNextFeatures drops that prefix, again? Do I overlook any evaluations of the prefixed nextFeatures?

We use the stack during the whole computation phase, since we might need to go up the stack once we encounter the end of a parser rule. Only the absolutely last step of the algorithm actually drops the stack again, since we only need the last elements of it for the completion provider.

I've reverted the breaking change and added a few more comments here and there 👍

packages/langium/src/lsp/completion/follow-element-computation.ts
```diff
@@ -146,7 +139,11 @@ function findFirstFeaturesInternal(options: { next: NextFeature, cardinalities:
             .map(e => modifyCardinality(e, feature.cardinality, cardinalities));
     } else if (ast.isAlternatives(feature) || ast.isUnorderedGroup(feature)) {
         return feature.elements.flatMap(e => findFirstFeaturesInternal({
-            next: { feature: e, new: false, type },
+            next: {
+                feature: e,
+                type,
+                property: next.property
```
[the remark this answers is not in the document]
msujew: Yes 👍

packages/langium/src/lsp/completion/completion-provider.ts
msujew: Kind of yes. I just like the `for` loop more, since it's easier to debug.

packages/langium/src/lsp/completion/completion-provider.ts
msujew: We sometimes have parser rules that look like this:

```
Ref: ref=[Element:ID];
```

Since we want the next element in our completion algorithm, we actually complete the `ref` assignment/cross reference although the current element cannot be of type `Ref`. In this case, the `NextFeature#type` property is set to `Ref` (since it's the new containing type), and we build a new synthetic element that can be used to compute the cross reference.

It's generally uninitialized when we don't enter a new parser rule during completion. I.e. like in the hello world grammar:

```
Hello |
```

Where the cross reference at `|` is part of the same parser rule as `Hello`.

sailingKieler: Thanks a lot 👍

sailingKieler (approved):
Thanks for the improvements! 🙏

PR #1234, opened 2023-10-20
Add semantic token type `decorator`

lars-reimann:
Adds the missing entry for `SemanticTokenTypes.decorator` to `AllSemanticTokenTypes`. Otherwise, `AllSemanticTokenTypes` and `AllSemanticTokenModifiers` were complete.


lars-reimann:
I've added the new entry at the end of the object, so the other entries don't need to be shifted. Would you prefer to keep the keys sorted alphabetically?

msujew (approved):
Thanks, seems like the `decorator` token was just recently added with version 3.17 of the protocol. Adding that at the end is perfectly fine.

Issue #1236, opened 2023-10-22
update `vscode-languageserver` and `vscode-languageclient` to `v9`


Yokozuna59:
Hi there, is it possible to update `vscode-languageserver` to `v9`? I'm aware it would be considered a breaking change, but it would be very helpful to my use case.
I plan to have [lsp-embedded-language-service](https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-embedded-language-service) in the same document, which means to use two language servers and switch between them back and forth according to position. One main issue is that the package that has the other language server has a different version of `vscode-languageserver`. That's why updating the version here would help me integrate both language servers.

> **_Note_**: I'm not sure if it's the best approach to do so, but that's what's on my mind right now.

PR #1237, opened 2023-10-24
Update language server libraries to version 9.0

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1236

Updates our used version of `vscode-languageserver` and everything related to their newest versions.

### DIFF
--- a/examples/requirements/package.json
+++ b/examples/requirements/package.json
@@ -73,8 +73,8 @@
         "commander": "~11.0.0",
         "langium": "~2.0.0",
         "lodash": "~4.17.21",
-        "vscode-languageclient": "~8.1.0",
-        "vscode-languageserver": "~8.1.0"
+        "vscode-languageclient": "~9.0.1",
+        "vscode-languageserver": "~9.0.1"
     },
     "devDependencies": {
         "langium-cli": "~2.0.0"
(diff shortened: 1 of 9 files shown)


examples/statemachine/package.json
```diff
@@ -60,7 +60,9 @@
         "chalk": "~5.3.0",
         "commander": "~11.0.0",
         "langium": "~2.0.0",
-        "lodash": "^4.17.21"
+        "lodash": "^4.17.21",
+        "vscode-languageclient": "~9.0.1",
```
Lotes, line 64: These deps were not needed before this change. Why now?

kaisalmen: That seems to have been missing all the time.

examples/statemachine/package.json
```diff
@@ -60,7 +60,9 @@
         "chalk": "~5.3.0",
         "commander": "~11.0.0",
         "langium": "~2.0.0",
-        "lodash": "^4.17.21"
+        "lodash": "^4.17.21",
```
kaisalmen, line 63: This is not needed in statemachine.

examples/requirements/package.json
```diff
@@ -73,8 +73,8 @@
         "commander": "~11.0.0",
         "langium": "~2.0.0",
         "lodash": "~4.17.21",
```
kaisalmen, line 75: Can be removed here as well. It is only required in domainmodel.

kaisalmen (commented):
See no issues with the dependency update. The lodash discovery was coincidence.

kaisalmen (approved):
LGTM

PR #1238, opened 2023-10-24
generator api ('joinToNode') supplements

sailingKieler:
Some improvements inspired by applications in productive projects

* extended definition of 'JoinOptions' used to configure executions of 'joinToNode'
* extended type definitions of `joinToNode` allowing to skip the `toGenerated` argument (conversion function)

### DIFF
--- a/packages/langium/src/generator/node-joiner.ts
+++ b/packages/langium/src/generator/node-joiner.ts
@@ -5,7 +5,7 @@
  ******************************************************************************/
 
 import type { AstNode, Properties } from '../syntax-tree.js';
-import type { Generated } from './generator-node.js';
+import { type Generated, isGeneratorNode } from './generator-node.js';
 import type { SourceRegion } from './generator-tracing.js';
 import { CompositeGeneratorNode, traceToNode } from './generator-node.js';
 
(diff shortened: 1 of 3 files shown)


packages/langium/src/generator/node-joiner.ts
```diff
@@ -50,25 +90,39 @@ export interface JoinOptions<T> {
...
+    iterable: Iterable<T> | T[],
+    toGeneratedOrOptions: ((element: T, index: number, isLast: boolean) => Generated) | JoinOptions<T> = defaultToGenerated,
+    options: JoinOptions<T> = {}
 ): CompositeGeneratorNode | undefined {
 
+    const toGenerated = typeof toGeneratedOrOptions === 'function' ? toGeneratedOrOptions : defaultToGenerated;
+    const { filter, prefix, suffix, separator, appendNewLineIfNotEmpty, skipNewLineAfterLastItem } = typeof toGeneratedOrOptions === 'object' ? toGeneratedOrOptions : options;
+
+    const prefixFunc = typeof prefix === 'function' ? prefix : undefined;
+    const suffixFunc = typeof suffix === 'function' ? suffix : undefined;
```
msujew, line 107: Suggestion: All the undefined-checking and casting to `Generated` below is a bit ugly, we can just define the function here a bit smarter.

```suggestion
    const prefixFunc = typeof prefix === 'function' ? prefix : (() => prefix);
    const suffixFunc = typeof suffix === 'function' ? suffix : (() => suffix);
```

msujew (commented):
Looks quite good, just a minor improvement, see below.

Issue #1216, opened 2023-10-04
Wrong AST hierarchy on completion


cdietrich:
given the grammar

```
grammar HelloWorld

entry Model:
    (myEnums+=MyEnum | ("call" expressions+=DotExpression))*
;
MyEnum:
    'enum' name=ID '{'
        literals+=MyEnumLiteral*
    '}'
;
MyEnumLiteral:
    name=ID
;

DotExpression infers Expression:
    Primary ({infer DotExpression.base = current} '.' target=DotTarget)*;

Expression:
    DotExpression;

Primary infers Expression:
    {infer Parenthesis} '(' expr=Expression ')' |
    RefExpression |
    NumberLiteral;

RefExpression:
    target=[MyEnum:ID];

NumberLiteral:
    value=INT
;

DotTarget:
   {infer DotTargetRef} ref=[MyEnumLiteral:ID] | {infer MyX} 'x' | {infer MyY} 'y';

hidden terminal WS: /\s+/;
ID returns string: ID1 | ID2;
terminal ID1: /[_a-zA-Z][\w_-]*/;
terminal ID2: /`[^`]*`/;
terminal INT returns number: /[0-9]+/;
terminal STRING: /"(\\.|[^"\\])*"|'(\\.|[^'\\])*'/;

hidden terminal ML_COMMENT: /\/\*[\s\S]*?\*\//;
hidden terminal SL_COMMENT: /\/\/[^\n\r]*/;

```

and this scope provider

```
import { EMPTY_SCOPE, DefaultScopeProvider, ReferenceInfo, Scope } from "langium";
import { isDotExpression, isDotTargetRef, isRefExpression } from "./generated/ast.js";

export class HelloWorldScopeProvider extends DefaultScopeProvider {

    override getScope(context: ReferenceInfo): Scope {
        if (context.property === 'ref' && isDotExpression(context.container)) {
            const memberCall = context.container;
            const previous = memberCall.base;
            if (!previous) {
                return super.getScope(context);
            }
            if (isRefExpression(previous)) {
                const t = previous.target.ref!
                return this.createScopeForNodes(t.literals)
            }
            return EMPTY_SCOPE;
        } else if (context.property === 'ref' && isDotTargetRef(context.container)) {
            const memberCall = context.container.$container;
            const previous = memberCall.base;
            if (!previous) {
                return super.getScope(context);
            }
            if (isRefExpression(previous)) {
                const t = previous.target.ref!
[...]

PR #1239, opened 2023-10-24
Add new completion context after previous token

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1216

The new completion context uses the previous AST node to compute the completion that should immediately follow the previous element. Automatic testing for this is quite difficult, since it requires a custom scope provider. But the issue mentioned above contains a manual reproduction instruction that can be used to confirm that this fix is effective.


cdietrich:
thx for the fix. will downport to our langium 1x prod env 2morrow.

cdietrich:
Can confirm that the patch also works in our actual DSL

insafuhrmann (approved):
Thanks @msujew, works in my test as well.

PR #1240, opened 2023-10-24
Remove vscode-uri bundler workaround

msujew:
Microsoft has finally released a node (and bundler) compatible ESM version of `vscode-uri` (3.0.8). See also https://github.com/microsoft/vscode-uri/pull/39.

This change removes the workaround that was needed in Langium to use the values exported by `vscode-uri`.


msujew:
cc @kaisalmen please retest with your vite config to ensure that the bundling behavior is correct in vite.

kaisalmen:
Thanks, will double check.

kaisalmen (approved):
Looking good. I double checked this. Bundling with rollup works. LGTM.

PR #1241, opened 2023-10-24
Treat lines with only comment symbols as empty in JSDoc comments

lars-reimann:
Previously, lines that only contained a (line) comment symbol did not trigger a new Markdown paragraph:

image

This PR adjusts this behavior, so a new paragraph gets created:

image


msujew (approved):
Thanks, I don't know how I missed that during testing 👍

PR #1245, opened 2023-10-27
Customize rendering of tags in JSDoc comments

### LINKED ISSUE #1242: Custom rendering of tags in JSDoc comments
The `JSDocDocumentationProvider` is a great default documentation provider. I'd like an easier means to customize the rendering of tags, however. For example, the first word after a `@param` tag might indicate the name of the parameter, which could be rendered as an inline code block.

A possible API would be to add a function

```ts
interface JSDocRenderOptions {
    renderTag?(tag: JSDocTag): string | undefined
}
```

This would also provide nice symmetry with the other properties of this interface (`tag`, `link`, `renderLink`).

Additionally, a method

```ts
protected documentationTagRenderer(node: AstNode, tag: JSDocTag): string | undefined
```

could be added to `JSDocDocumentationProvider`. By default, it could provide the prior behavior.

-----

If you are interested in this feature, I'd be willing to work on it.

lars-reimann:
Closes #1242

Add two means to customize the rendering of tags in JSDoc comments in the generated Markdown:

1. When calling the `toMarkdown` methods, a `renderTag` function can be passed in the options.
2. Inside the `JSDocDocumentationProvider`, the method `documentationTagRenderer` can be overridden.

### DIFF
--- a/packages/langium/src/documentation/jsdoc.ts
+++ b/packages/langium/src/documentation/jsdoc.ts
@@ -77,6 +77,11 @@ export interface JSDocRenderOptions {
      */
     link?: 'code' | 'plain'
     /**
+     * Custom tag rendering function.
+     * Return a markdown formatted tag or `undefined` if the tag is not valid.
+     */
+    renderTag?(tag: JSDocTag): string | undefined
+    /**
      * Custom link rendering function. Accepts a link target and a display value for the link.
      * Return a markdown formatted link with the format `[$display]($link)` or `undefined` if the link is not a valid target.
      */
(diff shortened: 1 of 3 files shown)


packages/langium/src/documentation/jsdoc.ts
```diff
@@ -562,6 +567,10 @@ class JSDocTagImpl implements JSDocTag {
     }
 
     toMarkdown(options?: JSDocRenderOptions): string {
+        return options?.renderTag?.(this) ?? this.toMarkdownDefault(options);
```
spoenemann, line 570: The comment above says the return value is `undefined` if the tag is not valid, but here it's treated the same as if `renderTag` is not defined. Shall we handle the two cases differently (`renderTag` undefined, `renderTag` returning undefined)?

lars-reimann: The comment

```
/**
 * Custom tag rendering function.
 * Return a markdown formatted tag or `undefined` if the tag is not valid.
 */
```

is misleading indeed. How about this instead:

```
/**
 * Custom tag rendering function.
 * Return a markdown formatted tag or `undefined` to fall back to the default rendering.
 */
```

spoenemann: Ok 👍

lars-reimann: I've changed the comment accordingly.

spoenemann (approved):

Issue #1251, opened 2023-11-01
cli command shown in "langium-quickstart.md" is not correct


tomsontom:
Langium version: 2.0.0
Package name: langium-generator

## Steps To Reproduce

1. generate hello-world sample
2. try running the advertised `./bin/cli` which leads to the following error (on OS-X)
```
./bin/cli
zsh: no such file or directory: ./bin/cli
```

## The current behavior

Command shown langium-quickstart.md does not work.

## The expected behavior

Command shown in langium-quickstart.md should work unversially

## Possible ways to solve

- chmod `cli.js` to be executable and change command to `./bin/cli.js`
- change command to `node ./bin/cli.js`

If not mistaken although I think shebang-lines don't work on Windows so I think the proper way is to change the readme to use the 2nd option

msujew:
@sailingKieler You fail to see it, because you have a Mac :)

I didn't work on Windows (as mentioned in the initial issue comment). Windows is fairly strict about what it accepts as an executable via the command line - which is why calling `node` first is necessary on Windows. Closing the issue.

sailingKieler:
Indeed, I thought in that direction but the description says:
<img width="612" alt="image" src="https://github.com/eclipse-langium/langium/assets/8513883/6c14d9b5-e5a8-443a-899b-cc38ff8aa891">

msujew:
Hm, I see. No idea, though using `node` should work anyway as expected ¯\\\_(ツ)\_/¯

tomsontom:
@sailingKieler - correct the error from above is produced on my OS-X
- I use `yo langium` to bootstrap
- In the bin-directory there's no file named `cli` hence the file not found error is at least to me expected
- And as the file has not set the executable bit I don't understand why you can execute it

@msujew - on windows (unless you run in WSL) I don't think windows can do anything with `#!/usr/bin/env node` - at least I would be very suprised and according to https://stackoverflow.com/questions/10396305/npm-package-bin-script-for-windows it is simply ignored but I confess I have not tried but as you wrote using "node" is the safeway and works just anywhere.

sailingKieler:
@tomsontom you're right, since v2.0 `./bin/cli` will be generated as `./bin/cli.js`. (I had v1.0.0 in my global installation.)

[...]

PR #1252, opened 2023-11-01
Fix cli commands in langium-quickstart.md (#1251)

tomsontom:
Closes https://github.com/eclipse-langium/langium/issues/1251

Fixed commands mentioned in to use the node-binary


msujew (approved):
Thanks, looks good. I'm wondering whether we need to somehow adjust the quickstart based on the selected options, but this change is good for now.

PR #1258, opened 2023-11-02
Add Separate Export for LSP Services

### LINKED ISSUE #1166: Export LSP services via dedicated ESM export
Currently, our LSP services (which rely heavily on the `vscode-languageserver` package) are exported through the main `langium` export. While this makes creating language servers very easy, it leads to a lot of unnecessary imports for adopters that don't want to build a language server, thereby increasing bundle size.

The long term plan (likely version 3.0, as it's a major breaking change) would be to export the LSP related services through a `langium/lsp` export. That way, adopters can choose to not to include those services.

montymxb:
Closes #1166 by providing a separate `langium/lsp` export for LSP-related services for Langium.

Note, this is a breaking change planned for introduction later in v3.0.0.

Changes (not exhaustive):
- Splits lsp related services into lsp-services & langium-lsp-module
- Furrther split some existing services into default & lsp variants (configuration provider, langium document factory, workspace manager, to name a few)
- Exposed related types like `LangiumServicesWithLSP` and `LangiumSharedServicesWithLSP` to make it pretty clear how to extend existing applications with the new type signatures
- Updates existing example languages to incorporate the new LSP module (in conjunction with the default one) to work correctly
- More or less leaves everything in 'grammar' that is tied to /lsp alone
- TextDocument(s) were pulled out into the LSP, but we may want to put this back, I'm open for thoughts here
- updates impacted tests
- inject can inject more

Open for refactors and changes based on feedback, and can go from there.

`createServicesForGrammarWithLSP` is marked with a TODO that is a effectively an RFC. Looking for feedback to see if I should generalize the approach a bit more to work with `createServicesForGrammar`, or leave it as is.


spoenemann:
We need to eliminate the internal dependencies to the lsp subfolder. This is only possible by splitting `default-module.ts`.

spoenemann (commented):
Thanks for this first spike!

I'm thinking about a different approach to structure the services and modules:

 - Introduce `LangiumCoreServices` and `LangiumLspServices` for the split, incl. respective shared services and default modules
 - Keep `LangiumServices` and `LangiumSharedServices` intact, i.e. the union of core and LSP (implying that these need to be defined in the `lsp` subfolder)

Rationale:
 - Reduce the impact of the change in terms of breaking existing code (just change the import)
 - Splitting core and LSP services is not something that many users will do, but rather a special feature for those using Langium as a library in the browser
 - We should keep things simple for the most common case, but make the more special cases possible

@msujew WDYT?

montymxb:
@spoenemann I'm on board with the core vs. lsp services 👍 . I was also wondering if we wanted to do something with an explicit 'core' set of services, I like that.

In addition retaining the structure of `LangiumServices` and `LangiumSharedServices` as being lsp-complete makes sense, given existing consumers of these services will generally expect them to be complete.

montymxb:
Also renamed the `DefaultServiceRegistry` to `CoreServiceRegistry`, still not 100% sure on this given the 'core' registry is also used in the lsp + core case -- just with a different interface.

montymxb:
Just a note, there's a single test failure after the last rebase on top of the new `./generate` export work. I'm not able to pin it down at this time, but the CI is a bit misleading here. There appears to be a separate error regarding vite being unable to import some functions, for example:
>  TypeError: __vite_ssr_import_0__.createDefaultSharedModule is not a function

Several other 'langium' imports in the domainmodel example seem to have a similar issue, but strangely not others like requirements or statemachine, which look like they should trigger the same issue. The plan is to keep looking into this more next week to resolve.

packages/langium/src/lsp/langium-lsp-module.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/**
+ * Creates Langium grammar services, enriched with LSP functionality
+ *
+ * @param context Shared module context, used to create additional shared modules
+ * @param sharedModule Existing shared module to inject together with new shared services
+ * @returns Shared services enriched with LSP services + Grammar services, per usual
+ */
+export function createLangiumGrammarServicesWithLSP(context: DefaultSharedModuleContext,
```
sailingKieler, line 92: This should probably go into some `langium-grammar-module.ts` in `langium/grammar/lsp`, right?
conflicts with #1171...

packages/langium/src/lsp/langium-lsp-module.ts
```diff
@@ -0,0 +1,115 @@
+/******************************************************************************
```
sailingKieler, line 1: I would prefer the name `default-lsp-module` to emphasize that it contributes defaults for all languages, not just the Langium grammar language (could be perceived as such)
(and to keep the name similarity to the `default-module`)

packages/langium/src/lsp/langium-lsp-module.ts
```diff
@@ -0,0 +1,115 @@
...
+ */
+export interface LSPModuleContext {
+    shared: LangiumSharedLSPServices;
+}
+
+/**
+ * Create a dependency injection module for the default LSP services. This is a
+ * set of services that are used by exactly one language.
+ */
+export function createLSPModule(context: LSPModuleContext): Module<LangiumServices, LangiumLSPServices> {
```
sailingKieler, line 45: similarly I would prefer the name `createDefaultLSPModule` (see comment on line 1)

packages/langium/src/lsp/workspace/workspace-manager.ts
```diff
@@ -0,0 +1,127 @@
...
+import type { MutexLock } from '../../utils/promise-util.js';
+import type { BuildOptions, DocumentBuilder } from './../../workspace/document-builder.js';
+import type { LangiumDocument, LangiumDocuments } from './../../workspace/documents.js';
+import type { FileSystemNode, FileSystemProvider } from './../../workspace/file-system-provider.js';
+import type { LangiumSharedServices } from '../lsp-services.js';
+
+/**
+ * Langium Workspace manager that uses LSP workspace folders
+ */
+export class LSPWorkspaceManager implements WorkspaceManager {
```
sailingKieler, line 22: I need to intervene at this point, we need the `workspace initialization` also in CLI apps in case the language offers built-in libs or allows to include dependency packages.

Importing types, e.g. from `vscode-languageserver` does hopefully not hurt, but we need to inject our own `CancellationToken.None` in the non-lsp case.

sailingKieler (commented):
Very good progress @montymxb!
I have a few detail remarks/wishes and one major thing: The workspace initialization should not got into the LSP related part.

I think special attention need to be paid to the stuff in `promise-util.ts` that need to be separated, too, doesn't it?

packages/langium/src/lsp/workspace/workspace-manager.ts
msujew: > Importing types, e.g. from `vscode-languageserver` does hopefully not hurt

It does hurt, because that the library we're hoping to exclude in the bundle if the adopter doesn't import from `langium/lsp`. However, I agree that the default workspace manager should still be available outside of the LSP export.

We should define our own `CancellationToken` interface in Langium.

examples/domainmodel/test/domainmodel-cli.test.ts
```diff
@@ -37,7 +37,7 @@ describe('Test the domainmodel CLI', () => {
     }
 
     test('Test action without CLI', async () => {
-        await generateAction(rawfileName, { destination, quiet: true });
+        await generateAction(rawfileName, { destination, quiet: false });
```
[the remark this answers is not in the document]
montymxb: Temporary, the quiet mode made it very difficult to actually see that there was an underlying error from vite otherwise. The plan is to switch that back.

packages/generator-langium/templates/vscode/src/language/main.ts
```diff
@@ -1,4 +1,4 @@
-import { startLanguageServer } from 'langium';
+import { startLanguageServer } from 'langium/lsp';
```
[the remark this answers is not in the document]
montymxb: Yep those still need to be updated. Thanks for pointing that out.

packages/langium/src/service-registry.ts
```diff
@@ -4,38 +4,41 @@
...
+import type { LangiumCoreServices } from './services.js';
 import { UriUtils, type URI } from './utils/uri-util.js';
 
 /**
- * The service registry provides access to the language-specific services. These are resolved
+ * The core service registry provides access to the core language-specific services. These are resolved
  * via the URI of a text document.
  */
-export interface ServiceRegistry {
+export interface CoreServiceRegistryI {
```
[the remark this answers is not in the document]
montymxb: I've added it since below we have a conflict with `CoreServiceRegistry` which implements this interface, but I'm not particularly fond of the name choices so far (I made a comment about this earlier as well). I wouldn't be opposed to moving the initial around, but I'm more curious if the choice of making a 'Core' service makes sense.

packages/langium/src/lsp/langium-lsp-module.ts
montymxb: Good eyes, it's not quite where it should be yet. I'm waiting to rebase on #1171 , and am planning to build off of that. If it's already been moved there, I'll follow that.

packages/langium/src/lsp/langium-lsp-module.ts
montymxb: I'm not opposed to this, but I imagine we should discuss this together. I have a feeling we might be renaming the default module to a 'core' module, which would change the case.

packages/langium/src/lsp/langium-lsp-module.ts
montymxb: See above

packages/langium/src/lsp/workspace/workspace-manager.ts
montymxb: I can factor the cancellation token out in another set of changes, and see how that works out.

sailingKieler: >> Importing types, e.g. from vscode-languageserver does hopefully not hurt
>
> It does hurt, because that the library we're hoping to exclude in the bundle if the adopter doesn't import from langium/lsp

Just to make sure we're on the same page here: Does
`import type { WorkspaceFolder } from 'vscode-languageserver';`
have an impact on the bundling (tree shaking)?

msujew: No, but `import { CancellationToken } from 'vscode-languageserver';` (which is in the file) has an impact.

Independent of whether the bundling can take care of that, I would much rather not import anything from `vscode-languageserver` outside of `langium/lsp`. That'll only invite trouble in the future.

sailingKieler: > No, but import { CancellationToken } from 'vscode-languageserver'; (which is in the file) has an impact.

That's for sure.

montymxb: Dito, although we could probably get away with just the types, it makes it fairly easy to import something that isn't a type by accident later.

examples/domainmodel/test/domainmodel-cli.test.ts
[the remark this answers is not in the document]
montymxb: Latest commit adds a TODO as an additional reminder, just so that doesn't sneak in before merging.

packages/langium/src/service-registry.ts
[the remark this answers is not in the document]
montymxb: Flipped the I around in the last changes.

packages/langium/src/service-registry.ts
```diff
@@ -4,38 +4,41 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { LangiumServices } from './services.js';
+import type { LangiumCoreServices } from './services.js';
 import { UriUtils, type URI } from './utils/uri-util.js';
 
 /**
- * The service registry provides access to the language-specific services. These are resolved
+ * The core service registry provides access to the core language-specific services. These are resolved
```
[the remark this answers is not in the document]
montymxb: Added the suggested wording.

packages/langium/src/lsp/workspace/workspace-manager.ts
montymxb: Last changes pull out and declare some of the necessary types from 'vscode-languageserver', which allows us to use CancelToken (and WorkspaceFolder) outside of the LSP context. The LSPWorkspaceManager is still present, but just extends upon the existing default, since that now provides ~90% of what we need anyways.

Some additional type changes were propagated using these internally defined types (CancellationToken & WorkspaceFolder) to some other services & example languages. We can still bend the type system as needed to determine where we want to switch between using ours vs. the vs-ls types, but I presume we'll assess in coming review passes.

msujew (commented):
Great, you've made good progress on this. We needsome better abstraction layers for some of these services, but otherwise this looks quite good.

montymxb:
Thanks for the feedback @msujew . I'm catching up on some things right now, and then should be coming back to reply & update in a day or two.

examples/domainmodel/test/domainmodel-cli.test.ts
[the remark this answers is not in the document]
montymxb: Underlying issue is resolved, so I've reverted this in an upcoming commit.

packages/langium/src/lsp/langium-lsp-module.ts
montymxb: This has been removed after rebasing on the grammar export PR.

packages/generator-langium/templates/vscode/src/language/main.ts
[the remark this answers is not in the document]
montymxb: Core generator templates have been updated

packages/langium/src/service-registry.ts
```diff
@@ -4,38 +4,41 @@
...
+import type { LangiumCoreServices } from './services.js';
 import { UriUtils, type URI } from './utils/uri-util.js';
 
 /**
- * The service registry provides access to the language-specific services. These are resolved
+ * The core service registry provides access to the language-specific core services. These are resolved
  * via the URI of a text document.
  */
-export interface ServiceRegistry {
+export interface ICoreServiceRegistry {
```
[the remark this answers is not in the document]
montymxb: Good point, I can update the base one accordingly.

packages/generator-langium/templates/cli/src/cli/cli-util.ts
```diff
@@ -1,10 +1,11 @@
-import type { AstNode, LangiumDocument, LangiumServices } from 'langium';
+import type { AstNode, LangiumCoreServices, LangiumDocument } from 'langium';
+import type { LangiumServices } from 'langium/lsp';
```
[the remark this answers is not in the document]
montymxb: Hmm probably not. Although it's used in the generated `extractAstNode` function below, it's just passed to `extractDocument`, which expects a core service set rather than an lsp-enriched on. Dropping it and updating the one usage should work just fine.

packages/langium-cli/src/generator/module-generator.ts
```diff
@@ -25,14 +25,16 @@ export function generateModule(grammars: Grammar[], config: LangiumConfig, gramm
...
+            import type { LangiumSharedServices, LangiumServices } from '../../lsp/lsp-services${config.importExtension}';
         `.appendTemplateIf(hasIParserConfigImport)`
 
             import type { IParserConfig } from '../../parser/parser-config${config.importExtension}';
         `
     ).appendTemplateIf(!config.langiumInternal)`
 
-        import type { LangiumGeneratedServices, LangiumGeneratedSharedServices, LangiumSharedServices, LangiumServices, LanguageMetaData, Module${hasIParserConfigImport ? ', IParserConfig' : ''} } from 'langium';
+        import type { LangiumGeneratedServices, LangiumGeneratedSharedServices, LanguageMetaData, Module${hasIParserConfigImport ? ', IParserConfig' : ''} } from 'langium';
+        import type { LangiumSharedServices, LangiumServices } from 'langium/lsp';
```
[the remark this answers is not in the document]
montymxb: I started with the core services, and then switched to the full ones, but I can't remember why. I think I had an issue with one of the generated examples, but it may have been pre-rebase. I'll go back over this and try the core service set again.

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -39,32 +118,24 @@ export class DefaultWorkspaceManager implements WorkspaceManager {
...
-        services.lsp.LanguageServer.onInitialized(_params => {
-            // Initialize the workspace even if there are no workspace folders
-            // We still want to load additional documents (language library or similar) during initialization
-            this.mutex.lock(token => this.initializeWorkspace(this.folders ?? [], token));
-        });
+        // for usage w/ LSP services, refer to the LSPWorkspaceManager instead
     }
 
     async initializeWorkspace(folders: WorkspaceFolder[], cancelToken = CancellationToken.None): Promise<void> {
+        // return Promise.resolve();
```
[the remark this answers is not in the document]
montymxb: Yup, I'll clean that up.

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -4,17 +4,96 @@
...
+    /**
+     * The name of the workspace folder. Used to refer to this
+     * workspace folder in the user interface.
+     */
+    name: string;
+}
+
+declare namespace WorkspaceFolder {
+    function is(value: unknown): value is WorkspaceFolder;
+}
```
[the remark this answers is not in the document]
montymxb: Dropped in the next commit. I believe previously there was an issue with leaving this out, but I don't see it in the current state.

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -4,17 +4,96 @@
...
+
+interface Disposable {
+    /**
+     * Dispose this object.
+     */
+    dispose(): void;
+}
```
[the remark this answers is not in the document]
montymxb: Oh nice, I'll _dispose_ of this in favor of that

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -4,17 +4,96 @@
...
+declare namespace Disposable {
+    function create(func: () => void): Disposable;
+}
```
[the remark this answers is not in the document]
montymxb: Dropped in an upcoming commit.

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -4,17 +4,96 @@
...
+ * Placeholder object for a no-cancel token.
+ * Replaces the usage of None from the namespace
+ * `CancellationToken` from `vscode-languageserver-protocol`.
+ */
+const CancellationToken = {
+    None: {
+        isCancellationRequested: false,
+        onCancellationRequested: Event.None
+    }
+};
```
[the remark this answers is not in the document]
montymxb: Pulling them into the workspace manager was done out of convenience while testing things out, but a proper spot in utils sounds good.

@msujew What do you think about the WorkspaceFolder? Given this is the workspace manager I could see it living here, but I'm not sure if there's a better spot for it, similar to the token & event.

packages/langium/src/lsp/workspace/workspace-manager.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export class LSPWorkspaceManager extends DefaultWorkspaceManager {
+
+    protected folders?: WorkspaceFolder[];
+
+    constructor(services: LangiumSharedServices) {
+        super(services);
+        services.lsp.LanguageServer.onInitialize(params => {
+            // convert lsp workspace folders to langium workspace folders (equivalent type plays nice with the existing non-lsp core services)
+            this.folders = (params.workspaceFolders as unknown as WorkspaceFolder[]) ?? undefined;
```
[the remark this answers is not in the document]
montymxb: You're right, hmm, when I was drilling down into `vscode-languageserver-types` I came across a def there w/ type `URI`, but we're definitely using string for ours. Thanks for catching that, I'll change that over accordingly.

packages/langium/src/workspace/workspace-manager.ts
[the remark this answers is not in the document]
msujew: Having `WorkspaceFolder` in the `WorkspaceManager` file is fine. For `token` and `event`, it makes sense to have `util` files for them.

packages/langium/src/workspace/configuration.ts
```diff
@@ -30,39 +28,22 @@ export interface ConfigurationProvider {
     updateConfiguration(change: DidChangeConfigurationParams): void;
 }
 
-export class DefaultConfigurationProvider implements ConfigurationProvider {
+/**
+ * Empty Configuration provider, used in the absence of an LSP supported configuration.
+ * Which is typically required to set this up properly.
+ */
+export class EmptyConfigurationProvider implements ConfigurationProvider {
```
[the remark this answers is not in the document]
montymxb: I can add in that additional abstraction layer, makes sense.

> The new LSPConfigurationProvider and EmptyConfigurationProvider share all of it's getter/setter logic.

For the point above, you mean the `InMemoryConfigurationProvider` here right?

packages/langium/src/lsp/workspace/documents.ts
```diff
@@ -0,0 +1,146 @@
...
+import type { FileSystemProvider } from './../../workspace/file-system-provider.js';
+import { DocumentState, type LangiumDocument, type LangiumDocumentFactory } from '../../workspace/documents.js';
+import { URI } from '../../utils/uri-util.js';
+import { TextDocument } from 'vscode-languageserver-textdocument';
+import type { LangiumSharedServices } from '../lsp-services.js';
+
+/**
+ * Document Factory that utilizes lsp TextDocuments (or the file system) to create/update LangiumDocuments
+ */
+export class LSPLangiumDocumentFactory implements LangiumDocumentFactory {
```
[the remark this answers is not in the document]
montymxb: True, I'll factor the duped logic out here too.

packages/langium/src/lsp/langium-lsp-module.ts
montymxb: Came back to this and decided to apply it, change will be up in a moment.

packages/langium/src/lsp/langium-lsp-module.ts
montymxb: Also applied this in an upcoming commit.

packages/langium/src/lsp/lsp-services.ts
```diff
@@ -0,0 +1,125 @@
...
+ * Combined Core + LSP services of Langium (total services)
+ */
+export type LangiumServices = LangiumCoreServices & LangiumLSPServices;
+
+/**
+ * Combined Core + LSP shared services of Langium (total services)
+ */
+export type LangiumSharedServices = LangiumSharedCoreServices & LangiumSharedLSPServices;
+
+export type LSPServices = {
```
spoenemann, line 52: Exporting both `LSPServices` and `LangiumLSPServices` can be confusing. Can we simply merge them to one definition?

spoenemann (changes requested):
Seeing the amount of additional complexity that is necessary to separate LSP from Core, I am no longer convinced that the benefits of the separation justify its costs.

What's the use case for the separation? Using Langium in a browser app to parse and process text (without editor support), with a minimal bundle size. That's a rather specific case.

Can we offer an alternative solution for that use case? For example, a recommended bundling configuration where imports from `vscode-languageserver` are replaced by some mocks?

Let's discuss this in January when everyone's back from vacations.

_Edit: I resolved my detail comments below – there's no need to address them before we clarify how to proceed in general._

packages/langium/src/lsp/default-lsp-module.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/**
+ * Creates Langium grammar services, enriched with LSP functionality
+ *
+ * @param context Shared module context, used to create additional shared modules
+ * @param sharedModule Existing shared module to inject together with new shared services
+ * @returns Shared services enriched with LSP services + Grammar services, per usual
+ */
+export function createLangiumGrammarServicesWithLSP(context: DefaultSharedModuleContext,
```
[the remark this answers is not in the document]
sailingKieler: > Can we move this to the `langium-vscode` package?

Would that be compatible with the Langium playground?

packages/langium/src/lsp/language-server.ts
```diff
@@ -566,7 +575,7 @@ export function createTypeHierarchyRequestHandler<P extends TypeHierarchySuperty
     const serviceRegistry = sharedServices.ServiceRegistry;
     return async (params: P, cancelToken: CancellationToken) => {
         const uri = URI.parse(params.item.uri);
-        const language = serviceRegistry.getServices(uri);
+        const language = serviceRegistry.getServices(uri) as LangiumServices;
```
sailingKieler, line 578: These casts are nasty!
Suggestion: Replace the declaration of `serviceCall`:
`serviceCall: (services: LangiumServices, params: P, cancelToken: CancellationToken) => HandlerResult<R, E>`
by
`serviceCall: (services: LangiumCoreServices & Partial<LSPServices>, params: P, cancelToken: CancellationToken) => HandlerResult<R, E>`

That way you can drop the cast. You just have to add more `?` to the references of `service.lsp....`
(which would be required anyway in order to not take the risk of unexpected crashes)

packages/langium/src/lsp/language-server.ts
```diff
@@ -38,6 +37,7 @@ import { mergeCompletionProviderOptions } from './completion/completion-provider
 import { DefaultSemanticTokenOptions } from './semantic-token-provider.js';
 import { mergeSignatureHelpOptions } from './signature-help-provider.js';
 import { URI } from '../utils/uri-util.js';
+import type { LangiumServices, LangiumSharedServices } from './lsp-services.js';
```
sailingKieler, line 40: An invocation of 'Organize imports` would by really nice 😉

packages/langium/src/service-registry.ts
```diff
@@ -4,38 +4,41 @@
...
+import type { LangiumCoreServices } from './services.js';
 import { UriUtils, type URI } from './utils/uri-util.js';
 
 /**
- * The service registry provides access to the language-specific services. These are resolved
+ * The core service registry provides access to the language-specific core services. These are resolved
  * via the URI of a text document.
  */
-export interface ServiceRegistry {
+export interface ServiceRegistry<T extends LangiumCoreServices> {
```
sailingKieler, line 14: AFAICS this type parameter is not necessary, Typescript allows to formulate the hand full of relevant occurrences differently 🙂

packages/generator-langium/templates/core/src/language/language-id-module.ts
```diff
@@ -23,7 +25,7 @@ export type <%= LanguageName %>Services = LangiumServices & <%= LanguageName %>A
  * declared custom services. The Langium defaults can be partially specified to override only
  * selected services, while the custom services must be fully specified.
  */
-export const <%= LanguageName %>Module: Module<<%= LanguageName %>Services, PartialLangiumServices & <%= LanguageName %>AddedServices> = {
+export const <%= LanguageName %>Module: Module<<%= LanguageName %>Services, PartialLangiumCoreServices & PartialLangiumLSPServices & <%= LanguageName %>AddedServices> = {
```
sailingKieler, line 28: I would prefer to revert this change and introduce `PartialLangiumServices` again.

sailingKieler (changes requested):
Excellent work @montymxb!!

I have some remarks to simplify some things, and I agree with Miro, that the LSP services should expect `LangiumServices` instead of `LangiumCoreServices`, but in general I'm very in favor of this contribution!

packages/langium/src/workspace/documents.ts
```diff
@@ -162,9 +160,7 @@ export class DefaultLangiumDocumentFactory implements LangiumDocumentFactory {
     }
 
     create<T extends AstNode = AstNode>(uri: URI, content?: string | TextDocument | { $model: T }): LangiumDocument<T> {
```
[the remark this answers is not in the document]
sailingKieler: This area has been changed by @msujew in the meantime, so more attention on this is required anyway.

packages/langium/src/lsp/lsp-services.ts
```diff
@@ -0,0 +1,125 @@
...
+    lsp: {
```
sailingKieler, line 53: A followup thought on my previous review: Can we turn `lsp` to `lsp?`? All it's members are optional and need to be checked for being defined, anyway. That would simplify things further.

packages/langium/src/workspace/configuration.ts
```diff
@@ -89,8 +64,33 @@ export class DefaultConfigurationProvider implements ConfigurationProvider {
...
+
     protected toSectionName(languageId: string): string {
         return `${languageId}`;
     }
 }
 
+/**
+ * An configuration provider implemented in-memory.
+ */
+export class InMemoryConfigurationProvider extends AbstractConfigurationProvider {
```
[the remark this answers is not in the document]
montymxb: Sure, that sounds fine to me.

packages/langium/src/utils/cancellation-token.ts
```diff
@@ -0,0 +1,59 @@
+/******************************************************************************
```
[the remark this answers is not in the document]
montymxb: I'm not sure, I haven't checked if that would pull in anything else as well that we don't want.

packages/langium/src/lsp/lsp-services.ts
```diff
@@ -0,0 +1,125 @@
...
+ * LSP Services shared between multiple languages where Langium provides default implementations.
+ */
+export type LangiumSharedLSPServices = {
+    ServiceRegistry: ServiceRegistry<LangiumServices>
+    lsp: {
+        Connection?: Connection
+        ExecuteCommandHandler?: ExecuteCommandHandler
+        FileOperationHandler?: FileOperationHandler
+        WorkspaceSymbolProvider?: WorkspaceSymbolProvider
+        DocumentUpdateHandler?: DocumentUpdateHandler,
```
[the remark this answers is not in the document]
montymxb: Likely a typo, I don't recall needing to make that optional for any concrete reason. It should be reverted back.

packages/langium/src/lsp/language-server.ts
```diff
@@ -224,8 +225,15 @@ export function startLanguageServer(services: LangiumSharedServices): void {
     connection.listen();
 }
 
+/**
+ * Adds a handler for document updates when content changes, or watch catches a change.
+ * In the case there is no handler service registered, this function does nothing.
+ */
 export function addDocumentUpdateHandler(connection: Connection, services: LangiumSharedServices): void {
     const handler = services.lsp.DocumentUpdateHandler;
+    if (!handler) {
```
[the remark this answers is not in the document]
montymxb: I agree due to the optional param, can be removed.

packages/langium/src/lsp/language-server.ts
montymxb: Yeah the casts are a stopgap. However to rectify we can just change `sharedServices` to `LangiumSharedLSPServices`. Not sure if we need to update the whole service call since the issue stems from just the serviceRegistry, although that may work as well.

packages/langium/src/lsp/language-server.ts
montymxb: Good point, I'll add that real quick.

packages/langium/src/service-registry.ts
montymxb: It's there to ensure that anything implementing the ServiceRegistry interface works with at least the core service set or greater. Although you can drop it & the `DefaultServicerRegistry` constraints will suffice here, you lose any typechecking that would flag a service registry violating this assumption (like a user customized variant using that interface).

packages/generator-langium/templates/core/src/language/language-id-module.ts
montymxb: Seems like this is the direction overall, and having `PartialLangiumServices` sourced from the lsp export instead.

packages/langium/src/lsp/lsp-services.ts
montymxb: Not if we had it like this before, but I guess it would be fine. Do we get anything by doing this, or is this just going to make us extra checks to verify `lsp` is present?

montymxb:
Thanks for the feedback @sailingKieler . Added some notes back to what you've mentioned.

packages/langium/src/workspace/documents.ts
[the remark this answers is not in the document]
sailingKieler: meanwhile invalid

packages/langium/src/workspace/configuration.ts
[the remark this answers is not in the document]
sailingKieler: meanwhile invalid

packages/langium/src/lsp/lsp-services.ts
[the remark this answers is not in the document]
sailingKieler: solved!

packages/langium/src/utils/cancellation-token.ts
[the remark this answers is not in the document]
sailingKieler: sticked to 'vscode-jsonrpc' or 'vscode-languageserver-protocol' in imports

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -4,16 +4,30 @@
...
 import type { BuildOptions, DocumentBuilder } from './document-builder.js';
 import type { LangiumDocument, LangiumDocuments } from './documents.js';
 import type { FileSystemNode, FileSystemProvider } from './file-system-provider.js';
 import type { WorkspaceLock } from './workspace-lock.js';
+import { CancellationToken } from '../utils/cancellation-token.js';
+
+/**
+ * A workspace folder inside a client.
+ */
+export interface WorkspaceFolder {
```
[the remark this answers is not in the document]
sailingKieler: sticked to a `import type .... from 'vscode-languageserver-types'` here, added a re-export `export type { WorkspaceFolder }` for convenience

packages/langium/src/lsp/language-server.ts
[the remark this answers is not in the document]
sailingKieler: solved.

packages/langium/src/lsp/default-lsp-module.ts
```diff
@@ -0,0 +1,117 @@
...
+        },
+        shared: () => context.shared
+    };
+}
+
+/**
+ * Create a dependency injection module for the LSP shared services. This is the set of
+ * services that are shared between multiple languages.
+ */
+export function createSharedLSPModule(context: DefaultSharedModuleContext): Module<LangiumSharedLSPServices & LangiumSharedCoreServices, LangiumSharedLSPServices> {
```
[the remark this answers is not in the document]
sailingKieler: solved.

packages/langium/src/lsp/language-server.ts
sailingKieler: fixed.

packages/langium/src/service-registry.ts
sailingKieler: dropped.

packages/generator-langium/templates/core/src/language/language-id-module.ts
sailingKieler: fixed.

packages/langium/src/lsp/lsp-services.ts
sailingKieler: meanwhile invalid.

sailingKieler:
I reworked Ben's changes and from my point of view it's done and complete.

For convenience in client code I added a re-export of `WorkspaceFolder` in `workspace-manager.ts`, as I had trouble with that type in a productive project and it's unit tests in the past, see https://github.com/eclipse-langium/langium/pull/1258/files#diff-9ef9cc03a159c7e21b4350faa99ddd28f97e4f27c934a51508e14531e4f4987bR18-R21
I hope that doesn't cause unintended side effects.

For a similar reason I added a separate commit adding a re-export of `* from 'vscode-languageserver-textdocument` to `documents.ts`, see https://github.com/eclipse-langium/langium/pull/1258/commits/6e356292b20a1d1cb0bbd1e793173979d3cd4ba7#diff-eb9d95b72e597e87b07fd5d97f6df837d272c1d3041d298ed617801ab670193aR7-R9
The footprint of that package is very low, just a few kilobytes. Correspondingly, I replaced all references of imports of that package. In case of concerns that change can be dropped easily. Wrt. to the API surface I'm open for an alternative solution.

Looking forward to reviews. 👀

packages/langium/src/workspace/documents.ts
```diff
@@ -98,6 +102,14 @@ export interface DocumentSegment {
     readonly end: number
 }
 
+/**
+ * Surrogate definition of the `TextDocuments` interface from the `vscode-languageserver` package.
+ * No implementation object is expected to be offered by `LangiumCoreServices`, but only by `LangiumLSPServices`.
+ */
+export type LangiumCoreTextDocuments = {
```
msujew, line 109: Naming: Given that this interface isn't extended by another interface (i.e. there's no `LangiumTextDocuments` in `langium/lsp`, it probably makes sense to rename this:
```suggestion
export type TextDocumentProvider = {
```

packages/langium/src/workspace/document-builder.ts
```diff
@@ -4,19 +4,19 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { URI } from '../utils/uri-utils.js';
+import { CancellationToken, Disposable } from 'vscode-jsonrpc';
```
msujew, line 7: Suggestion: The `vscode-jsonrpc` library is roughly 50kb in size. Seems like quite a lot if we only need to import some utilities from it. We should implement these ourselves. We actually already have a `Disposable` implementation on our own.

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -4,22 +4,26 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { CancellationToken } from 'vscode-languageserver';
+import { CancellationToken } from 'vscode-jsonrpc';
+import type { WorkspaceFolder } from 'vscode-languageserver-types';
```
msujew, line 8: Question: Why import at all here, given that the `WorkspaceFolder` type is fairly trivial? IMO we shouldn't import from `vscode-languageserver-types` at all, even if it's just a `type` import.

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -46,22 +50,22 @@ export class DefaultWorkspaceManager implements WorkspaceManager {
...
         this.fileSystemProvider = services.workspace.FileSystemProvider;
         this.mutex = services.workspace.WorkspaceLock;
+    }
 
-        services.lsp.LanguageServer.onInitialize(params => {
-            this.folders = params.workspaceFolders ?? undefined;
-        });
+    initialize(params: InitializeParams): void {
+        this.folders = params.workspaceFolders ?? undefined;
+    }
```
msujew, line 63: We explicitly went away from this style of initialization, see https://github.com/eclipse-langium/langium/pull/593. I don't think we should reintroduce it.

msujew (commented):
Looks generally quite good. However, I'm not a big fan of type imports from packages that are not allowed to be "real" imports. I think this makes maintainability of the code way harder, as it isn't quite clear where you're allowed to import from certain packages. I think this will become even more exaggerated for less experienced contributors to our code base.

packages/langium/src/lsp/language-server.ts
```diff
@@ -171,9 +183,28 @@ export class DefaultLanguageServer implements LanguageServer {
...
+        await this.fireInitializedOnDefaultServices(params);
+
         this.onInitializedEmitter.fire(params);
         this.onInitializedEmitter.dispose();
     }
+
+    protected fireInitializeOnDefaultServices(params: InitializeParams): void {
+        this.services.workspace.ConfigurationProvider.initialize(params);
+        this.services.workspace.WorkspaceManager.initialize(params);
+    }
```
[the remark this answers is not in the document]
sailingKieler: True, but then the constructors would have to rely on `services.lsp.LanguageServer....`, and that's not possible if the LSP part is not available. And we didn't want to have a core workspace manager and an LSP-empowered specialized one.

packages/langium/src/workspace/workspace-manager.ts
sailingKieler: Personally, I'm always confused is several "copies" of the same type definition exist, and I'm always wondering which one to import. `WorkspaceFolder` is (transitively) available via `vscode-languageserver` and also via `vscode-languageclient`, which can bite you at runtime or during bundling, esp. if the developer isn't aware of that issue. That's why I think offering that interface via Langium and have stated and document the proper unique source is a good thing. However, if `vscode-languageserver-types` brings a too heavy overhead, we can revise that.

packages/langium/src/workspace/document-builder.ts
sailingKieler: Wouldn't tree-shaking drop all the unused overhead?

msujew: `vscode-jsonrpc` is a commonjs package, tree-shaking usually doesn't work on those.

packages/langium/src/lsp/language-server.ts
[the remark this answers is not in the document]
msujew: > And we didn't want to have a core workspace manager and an LSP-empowered specialized one.

I probably didn't see the discussion on this. Can you link it? I'm personally in favor of having a dedicated LSP workspace service, since this introduces a lot of unexpected pollution in the default workspace manager, i.e. the `initialize` message, which has an argument type coming from the `vscode-languageserve-types` package.

packages/langium/src/workspace/workspace-manager.ts
msujew: > However, if vscode-languageserver-types brings a too heavy overhead, we can revise that.

It's not necessarily overhead but maintenance effort. We have to ensure on every change that we don't accidentally change this import from a `type` import to a real import.

packages/langium/src/workspace/document-builder.ts
msujew: I'm not sure how well bundlers can optimize commonjs packages. See [here](https://web.dev/articles/commonjs-larger-bundles) for example.

packages/langium/src/lsp/language-server.ts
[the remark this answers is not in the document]
msujew: *sigh*, alright, I guess I'm fine with this. We should write some custom eslint rules to prevent people from accidentally importing things they are not allowed to import.

packages/langium/src/workspace/documents.ts
msujew: I get that, but I think `TextDocumentProvider` is good enough.

packages/langium/src/workspace/workspace-manager.ts
sailingKieler: Would it be an option to import the interfaces and symbols (like `CancellationToken.None`) via one or tow TS modules, pre-bundle them as part of `npm run build`, and than rely on those pre-processed definitions?
Than we wouldn't copy-paste-clones of definitions from `vscode-languageserver....` but (ideally) have full control over the set of types and symbols being used.

msujew:
@sailingKieler Alright, I'm backing down on the `import type` issue. However, we definitely shouldn't import from `vscode-jsonrpc`, as it likely won't be tree-shaken correctly due to commonjs. It should be enough to implement `CancellationToken(Source)` ourselves. This is a good think in general probably.

msujew:
@Yokozuna59 Is this still interesting for you guys over at Mermaid? This change should be able to reduce the bundled dependencies by roughly ~150kb (assuming you only use the Langium parser).

Yokozuna59:
> @Yokozuna59 Is this still interesting for you guys over at Mermaid?

@msujew Yes, we are currently bumping a major release (`v11`), which should consist of merging the existing langium parsers with main steam. Our next plan is to support the flowchart diagram. Since it's the largest and most complex diagram we have, if it is able to replace the old parser, then the other diagrams will be easy. I'm not quite sure what would happen if we weren't able to replace the flowchart diagram, but it would be a problem for sure.

I'm currently waiting a PR to be merged so I can work on the flowchart parser.
 

> This change should be able to reduce the bundled dependencies by roughly ~150kb (assuming you only use the Langium parser).

Awesome! I'm working on the LS on a separate repo (though I have been less active on GitHub and coding for some time).

sailingKieler:
Hey @msujew,

I added another commit contributing [packages/langium/src/utils/cancellation.ts](https://github.com/eclipse-langium/langium/commit/81682ca4026be27a60446693da6a3ada09922c12#diff-09bbb7284b44bf9e06f3bfbbf7c9a318dd3295ba1c6fe41148333900b6df7461) just containing
`export * from 'vscode-jsonrpc/lib/common/cancellation.js';` and updating all related imports.

I added it to `utils/index.ts` as a named export similar to the `AstUtils` to not pollute Langium's namespace further.

Technically, the results will be of the same size as with copying the code.
How likely is it that `vscode-jsonrpc` will also get node16-exports in near future? (like `vscode-languageserver-types` has, for example). Should I add an export of `vscode-jsonrpc/lib/common/event.js`, too?
What do you think about that?

sailingKieler:
I did another refinement, now there's no `import ... from 'vscode-jsonrpc';` in the code base anymore, except for
`export * from 'vscode-jsonrpc/lib/common/cancellation.js';` in `utils/cancellation.ts`, as mentioned above.

I also double-checked on exporting `Event` (see above), but that doesn't bring any value IMO, so I discarded that.

I have one more commit in the pipeline replacing all type imports of `vscode-languageserver` by `vscode-languageserver-types` or `vscode-languageserver-protocol`, but that would be too much for this PR I think.

msujew:
@sailingKieler

> How likely is it that vscode-jsonrpc will also get node16-exports in near future? (like vscode-languageserver-types has, for example). Should I add an export of vscode-jsonrpc/lib/common/event.js, too?

Good question - no idea. I assume at the same time that vscode switches over to ESM, which might take a while. I think the new approach is just fine.

packages/langium/src/workspace/workspace-manager.ts
Lotes: @montymxb Why is this resolved?

sailingKieler: We had an internal discussion on that.

Lotes:
Why was `InitializableService` reintroduced? Is it not better to register an event listener? Adding super types to make this happen is a solution. But I would prefer avoiding changes on the type hierarchy, since is already doable with listeners.

Or do I miss something?

packages/langium/src/workspace/workspace-manager.ts
msujew: See https://github.com/eclipse-langium/langium/pull/1258#discussion_r1461918831

sailingKieler:
> Why was `InitializableService` reintroduced? Is it not better to register an event listener? Adding super types to make this happen is a solution. But I would prefer avoiding changes on the type hierarchy, since is already doable with listeners.
> 
> Or do I miss something?

As you can see in https://github.com/eclipse-langium/langium/pull/1258/files#diff-9ef9cc03a159c7e21b4350faa99ddd28f97e4f27c934a51508e14531e4f4987bL49-L58 the `DefaultWorkspaceManager` (a core service) access the `LanguageServer` -- an LSP service --, which is to be avoided.
Same was the case for the `DefaultConfigurationProvider`. Moreover, the latter relied on API of `vscode-languageserver` that is also to be avoided.

sidharthv96:
Are there any migration docs in the works, that I can refer to and try to make changes in our parser?

sailingKieler:
> Are there any migration docs in the works, that I can refer to and try to make changes in our parser? Or an example of how we can get from a string to AST without LSP. I think the current examples rely on LSP.

@sidharthv96 the changes of this will mainly require you to update/refactor `import`s within your custom code, and to re-generate the generated code with the `langium cli`.

In addition, to take advantage of this restructuring, you have to rework the preparations in [your langium config module](https://github.com/mermaid-js/mermaid/blob/b7c72cb38ab953cad4f405345a2be1db3bdfe41b/packages/parser/src/language/info/module.ts#L57-L72), precisely replace the calls of `createDefault[Shared]Module` by `createDefault[Shared]CoreModule`, and adapt the imports accordingly. That's basically it wrt. to this PR.

However, the changes of #1320 will most likely induce more manual migration effort.
For sure, there will be some migration guide.

packages/langium/src/workspace/documents.ts
```diff
@@ -4,18 +4,22 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { Diagnostic, Range, TextDocuments } from 'vscode-languageserver';
-import { CancellationToken } from 'vscode-languageserver';
+// re-export 'TextDocument' & friends from 'vscode-languageserver-textdocument' for convenience,
+//  the overhead is very small, just a few kilobytes
+export * from 'vscode-languageserver-textdocument';
```
spoenemann, line 9: What's the reason for reexporting this dependency here?

spoenemann (commented):
This is now in a really good shape, good work! 💪 

The [poll results](https://github.com/eclipse-langium/langium/discussions/1346) don't suggest that this effort is needed: at the time of writing, none of the 26 participants selected "Parsing only". However, we have statements from @sidharthv96 (https://github.com/eclipse-langium/langium/pull/1258#issuecomment-1907410628) and @jindong-zhannng (https://github.com/eclipse-langium/langium/discussions/1346#discussioncomment-8203486) confirming that there are at least some applications where bundle size is crucial. I see this as a sufficient sign that we can proceed, provided that the changes are not too disruptive for other users.

I think we found good solutions to avoid raising the overall complexity for "normal" Langium users. We still need to write a migration guide, but in most cases it's just about updating imports (which is necessary for other v3 changes, too).

@sailingKieler please have a look at my questions and remarks below. In addition, I found imports from `'vscode-languageserver'` in the following source files, which should be changed to `type` imports from `'vscode-languageserver-types'`:
 - [ ] ast-utils.ts
 - [ ] cst-utils.ts
 - [ ] jsdoc.ts

packages/langium/src/workspace/documents.ts
spoenemann: I'm in favor of `TextDocumentProvider` too: it describes better what that type is about.

examples/domainmodel/src/language-server/domain-model-scope.ts
```diff
@@ -24,7 +23,7 @@ export class DomainModelScopeComputation extends DefaultScopeComputation {
     /**
      * Exports only types (`DataType or `Entity`) with their qualified names.
      */
-    override async computeExports(document: LangiumDocument, cancelToken = CancellationToken.None): Promise<AstNodeDescription[]> {
+    override async computeExports(document: LangiumDocument, cancelToken = Cancellation.CancellationToken.None): Promise<AstNodeDescription[]> {
```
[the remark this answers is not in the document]
sailingKieler: Exactly! `vscode-jsonrpc` does not export an ESM version of it's content (yet) that can be properly tree-shaken. Therefore, I added the following module re-exporting only the cancellation parts: https://github.com/eclipse-langium/langium/pull/1258/files#diff-09bbb7284b44bf9e06f3bfbbf7c9a318dd3295ba1c6fe41148333900b6df7461

Effectively this is like copying over the code.

packages/langium/src/default-module.ts
```diff
@@ -50,19 +33,20 @@ import { DefaultCommentProvider } from './documentation/comment-provider.js';
 import { LangiumParserErrorMessageProvider } from './parser/langium-parser.js';
 import { DefaultAsyncParser } from './parser/async-parser.js';
 import { DefaultWorkspaceLock } from './workspace/workspace-lock.js';
+import { DefaultConfigurationProvider } from './index.js';
```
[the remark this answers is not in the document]
sailingKieler: Good catch! 🙏

packages/langium/src/lsp/language-server.ts
```diff
@@ -46,6 +47,13 @@ export interface LanguageServer {
     onInitialized(callback: (params: InitializedParams) => void): Disposable
 }
 
+/**
+ * Language-specific core and optional LSP services.
+ * To be used while accessing the language-specific services via the service registry without a-priori knowledge about the presence of LSP services for the particular languages.
+ * Shared services should be accessed via the language server's `services` property.
+ */
+export type LangiumCoreAndPartialLSPServices = Omit<LangiumCoreServices & PartialLangiumLSPServices, 'shared'>
```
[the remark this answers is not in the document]
sailingKieler: I also tend to keep it internal. However, it's used in some APIs like
https://github.com/eclipse-langium/langium/pull/1258/files#diff-3216158ebf263e0a3ffbc838b9c1ba3f8419c253eff80fad2854b6cede80c5efR95-R98
and
https://github.com/eclipse-langium/langium/pull/1258/files#diff-3216158ebf263e0a3ffbc838b9c1ba3f8419c253eff80fad2854b6cede80c5efR582-R601.
Exposing those functions but keeping the referenced argument type internal feels weird to me.

packages/langium/src/validation/document-validator.ts
```diff
@@ -8,11 +8,12 @@ import type { MismatchedTokenException } from 'chevrotain';
 import type { Diagnostic } from 'vscode-languageserver';
 import type { LanguageMetaData } from '../languages/language-meta-data.js';
 import type { ParseResult } from '../parser/langium-parser.js';
-import type { LangiumServices } from '../services.js';
+import type { LangiumCoreServices } from '../services.js';
 import type { AstNode, CstNode } from '../syntax-tree.js';
 import type { LangiumDocument } from '../workspace/documents.js';
 import type { DiagnosticData, DiagnosticInfo, ValidationAcceptor, ValidationCategory, ValidationRegistry } from './validation-registry.js';
-import { CancellationToken, DiagnosticSeverity, Position, Range } from 'vscode-languageserver';
+import { DiagnosticSeverity, Position, Range } from 'vscode-languageserver';
```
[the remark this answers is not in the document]
sailingKieler: I'll improve the `import`. But do we need to copy the `create` functions?
Note that `vscode-languageserver-types` provides an ESM/Node16-based export and I would assume that bundlers can properly tree-shake unused symbols, see https://github.com/microsoft/vscode-languageserver-node/blob/4f782ceac1b4444d335a32561bda0ded305c401e/types/package.json#L17-L23

packages/langium/src/workspace/documents.ts
sailingKieler: As soon as one has to customize the document creation/loading stuff clients have to deal the `TextDocument` API.
The import from `vscode-languageserver-textdocument` always felt quite alien to me, especially since `TextDocument` is usually the only thing to be taken from there.
Therefore, I advocate to delegate the issue of from where to import the correct stuff to Langium's API (the foot print is minimal, and `...server-textdocument` provides an ESM/Node16 export).

However, I would be in for wrapping the content into a named export to not pollute Langium's API surface with symbols like `Position`, `Range`, and `TextEdit`.

packages/langium/src/validation/validation-registry.ts
```diff
@@ -4,8 +4,9 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { CancellationToken, CodeDescription, DiagnosticRelatedInformation, DiagnosticTag, integer, Range } from 'vscode-languageserver';
-import type { LangiumServices } from '../services.js';
+import type { CodeDescription, DiagnosticRelatedInformation, DiagnosticTag, integer, Range } from 'vscode-languageserver';
```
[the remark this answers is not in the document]
sailingKieler: done.

packages/langium/src/default-module.ts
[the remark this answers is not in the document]
sailingKieler: done.

packages/langium/src/workspace/documents.ts
sailingKieler: done!

packages/langium/src/validation/document-validator.ts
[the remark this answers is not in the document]
spoenemann: There's no need to copy the functions. It's just JSON data, it can be created with plain objects.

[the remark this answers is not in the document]
sailingKieler: Okay, got it.
In addition to the type related functions, `DiagnosticSeverity` constants are referenced.
Do you prefer copying over those constant definitions (with primitive values) from https://github.com/microsoft/vscode-languageserver-node/blob/release/types/3.17.5/types/src/main.ts#L617-L639

packages/langium/src/workspace/documents.ts
spoenemann: I'm not sure whether reexporting really helps to avoid confusion: it creates one more package from which users can choose to import, and they might wonder whether the definitions are compatible.

In any case, I would avoid reexporting `*` from a foreign package: they might introduce stuff that we don't want to have in our API. Please list the exported symbols explicitly. Is it more than just `TextDocument`?

packages/langium/src/validation/document-validator.ts
```diff
@@ -232,13 +233,13 @@ export function getDiagnosticRange<N extends AstNode>(info: DiagnosticInfo<N, st
 export function toDiagnosticSeverity(severity: 'error' | 'warning' | 'info' | 'hint'): DiagnosticSeverity {
     switch (severity) {
         case 'error':
-            return DiagnosticSeverity.Error;
+            return 1; // according to vscode-languageserver-types/lib/esm/main.js#DiagnosticSeverity.Information
```
sailingKieler, line 236: I now replaced the constants like `DiagnosticSeverity.Information` from `vscode-languageserver-types` by their values, and left an info on their original definition.

packages/langium/src/validation/document-validator.ts
```diff
@@ -232,13 +233,13 @@ export function getDiagnosticRange<N extends AstNode>(info: DiagnosticInfo<N, st
 export function toDiagnosticSeverity(severity: 'error' | 'warning' | 'info' | 'hint'): DiagnosticSeverity {
     switch (severity) {
         case 'error':
-            return DiagnosticSeverity.Error;
+            return 1; // according to vscode-languageserver-types/lib/esm/main.js#DiagnosticSeverity.Error
```
sailingKieler, line 236: I now replaced the constants like `DiagnosticSeverity.Information` from `vscode-languageserver-types` by their values, and left an info on their original definition.

msujew: Not a big fan of this - users that don't want to include `vscode-languageserver-types` in their build, are now only allowed to use the numbers themselves (or have to redeclare the enum themselves). We should offer the enum ourselves.

sailingKieler: Actually this peace of code is only used to convert the Langium notion of `severity` being defined as `'error' | 'warning' | 'info' | 'hint'` into the LSP notion while composing LSP diagnostics, right?
Our code is just converting the values according to the protocol definition, and therefore I don't see why Langium should export copies of those definitions.
Do you have concrete use cases in mind? Why should users care about those constants?

msujew: Ah, right I forgot. Alright, that's fine then 👍

spoenemann (approved):
Let's ship it! 🚀

packages/langium/src/services.ts
```diff
@@ -35,42 +26,33 @@ import type { ScopeProvider } from './references/scope-provider.js';
...
+import type { WorkspaceManager } from './workspace/workspace-manager.js';
+
+export type {
+    InitializeParams, InitializedParams
+};
+
+export interface InitializableService {
+    initialize(params: InitializeParams): void
+    initialized(params: InitializedParams): MaybePromise<void>
+}
```
spoenemann, line 49: Looking at this again, I'd rather remove this interface: we already have `onInitialize` and `onInitialized` events in the LanguageServer service, which are the recommended ways to hook into the LSP lifecycle. The interface is only used by two core framework classes which require to invert the dependency. The interface is not necessary because LanguageServer hard-codes the calls to these classes anyway.

packages/langium/src/service-registry.ts
```diff
@@ -4,38 +4,41 @@
...
      */
-    register(language: LangiumServices): void;
+    register(language: LangiumCoreServices): void;
 
     /**
      * Retrieve the language-specific services for the given URI. In case only one language is
      * registered, it may be used regardless of the URI format.
      */
-    getServices(uri: URI): LangiumServices;
+    getServices(uri: URI): LangiumCoreServices;
```
Yokozuna59, line 25: How can I access the LSP services if this method only returns `LangiumCoreServices`?

I'm currently trying to run two language servers in the same document, one of which is the Langium language server and the other not.
The Langium language server is the default one, and if the user typed some specific text, it would use the other language server suggestions.

Anyway, the approach I'm taking is to create two modes; according to mode, the language server would use either the language server or the other language server methods to suggest, but now I can't use the language server since I can't directly access the LSP services.


Here is an example:

```typescript
export class LangiumMode implements LanguageMode {
  private getServices({ uri }: TextDocument): LangiumServices /* error in 3.0.0 */ { 
    return this.shared.ServiceRegistry.getServices(URI.parse(uri));
  }

  public doComplete(
    document: TextDocument,
    params: CompletionParams,
    cancelToken?: CancellationToken | undefined,
  ): MaybePromise<CompletionList | undefined> {
    return this.getServices(document).lsp.CompletionProvider?.getCompletion(
      this.textDocumentToLangiumDocument(document),
      params,
      cancelToken,
    );
  }
  
  // ...
}
```

Yokozuna59 (commented):
Hi there, thanks for your work!

I'm currently facing an issue regarding the `ServiceRegistry#getServices`, see the comment below.

packages/langium/src/service-registry.ts
cdietrich: for me it worked to downcast

sailingKieler: @Yokozuna59  A quick answer: Look https://github.com/eclipse-langium/langium/issues/1166#issuecomment-1945711393 and below.

Yokozuna59: I see, thanks @cdietrich @sailingKieler.

Issue #1262, opened 2023-11-03
Parsing EOF leads to invalid CST ranges


msujew:
Langium version: 2.1.0
Package name: `langium`

## Steps To Reproduce

1. Create a grammar that uses the new `EOF` token.
2. Assert that subrules using this token never get assigned

## The current behavior

The EOF token result in an invalid CST tree where properties like `length` are `NaN`

## The expected behavior

EOF tokens are generally ignored in the CST computation.

PR #1263, opened 2023-11-04
Check consumed parser tokens for validity

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1262

This is effectively a second (but better) attempt for https://github.com/eclipse-langium/langium/pull/1135. This change validates any parser token before putting it in the CST and AST. Invalid tokens are simply ignored after consuming them from the token stack.

This prevents the issue of invalid CST ranges (and therefore issues in LSP and value assignments) completely.


msujew:
cc @Yokozuna59 This should fix the EOF issue you were experiencing.

spoenemann (approved):
👍

Issue #1266, opened 2023-11-04
Inconsistent behavior in the completion provider service


theogiraudet:
The context node given to the `completionFor` function of the `DefaultCompletionProvider` is inconsistent.

Langium version: 2.1.0
Package name: `langium`.

## Steps To Reproduce

1. Use this grammar:
```
grammar MinExample

entry Model:
    '{' variables+=Variable* '}';

Variable:
    'let' id=ID ':' type=Type ('=' value=Value)? ';';

Type: {IntType} 'int' | {BoolType} 'bool';

Value: IntValue | BoolValue;

IntValue:
    value=INT;

BoolValue:
    value?='true' | 'false';

interface IntType {}
interface BoolType {}

hidden terminal WS: /\s+/;
terminal ID: /[_a-zA-Z][\w_]*/;
terminal INT returns number: /[0-9]+/;

hidden terminal ML_COMMENT: /\/\*[\s\S]*?\*\//;
hidden terminal SL_COMMENT: /\/\/[^\n\r]*/;
```
3. Override the default completion provider with this:
```ts
import { CompletionAcceptor, CompletionContext, DefaultCompletionProvider, GrammarAST, MaybePromise, NextFeature } from "langium";
import { BoolType, isVariable } from "./language/generated/ast.js";

export class MinExampleCompletionProvider extends DefaultCompletionProvider { 
    override completionFor(context: CompletionContext, next: NextFeature, acceptor: CompletionAcceptor): MaybePromise<void> {
        console.log(context.node?.$type)
    }
}
```
4. Observe the value printed in the debug for these two programs:
```
{
let foo: int = 0;
let bar: int = <trigger completion here>
}
```
The value displayed is "Model".
```
{
let foo: int = 0;
let bar: int = <trigger completion here>
// (or any other valid line you want)
}
```
The value displayed is "Variable".

Link to code example:

You can also find the example in this repository: https://github.com/theogiraudet/Langium-Completion-Bug-Minimal-Example.

## The current behavior

The value displayed for the two programs is different even if these two programs are semantically identical.
This behavior is disturbing when we want to change the default behavior of the completion provider (in my case, I want to disable the keyword completion when the type of the current variable is `int`).
The bug isn't present when I remove the bracket tokens from the grammar.

This bug seems to be really close to #1216.

## The expected behavior

The two context nodes are identical.

PR #1267, opened 2023-11-05
Improve completion AST node accuracy

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1266

Improves the accuracy with which the completion provider picks the AST node:

1. Fixes a very minor issue in the `findLeafNodeAtOffset` method, which sometimes gave back a CST node that appeared after the specified offset.
2. Only pick one AST node, ideally at the position before the current completion position.

Also reduces the amount of false-positive completion proposals by implementing a `performNextTokenCompletion` method. In cases where we definitely first want to complete the current ID/word/keyword/etc, the completion provider should not yield more completion contexts. This method controls that behavior by checking whether we have a word at the current offset position.

### DIFF
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -248,9 +248,14 @@ export class DefaultCompletionProvider implements CompletionProvider {
         }
         // For all other purposes, it's enough to jump to the start of the current/previous token
         const { nextTokenStart, nextTokenEnd, previousTokenStart, previousTokenEnd } = this.backtrackToAnyToken(text, offset);
-        let astNode: AstNode | undefined;
+        let astNodeOffset = nextTokenStart;
+        if (offset <= nextTokenStart && previousTokenStart !== undefined) {
+            // This check indicates that the cursor is still before the next token, so we should use the previous AST node (if it exists)
+            astNodeOffset = previousTokenStart;
+        }
+        const astNode = findLeafNodeAtOffset(cst, astNodeOffset)?.astNode;
+        let performNextCompletion = true;
         if (previousTokenStart !== undefined && previousTokenEnd !== undefined && previousTokenEnd === offset) {
-            astNode = findLeafNodeAtOffset(cst, previousTokenStart)?.astNode;
             // This context aims to complete the current feature
             yield {
                 ...partialContext,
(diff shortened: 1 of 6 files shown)


packages/langium/src/utils/cst-util.ts
```diff
@@ -155,30 +155,53 @@ export function isCommentNode(cstNode: CstNode, commentNames: string[]): boolean
...
-                firstChild = middleChild + 1;
-            } else {
-                return findLeafNodeAtOffset(n, offset);
-            }
-        }
-        if (firstChild === lastChild) {
-            return findLeafNodeAtOffset(node.content[firstChild], offset);
+        const searchResult = binarySearch(node, offset);
+        if (searchResult) {
+            return findLeafNodeAtOffset(searchResult, offset);
```
sailingKieler, line 172: Code looks very good to me, I'm just surprised for the new test in [cst-utils.test.ts](https://github.com/eclipse-langium/langium/pull/1267/files#diff-5d9ca30c93ce56e1c0049557d3bf3df7e9f66ee91a290239f3baaf036efc325f) it runs into a recursion of depth 12 😳 
Would it make sense to add a special treatment for case of `node.content.length === 1`? 

image

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -296,6 +309,12 @@ export class DefaultCompletionProvider implements CompletionProvider {
         }
     }
 
+    protected performNextTokenCompletion(document: LangiumDocument, text: string, _offset: number, _end: number): boolean {
+        // This regex returns false if the text ends with a letter
+        // We don't want to complete new text immediately after a keyword, ID etc.
+        return /\P{L}$/u.test(text);
```
sailingKieler, line 315: Please add a comment on the meaning of the regEx, and motivate why not `/\w$/`,
and that the pattern is expected to care only for the last character within `text` (although it's implicitly mentioned within the first comment line)

sailingKieler (approved):
I'm basically fine with theses changes, and I would happy if my remarks would somehow make it into the code 😉

packages/langium/src/utils/cst-util.ts
msujew: The recursion depth is exactly the same depth as the CST depth at the specific position, since it's a tree based recursion algorithm. We could probably improve that by making the function work iteratively instead of recursively. Though that might be something for a separate PR.

> Would it make sense to add a special treatment for case of node.content.length === 1?

I've tried that just now, but the amount of comparisons we need is actually larger than without it as we have a lot of cases where `node.content.length > 1`. It only makes the method more verbose in my opinion.

sailingKieler: >  It only makes the method more verbose in my opinion.
That's probably true.

I just though that this method is a prominent one and performance could be a relevant topic in case of giant input files. The recursion depth is maybe not reducible, but calling `binarySearch` and doing one or two iterations in its loop makes little sense in case there's just a single child.

msujew: Performance is relevant, but the amount of comparisons we're doing for n = 1 is exactly the same if we put a `if (n = 1)` in the front of the method, or whether we just let the method run into the loop. In fact, it leads to more comparison due to the new `if (n = 1)` at the start of the method. We still need to perform the "in bounds" checking of the child node even if its the only child.

sailingKieler: Makes sense.
> We still need to perform the "in bounds" checking of the child node even if its the only child.

I wasn't sure on which assumptions we can rely at this point.

PR #1269, opened 2023-11-06
Added validate-exports script to check the consistency of package exports

### LINKED ISSUE #1261: Enforce isolation of package exports
It would be great to enforce that the main export of the `langium` package doesn't have any transitive dependencies to code from `test`, `grammar` (#1171) and `lsp` (#1258).

Can we write a test to detect such unwanted dependencies?

spoenemann:
Fixes #1261.

#1171 and #1258 should be adapted to include additional exports validation and update the exclude paths of main.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -29,6 +29,7 @@ jobs:
       run: |
         npm ci
         npm run build:clean
+        npm run validate-exports
     - name: Test
       if: success() || failure()
       shell: bash
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "test": "vitest",
     "test-ui": "vitest --ui",
     "coverage": "vitest run --coverage",
+    "validate-exports": "npm run validate-exports --workspace=langium",
     "version:dependencies": "node ./scripts/update-version.js && npm install",
     "langium:generate": "npm run langium:generate --workspace=langium --workspace=examples/domainmodel --workspace=examples/arithmetics --workspace=examples/statemachine --workspace=examples/requirements",
     "dev-build": "npm run dev-clean && npm install && npm link ./packages/langium && npm link ./packages/langium-cli && npm link ./packages/generator-langium",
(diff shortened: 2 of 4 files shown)


packages/langium/test/package-main.tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": "../tsconfig.src.json",
+    "compilerOptions": {
+      "noEmit": true
+    },
+    "exclude": [
+        "../src/test/**/*"
+    ]
+}
```
sailingKieler, line 9: That's very smart 🎩

I'm just confused by the name and location of the file. What about
`tsconfig.src-main.json`,  `tsconfig.src-main-only.json`, `tsconfig.src-plain.json`, `tsconfig.src-plain-only.json`? (to have it inline with the other `tsconfig`s).

I would love to see a comment like
`// only used validating to not import modules from 'src/test', ... within base modules like 'src/parser/*'
in line 1 😉

sailingKieler (approved):

packages/langium/test/package-main.tsconfig.json
spoenemann: > I'm just confused by the name and location of the file. What about
tsconfig.src-main.json, tsconfig.src-main-only.json, tsconfig.src-plain.json, tsconfig.src-plain-only.json? (to have it in line with the other tsconfigs).

I'd prefer to keep it separate to highlight that this is part of the framework tests and that it's not used for building. Having too many tsconfigs next to each other could be confusing.

> I would love to see a comment like `// only used for validating to not import modules from 'src/test', ... within base modules like 'src/parser/*' in line 1 😉

I got an error in the JSON editor when I added a comment (comments are not part of the standard JSON syntax).

sailingKieler: That's another pro arg for renaming, as `tsconfig. ... .json` files are handling differently. With such a name, comments are permitted, see
image

spoenemann: Ok, renamed and added a comment. I still think it's better to keep the file in a subfolder to avoid confusion.

sailingKieler: 🎉

PR #1276, opened 2023-11-08
Remove explicit EOF token from token builder

msujew:
Chevrotain actually doesn't require to put the EOF token in the lexer definition. The EOF token is always implicitly part of the token stream.

Instead, having it in there might lead to issues instead. See also https://github.com/eclipse-langium/langium/discussions/1048#discussioncomment-7503713.

### DIFF
--- a/packages/langium/src/parser/token-builder.ts
+++ b/packages/langium/src/parser/token-builder.ts
@@ -38,6 +38,8 @@ export class DefaultTokenBuilder implements TokenBuilder {
                 tokens.push(terminalToken);
             }
         });
+        // We don't need to add the EOF token explicitly.
+        // It is automatically available at the end of the token stream.
         return tokens;
     }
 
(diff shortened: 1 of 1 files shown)


packages/langium/src/parser/token-builder.ts
```diff
@@ -38,11 +38,6 @@ export class DefaultTokenBuilder implements TokenBuilder {
                 tokens.push(terminalToken);
             }
         });
-
-        //reminder: EOF should always be the last token, because it is very unlikely that it will be matched within the input
-        if (reachableRules.some(r => streamAst(r.definition).some(isEndOfFile))) {
-            tokens.push(EOF);
-        }
```
sailingKieler, line 45: I think this is worth a comment, maybe with a reference to a statement in chevrotain docs or code.
Otherwise, nobody except you will have a clue why it is like it is.

sailingKieler (approved):
👍

PR #1278, opened 2023-11-08
Add type hierarchy provider as Language Server capability

### LINKED ISSUE #1275: Type hierarchy provider
LSP 3.17 added a way to get the subtypes/supertypes of an entity, based on three requests:
* [textDocument/prepareTypeHierarchy](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareTypeHierarchy)
* [typeHierarchy/supertypes](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#typeHierarchy_supertypes)
* [typeHierarchy/subtypes](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#typeHierarchy_subtypes)

It would be nice to be able to easily hook into this, similar to the `CallHierarchyProvider`.

I'm currently working on this for a Langium-based DSL:
* [TypeHierarchyProvider](https://github.com/Safe-DS/DSL/blob/bdd3116c7e4d3a4da15006b7ed2822826b39a621/packages/safe-ds-lang/src/language/lsp/safe-ds-type-hierarchy-provider.ts#L27C21-L27C21)
* [AbstractTypeHierarchyProvider](https://github.com/Safe-DS/DSL/blob/bdd3116c7e4d3a4da15006b7ed2822826b39a621/packages/safe-ds-lang/src/language/lsp/safe-ds-type-hierarchy-provider.ts#L39C23-L39C52)
* [request handler](https://github.com/Safe-DS/DSL/blob/bdd3116c7e4d3a4da15006b7ed2822826b39a621/packages/safe-ds-lang/src/language/lsp/safe-ds-language-server.ts#L123)

The interface is basically a copy of the `CallHierarchyProvider`. If you'd accept such a PR, I'd like to move the interface, the no-op implementation, and the request handler up into Langium.

lars-reimann:
Closes #1275

* Handle type hierarchy requests in the Language Server
* Add `TypeHierarchyProvider` as a new LSP service
* Add the interface `TypeHierarchyProvider` and a no-op abstract implementation `AbstractTypeHierarchyProvider` to easily customize the responses for a given language
* Add a type hierarchy provider for interfaces in Langium grammars

**Test instructions:**

Some automated tests are included. To check functionality in VS Code, create a Langium grammar with the following content:

```ts
interface A {}
interface B extends A {}
```

Place the cursor on the name of either interface and invoke the action "Types: Show Type Hierarchy". Via the button 
image
you can switch between showing subtypes and supertypes of the selected interface.


packages/langium/src/lsp/language-server.ts
```diff
@@ -519,6 +526,47 @@ export function createCallHierarchyRequestHandler<P extends CallHierarchyIncomin
     };
 }
 
+export function addTypeHierarchyHandler(connection: Connection, sharedServices: LangiumSharedServices): void {
+    connection.languages.typeHierarchy.onPrepare(
```
spoenemann, line 530: Here we could check whether there is a language with a type hierarchy provider before we register the handlers.

packages/langium/src/lsp/type-hierarchy-provider.ts
```diff
@@ -0,0 +1,134 @@
...
+                name,
+                range: targetNode.$cstNode.range,
+                selectionRange: nameNode.range,
+                uri: document.uri.toString(),
+                ...this.getTypeHierarchyItem(targetNode),
+            },
+        ];
+    }
+
+    protected getTypeHierarchyItem(_targetNode: AstNode): Partial<TypeHierarchyItem> | undefined {
```
spoenemann, line 89: Please add a comment about how this template method is used.

packages/langium/src/lsp/type-hierarchy-provider.ts
```diff
@@ -0,0 +1,134 @@
...
+        const targetNode = findDeclarationNodeAtOffset(
+            rootNode.$cstNode,
+            document.textDocument.offsetAt(params.item.range.start),
+            this.grammarConfig.nameRegexp,
+        );
+        if (!targetNode) {
+            return undefined;
+        }
+
+        const references = this.references.findReferences(targetNode.astNode, {
```
spoenemann, line 124: It's true that type hierarchies are usually defined with references from the subtypes to the supertypes, but you could also define a language where it's the other way around. I suggest moving this call to the specific implementation instead of making assumptions whether you might need it.

spoenemann (commented):
Excellent contribution! 🎉 

I have only a few remarks:

lars-reimann:
Thanks for the feedback. I've adjusted the PR accordingly.

spoenemann (approved):
Thanks for the contribution!

PR #1284, opened 2023-11-09
Flush semantic token builder cache on request

msujew:
Related to https://github.com/eclipse-langium/langium/discussions/1283

I wasn't able to reproduce this - it might only appear in the monaco-vscode-api package, which probably slightly differs in how vscode invokes the semantic token provider.

Anyway, flushing the cache when we perform a non-delta updates seems pretty reasonable on my side. The main issue was that the semantic token builder from the `vscode-languageserver` package is pretty undocumented and gives little help on how to use it.

It also seems to fix the issue that @cdietrich was experiencing with a web based Langium LSP worker.

### DIFF
--- a/packages/langium/src/lsp/semantic-token-provider.ts
+++ b/packages/langium/src/lsp/semantic-token-provider.ts
@@ -205,6 +205,7 @@ export abstract class AbstractSemanticTokenProvider implements SemanticTokenProv
         this.currentRange = undefined;
         this.currentDocument = document;
         this.currentTokensBuilder = this.getDocumentTokensBuilder(document);
+        // Flush the current state - we don't want to perform a delta highlighting
         this.currentTokensBuilder.previousResult(this.currentTokensBuilder.id);
         await this.computeHighlighting(document, this.createAcceptor(), cancelToken);
         return this.currentTokensBuilder.build();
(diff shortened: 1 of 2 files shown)


cdietrich:
@msujew this is quite easy to reproduce.
after yo langium gen just adjust classic like this

```
editorOptions: {
                    'semanticHighlighting.enabled': true,
                    //theme: 'vs-dark'
                }
```

@kaisalmen is it intentional that with theme delta highlighting is done, otherwise full?

packages/langium/src/lsp/semantic-token-provider.ts
```diff
@@ -205,6 +205,8 @@ export abstract class AbstractSemanticTokenProvider implements SemanticTokenProv
         this.currentRange = undefined;
         this.currentDocument = document;
         this.currentTokensBuilder = this.getDocumentTokensBuilder(document);
+        // Flush the current state - we don't want to perform a delta highlighting
+        this.currentTokensBuilder.previousResult(this.currentTokensBuilder.id);
```
Lotes, line 209: This line looks cryptic to me, because I have not so much knowledge about Langium. Also the comment isn't really helpful.

Could you add a FLUSH method to the token builder instead? All what is needed comes from the token builder, so I guess it could be a good idea.

Lotes (approved):

Issue #1230, opened 2023-10-18
Langium documents don't get removed after the containing directory is renamed


lars-reimann:
After renaming a directory, all DSL files contained inside are listed twice in `services.shared.workspace.LangiumDocuments`. This can affect various downstream features, like linking, type computation, or validation.

Langium version: 2.0.2
Package name: langium

## Steps To Reproduce

1. In a Langium project, observe the contents of `services.shared.workspace.LangiumDocuments`. In the repo linked below, it gets logged in the output panel.
2. Create a folder.
3. Create a DSL file inside (extension `.hello` for the repo linked below).
4. Rename the folder.

Link to code example: https://github.com/lars-reimann/langium-cst-range

## The current behavior

Before renaming the directory, the document is included (correct):
![image](https://github.com/eclipse-langium/langium/assets/2501322/6b3c91fc-c2fb-4396-84af-868b0838f3de)

After renaming (third array), the new document is shown (correct), but the old one is not deleted (wrong):
![image](https://github.com/eclipse-langium/langium/assets/2501322/65c1a511-90fb-461b-b2f2-cc5f8d13138a)

## The expected behavior

Only the new document should be included.

msujew:
@lars-reimann Thanks for the report. I believe this is part of a larger issue where we don't actually listen to a few of the workspace related LSP notifications, mainly:

* `workspace/didChangeWorkspaceFolders`
* `workspace/willRenameFiles`
* `workspace/didRenameFiles`
* `workspace/willCreateFiles`
* `workspace/didCreateFiles`
* `workspace/willDeleteFiles`
* `workspace/didDeleteFiles`

Some of these are not interesting for the default behavior of Langium, but we should definitely start handling the `didChangeWorkspaceFolders`, `didRenameFiles`, `didCreateFiles` and `didDeleteFiles` notifications.

spoenemann:
I had a closer look at this. We already get notifications of most changes through `workspace.didChangeWatchedFiles` notifications. So the only case where we need additional notifications are when folders are renamed, as those are not covered by the glob pattern `**/*.hello`. We could do this by explicitly watching all folders in the language server, but what I don't like about it is that we have a distribution of responsibilities between language client and language server code. It would be better to find a solution where watched file system events are specified fully on the client or server side.

cdietrich:
see also didClose in https://github.com/eclipse-langium/langium/discussions/1281

spoenemann:
[...]

PR #1286, opened 2023-11-09
Introduce DocumentUpdateHandler service

### LINKED ISSUE #1272: Add service for text document handling
Currently, the `language-server.ts` has a function to register a text document handler. This function is hardly overridable, and doesn't allow to query whether certain operations are supported (such as `onWillSave`).

It would be a good idea to extract this handler method into an interface + default implementation that can be used to provide better server capability info and more fine grained text document handling.

spoenemann:
Fixes #1272.

In addition, I stumbled over this statement from the LSP specification, [didChangeWatchedFiles notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles):

> It is recommended that servers register for these file system events using the registration mechanism. In former implementations clients pushed file events without the server actively asking for it.

The default implementation of the new DocumentUpdateHandler service, which listens and reacts to `didChangeWatchedFiles` notifications, takes care of registering a file watcher so this doesn't have to be done by the client. This solves the problem I described in https://github.com/eclipse-langium/langium/issues/1230#issuecomment-1785356096:

> ...what I don't like about it is that we have a distribution of responsibilities between language client and language server code. It would be better to find a solution where watched file system events are specified fully on the client or server side.

We should test this thoroughly, and also check what happens if the client still creates a FileSystemWatcher – does the language server receive the notifications twice? In that case, this is a breaking change that should be well documented in the CHANGELOG.

### DIFF
--- a/packages/langium/src/lsp/language-server.ts
+++ b/packages/langium/src/lsp/language-server.ts
@@ -87,7 +87,7 @@ export class DefaultLanguageServer implements LanguageServer {
 
     protected buildInitializeResult(_params: InitializeParams): InitializeResult {
         const languages = this.services.ServiceRegistry.all;
-        const fileOperationOptions = this.services.lsp.DocumentUpdateHandler.fileOperationOptions;
+        const fileOperationOptions = this.services.lsp.FileOperationHandler?.fileOperationOptions;
         const hasFormattingService = this.hasService(e => e.lsp.Formatter);
         const formattingOnTypeOptions = languages.map(e => e.lsp.Formatter?.formatOnTypeOptions).find(e => Boolean(e));
         const hasCodeActionProvider = this.hasService(e => e.lsp.CodeActionProvider);
@@ -183,6 +183,7 @@ export function startLanguageServer(services: LangiumSharedServices): void {
     }
 
     addDocumentsHandler(connection, services);
+    addFileOperationHandler(connection, services);
     addDiagnosticsHandler(connection, services);
     addCompletionHandler(connection, services);
     addFindReferencesHandler(connection, services);
@@ -231,6 +232,13 @@ export function addDocumentsHandler(connection: Connection, services: LangiumSha
     }
     if (handler.didChangeWatchedFiles) {
         connection.onDidChangeWatchedFiles(params => handler.didChangeWatchedFiles!(params));
+    }
+}
+
+export function addFileOperationHandler(connection: Connection, services: LangiumSharedServices): void {
+    const handler = services.lsp.FileOperationHandler;
+    if (!handler) {
+        return;
     }
     if (handler.didCreateFiles) {
         connection.workspace.onDidCreateFiles(params => handler.didCreateFiles!(params));
--- a/packages/langium/src/lsp/language-server.ts
+++ b/packages/langium/src/lsp/language-server.ts
@@ -30,7 +30,7 @@ import type {
 } from 'vscode-languageserver';
 import type { LangiumServices, LangiumSharedServices } from '../services.js';
 import type { LangiumDocument } from '../workspace/documents.js';
-import { Emitter, FileChangeType, LSPErrorCodes, ResponseError, TextDocumentSyncKind } from 'vscode-languageserver';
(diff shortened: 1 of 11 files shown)


spoenemann:
After rethinking, I realized the workspace file operations shouldn't be used for triggering DocumentBuilder updates because they are called only before / after operations done by the user in the client (VS Code). I extracted those operations into a new service `FileOperationHandler`.

I think issues such as #1230 need to be addressed with the file system watcher, which we can control from the language server after this change.

sailingKieler (commented):
Makes I sense for me.
I'm just wondering whether Langium should bring a handler for `didDeleteFiles` notifications by default, as that might (should) have direct impact on the content of the index, doesn't it? (`didCreateFiles` is probably less critical as newly created files are probably empty initially)

packages/langium/src/lsp/language-server.ts
```diff
@@ -215,30 +218,39 @@ export function startLanguageServer(services: LangiumSharedServices): void {
...
 
 export function addDocumentsHandler(connection: Connection, services: LangiumSharedServices): void {
-    const documentBuilder = services.workspace.DocumentBuilder;
-    const mutex = services.workspace.MutexLock;
-
-    function onDidChange(changed: URI[], deleted: URI[]): void {
-        mutex.lock(token => documentBuilder.update(changed, deleted, token));
+    const handler = services.lsp.DocumentUpdateHandler;
+    const documents = services.workspace.TextDocuments;
+    if (handler.didChangeContent) {
```
sailingKieler, line 223: `didChangeContent` and ` didChangeWatchedFiles` are mandatory according you definitions in `DocumentUpdateHandler`, right? So these guards here and below could be dropped.

sailingKieler:
By the way: I tested the ` didChangeWatchedFiles` functionality in Gitpod with the arithmetics language:
I appended a non-correct string to the example file by executing `echo "7*e;" >> example.calc` in a shell with having `example.calc` both closed and opened, and error markers showed up immediately in both cases.

spoenemann:
> I'm just wondering whether Langium should bring a handler for didDeleteFiles notifications by default, as that might (should) have direct impact on the content of the index, doesn't it? (didCreateFiles is probably less critical as newly created files are probably empty initially)

Those events are covered by the file watcher. The only events we're currently missing are operations on the folder level (#1230).

packages/langium/src/services.ts
```diff
@@ -160,11 +162,13 @@ export type LangiumDefaultSharedServices = {
     ServiceRegistry: ServiceRegistry
     lsp: {
```
Lotes, line 163: Idea: Order by name? Then, we can find services easier. Don't know if that makes sense...

spoenemann:
I don't know what to test here. The default implementation doesn't do much. Regarding documentation, we should mention the new service in https://langium.org/docs/document-lifecycle/

sailingKieler:
@spoenemann
> We should test this thoroughly, and also check what happens if the client still creates a FileSystemWatcher – does the language server receive the notifications twice? In that case, this is a breaking change that should be well documented in the CHANGELOG.

I tested this: The (VS Code) language client indeed sends two independent file change events, see:
image

Wrt. to Langium's current implementation this seems to be fine, mainly because of https://github.com/eclipse-langium/langium/blob/d2cfea2effb0f24f7a5b9383d6935f68016f37ec/packages/langium/src/workspace/document-builder.ts#L167

However, I would prefer to add a deduplication in https://github.com/eclipse-langium/langium/blob/d2cfea2effb0f24f7a5b9383d6935f68016f37ec/packages/langium/src/lsp/document-update-handler.ts#L82-L84 (and document it's purpose clearly).

sailingKieler:
Btw. the same happens when moving files, all fs events are collected and sent via a single notification

image

sailingKieler:
My guess is that in a few weeks/month we'll have forgotten about that detail, and once the area in the document builder is somehow changed, potential issues pop up quickly.

Another option would be to have test enforcing the tolerance of such duplicate URIs s.t. we make sure we can deal such cases.

spoenemann:
Ok I used streams so the deduplication doesn't get too verbose.

sailingKieler (approved):
👍

Issue #1293, opened 2023-11-16
Non-linear performance during code generation


msujew:
As outlined in https://github.com/eclipse-langium/langium/discussions/1292, there seems to be a performance issue in our generator functions.

The generator `toString()` function seems to run in quadratic (`O(n^2)`) time. We should be able to get this down to linear time.

<details>
<summary>Reproducible Example</summary>

```ts
import { CompositeGeneratorNode, NL, toString } from 'langium';

console.log('perfromance test');
const node = new CompositeGeneratorNode();

function prepare(node: CompositeGeneratorNode, l: number, r: number): void {
    if (r<=0) {
        for (let i=0;i<l;i++) {
            node.append(`i=${i} this is a demo line...`, NL);
        }    
    }
    else {
        for (let i=0;i<3;i++) {
            node.append(`open=${i}`, NL);
            node.indent(node=>{
                prepare(node, l, r-1);
            });
            node.append(`close=${i}`, NL);
        }    
    }
}

const t0 = new Date().getTime();
prepare(node, 200, 4); // <-------------------------------- play with the second argument
const t1 = new Date().getTime();


console.log('data prepared');
console.log('convert to text:');
const t2 = new Date().getTime();
const text = toString(node);
const t3 = new Date().getTime();
console.log('print text:');
console.log(text);
console.log(`nb lines: ${text.split('\n').length}`)
console.log(`time to generate text:  ${(t1-t0)/1000} sec`);
console.log(`time to convert text:  ${(t3-t2)/1000} sec`);
console.log('end');
```

</details>

spoenemann:
Fixed with #1294.

PR #1294, opened 2023-11-17
generator support: eliminated performance sink in 'node-processor', solves #1293

sailingKieler:
* added content length tracking to avoid repeated string concatenation under way

### DIFF
--- a/packages/langium/src/generator/node-processor.ts
+++ b/packages/langium/src/generator/node-processor.ts
@@ -22,6 +22,7 @@ class Context {
     private traceData: InternalTraceRegion[] = [];
 
     private lines: string[][] = [[]];
+    private length: number = 0;
 
     constructor(defaultIndent?: string | number) {
         if (typeof defaultIndent === 'string') {
@@ -35,6 +36,10 @@ class Context {
         return this.lines.map(e => e.join('')).join('');
     }
 
+    get contentLength(): number {
+        return this.length;
+    }
+
     get currentLineNumber(): number {
         return this.lines.length - 1;
     }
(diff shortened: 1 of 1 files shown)


packages/langium/src/generator/node-processor.ts
```diff
@@ -35,6 +36,10 @@ class Context {
         return this.lines.map(e => e.join('')).join('');
```
dhuebner, line 36: That why I prefer to have a `get` or `compute` in names to make it clear there are things happening. Where `this.content.length` looks like a simple field access.

dhuebner (approved):

goto40 (approved):
Good work @sailingKieler !!

The patch **fixes the performance effect**:

* Prior to the patch the performance was *very slow* (see #1292) and hat a O(N^2) characteristic regarding the number of generated lines (in presence of indentation).
* Now it seems to be linear (as expected) and much faster.

Timings with my example now:

| nb lines | time in sec to produce input | time for toString |
| :---: | :---: | :---: |
| 3'200'000 | 0.7 | 2.7 |
| 1'600'000 | 0.3 | 1.25 |
| 800'000 | 0.15 | 0.7 |
| 400'000 | 0.1   | 0.4 | 

This is much better than before:

| nb lines | time for toString |
| :---: | :---: |
| 8'000 | 4.0 |
| 16'000 | 16.0 |

goto40:
I can also confirm that the output is identical for the example sketched in #1292:
```
wc -l ~/out_old.txt 
16449 ~/out_old.txt

diff ~/out_old.txt ~/out_new.txt 
16447,16448c16447,16448
< time to generate text:  0.004 sec
< time to convert text:  16.945 sec
---
> time to generate text:  0.006 sec
> time to convert text:  0.037 sec
```

PR #1296, opened 2023-11-20
Add validation against function call recursion #1019

### LINKED ISSUE #1019: Recursion should not be allowed in arithmetics
Currently, in arithmetics you are able to call the function inside of the function recursive. This should not be possible considering conditional statements don't exists.

## Example
```
func(1,2);
def func(t, x):
    func(t, x);
```

example code

When running this the expression returns **null** and the memory usage rises. 

## The expected behavior
Show an error that this is not possible.

hpopov:
Resolves #1019.
Recursion validation implemented using BFS algorithm for nested function calls as graph nodes.

Implemented scenarios:

1. Function calls itself `def fun1(x): fun1(x);`
2. Several functions calls each other in a cycle, e.g.:

```arithmetics
def fun1(x):
    fun2(x) + fun2(x) + fun3(x);

def fun2(x):
    fun3(x) + fun4(x);

def fun3(x):
    3*x;

def fun4(x):
    fun3(x) + fun1(x);
```
Here every affected call in the cycle gets a validation error (fun1::fun2, fun2::fun4, and fun4::fun1). In case a function B is called several times from a function A, only first call instance is marked with error (e.g., only first fun1::fun2).

3. If a Module has multiple recursion cycles, all of them are identified. See example scenarios in `example/recursions.calc`.

### DIFF
--- a/examples/arithmetics/src/language-server/arithmetics-validator.ts
+++ b/examples/arithmetics/src/language-server/arithmetics-validator.ts
@@ -84,46 +84,22 @@ export class ArithmeticsValidator {
         }
 
         const traversedFunctions = new Set<Definition>();
-        function* getNestedCallsIfUnprocessed(func: Definition): Generator<NestedFunctionCall> {
+        function* getNotTraversedNestedCalls(func: Definition): Generator<NestedFunctionCall> {
             if (!traversedFunctions.has(func)) {
                 traversedFunctions.add(func);
                 yield* getNestedCalls(func);
             }
         }
 
-        const callsTree = new Map<ResolvedFunctionCall, NestedFunctionCall>();
-        const getCycle = (to: NestedFunctionCall): [NestedFunctionCall, NestedFunctionCall] | undefined => {
-            const referencedFunc = to.call.func.ref;
-            let parent = callsTree.get(to.call);
-            while (parent) {
-                if (parent.host === referencedFunc) return [parent, to];
-                parent = callsTree.get(parent.call);
-            }
-            return undefined;
-        };
-
-        const callCycles: NestedFunctionCall[][] = [];
-        const printCycle = (cycle: NestedFunctionCall[]): string => {
-            const start = cycle[0];
-            const end = cycle[cycle.length - 1];
-            if (start === end) return printNestedFunctionCall(start);
-            let printedCycle = printNestedFunctionCall(end);
-            let parent = callsTree.get(end.call);
-            while (parent) {
-                printedCycle = printNestedFunctionCall(parent) + '->' + printedCycle;
-                if (parent.call === start.call) break;
-                parent = callsTree.get(parent.call);
-            }
-            return printedCycle;
-        };
-
+        const callsTree: FunctionCallTree = new Map<ResolvedFunctionCall, NestedFunctionCall>();
+        const callCycles: FunctionCallCycle[] = [];
         const bfsStep = (parent: NestedFunctionCall): NestedFunctionCall[] => {
             const referencedFunc = parent.call.func.ref;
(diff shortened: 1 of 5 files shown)


hpopov:
Implements #1019

examples/arithmetics/src/language-server/arithmetics-validator.ts
```diff
@@ -4,21 +4,22 @@
...
+import { isBinaryExpression, isDefinition, isFunctionCall, isNumberLiteral } from './generated/ast.js';
 
 export function registerValidationChecks(services: ArithmeticsServices): void {
     const registry = services.validation.ValidationRegistry;
     const validator = services.validation.ArithmeticsValidator;
     const checks: ValidationChecks<ArithmeticsAstType> = {
         BinaryExpression: validator.checkDivByZero,
-        Definition: validator.checkNormalisable,
-        Module: validator.checkUniqueDefinitions,
+        Definition: [validator.checkUniqueParmeters, validator.checkNormalisable],
```
spoenemann, line 21: Could you fix the type along the way? `checkUniqueParameters`

examples/arithmetics/src/language-server/arithmetics-validator.ts
```diff
@@ -73,6 +72,51 @@ export class ArithmeticsValidator {
...
+                yield* NestedFunctionCall.selectCalls(func);
+            }
+        }
+
+        const callsTree: FunctionCallTree = new Map<ResolvedFunctionCall, NestedFunctionCall>();
+        const callCycles: FunctionCallCycle[] = [];
+        const bfsStep = (parent: NestedFunctionCall): NestedFunctionCall[] => {
+            const referencedFunc = parent.call.func.ref;
+            const uncycledChildren: NestedFunctionCall[] = [];
+            if (parent.host === referencedFunc) callCycles.push([parent]);
```
spoenemann, line 89: I would always write a `{ ... }` block in control structures for consistency and readability.

examples/arithmetics/src/language-server/arithmetics-validator.ts
```diff
@@ -93,4 +137,59 @@ export class ArithmeticsValidator {
...
+            if (isResolvedFunctionCall(expression)) yield { call: expression, host };
+        } else if (isBinaryExpression(expression)) {
+            for (const expr of [expression.left, expression.right]) {
+                if (expr) yield* selectCalls(host, expr);
+            }
+        }
+    }
+
+    export function toString({ host, call }: NestedFunctionCall): string {
+        return `${host.name}::${call.func.ref.name}()`;
```
spoenemann, line 193: This `::` syntax is not used in the DSL, so its meaning could be harder to understand; I would simplify it by removing the "host".

spoenemann (commented):
Thanks for the contribution! I have only a few minor remarks.

examples/arithmetics/src/language-server/arithmetics-validator.ts
hpopov: Fixed that (you probably meant `checkMatchingParameters`)

examples/arithmetics/src/language-server/arithmetics-validator.ts
hpopov: Agree, the meaning can be inferred from the previous call in `fun1()->fun2()->fun3()` cycle.

examples/arithmetics/src/language-server/arithmetics-validator.ts
hpopov: Got it, fixed

spoenemann (approved):
👍

Issue #1282, opened 2023-11-09
Testing facilities generated out of the box 


acherm:
Hi, 

I hope new versions of Langium generator will offer the ability to generate testing facilities out of the box, as part of the DSL project. 
It was a bit the case with Xtext, with an `***/test` package and a `parseHelper` ready to use https://eclipse.dev/Xtext/documentation/103_domainmodelnextsteps.html#tutorial-unit-tests 

As part of my experience with setting such testing facilities (see https://dev.to/diverse_research/testing-your-dsls-in-langium-gp9), I encountered two issues: 
 * lack of understanding of the testing API... there will be a tutorial: https://github.com/eclipse-langium/langium-website/issues/203 and it's super nice!
 * an integrated and ready-to-use solution with `package.json` containing a `test` command, a configuration file of `vite`, and the boilerplate almost described here https://dev.to/lotes/comment/2aiog or in the post
 
Not only DSL developers will enjoy a testing facility out of the box and ready-to-use, but it will enforce best practices, and not forcing developers to dig into the API/documentation. 

I am seeing it as a "feature" of the langium generator, and perhaps a flag or a configuration option somewhere should be introduced to generate or not testing facilities. 

Kudos for the great work! 

Mathieu

sailingKieler:
Contributed with #1298.

PR #1298, opened 2023-11-20
generator-langium: extended yeoman generator extension to offer parsing, linking, and validation test stubs (#1282)

sailingKieler:
* providing additional automated tests for the cases of 'stubsOnly', 'vitest', and 'jest'


dhuebner:
@sailingKieler 
Running `npm run test` currently fails for me with:
```
 FAIL  packages/generator-langium/test/yeoman-generator.test.ts > Check yeoman generator works > 3 Should produce files for Vitest
Error: EACCES: permission denied, mkdir '/__yo-gen-testing'
```

Could it be that there are too many `..`? 

Dir name in my case is: `__dirname: /workspace/langium/packages/generator-langium/`
Test tries to write to: `Generating into directory: /__yo-gen-testing`

sailingKieler:
Hi Dennis,

> @sailingKieler Running `npm run test` currently fails for me with:
> 
> ```
>  FAIL  packages/generator-langium/test/yeoman-generator.test.ts > Check yeoman generator works > 3 Should produce files for Vitest
> Error: EACCES: permission denied, mkdir '/__yo-gen-testing'
> ```
> 
> Could it be that there are too many `..`?
> 
> Dir name in my case is: `__dirname: /workspace/langium/packages/generator-langium/` Test tries to write to: `Generating into directory: /__yo-gen-testing`

The test is supposed to write `__yo-gen-testing` next to `langium`, i.e. to `/workspace/__yo-gen-testing` in order to not see the langium root `node_modules`. Did you test it on Gitpod?

dhuebner:
@sailingKieler 
Yes, I tested it with GitPod. I think the tests should not create files outside the workspace in any case. However it seems that there is an additional `..`  in `'../../../../__yo-gen-testing'`  that can me omitted. Can you confirm?

sailingKieler:
@dhuebner you're right, and I guess that's due to the introduction of https://github.com/eclipse-langium/langium/blob/02b3c92388bf6740cf1dc879f22c2b4ae5cd73f6/packages/generator-langium/test/yeoman-generator.test.ts#L13 I had to rebase on. I'll fix that! Thanks a lot for testing! 🙏

packages/generator-langium/templates/test/test/parsing/parsing.test.ts
```diff
@@ -0,0 +1,56 @@
...
+describe('Parsing tests', () => {
+
+    test('parse simple model', async () => {
+        document = await parse(`
+            person Langium
+            Hello Langium!
+        `);
+
+        // check for absensce of parser errors the classic way:
+        //  deacivated, find a much more human readable way below!
```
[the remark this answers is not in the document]
spoenemann: I propose to keep the code as it it, but add more explanation why it's useful to optimize for test output readability.

packages/generator-langium/src/index.ts
```diff
@@ -129,6 +138,37 @@ class LangiumGenerator extends Generator {
...
+                        name: 'Yes, test stubs only, no additional project settings',
+                        value: 'stubsOnly'
+                    },
+                    {
+                        key: 'n',
+                        name: 'No test stubs and no additional project settings',
+                        value: 'none'
+                    }
+                ],
+                default: 'vitest'
```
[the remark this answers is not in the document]
spoenemann: To explain my suggestion: I would like to optimize the user experience for first-time users. In that context, we should be cautious not to overwhelm them.

Other opinions?

packages/generator-langium/templates/test/test/parsing/parsing.test.ts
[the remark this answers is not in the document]
sailingKieler: It's basically everybody's choice to switch back to the one liner version.
Regrettably there's no way of attaching custom message AFAIK.

I think it's a chance to show users alternatives. 😉

packages/generator-langium/src/index.ts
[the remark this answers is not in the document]
sailingKieler: I just reduced the wording to:
```
You can add the setup for language tests using Vitest (default) or Jest.
? Include language tests? (Vjsnh) 
  v) Tests + Vitest-based project settings (recommended)
  j) Tests + Jest-based project settings
  s) Tests only
  n) nothing
  h) Help, list all options
```

packages/generator-langium/templates/test/test/parsing/parsing.test.ts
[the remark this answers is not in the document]
sailingKieler: Added some explanation: https://github.com/eclipse-langium/langium/blob/01c9028302ad1e3d46e5f537237d942933a72bea/packages/generator-langium/templates/test/test/parsing/parsing.test.ts#L28-L37

sailingKieler:
@dhuebner feel free to give another try

packages/generator-langium/templates/test/test/linking/linking.test.ts
```diff
@@ -0,0 +1,53 @@
...
+let parse:    ReturnType<typeof parseHelper<Model>>;
+let document: LangiumDocument<Model> | undefined;
+
+beforeAll(async () => {
+    services = create<%= LanguageName %>Services(EmptyFileSystem);
+    parse = parseHelper<Model>(services.<%= LanguageName %>);
+
+    // activate the following if your linking test requires elements from a built-in library, for example
+    // await services.shared.workspace.WorkspaceManager.initializeWorkspace([]);
+});
```
msujew, line 18: Suggestion: We can simplify this quite a bit. 

```suggestion
const services = create<%= LanguageName %>Services(EmptyFileSystem);
const parse = parseHelper<Model>(services.<%= LanguageName %>);
// activate the following if your linking test requires elements from a built-in library, for example
/* beforeAll(async () => {
    await services.shared.workspace.WorkspaceManager.initializeWorkspace([]);
}); */

File: packages/generator-langium/templates/test/test/linking/linking.test.ts [hunk 4]
```diff
@@ -0,0 +1,53 @@
...
+
+afterEach(async () => {
+    document && await services.shared.workspace.DocumentBuilder.update([], [ document?.uri ]);
```
msujew, line 21: Suggestion: We have a method in our testing functions for that.

```suggestion
    clearDocuments(services);
```

packages/generator-langium/templates/test/test/parsing/parsing.test.ts
```diff
@@ -0,0 +1,60 @@
+import { beforeAll, describe, expect, test } from "<%= JestOrVitest %>";
+import { EmptyFileSystem, type LangiumDocument } from "langium";
+import { expandToString as s} from "langium";
+import { parseHelper } from "langium/test";
+import { create<%= LanguageName %>Services } from "../../src/language/<%= language-id %>-module.js";
+import { Model, isModel } from "../../src/language/generated/ast.js";
+
+let services: ReturnType<typeof create<%= LanguageName %>Services>;
+let parse:    ReturnType<typeof parseHelper<Model>>;
+let document: LangiumDocument<Model> | undefined;
```
msujew, line 10: See above.

packages/generator-langium/templates/test/test/validating/validating.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import { EmptyFileSystem, type LangiumDocument } from "langium";
+import { expandToString as s} from "langium";
+import { parseHelper } from "langium/test";
+import type { Diagnostic } from "vscode-languageserver-types";
+import { create<%= LanguageName %>Services } from "../../src/language/<%= language-id %>-module.js";
+import { Model, isModel } from "../../src/language/generated/ast.js";
+
+let services: ReturnType<typeof create<%= LanguageName %>Services>;
+let parse:    ReturnType<typeof parseHelper<Model>>;
+let document: LangiumDocument<Model> | undefined;
```
msujew, line 11: See above

packages/generator-langium/templates/test/test/parsing/parsing.test.ts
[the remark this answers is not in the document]
msujew: > Regrettably there's no way of attaching custom messages in case of failures AFAIK.

There is, just use `expect(value, failureMessage)...`

packages/generator-langium/src/index.ts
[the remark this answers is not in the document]
msujew: I think it's fine how it is 👍

packages/generator-langium/templates/test/test/linking/linking.test.ts
sailingKieler: This is how I used to do it in practice. However, the top-level statements as well as the `describe` lambdas might be executed by IDE feature to produce overviews like following, right?
That's why I used the super correct version here.

image

msujew: I see. And those test controllers don't perform a new import, right? So the services don't get reloaded. That's fine then.

packages/generator-langium/src/index.ts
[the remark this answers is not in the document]
msujew: On second thought, why not just have the Vitest setup and make it a y/n question? I believe Vitest is the only relevant testing framework for Langium now, especially since Jest doesn't support ESM natively.

packages/generator-langium/templates/test/test/parsing/parsing.test.ts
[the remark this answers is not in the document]
sailingKieler: 😱 Since when is that?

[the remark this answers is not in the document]
msujew: Since we switched to vitest :)

sailingKieler:
okay I dropped the options, and the Jest support, apparently I'm the only lonely wolf using it (in a commercial project).

Should it then be a question, at all? Maybe not.

packages/generator-langium/templates/test/test/linking/linking.test.ts
sailingKieler: Added.

packages/generator-langium/templates/test/test/linking/linking.test.ts
```diff
@@ -0,0 +1,53 @@
...
+            //  the referenced AST element as well as for a potential error message;
+            checkDocumentValid(document)
+                || document.parseResult.value.greetings.map(g => g.person.ref?.name || g.person.error?.message).join('\n')
+        ).toBe(s`
+            Langium
+        `);
+    });
+});
+
+function checkDocumentValid(document: LangiumDocument): string | undefined {
```
Lotes, line 45: This could be a function in 'langium/test'...

I am normally writing it like this:

```ts
function expectNoErrors(doc: LangiumDocument) {
  expect(doc.parseResult.lexerErrors, firstOrUndefined(doc.parseResult.lexerErrors, e => e.message)).toHaveLength(0);
  expect(doc.parseResult.parserErrors, firstOrUndefined(doc.parseResult.parserErrors, e => e.message)).toHaveLength(0);
  expect(doc.diagnostics ?? [], firstOrUndefined(doc.diagnostics ?? [], e => e.message)).toHaveLength(0);
}
function firstOrUndefined<T>(array: T[]|undefined, convert: (t: T) => string): string|undefined {
  return (array?.length??0) > 0 ? convert(array![0]) : undefined;
}
```

You do not have to use it, my hint is: do not so much in one command and split into multiple expects if possible to increase readability.

packages/generator-langium/templates/test/test/linking/linking.test.ts
```diff
@@ -0,0 +1,53 @@
...
+            Hello Langium!
+        `);
+
+        expect(
+            // here we first check for validity of the parsed document object by means of the reusable function
+            //  'checkDocumentValid()' to sort out (critical) typos first,
+            // and then evaluate the cross references we're interested in by checking
+            //  the referenced AST element as well as for a potential error message;
+            checkDocumentValid(document)
+                || document.parseResult.value.greetings.map(g => g.person.ref?.name || g.person.error?.message).join('\n')
```
Lotes, line 38: This line is too generic, it covers all cases where a person ref is not resolved correctly. I would recommend to address the exact element in the model with multiple expect().

```ts
expectNoErrors(document); //see below
expect(document.parseResult.value.greetings).toHaveLength(1);
expect(document.parseResult.value.greetings[0].person.ref?.name, document.parseResult.value.greetings[0].person.error?.message).toBeDefined();
```

sailingKieler: First of all: Don't forget it's just an example!

> This line is too generic

Why? The number of expected Greetings is (implicitly) specified in the expected string representation.
It just doesn't follow the "established pattern".

> `expect(document.parseResult.value.greetings).toHaveLength(1);`

Actually I don't like such kinds of tests: Why? It gives you failure messages like this:
```
AssertionError: expected [ { '$type': 'Greeting', …(5) }, …(1) ] to have a length of 1 but got 2
```
IMO, this is not developer friendly, esp. if the test pops up half a year after writing for somebody else than the original author. Imagine a scenario that is not as trivial as this one.

Another aspect is that the test stops immediately at the cardinality failure, nothing else is executed.
You have to investigate and fix it (or deactivate) to come to the next stumble stone.

My proposed solution is more tolerant on that, but - agreed - you have take a few seconds and think different.

image

Give it try 😉

packages/generator-langium/templates/test/test/linking/linking.test.ts
sailingKieler: > split into multiple expects if possible to increase readability

Maybe it simplifies reviewing the test, ok.
But does it improve anything when you run into spontaneous test failures (while working at something totally different)?

sailingKieler: And yes, I want to offer a convenient solution to developers that cannot rely on `vitest`'s flavor of `expect`.
Maybe I'm to kind at this point...

Lotes: I like the idea of having a generic checker for document validity. I also like the cool error formatting in this case. It is also good to not bind the user to the testing framework.
When you want to offer it to the user, it should live somewhere in our `langium/test` namespace (can also be a separate PR if you want). Otherwise we invent this wheel again and again.
Plus, I was wondering, why only parser errors? We also have lexer and validation errors.

packages/generator-langium/templates/test/test/linking/linking.test.ts
Lotes: I see, look at my other comment. I will resolve this thread.

spoenemann (commented):
Thanks, it looks really good. Just two issues:

- I see this in the editor for tsconfig.json:
```
File '/workspace/langium/examples/hello-world/test/linking/linking.test.ts' is not under 'rootDir' '/workspace/langium/examples/hello-world/src'. 'rootDir' is expected to contain all source files.
  The file is in the program because:
    Matched by include pattern 'test/**/*.ts' in '/workspace/langium/examples/hello-world/tsconfig.json'
  File is ECMAScript module because '/workspace/langium/examples/hello-world/package.json' has field "type" with value "module"
```
- I get the following error when I answer "Yes" to all questions:
```
"initialize" is not exported by "../../node_modules/vscode/extensions.js", imported by "../../node_modules/monaco-languageclient/lib/monaco-vscode-api-services.js".
file: /workspace/langium/node_modules/monaco-languageclient/lib/monaco-vscode-api-services.js:6:9
4:  * ------------------------------------------------------------------------------------------ */
5: import { ILogService, initialize, StandaloneServices } from 'vscode/services';
6: import { initialize as initializeVscodeExtensions } from 'vscode/extensions';
            ^
7: import getLanguagesServiceOverride from '@codingame/monaco-vscode-languages-service-override';
8: import getModelServiceOverride from '@codingame/monaco-vscode-model-service-override';
error during build:
RollupError: "initialize" is not exported by "../../node_modules/vscode/extensions.js", imported by "../../node_modules/monaco-languageclient/lib/monaco-vscode-api-services.js".
```

kaisalmen:
@spoenemann the second problem I will fix with a separate PR. The minor updates of `monaco-vscode-api` are pretty major and therefore this error surfaces here. There is https://github.com/CodinGame/monaco-vscode-api/issues/291

kaisalmen:
⬆️ Fix is here: https://github.com/eclipse-langium/langium/pull/1317

@sailingKieler consider upgrading to vite v5 / vitest v1 before merging this PR.

kaisalmen:
> Wouldn't that be out of scope for this PR?

Only for the generated output, not the whole project

sailingKieler:
> Only for the generated output, not the whole project

Of course 🤦🏽‍♂️ 😅

sailingKieler:
> consider upgrading to vite v5 / vitest v1 before merging this PR.

Tested upgrading to `vitest` ~1.0.0 but that requires `Node.js` v18, see
https://github.com/vitest-dev/vitest/blob/66933c3a590fa5cff35e252a5f107b0f41d9248b/packages/vitest/package.json#L103-L105 and
https://github.com/vitest-dev/vitest/blob/66933c3a590fa5cff35e252a5f107b0f41d9248b/packages/vitest/package.json#L111-L113

The latter causes a conflict during `npm i`:
```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: vitest@1.0.4
npm ERR! Found: @types/node@16.18.68
npm ERR! node_modules/@types/node
npm ERR!   dev @types/node@"~16.18.41" from the root project
npm ERR!   peerOptional @types/node@">= 14" from vite@4.4.12
npm ERR!   node_modules/vite
npm ERR!     dev vite@"~4.4.11" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peerOptional @types/node@"^18.0.0 || >=20.0.0" from vitest@1.0.4
npm ERR! node_modules/vitest
npm ERR!   dev vitest@"~1.0.0" from the root project
npm ERR! 
npm ERR! Conflicting peer dependency: @types/node@20.10.4
npm ERR! node_modules/@types/node
npm ERR!   peerOptional @types/node@"^18.0.0 || >=20.0.0" from vitest@1.0.4
npm ERR!   node_modules/vitest
npm ERR!     dev vitest@"~1.0.0" from the root project
```

Do you wanna switch to v18 as minimally required node.js version?

kaisalmen:
> Do you wanna switch to v18 as minimally required node.js version?

The question is why not. node 18 was the last LTS version. 20 is the current LTS version. 16 is outdated.

sailingKieler:
>> Do you wanna switch to v18 as minimally required node.js version?
>
> The question is why not. node 18 was the last LTS version. 20 is the current LTS version. 16 is outdated.

Filed dedicated PR https://github.com/eclipse-langium/langium/pull/1319

spoenemann (approved):
Thanks, looks good!

sailingKieler:
@msujew @kaisalmen I upgraded the versions of node.js & vitest in the generated `package.json`, and fixed some further tiny bugs.

Can you guys have a final look?

packages/generator-langium/templates/test/test/validating/validating.test.ts
```diff
@@ -0,0 +1,66 @@
+import { beforeAll, describe, expect, test } from "vitest";
+import { EmptyFileSystem, type LangiumDocument } from "langium";
+import { expandToString as s} from "langium/generate";
```
msujew, line 3: Minor

```suggestion
import { expandToString as s } from "langium/generate";
```

packages/generator-langium/templates/test/test/parsing/parsing.test.ts
```diff
@@ -0,0 +1,60 @@
+import { beforeAll, describe, expect, test } from "vitest";
+import { EmptyFileSystem, type LangiumDocument } from "langium";
+import { expandToString as s} from "langium/generate";
```
msujew, line 3:
```suggestion
import { expandToString as s } from "langium/generate";
```

packages/generator-langium/templates/test/test/linking/linking.test.ts
```diff
@@ -0,0 +1,53 @@
+import { afterEach, beforeAll, describe, expect, test } from "vitest";
+import { EmptyFileSystem, type LangiumDocument } from "langium";
+import { expandToString as s} from "langium/generate";
```
msujew, line 3:
```suggestion
import { expandToString as s } from "langium/generate";
```

.vscode/launch.json
```diff
@@ -87,6 +87,21 @@
                 "${workspaceFolder}/{packages,examples}/*/{lib,out}/**/*.js"
             ]
         },
+        {
+            "name": "Run Yoeman Generator",
```
msujew, line 91: Typo:

```suggestion
            "name": "Run Yeoman Generator",
```

msujew (approved):
Looks good to me. Just minor typos, see below 👍

packages/generator-langium/templates/test/test/validating/validating.test.ts
sailingKieler: Fixed.

packages/generator-langium/templates/test/test/parsing/parsing.test.ts
sailingKieler: Fixed.

packages/generator-langium/templates/test/test/linking/linking.test.ts
sailingKieler: Fixed.

.vscode/launch.json
sailingKieler: Fixed.

PR #1299, opened 2023-11-22
Enable array of validations for AstNodes

JohannesMeierSE:
At the moment, it is possible to register a single function for checking `AstNode` like this:

```ts
const checks: ValidationChecks<MyAstType> = {
  // ... checks for language-dependent types ...
  AstNode: checkAstNode
};
```
 
But it is not possible to register an array of functions for checking `AstNode`. This PR enables this feature:

```ts
const checks: ValidationChecks<MyAstType> = {
  // ... checks for language-dependent types ...
  AstNode: [ checkAstNodeOne, checkAstNodeTwo ]
};
```

This behaviour is already supported for language-dependent types, but not for `AstNode` yet.


spoenemann (approved):
Thanks! 👍

PR #1304, opened 2023-11-28
Reparse documents only when content changed

msujew:
When opening a workspace, we currently parse documents twice, even if their content hasn't changed.

1. Documents are parsed the first time when the `initializeWorkspace` method of the `WorkspaceManager` is called.
2. When a document has been opened in the editor, a `textDocument/onDidChange` notification is sent to the language server. This leads to a `LangiumDocumentFactory#update` call that tries to reparse the document.

This reparsing step can be quite expensive in case our documents are very large, as the parsing step cannot be interrupted, unlike every other builder step.

This change addresses point 2 by skipping the reparse in case the text document hasn't changed since the last parse. See tests on the effect (and potential side effects) of this change.


spoenemann (approved):
Looks good!

PR #1306, opened 2023-11-29
Refactor document factory to support async parsing

### LINKED ISSUE #273: URI resolving: sync / async, handle errors
The `TextDocumentFactory` currently uses synchronous Node.js file system API to load document contents. The synchronous method is required for resolving cross-references, otherwise we would force all AST processing to be async. We should define an additional asynchronous variant to be used in contexts that are already asynchronous.

Open question: can we guarantee that we can provide synchronous URI resolving in all contexts? What if documents are fetched from a remote service? We might be able to solve that by requiring all linked documents to be fetched into memory before linking happens.

Additionally, we should document all methods that can throw file system errors, and handle these errors at proper levels.

msujew:
Closes #273

This change allows adopters to create asynchronous parsers to prevent blocking the event loop when parsing large files. Note that the default implementation is still synchronous. Adopters have to supply the necessary worker threads/webworkers themselves.

Most things in this change are pretty backwards compatible. The only really breaking change is that `getOrCreateDocument` is now async and returns a promise. In addition to that method, the `LangiumDocuments` now expose `getDocument(URI): LangiumDocument | undefined` and `createDocument` (both sync and async).

### DIFF
--- a/packages/langium/src/lsp/call-hierarchy-provider.ts
+++ b/packages/langium/src/lsp/call-hierarchy-provider.ts
@@ -13,6 +13,7 @@ import type { AstNode } from '../syntax-tree.js';
 import type { Stream } from '../utils/stream.js';
 import type { ReferenceDescription } from '../workspace/ast-descriptions.js';
 import type { LangiumDocument, LangiumDocuments } from '../workspace/documents.js';
+import type { MaybePromise } from '../utils/promise-util.js';
 import { SymbolKind } from 'vscode-languageserver';
 import { findDeclarationNodeAtOffset } from '../utils/cst-util.js';
 import { URI } from '../utils/uri-util.js';
@@ -21,11 +22,11 @@ import { URI } from '../utils/uri-util.js';
  * Language-specific service for handling call hierarchy requests.
  */
 export interface CallHierarchyProvider {
-    prepareCallHierarchy(document: LangiumDocument, params: CallHierarchyPrepareParams, cancelToken?: CancellationToken): CallHierarchyItem[] | undefined;
+    prepareCallHierarchy(document: LangiumDocument, params: CallHierarchyPrepareParams, cancelToken?: CancellationToken): MaybePromise<CallHierarchyItem[] | undefined>;
 
-    incomingCalls(params: CallHierarchyIncomingCallsParams, cancelToken?: CancellationToken): CallHierarchyIncomingCall[] | undefined;
+    incomingCalls(params: CallHierarchyIncomingCallsParams, cancelToken?: CancellationToken): MaybePromise<CallHierarchyIncomingCall[] | undefined>;
 
-    outgoingCalls(params: CallHierarchyOutgoingCallsParams, cancelToken?: CancellationToken): CallHierarchyOutgoingCall[] | undefined;
+    outgoingCalls(params: CallHierarchyOutgoingCallsParams, cancelToken?: CancellationToken): MaybePromise<CallHierarchyOutgoingCall[] | undefined>;
 }
 
 export abstract class AbstractCallHierarchyProvider implements CallHierarchyProvider {
@@ -41,7 +42,7 @@ export abstract class AbstractCallHierarchyProvider implements CallHierarchyProv
         this.references = services.references.References;
     }
 
-    prepareCallHierarchy(document: LangiumDocument<AstNode>, params: CallHierarchyPrepareParams): CallHierarchyItem[] | undefined {
+    prepareCallHierarchy(document: LangiumDocument<AstNode>, params: CallHierarchyPrepareParams): MaybePromise<CallHierarchyItem[] | undefined> {
         const rootNode = document.parseResult.value;
         const targetNode = findDeclarationNodeAtOffset(
             rootNode.$cstNode,
@@ -81,11 +82,8 @@ export abstract class AbstractCallHierarchyProvider implements CallHierarchyProv
         return undefined;
     }
 
-    incomingCalls(params: CallHierarchyIncomingCallsParams): CallHierarchyIncomingCall[] | undefined {
-        const document = this.documents.getDocument(URI.parse(params.item.uri));
-        if (!document) {
-            return undefined;
-        }
+    async incomingCalls(params: CallHierarchyIncomingCallsParams): Promise<CallHierarchyIncomingCall[] | undefined> {
+        const document = await this.documents.getOrCreateDocument(URI.parse(params.item.uri));
         const rootNode = document.parseResult.value;
         const targetNode = findDeclarationNodeAtOffset(
             rootNode.$cstNode,
@@ -108,13 +106,10 @@ export abstract class AbstractCallHierarchyProvider implements CallHierarchyProv
     /**
      * Override this method to collect the incoming calls for your language
      */
-    protected abstract getIncomingCalls(node: AstNode, references: Stream<ReferenceDescription>): CallHierarchyIncomingCall[] | undefined;
+    protected abstract getIncomingCalls(node: AstNode, references: Stream<ReferenceDescription>): MaybePromise<CallHierarchyIncomingCall[] | undefined>;
 
-    outgoingCalls(params: CallHierarchyOutgoingCallsParams): CallHierarchyOutgoingCall[] | undefined {
-        const document = this.documents.getDocument(URI.parse(params.item.uri));
-        if (!document) {
-            return undefined;
-        }
+    async outgoingCalls(params: CallHierarchyOutgoingCallsParams): Promise<CallHierarchyOutgoingCall[] | undefined> {
+        const document = await this.documents.getOrCreateDocument(URI.parse(params.item.uri));
         const rootNode = document.parseResult.value;
         const targetNode = findDeclarationNodeAtOffset(
             rootNode.$cstNode,
@@ -130,5 +125,5 @@ export abstract class AbstractCallHierarchyProvider implements CallHierarchyProv
     /**
      * Override this method to collect the outgoing calls for your language
      */
-    protected abstract getOutgoingCalls(node: AstNode): CallHierarchyOutgoingCall[] | undefined;
+    protected abstract getOutgoingCalls(node: AstNode): MaybePromise<CallHierarchyOutgoingCall[] | undefined>;
 }
--- a/packages/langium/src/lsp/call-hierarchy-provider.ts
+++ b/packages/langium/src/lsp/call-hierarchy-provider.ts
@@ -82,7 +82,10 @@ export abstract class AbstractCallHierarchyProvider implements CallHierarchyProv
     }
 
     incomingCalls(params: CallHierarchyIncomingCallsParams): CallHierarchyIncomingCall[] | undefined {
-        const document = this.documents.getOrCreateDocument(URI.parse(params.item.uri));
+        const document = this.documents.getDocument(URI.parse(params.item.uri));
+        if (!document) {
+            return undefined;
+        }
         const rootNode = document.parseResult.value;
         const targetNode = findDeclarationNodeAtOffset(
             rootNode.$cstNode,
@@ -108,7 +111,10 @@ export abstract class AbstractCallHierarchyProvider implements CallHierarchyProv
     protected abstract getIncomingCalls(node: AstNode, references: Stream<ReferenceDescription>): CallHierarchyIncomingCall[] | undefined;
 
     outgoingCalls(params: CallHierarchyOutgoingCallsParams): CallHierarchyOutgoingCall[] | undefined {
-        const document = this.documents.getOrCreateDocument(URI.parse(params.item.uri));
+        const document = this.documents.getDocument(URI.parse(params.item.uri));
+        if (!document) {
+            return undefined;
+        }
         const rootNode = document.parseResult.value;
         const targetNode = findDeclarationNodeAtOffset(
             rootNode.$cstNode,
--- a/packages/langium/src/lsp/language-server.ts
+++ b/packages/langium/src/lsp/language-server.ts
@@ -483,55 +483,37 @@ export function addWorkspaceSymbolHandler(connection: Connection, services: Lang
 
 export function addCallHierarchyHandler(connection: Connection, services: LangiumSharedServices): void {
     connection.languages.callHierarchy.onPrepare(createServerRequestHandler(
-        (services, document, params, cancelToken) => {
+        async (services, document, params, cancelToken) => {
             if (services.lsp.CallHierarchyProvider) {
-                return services.lsp.CallHierarchyProvider.prepareCallHierarchy(document, params, cancelToken) ?? null;
+                const result = await services.lsp.CallHierarchyProvider.prepareCallHierarchy(document, params, cancelToken);
+                return result ?? null;
             }
             return null;
         },
         services
     ));
 
-    connection.languages.callHierarchy.onIncomingCalls(createCallHierarchyRequestHandler(
-        (services, params, cancelToken) => {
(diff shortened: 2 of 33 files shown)


packages/langium/src/workspace/documents.ts
```diff
@@ -282,10 +306,28 @@ export interface LangiumDocuments {
...
+     * The new document is automatically added to this service and can be retrieved using {@link getDocument}.
+     */
+    createDocument(uri: URI, text: string): LangiumDocument;
+
+    /**
+     * Creates a new document with the given URI and text content asynchronously.
+     * The process can be interrupted with a cancellation token.
+     * The new document is automatically added to this service and can be retrieved using {@link getDocument}.
      */
-    getOrCreateDocument(uri: URI): LangiumDocument;
```
cdietrich, line 288: is there a reason not to keep the old api.
writing

`hasDocument ? getDocument : createDocument` might be a bit tedious

msujew: That would require something like this:

```ts
getOrCreateDocument(uri: URI, text: string | () => string): LangiumDocument;
```

Would that be fine for you? I aim to deprecate the synchronous `readFileSync` method of the file system access.

cdietrich: hmmm why do i need to pass a string? when i tried to call hasDocument i get falses in our tests in 2.1.3
=> who put the documents into the map?

msujew: Maybe You've overseen the new `getOrCreateDocument(uri: URI, cancellationToken?: CancellationToken): Promise<LangiumDocument>` method? The method still exists, it's just async now. The signature above would be necessary to keep it sync, because we cannot load from disk in a sync way anymore.

packages/langium/test/serializer/json-serializer.test.ts
```diff
@@ -217,7 +217,7 @@ describe('JsonSerializer', async () => {
         `;
         const model = serializer.deserialize<Entry>(json);
         expect(model.elements).toHaveLength(1);
-        expect(model.elements[0].other?.error?.message).toEqual('Error: No file system is available.');
+        expect(model.elements[0].other?.error?.message).toEqual('Could not resolve URI: file:///does-not-exist.langium#/elements@0');
```
dhuebner, line 220: This error message is misleading to me because of fragment in the URI. Is there a way  to report the actual cause?

msujew:
@sailingKieler The intended changes are:

Allow for async parsing in the document factory. While the parsing process isn't async in itself, we can improve performance of Langium by moving the parsing process over to worker threads/web workers. By doing this, parsing becomes async. Additionally, this gives us the option to interrupt parsing if we need to. This is very important for large files that might take multiple seconds to parse.

This means that all factory methods that somehow produce an AST from a string can now be async as well, assuming they use the new `AsyncParser` service. I've kept the sync versions in there for now.

As a consequence, we don't need the sync file loading anymore. If parsing is async, file loading can be async as well. It has always irked me that we had sync file loading in the framework in the first place. See also #273. It prevented parallel file loading during workspace startup and is generally viewed as an anti-pattern in modern JS development. Other file systems (such as loading a workspace via HTTP), also were async only, so you would always get an error if you tried to load a document through `getOrCreateDocument` as it relied on the sync file loading mechanism.

sailingKieler:
> @sailingKieler The intended changes are:
> 
> Allow for async parsing in the document factory. While the parsing process isn't async in itself, we can improve performance of Langium by moving the parsing process over to worker threads/web workers. By doing this, parsing becomes async. Additionally, this gives us the option to interrupt parsing if we need to. This is very important for large files that might take multiple seconds to parse.
> 
> This means that all factory methods that somehow produce an AST from a string can now be async as well, assuming they use the new `AsyncParser` service. I've kept the sync versions in there for now.
> 
> As a consequence, we don't need the sync file loading anymore. If parsing is async, file loading can be async as well. It has always irked me that we had sync file loading in the framework in the first place. See also #273. It prevented parallel file loading during workspace startup and is generally viewed as an anti-pattern in modern JS development. Other file systems (such as loading a workspace via HTTP), also were async only, so you would always get an error if you tried to load a document through `getOrCreateDocument` as it relied on the sync file loading mechanism.

@msujew Thank's a lot for these explanations. 🙏
What I was also interested in was where you basically put on the saw (probably in the document builder) and what the resulting consequences are.

msujew:
@sailingKieler Aside from the changes in the document factory, there is very little change. The most major one is the change to `getOrCreateDocument` in the `LangiumDocuments` service, which is now just async. All sync callers can use the new `getDocument(): LangiumDocument | undefined` instead and call `createDocument()` explicitly if they need to.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -157,10 +157,7 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
         this.indexManager.remove(deleted);
         // Set the state of all changed documents to `Changed` so they are completely rebuilt
         for (const changedUri of changed) {
-            const invalidated = this.langiumDocuments.invalidateDocument(changedUri);
-            if (!invalidated) {
-                this.langiumDocuments.getOrCreateDocument(changedUri);
```
spoenemann, line 162: This line was important in order to load documents that are new (the name `changed` is misleading in this regard: it includes new documents). I confirmed that the change introduces a bug: when I create a new file, it's not considered by the document builder.

We have to load these new documents before the first call to `interruptAndCheck` to ensure that we don't lose them.

Idea: create "empty" documents with state `Changed` and no AST (this can be done synchronously). The document builder will parse the actual content in step 0 of `buildDocuments`.

packages/langium/src/lsp/type-hierarchy-provider.ts
```diff
@@ -99,7 +99,10 @@ export abstract class AbstractTypeHierarchyProvider implements TypeHierarchyProv
     }
 
     supertypes(params: TypeHierarchySupertypesParams, _cancelToken?: CancellationToken): TypeHierarchyItem[] | undefined {
-        const document = this.documents.getOrCreateDocument(URI.parse(params.item.uri));
+        const document = this.documents.getDocument(URI.parse(params.item.uri));
```
spoenemann, line 102: LSP services can always be async. Shall we change the service API to support async calls?

packages/langium/src/lsp/language-server.ts
```diff
@@ -610,7 +611,7 @@ export function createRequestHandler<P extends { textDocument: TextDocumentIdent
             console.error(`Could not find service instance for uri: '${uri.toString()}'`);
             return null;
         }
-        const document = documents.getOrCreateDocument(uri);
+        const document = documents.getDocument(uri);
```
spoenemann, line 614: This function is `async`, so we could `await documents.getOrCreateDocument(uri)`?

packages/langium/src/lsp/call-hierarchy-provider.ts
```diff
@@ -82,7 +82,10 @@ export abstract class AbstractCallHierarchyProvider implements CallHierarchyProv
     }
 
     incomingCalls(params: CallHierarchyIncomingCallsParams): CallHierarchyIncomingCall[] | undefined {
-        const document = this.documents.getOrCreateDocument(URI.parse(params.item.uri));
+        const document = this.documents.getDocument(URI.parse(params.item.uri));
```
spoenemann, line 85: Same as for the type hierarchy: shall we change the service API to async?

spoenemann (commented):
This is a heavily breaking change, but the new API of LangiumDocumentFactory and LangiumDocuments is better now, so I think it's worth it.

Shall we add an implementation of AsyncParser with Node.js worker threads in `src/node` to make the benefit of this change more apparent?

packages/langium/src/workspace/document-builder.ts
msujew: I've kept the same logic. The contract of our `LangiumDocument` doesn't allow for an `undefined` parse value, so it didn't feel right to create an empty document. We now parse new documents in parallel though.

msujew:
> Shall we add an implementation of AsyncParser with Node.js worker threads in src/node to make the benefit of this change more apparent?

@spoenemann I'd love to, but I don't have time for this before the holidays/the 3.0 release. I'd postpone this implementation until 3.1.0. This won't be a breaking change, as it only adds a new implementation.

Otherwise, I've incorporated all your changes 👍

packages/langium/src/workspace/document-builder.ts
spoenemann: I'm thinking about changing the contract. The problem is that the initial parsing is done during the non-interruptible phase. The same applies to the WorkspaceManager: we have to parse the whole workspace before we pass it to the DocumentBuilder.

Making `parseResult` optional could be a quite heavy (and inconvenient) change. WDYT of creating a synthetic root node as placeholder?

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -116,7 +116,8 @@ export class DefaultWorkspaceManager implements WorkspaceManager {
                 if (entry.isDirectory) {
                     await this.traverseFolder(workspaceFolder, entry.uri, fileExtensions, collector);
                 } else if (entry.isFile) {
-                    const document = this.langiumDocuments.getOrCreateDocument(entry.uri);
+                    const content = await this.fileSystemProvider.readFile(entry.uri);
+                    const document = await this.langiumDocuments.createDocument(entry.uri, content, CancellationToken.None);
```
spoenemann, line 120: In case we keep the current practice of parsing all documents upfront, we should change this method to first collect the URIs and then parse them in parallel. The current approach still parses them sequentially.

spoenemann: But first we need to clarify the basic approach to LangiumDocuments (see discussion above).

msujew: The current approach actualy parses them in parallel. See the `Promise.all(entry.map(...))` call.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -155,14 +155,22 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+                // This will be parsed as soon as we reach the first document builder phase.
+                // This allows to cancel the parsing process later in case we need it.
+                const newDocument = this.langiumDocumentFactory.fromModel({ $type: 'INVALID' }, changedUri);
+                newDocument.state = DocumentState.Changed;
+                this.langiumDocuments.addDocument(newDocument);
             }
             this.buildState.delete(changedUri.toString());
         }
+        // Parse documents in parallel (if possible)
+        await Promise.all(newDocuments);
```
spoenemann, line 173: The `newDocuments` array is no longer needed.

packages/langium/src/workspace/documents.ts
```diff
@@ -288,10 +312,28 @@ export interface LangiumDocuments {
...
+    /**
+     * Retrieve the document with the given URI. If not present, a new one will be created using the file system access.
+     * The new document will be added to the list of documents managed under this service.
+     */
+    getOrCreateDocument(uri: URI, cancellationToken?: CancellationToken): Promise<LangiumDocument>;
+
+    /**
+     * Creates a new document with the given URI and text content.
+     * The new document is automatically added to this service and can be retrieved using {@link getDocument}.
+     */
```
spoenemann, line 328: Add a hint that this method throws an error if a document with the given URI is already present (also for the other overload).

spoenemann (approved):
Great, just two minor things left

spoenemann:
>> Shall we add an implementation of AsyncParser with Node.js worker threads in src/node to make the benefit of this change more apparent?

> @spoenemann I'd love to, but I don't have time for this before the holidays/the 3.0 release. I'd postpone this implementation until 3.1.0. This won't be a breaking change, as it only adds a new implementation.

The only issue with this is: are we sure that we won't run into blockers when we implement a parser worker pool? For example, how will we solve the data transfer of AST and CST?

msujew:
@spoenemann Since we have postponed 3.0 to after the holidays, I'll probably integrate the implementation before the release. However, I would prefer to do this in a separate PR. This PR is large enough as it is.

packages/langium/src/workspace/documents.ts
```diff
@@ -339,16 +383,34 @@ export class DefaultLangiumDocuments implements LangiumDocuments {
...
-            return langiumDoc;
+        return this.documentMap.get(uriString);
+    }
+
+    async getOrCreateDocument(uri: URI, cancellationToken?: CancellationToken): Promise<LangiumDocument> {
+        let document = this.getDocument(uri);
+        if (document) {
+            return document;
+        }
+        const text = await this.fileSystemProvider.readFile(uri);
```
sailingKieler, line 396: Is there a reason for moving the file content reading from the documentFactory to here?
This feels weird to me, as `update` of `DefaultDocumentFactory` does the FS access, too.

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -116,7 +116,8 @@ export class DefaultWorkspaceManager implements WorkspaceManager {
                 if (entry.isDirectory) {
                     await this.traverseFolder(workspaceFolder, entry.uri, fileExtensions, collector);
                 } else if (entry.isFile) {
-                    const document = this.langiumDocuments.getOrCreateDocument(entry.uri);
+                    const content = await this.fileSystemProvider.readFile(entry.uri);
```
sailingKieler, line 119: Here's a third place of `await this.fileSystemProvider.readFile(entry.uri)`.
I would prefer to encapsulate them within a single service, and `DocumentFactory` is the natural candidate (and it was like that before I think)

sailingKieler:
I'm fine with the changes except for the distributed file system accesses.
I would appreciate if you could consolidate them, ideally within the `DocumentFactory`.

msujew:
@sailingKieler I've added a new `LangiumDocumentFactory#fromUri` method that consolidates all the disk access into one place.

sailingKieler (approved):
🚀

cdietrich:
i wonder if we can have a readFileSync readDirSync back in the FileSystem provider
this would at least allow to mimic the old getOrCreateDocument more easily

spoenemann:
We could, but those operations are harder to implement in the frontend or other context where async is the default.

cdietrich:
my problem is that exchaning the FileSystemProvider with a own subclass is pita.
or should we subclass 

EmptyFileSystemProvider, NodeFileSystemProvider,

create respective copies of NodeFileSystem, EmptyFileSystem.

and at usageplace just downcast to our subinterface?

cdietrich:
another question: why is there no userdata in index entires. our main usecase to resolve during scoping is cause we can filter pure on what is in AstNodeDescription

msujew:
> why is there no userdata in index entires. our main usecase to resolve during scoping is cause we can not filter pure on what is in AstNodeDescription

We had no need for it yet, and since the `AstNodeDescriptions` are just plain JS objects, you can always extend/cast them to add new fields. I'd be fine with adding a `data: any` field to all `AstNodeDescriptions`.

> i wonder if we can have a readFileSync readDirSync back in the FileSystem provider

I'd be open for it. I'm wondering though in which case you need a sync read, when the document builder itself is async? Can you explain that a bit?

cdietrich:
no in the scope provider we do:

for astdescription in index, call getOrCreateDocument,
but the ast nodes into a stream.
filter based on our critera,
create scope based on that.

msujew:
Isn't `getDocument` enough for that case? Do you have `AstDescription` objects in your index for unloaded documents?

cdietrich:
Get document might not find it. At least in some cases. Did not look into that in detail yet.

cdietrich:
looks like so far it was a set of test data error anyway (that somehow worked in 2x). works as expected if that one is fixed
need to do more tests though

Issue #1291, opened 2023-11-15
Add read queue to workspace mutex


msujew:
Right now, the workspace mutex is only able to lock on write operations - any read operation will abort the current write operation. We want to extend the mutex to allow to perform a read operation (or multiple ones in parallel) after finishing all queued write operations. 

This could also resolve the issue experienced in https://github.com/eclipse-langium/langium/discussions/1290, since it's probably related to an LSP service trying to resolve a cross reference during the build process.

spoenemann:
We already had a read-write lock in a previous PR: https://github.com/eclipse-langium/langium/pull/549/commits/56966f6cdf13ab33a03daec10895d30f8eaaaa12#diff-095a6181fde780365fcff35aa085c4b5aff0ec9e081d26f716bd29bca6575974

We decided to reduce it to a simple mutex because we wanted to keep things simple. At that time, I didn't see the case where a language developer adds a custom operation that needs read access at a point in time where the document is in a clean state. We could revive that old code now.

PR #1310, opened 2023-11-30
Extend mutex lock to support read operations

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1291

Renames the existing `lock` operation to `write`. Adds a new `read` lock that only runs after all `write` operations have finished without interrupting them.


packages/langium/src/utils/promise-util.ts
```diff
@@ -86,34 +86,119 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
...
 /**
- * Utility class to execute mutually exclusive actions.
+ * Utility service to execute mutually exclusive actions.
  */
-export class MutexLock {
-
-    private previousAction = Promise.resolve();
-    private previousTokenSource = new CancellationTokenSource();
-
+export interface MutexLock {
```
spoenemann, line 91: Since this is a workspace-related service with a clear intent, rather than a reusable utility, I'd propose to rename it to `WorkspaceLock` and move the interface and default implementation to `src/workspace/workspace-lock.ts`.

packages/langium/src/utils/promise-util.ts
```diff
@@ -86,34 +86,119 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
...
+     *
+     * Note that read actions are not allowed to modify anything in the workspace. Please use {@link write} instead.
+     *
+     */
+    read<T>(action: () => MaybePromise<T>): MaybePromise<T>;
+
+    /**
+     * Cancels the currently executed write action.
+     */
+    cancel(): void;
```
spoenemann, line 116: Rename to `cancelWriters`?

packages/langium/src/utils/promise-util.ts
```diff
@@ -86,34 +86,119 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
...
+
+export class DefaultMutexLock implements MutexLock {
+
+    private previousTokenSource = new CancellationTokenSource();
+    private writeQueue: MutexEntry[] = [];
+    private readQueue: MutexEntry[] = [];
+    private done = true;
+
+    write(action: (token: CancellationToken) => MaybePromise<void>): Promise<void> {
         this.cancel();
```
spoenemann, line 135: Should `cancel` (or `cancelWriters`) also clear the write queue?

packages/langium/src/workspace/workspace-lock.ts
```diff
@@ -0,0 +1,130 @@
...
+                deferred.resolve(result);
+            }).catch(err => {
+                if (isOperationCancelled(err)) {
+                    // If the operation was cancelled, we don't want to reject the promise
+                    deferred.resolve(undefined);
+                } else {
+                    deferred.reject(err);
+                }
+            }).finally(() => {
+                goToNext();
```
spoenemann, line 122: Minor: Why not move the content of this function directly into this block?

spoenemann (approved):
👍

packages/langium/src/workspace/workspace-lock.ts
```diff
@@ -0,0 +1,129 @@
...
+        let completed = 0;
+        const goToNext = () => {
+            completed += 1;
+            // If all specified actions have been completed, we can perform the next operation
+            if (completed === entries.length) {
+                this.done = true;
+                this.performNextOperation();
+            }
+        };
+        for (const entry of entries) {
```
sailingKieler, line 110: Would `Promise.all(...)` work here, too?
Is there a benefit of this sequential loop?

sailingKieler: My thought ist

With returning the result of `Promise.all(...)` the `goToNext` construction could be skipped and `performNextOperation()` could be written like

```
private performNextOperation(): void {
        let promise: Promise;
        if (this.writeQueue.length > 0) {
            // Just perform the next write action
            promise = this.performOperation([this.writeQueue.shift()!]);
        } else if (this.readQueue.length > 0) {
            // Empty the read queue and perform all actions in parallel
            const entries = this.readQueue.splice(0, this.readQueue.length);
            promise = this.performOperation(entries);
        }
        if (promise) {
            promise.finally(() => performNextOperation());
        }
}
```

which shows the 'run to completion' characteristics better.

sailingKieler (approved):

PR #1319, opened 2023-12-13
declare minimal node engine version for development to be >= 18.0.0, upgrade vitest to '~1.0.0'

sailingKieler:
github workflow is already executed with node v18


sailingKieler:
Is there anything else (documentation?) to be updated?

packages/langium-vscode/package.json
```diff
@@ -17,7 +17,7 @@
     "theme": "dark"
   },
   "engines": {
-    "vscode": "^1.67.0"
+    "vscode": "^1.84.0"
```
kaisalmen, line 20: I am not sure if this "target" is required to be as new is this, but on the other hand we now depend on vscode-languageclient/server 9.x. I don't know since when vscode requires it itself.

packages/langium/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "A language engineering tool for the Language Server Protocol",
   "homepage": "https://langium.org",
   "engines": {
-    "node": ">=16.0.0"
+    "node": ">=18.18.0"
```
kaisalmen, line 7: We may need to name an npm version here as well, because the `package-lock.json` version is already `"lockfileVersion": 3` which requires I guess npm v9 (need to find that out)

examples/arithmetics/package.json
```diff
@@ -77,8 +77,8 @@
         "langium-cli": "~2.1.0"
     },
     "volta": {
-        "node": "18.17.1",
-        "npm": "9.6.7"
+        "node": "18.18.2",
```
kaisalmen, line 80: Node `18.19.0` is already out 🙈 , but it contains npm v10 now (to my surprise).

kaisalmen (commented):
I noted some things to consider or discuss. Nailing the env is not necessarily easy. 😅

packages/langium-vscode/package.json
sailingKieler: Exactly that was my reason to upgrade `vscode`. And the VS Code IDE usually checks for updates itself so users probably won't have a (very) out-dated version,

examples/arithmetics/package.json
sailingKieler: Yes, I know. I, however, don't wanna yell at users (like me) having 18.18.... installed. (18.18.0 was released mid of September '23)
It shouldn't make a difference for Langium.

packages/langium/package.json
sailingKieler: That's declared in the root `package.json`, see https://github.com/eclipse-langium/langium/blob/79e430127324c5ff45a2c753744ac1aa7c5237e0/package.json#L6

examples/arithmetics/package.json
kaisalmen: Alright

packages/langium/package.json
kaisalmen: Overlooked that one.

packages/langium-vscode/package.json
kaisalmen: I agree. Does anyone else have a reason why we shouldn't?

msujew (changes requested):

packages/langium/package.json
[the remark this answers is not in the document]
sailingKieler: The change is motivated by https://github.com/eclipse-langium/langium/pull/1298#issuecomment-1852111534

packages/langium-vscode/package.json
msujew: One word: Theia. If we don't benefit from any API changes (or need to update due to `vscode-languageclient`), we shouldn't update our required vscode version. This will only break compability with Theia users and accomplish nothing else.

packages/langium/package.json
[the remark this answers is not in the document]
msujew: Ok, I'm fine with the EOL change. But shouldn't we just use `18.0.0` instead then? Just because someone is running 18.14.0 (or similar), they shouldn't face an dependency issue.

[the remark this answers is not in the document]
montymxb: I agree with @sailingKieler here. This seems like a good time to bump to 18, given [Node 16 end-of-life already passed in Sept. 11, 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol). Plus the timing of the change would align with the upcoming major release. Although we may not have issues with 16 right now, we can't expect support on 16 going forward, and we would need to release another major version anyways once we do another major node version change. Also, node points out that production apps should really [strive to be on the active or maintenance LTS](https://nodejs.org/en/about/previous-releases), which makes sense for Langium.

I don't have as much to add regarding which version of 18.X.X we should target, so long as we're at least on 18.

spoenemann:
> Is there anything else (documentation?) to be updated?

See #1277

packages/langium/package.json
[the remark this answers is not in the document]
sailingKieler: I don't care whether to chose `18.0.0` or `18.18.0`, it just should be in sync with the devDependency on `@types/node`. `18.0.0` is from April-22, while `18.18.0` is from September-23. Functionality-wise, I can't say anything on differences.

packages/langium-vscode/package.json
sailingKieler: @msujew Actually, I have no clue what Theia does with this setting. Could you elaborate a bit?
Without any context the current settings looks like a magic number and appears just out-dated.

sailingKieler:
It's quite a different PR now: Here're the updates as proposed in today's dev meeting.

kaisalmen:
@sailingKieler sorry I wasn't able to participate in today's meeting. This seems to be a good compromise.

kaisalmen (approved):
Just tested the new solution and it looks good.

spoenemann (approved):
👍 the only thing I noticed is this message when running the tests:

> The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.

kaisalmen:
@spoenemann I missed that, but have a solution for this. Either the base `package.json` or the vite config file must have the ".mts" extension.

sailingKieler:
> @spoenemann I missed that, but have a solution for this. Either the base `package.json` must be of type module or the vite config file must have the ".mts" extension. We did the later one in sprotty.

done!

@msujew can you have a look again?

package.json
```diff
@@ -26,16 +27,19 @@
...
     "eslint": "~8.47.0",
     "eslint-plugin-header": "~3.1.1",
     "editorconfig": "~2.0.0",
     "shx": "~0.3.4",
     "typescript": "~5.1.6",
-    "vitest": "~0.34.2"
+    "vitest": "~1.0.0"
+  },
+  "overrides": {
+   "@types/node": "~16.18.41"
```
msujew, line 42: Question: What is this override for? We already use `@types/node@16.18.41` in our `devDependencies`.

sailingKieler: I stated it in the commit msg for the records:
> added version override on '@types/node' to cope with vitest's (transitive) peer dependency on '@types/node@"^18.0.0 || >=20.0.0"'

I think vitest's dependency is even flagged as optional, appears to work nicely.

msujew: I don't really understand why they have a `@types/*` dependency as a peer dependency (they really shouldn't), but I understand now why you did what you did. Thanks.

msujew (approved):
Alright thanks. Looks good to me as well 👍

PR #1328, opened 2024-01-02
`AbstractType`s need to include `InferredType`s

JohannesMeierSE:
... since cross-references might use `InferredType`s as their types.

This seems to be an edge-case in the Langium grammar:
The type of a cross-reference does not point to the parser rule itself, if this parser rule has an explicitly declared inferred type.
Instead, the cross-reference points to the corresponding `InferredType`.

But `InferredType` is not supported by `AbstractType`, which is used as type for cross-references in the Langium grammar:
```
type AbstractType = Interface | Type | Action | ParserRule;

CrossReference infers AbstractElement:
    {infer CrossReference} '[' type=[AbstractType] ((deprecatedSyntax?='|' | ':') terminal=CrossReferenceableTerminal )? ']';
```

The example from the created test case is available [here in the playground](https://langium.org/playground?grammar=OYJwhgthYgBAEgUwDbIPYHU0mQEwFD6IB2ALiAJ6wCyauKAXLGANQC8AggFQDchHsAJbEAZohABnWACEmAcjBzYxSIjYBJACKwAFGlIALcRzYBtaQF0AlAH4%2BBwbnrFYpcRGFhksDAGUmAPQAOhIsAXxuIB4q3lqBpmAAtABeHIkAWgD6FgkpaekADIkAnNlc4fhAA&content=IYAgsgngcsC2CmBGAUKSMECZzTk5QA).

At the moment, the TS compiler complains about this issue:
Screenshot 2024-01-02 at 11 55 50

Resolving this cross-reference to the InferredType works at runtime, since it is explicitly exported: https://github.com/eclipse-langium/langium/blob/2ca2b3adb452ad5048135af85e1510f84a71fd7b/packages/langium/src/grammar/references/grammar-scope.ts#L106

I am proposing to mark InferredTypes as AbstractTypes ...
```
type AbstractType = Interface | Type | Action | ParserRule | InferredType;
```
... but there might be a better solution.

Update 2024-01-05:
- I decided to export implicitly and explicitly inferred types (i.e. the `InferredType` nodes) with the type `InferredType` into the scopes, since that fits to the $type of the exported AST node and that clearly distinguishes explicit types and interfaces from inferred types. @spoenemann In my eyes, that is more consistent than the current implementation.
- Since InferredTypes are AbstractTypes now, it is not required for `Action`s to be `AbstractType`s anymore: Actions either reuse already declared types or contribute a new type with an `InferredType` and InferredTypes are AbstractTypes now. Additionally, `Action`s are not exported into scopes in `grammar-scope.ts` and therefore cannot be resolved for cross-references to `AbstractType`.

@pluralia Since we discussed about these topics, I am looking forward to your review as well!

Update 2024-01-12:
- I fixed a bug, since I missed that Actions with explicitly inferred types are still exported into scopes and not their InferredType.
- Additionally, I fixed a related bug, since the identification for the name-assignment for a type used in cross-references which is explicitly inferred from an Action didn't work as expected. The example for the new test cases can be found in the [playground as well](https://langium.org/playground?grammar=OYJwhgthYgBAEgUwDbIPYHU0mQEwFD6IB2ALiAJ6wCyauKAXLGANQC8AggFQDchA9Pw6wAlsQBmiEAGdYAISYByMItjFIiNgEkAIrAAUaUgAspHNgG05AXQCUAfh6xBsAO7YA1tPwcmAbzFJODkAX1hlVXUITV0DI1MQcys7R2d%2BN09ZYiNCYxFcemJYUikIMTBkWAwAZSZ%2BAB1pFn4%2BEpAy9UrdOoswAFoALw4%2BgC0AfWteweGRgAY%2BgE4Jrhb8IA&content=IYAgsgngcsC2CmBGAUKSMECZzTk5QA).

### DIFF
--- a/packages/langium/src/languages/generated/ast.ts
+++ b/packages/langium/src/languages/generated/ast.ts
@@ -25,7 +25,7 @@ export function isAbstractRule(item: unknown): item is AbstractRule {
     return reflection.isInstance(item, AbstractRule);
 }
 
-export type AbstractType = Action | Interface | ParserRule | Type;
+export type AbstractType = Action | InferredType | Interface | ParserRule | Type;
 
 export const AbstractType = 'AbstractType';
 
@@ -684,6 +684,7 @@ export class LangiumGrammarAstReflection extends AbstractAstReflection {
             case ParameterReference: {
                 return this.isSubtype(Condition, supertype);
             }
+            case InferredType:
             case Interface:
             case Type: {
                 return this.isSubtype(AbstractType, supertype);
--- a/packages/langium/src/languages/generated/ast.ts
+++ b/packages/langium/src/languages/generated/ast.ts
@@ -25,7 +25,7 @@ export function isAbstractRule(item: unknown): item is AbstractRule {
     return reflection.isInstance(item, AbstractRule);
 }
 
-export type AbstractType = Action | InferredType | Interface | ParserRule | Type;
+export type AbstractType = InferredType | Interface | ParserRule | Type;
 
 export const AbstractType = 'AbstractType';
 
@@ -643,9 +643,7 @@ export class LangiumGrammarAstReflection extends AbstractAstReflection {
 
     protected override computeIsSubtype(subtype: string, supertype: string): boolean {
         switch (subtype) {
-            case Action: {
-                return this.isSubtype(AbstractElement, supertype) || this.isSubtype(AbstractType, supertype);
-            }
+            case Action:
             case Alternatives:
             case Assignment:
             case CharacterRange:
--- a/packages/langium/src/utils/grammar-utils.ts
+++ b/packages/langium/src/utils/grammar-utils.ts
@@ -237,6 +237,10 @@ export function findAssignment(cstNode: CstNode): ast.Assignment | undefined {
 export function findNameAssignment(type: ast.AbstractType): ast.Assignment | undefined {
     if (ast.isInferredType(type)) {
         // inferred type is unexpected, extract AbstractType first
+        if (ast.isAction(type.$container)) {
+            // a type which is explicitly inferred by an action
+            return undefined;
+        }
         type = type.$container;
     }
     return findNameAssignmentInternal(type, new Map());
@@ -254,7 +258,9 @@ function findNameAssignmentInternal(type: ast.AbstractType, cache: Map<ast.Abstr
         return childAssignment;
     }
 
-    if (cache.has(type)) return cache.get(type);
+    if (cache.has(type)) {
+        return cache.get(type);
+    }
     cache.set(type, undefined);
     for (const node of streamAllContents(type)) {
         if (ast.isAssignment(node) && node.feature.toLowerCase() === 'name') {
@@ -395,7 +401,7 @@ export function getExplicitRuleType(rule: ast.ParserRule): string | undefined {
     return undefined;
 }
 
-export function getTypeName(type: ast.AbstractType): string {
+export function getTypeName(type: ast.AbstractType | ast.Action): string {
     if (ast.isParserRule(type)) {
         return isDataTypeRule(type) ? type.name : getExplicitRuleType(type) ?? type.name;
     } else if (ast.isInterface(type) || ast.isType(type) || ast.isReturnType(type)) {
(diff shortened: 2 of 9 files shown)


JohannesMeierSE:
Yes, I was a bit surprised by the use of `Interface` in all cases, but I don't see a problem/bug in practise so far.

Additionally, I found another point: There might be an issue with multiple `InferredType`s with the same name, since all cross-references for this type point to the first `InferredType` in the grammar. It seems to work, but that looks a bit strange to me.

Maybe, the source for these discussions is the terminology/understanding of the various "kinds of types"?
1. We have "language types" like `Person` and `Greeting` in the hello world example.
2. `InferredType`, `ReturnType`, ... define the wanted language types _and_ are types for the AST nodes which represent the current Langium grammar.
3. The language types are realized in TypeScript as `type` or as `interface`.

As an example, when we have an `InferredType` in a parser rule, it is an AST node of type "InferredType" which explicitly defines the name of the inferred (language) type of the parser rule.

As another consequence, when we want to store the _language types_ in the global scope, all these types have the same type, which reflects the current situation (whether "interface" is the best name for this type is another discussion).

A problem might be, that the same language types can be defined multiple times.

spoenemann:
> Additionally, I found another point: There might be an issue with multiple InferredTypes with the same name, since all cross-references for this type point to the first InferredType in the grammar. It seems to work, but that looks a bit strange to me.

Yes, cross-references can be resolved only to one target element, but the same type can be inferred by multiple rules or actions in the grammar.

JohannesMeierSE:
> For implicitly inferred types (a rule with no `returns` and no `infers`), we could set it to `InferredType` – here we need to make a distinction between the parser rule and the corresponding type.

I am not yet convinced, since AST nodes of $type `InferredType` are used for implicitly _and_ explicitly inferred types: Why should we use `InferredType` in scopes _only_ for implicitly inferred types? That would confuse me.

As an alternative, we could use `InferredType` in scopes for _all_ inferred types, i.e. explicitly inferred types in actions ("{infer ...}"), explicitly inferred types in parser rules ("infers"), and implicitly inferred types in parser rules (neither "infers" nor "returns"). That would distinguish inferred types from explicitly declared `Type`s and `Interface`s.

packages/langium/src/languages/generated/ast.ts
```diff
@@ -25,7 +25,7 @@ export function isAbstractRule(item: unknown): item is AbstractRule {
     return reflection.isInstance(item, AbstractRule);
 }
 
-export type AbstractType = Action | Interface | ParserRule | Type;
```
Lotes, line 28: Why was Action removed here?

packages/langium/src/utils/grammar-utils.ts
```diff
@@ -234,29 +234,41 @@ export function findAssignment(cstNode: CstNode): ast.Assignment | undefined {
...
+    if (ast.isInferredType(startNode)) {
+        // for inferred types, the location to start searching for the name-assignment is different
+        if (ast.isAction(startNode.$container)) {
+            // a type which is explicitly inferred by an action: investigate the sibbling of the Action node, i.e. start searching at the Action's parent
+            startNode = startNode.$container.$container!;
+        } else if (ast.isParserRule(startNode.$container)) {
+            // investigate the parser rule with the explicitly inferred type
+            startNode = startNode.$container;
+        } else {
+            throw new Error('impossible');
```
Lotes, line 248: May use assertUnreachable?

packages/langium/src/languages/generated/ast.ts
JohannesMeierSE: In my eyes, an action is no type. An action changes the type of a node, but an action is no type itself.

Looking at actions and types, there are two different cases:
- The action uses an already declared type (`{WantedType}`). In this case, the action don't need to be covered by AbstractType, since `Interface | Type | ParserRule` are already covered, who declared the type used in the action.
- The action explicitly infers the wanted type now, since the wanted type is not yet declared (`{infer WantedType}`). In this case, the `InferredType` of the action is used as type and submitted to the global index, not the action itself. (Note, that this PR proposes `InferredType`s to be `AbstractType`s. Previously, the action was submitted to the global scope as type.)

Therefore, Actions don't need to be AbstractTypes anymore.

With other words, there are no changes in the behavior of abstract types and actions, only the wording changed.

packages/langium/src/utils/grammar-utils.ts
JohannesMeierSE: Yes, that is much better, thanks for the hint!

spoenemann (approved):
The changes make sense, thank you!

PR #1330, opened 2024-01-05
Improvements in IndexManager

spoenemann:
- Don't remove the reference to the AstNode
- `remove` takes a single URI for consistency with other services
- Renamed protected properties for more clarity

### DIFF
--- a/packages/langium/src/workspace/document-builder.ts
+++ b/packages/langium/src/workspace/document-builder.ts
@@ -153,8 +153,8 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
         for (const deletedUri of deleted) {
             this.langiumDocuments.deleteDocument(deletedUri);
             this.buildState.delete(deletedUri.toString());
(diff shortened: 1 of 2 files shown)


spoenemann:
Some context about deleting the `node` property in AstNodeDescriptions:

- We did that in the initial version of the IndexManager out of fear that such a reference could become outdated. But we're clearing the index anyway when a document needs to be reindexed, so it shouldn't be a problem in practice.
- We also had in mind that one might want to unload a document from memory to reduce the overall memory consumption of a workspace. But we never tried that and so far we haven't seen the need for it, so we should leave that topic for later (seeing a real example where unloading is necessary). Enabling such unloading will require more work, anyway, because linking is currently synchronous, so we can't load documents during linking.
- Serializing the index is also a topic, but that can be done with explicit IndexManager methods to be added in a future version.

msujew (approved):
Great, looks good to me 👍

PR #1334, opened 2024-01-08
Ensure expected document state in LSP responses

msujew:
Related to https://github.com/eclipse-langium/langium/discussions/1290

In the discussion in particular, the console warning appears due to the GLSP handler not awaiting the linking phase. However, all of our own LSP handlers haven't been awaiting that phase, instead opting to respond as quickly as possible. This sometimes leads to unexpected behavior in the response or tries to resolve a cross reference before it is ready.

This change attempts to mitigate the risk of this issue appearing by adding a `waitUntil` method to the `DocumentBuilder`. We use that method in the language server to wait until the required data is available for the respective service to do their work. This is usually either `IndexedReferences` (for any service that requires to resolve cross references) and `Parsed` (for services that operate on the AST/CST directly).

The new `waitUntil` method can also be interrupted in case it is no longer required. It will then `reject` (see tests).

### DIFF
--- a/packages/langium/src/workspace/document-builder.ts
+++ b/packages/langium/src/workspace/document-builder.ts
@@ -80,6 +80,11 @@ export interface DocumentBuilder {
      * Notify the given callback when a set of documents has been built reaching a desired target state.
      */
     onBuildPhase(targetState: DocumentState, callback: DocumentBuildListener): Disposable;
+
+    /**
+     * Wait until the workspace has reached the specified state for all documents.
+     */
+    waitUntil(state: DocumentState): Promise<void>;
 }
 
 export type DocumentUpdateListener = (changed: URI[], deleted: URI[]) => void | Promise<void>
@@ -100,6 +105,7 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
     protected readonly updateListeners: DocumentUpdateListener[] = [];
     protected readonly buildPhaseListeners: MultiMap<DocumentState, DocumentBuildListener> = new MultiMap();
     protected readonly buildState: Map<string, DocumentBuildState> = new Map();
+    protected currentState: DocumentState = DocumentState.Changed;
 
     constructor(services: LangiumSharedCoreServices) {
         this.langiumDocuments = services.workspace.LangiumDocuments;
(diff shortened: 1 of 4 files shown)


packages/langium/test/workspace/document-builder.test.ts
```diff
@@ -238,6 +239,88 @@ describe('DefaultDocumentBuilder', () => {
...
+        const services = await createServices();
+        const documentFactory = services.shared.workspace.LangiumDocumentFactory;
+        const documents = services.shared.workspace.LangiumDocuments;
+        const builder = services.shared.workspace.DocumentBuilder;
+        const document = documentFactory.fromString('', URI.parse('file:///test1.txt'));
+        documents.addDocument(document);
+
+        const cancelTokenSource = new CancellationTokenSource();
+        builder.waitUntil(DocumentState.IndexedReferences, cancelTokenSource.token).then(() => {
+            fail('This should have been cancelled');
```
sailingKieler, line 314: So, this is assumed to be never reached? Please leave a comment, or make that explicit in the message.

vite.config.mts
```diff
@@ -18,7 +18,7 @@ export default defineConfig({
             interopDefault: true
         },
         include: ['**/test/**/*.test.ts'],
-        exclude: ['**/node_modules/**', '**/dist/**', '**/generated/**', '**/templates/**'],
+        exclude: ['**/node_modules/**', '**/dist/**', '**/generated/**', '**/templates/**', '**/examples/hello*/**'],
```
sailingKieler, line 21: That's most likely not related to this PR, right?
Is the code `examples/hello*/` a left-over of a previous run of `generator-langium` tests?

sailingKieler (commented):
I like the contribution of this PR, as it provides an explicit way of awaiting the completion of the builder. On the other hand, it's kind of a sugar API based on the build listeners, right, with the additional support for cancelation checking?

Actually, I'm undecided whether it's worth exposing it on the document builder. It could also be added to the `LanguageServer` as it's (mainly) used there, couldn't it?

vite.config.mts
msujew: Not necessarily, but I've found that the `npm run test` command restarts once the yeoman generator test has passed, since it updates the files. Adding this exclude rule seems to fix it.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -79,6 +80,25 @@ export interface DocumentBuilder {
...
+     * @param cancelToken Optionally allows to cancel the wait operation, disposing any listeners in the process
+     * @throws `OperationCancelled` if cancellation has been requested before the state has been reached
+     */
+    waitUntil(state: DocumentState, cancelToken?: CancellationToken): Promise<void>;
+
+    /**
+     * Wait until the document specified by the {@link uri} has reached the specified state.
+     *
+     * @param state The desired state. The promise won't resolve until the document has reached this state
+     * @param uri The specified URI that points to the document. If the URI does not exist, the promise will resolve once the workspace has reached the specified state.
```
sailingKieler, line 97: I just stumbled over
> If the URI does not exist, the promise will resolve once the workspace has reached the specified state.

Would it be possible to reflect that in the result type, like `Promise<URI | undefined>` returning the URI if a corresponding document exists?

packages/langium/src/workspace/document-builder.ts
```diff
@@ -290,6 +314,7 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
             document.state = targetState;
         }
         await this.notifyBuildPhase(filtered, targetState, cancelToken);
+        this.currentState = targetState;
```
spoenemann, line 317: Shall we set the workspace state before notifying listeners?

spoenemann (approved):
Good improvement!

PR #1335, opened 2024-01-08
Generate grammar comments

msujew:
In https://github.com/eclipse-langium/langium/pull/1095 we actually forgot to enable the serializer to also serialize the comments. This change rectifies that oversight. See also https://github.com/eclipse-langium/langium/discussions/1333.


packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -1,8 +1,9 @@
-/******************************************************************************
- * Copyright 2021 TypeFox GmbH
- * This program and the accompanying materials are made available under the
- * terms of the MIT License, which is available in the project root.
- ******************************************************************************/
+// ******************************************************************************
```
sailingKieler, line 1: I guess this comment formatting is changed to be not recognized as JSdoc, right?

sailingKieler (commented):
LGTM. However, I would prefer to have a dedicated unit test "formally" documenting the feature in addition to the 'integration' test in the domain model language.

packages/langium/src/grammar/langium-grammar.langium
msujew: Correct 👍

sailingKieler (approved):
🚀

PR #1343, opened 2024-01-16
Allow hidden terminals reference in terminal lookahead

### LINKED ISSUE #1336: Allow usage of hidden tokens in terminal lookahead
Right now, the Langium validation marks usage of hidden tokens inside of the terminal lookahead (`?=...`) with an error. This makes writing these lookaheads quite annoying, especially if you want to allow comments inside of them.

gfontorbe:
Closes #1336 

Updated the validation rule `checkUsedHiddenTerminalRule` to not send an error if the referenced hidden terminal is part of a lookahead.

This should also work for hidden terminal present in arbitrarily nested `TerminalGroup`s


msujew (approved):
Thanks, looks good to me 👍

Issue #1358, opened 2024-01-25
Data type using one terminal behaves differently from that terminal


msujew:
### Discussed in https://github.com/eclipse-langium/langium/discussions/1357

<div type='discussions-op-text'>

<sup>Originally posted by **drhagen** January 25, 2024</sup>
If I have this grammar

```langium
grammar HelloWorld

entry Model:
    ((persons+=Person) EOL)*;

Person:
    'person' name=ID;

hidden terminal WS: /[ \t]+/;
terminal ID: /[_a-zA-Z][\w_]*/;
terminal NEWLINE: '\n';
EOL returns string:
    NEWLINE;
```

and try to parse this file

```
person John
person Jane
```

I get an error "Expecting end of file but found 'person'".

![image](https://github.com/eclipse-langium/langium/assets/2952478/2c78c6d1-7274-410e-814d-ee424d5888aa)

If I replace `EOL` with `NEWLINE` and delete `EOL` entirely

```langium
grammar HelloWorld

entry Model:
    ((persons+=Person) NEWLINE)*;

Person:
    'person' name=ID;

hidden terminal WS: /[ \t]+/;
terminal ID: /[_a-zA-Z][\w_]*/;
terminal NEWLINE: '\n';
```

the file parses successfully. What am I missing something about Langium because I would have expected those to be equivalent?</div>

msujew:
Seems like we only ever assume that data type rules call other data type rules in an unassigned way, see: https://github.com/eclipse-langium/langium/blob/0239db7b353bf06f5c0710e40bd264fe17fccb55/packages/langium/src/parser/langium-parser.ts#L236-L251

I assume the solution would be to just ignore the result and store the CST created in the current AST node.

PR #1367, opened 2024-01-31
Fix unassigned data type rule usage

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1358

Adds a check that the value we're trying assign is actually an object and not the result of a data type rule (a primitive).


packages/langium/src/parser/langium-parser.ts
```diff
@@ -19,6 +19,7 @@ import { isAssignment, isCrossReference, isKeyword } from '../languages/generate
 import { getTypeName, isDataTypeRule } from '../utils/grammar-utils.js';
 import { assignMandatoryProperties, getContainerOfType, linkContentToContainer } from '../utils/ast-utils.js';
 import { CstNodeBuilder } from './cst-node-builder.js';
+import { isAstNode } from '../../lib/syntax-tree.js';
```
spoenemann, line 22:
```suggestion
import { isAstNode } from '../syntax-tree.js';
```

spoenemann (approved):
👍

PR #1381, opened 2024-02-16
Remove the InitializableService interface

spoenemann:
Repeating my comment from https://github.com/eclipse-langium/langium/pull/1258#discussion_r1492320068:

Looking at it again, I'd rather remove the InitializableService interface: we already have `onInitialize` and `onInitialized` events in the LanguageServer service, which are the recommended ways to hook into the LSP lifecycle. The interface is only used by two core framework classes which require to invert the dependency. The interface is not necessary because LanguageServer hard-codes the calls to these classes anyway.


msujew (approved):
I'm in favor of this change. Looks good to me 👍

sailingKieler:
Actually I didn't want to pollute the service interface definitions like `ConfigurationProvider` and `WorkspaceManager` with the initialization stuff. In my customer project there's the need for more non-LSP services to be initialized with the workspace folders, so from my point of view the interface describes a standard way to implement that.

I also still have an idea in mind that is about providing an initialization hook via the constructors of `DefaultConfigurationProvider` and `DefaultWorkspaceManager` (and probably others), similar to the language server-based approach, but I didn't get to trying that, yet.

I don't have a strong opinion here.

sailingKieler (commented):
see comment above

spoenemann:
> In my customer project there's the need for more non-LSP services to be initialized with the workspace folders, so from my point of view the interface describes a standard way to implement that.

Do you need to separate the core part from the lsp part in that project? If not, you can simply inject `LanguageServer` in those non-LSP services and use the `onInitialize` event.

spoenemann:
I added a change to avoid problems during initialization: WorkspaceManager and ConfigurationProvider run their `initialized` hook in parallel.

Issue #1371, opened 2024-02-05
getTypeMetaData does not consider returned types


cdietrich:
Given the grammar

```
interface IAmArray {
   elements: ArrayContent[];
}
interface DeclaredArray extends IAmArray{ }

DeclaredArray returns DeclaredArray:
     'declared' (elements+=ArrayContent)* ';';

ArrayContent:
    name=ID;
```

`AstReflection.getTypeMetaData` wont flag the elements of DeclaredArray as array and thus it wont have a default value

cdietrich:
alternatively: `assignMandatoryAstProperties` should also check return types.

msujew:
The issue is that the generated ast reflection currently only considers properties of the current interface, not of the whole interface chain:

https://github.com/eclipse-langium/langium/blob/1f5d1246692aafb70672fb6ce695fefc4e54767c/packages/langium-cli/src/generator/ast-generator.ts#L84

cdietrich:
nice. thx for the fix. works now as expected

PR #1383, opened 2024-02-19
Include inherited properties in type meta data

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1371


spoenemann (approved):
That was an easy fix 😁

cdietrich:
nice. will there be another alpha before the release?

montymxb:
@cdietrich Most likely, we were discussing it in the Langium dev meeting yesterday. Another next should be coming up shortly, probably before the end of the week.

PR #1385, opened 2024-02-20
langium/lsp: Added 'getReferenceCandidates(...)' hook to 'DefaultCompletionProvider'

sailingKieler:
… allowing to add the completion-specific scope computation refinements.

### DIFF
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -24,7 +24,7 @@ import * as ast from '../../languages/generated/ast.js';
 import { assignMandatoryProperties, getContainerOfType } from '../../utils/ast-utils.js';
 import { findDeclarationNodeAtOffset, findLeafNodeBeforeOffset } from '../../utils/cst-utils.js';
 import { getEntryRule, getExplicitRuleType } from '../../utils/grammar-utils.js';
-import { stream } from '../../utils/stream.js';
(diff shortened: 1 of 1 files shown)


spoenemann (commented):
I think this makes sense and I confirm that it has been often useful in Xtext.

Thought: shall the template method also include the call to `scope.getAllElements()`? That means it would return a Stream, making it very easy to filter elements from the scope without the need to manipulate the scope hierarchy.

sailingKieler:
> Thought: shall the template method also include the call to scope.getAllElements()? That means it would return a Stream, making it very easy to filter elements from the scope without the need to manipulate the scope hierarchy.

Hmm ... I don't have strong opinion here, we could do that yes.

In my case I only manipulated the `ReferenceInfo` depending on some conditions wrt. the `CompetionContext`, and then delegated to the scope provider and returned its result.

We could also take `{ getAllElements: () => Stream<AstNodeDescription> }` as return type.
Then a `Scope` would be a valid result, but something like the following is also possible.

```
    .
    .
    .
    return {
        getAllElements: () => scope.getAllElements().filter(e => ... ),
    };
```

sailingKieler:
You may be right, but then we need a better template method name.

spoenemann (approved):
Looks good! I totally overlooked that we already had `filterCrossReference`, but the new API is better IMO, unifying the two template methods into one.

PR #1386, opened 2024-02-22
re-fix configuration provider initialization

sailingKieler:
This change restores the original configuration provider initialization approach contributed in #519.

To be tested:
* Proper application of the `langium.build.ignorePatterns` during workspace init of the grammar language server (see https://github.com/eclipse-langium/langium/blob/1bc7009287d46ef82b5432f8eabd5f924cbd658c/packages/langium-vscode/src/language-server/grammar-workspace-manager.ts#L33-L38), expected default values are defined here: https://github.com/eclipse-langium/langium/blob/main/packages/langium-vscode/package.json#L69-L78.
--> Looks good AFAICS.

### DIFF
--- a/packages/langium/src/lsp/language-server.ts
+++ b/packages/langium/src/lsp/language-server.ts
@@ -199,13 +199,16 @@ export class DefaultLanguageServer implements LanguageServer {
         const configurationParams = connection ? <ConfigurationInitializedParams>{
             ...params,
             register: params => connection.client.register(DidChangeConfigurationNotification.type, params),
-            getConfiguration: params => connection.workspace.getConfiguration(params)
+            fetchConfiguration: params => connection.workspace.getConfiguration(params)
         } : params;
 
-        this.services.workspace.WorkspaceManager.initialized(params)
-            .catch(err => console.error('Error in WorkspaceManager initialization:', err));
+        // do not await the promises of the following calls, as they must not block the initialization process!
+        //  otherwise, there is the danger of out-of-order processing of subsequent incoming messages from the language client
+        // however, awaiting should be possible in general, e.g. in unit test scenarios
         this.services.workspace.ConfigurationProvider.initialized(configurationParams)
             .catch(err => console.error('Error in ConfigurationProvider initialization:', err));
+        this.services.workspace.WorkspaceManager.initialized(params)
+            .catch(err => console.error('Error in WorkspaceManager initialization:', err));
     }
 }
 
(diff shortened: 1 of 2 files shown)


spoenemann (approved):
The changes look good to me. @dhuebner could you test this with the grammar language?

msujew (commented):
All of this starts feeling very convoluted. I think this change is a symptom of another problem in our language server logic. I'd like to present a different flow of data to improve on this and clean up the implementations a bit:

1. Make both `initialized` methods async again 
2. Add `ready` promises to the `ConfigurationProvider` and `WorkspaceManager` interfaces. These are resolved as soon as the configuration is available and the workspace has been loaded (but not built!) respectively.
3. Await these promises at the appropriate times (i.e. for example in the `textDocument/didChange` handler) to ensure that everything actually waits for the required/expected events before continuing their work.

What is your opinion on that @sailingKieler and @spoenemann? This PR in its current implementation very much feels like a workaround for design issues in our language server, instead of an actual improvement.

sailingKieler:
> Especially the "switch the configuration and workspace manager initialized calls around" smells like a badly designed system to me. That should never be the fix in a system that is well designed.

Actually, the order shouldn't matter.
@spoenemann switched and fixed the invocations [here](https://github.com/eclipse-langium/langium/commit/1bc7009287d46ef82b5432f8eabd5f924cbd658c#diff-3216158ebf263e0a3ffbc838b9c1ba3f8419c253eff80fad2854b6cede80c5efL205-R208), and I just flipped them back for testing purposes (and a bit because of the ConfigProvider shouldn't depend on anything, while the WorkspaceManager might depend on the former), and committed that by accident.

msujew:
> @spoenemann switched and fixed the invocations [here](https://github.com/eclipse-langium/langium/commit/1bc7009287d46ef82b5432f8eabd5f924cbd658c#diff-3216158ebf263e0a3ffbc838b9c1ba3f8419c253eff80fad2854b6cede80c5efL205-R208)

I know, and I wasn't a big fan of that either, it was just merged before I was able to complain about it :)

spoenemann:
Switching was not the fix, removing `await` was the fix. In the general case, there is no dependency between WorkspaceManager and ConfigurationProvider. Languages that need the configuration to initialize the workspace can just use ConfigurationProvider, which awaits its initialization before returning a value.

> Await these promises at the appropriate times (i.e. for example in the textDocument/didChange handler) to ensure that everything actually waits for the required/expected events before continuing their work.

That sounds good to make this logical dependency more explicit.

I propose to merge this and add Mark's proposal in a follow-up PR.

PR #1388, opened 2024-02-26
Improve workspace manager startup behavior

msujew:
Updates our `ConfigurationProvider` and `WorkspaceManager` to expose `ready` promises. Downstream services can use these promises to ensure that the services are ready to be used.

### DIFF
--- a/packages/langium/src/lsp/document-update-handler.ts
+++ b/packages/langium/src/lsp/document-update-handler.ts
@@ -72,11 +72,15 @@ export class DefaultDocumentUpdateHandler implements DocumentUpdateHandler {
         }
     }
 
-    protected async fireDocumentUpdate(changed: URI[], deleted: URI[]): Promise<void> {
+    protected fireDocumentUpdate(changed: URI[], deleted: URI[]): void {
         // Only fire the document update when the workspace manager is ready
         // Otherwise, we might miss the initial indexing of the workspace
-        await this.workspaceManager.ready;
-        this.workspaceLock.write(token => this.documentBuilder.update(changed, deleted, token));
+        this.workspaceManager.ready.then(() => {
+            this.workspaceLock.write(token => this.documentBuilder.update(changed, deleted, token));
+        }).catch(err => {
+            // This should never happen, but if it does, we want to know about it
+            console.error('Workspace initialization failed. Could not perform document update.', err);
+        });
     }
 
     didChangeContent(change: TextDocumentChangeEvent<TextDocument>): void {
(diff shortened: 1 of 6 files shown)


packages/langium/src/lsp/document-update-handler.ts
```diff
@@ -69,7 +72,10 @@ export class DefaultDocumentUpdateHandler implements DocumentUpdateHandler {
         }
     }
 
-    protected fireDocumentUpdate(changed: URI[], deleted: URI[]): void {
+    protected async fireDocumentUpdate(changed: URI[], deleted: URI[]): Promise<void> {
```
spoenemann, line 75: When an async method is called in a synchronous context, `catch` blocks must be appended to avoid uncaught rejected promises. It might be easier to leave the method synchronous and use `then` with two callbacks in here.

packages/langium/src/workspace/configuration.ts
```diff
@@ -77,34 +87,23 @@ export class DefaultConfigurationProvider implements ConfigurationProvider {
                 });
             }
 
-            // params.fetchConfiguration(...) is a function to be provided by the calling language server for the sake of
-            //  decoupling this implementation from the concrete LSP implementations, specifically the LSP Connection
```
spoenemann, line 81: Please copy this comment to `initializeConfiguration`

packages/langium/src/workspace/configuration.ts
```diff
@@ -77,34 +87,25 @@ export class DefaultConfigurationProvider implements ConfigurationProvider {
...
-            // consequently, we fetch the initial configuration data on blocking manner on-demand, when the first configuration value is read,
-            //  see below in #getConfiguration(...);
+            await this.initializeConfiguration(params);
         }
+        this._ready.resolve();
     }
 
-    protected async initializeConfiguration(): Promise<void> {
-        if (this.fetchConfiguration) {
+    protected async initializeConfiguration(params: ConfigurationInitializedParams): Promise<void> {
```
sailingKieler, line 95: If you implement it this way you then could merge the two methods, i.e. inline the content of `initializeConfiguration()`, or do you expect any need for the latter as a customization hook?

packages/langium/src/workspace/document-builder.ts
```diff
@@ -254,8 +254,8 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
     protected async buildDocuments(documents: LangiumDocument[], options: BuildOptions, cancelToken: CancellationToken): Promise<void> {
         this.prepareBuild(documents, options);
         // 0. Parse content
-        await this.runCancelable(documents, DocumentState.Parsed, cancelToken, doc => {
-            this.langiumDocumentFactory.update(doc, cancelToken);
+        await this.runCancelable(documents, DocumentState.Parsed, cancelToken, async doc => {
+            await this.langiumDocumentFactory.update(doc, cancelToken);
```
spoenemann, line 258: A bit simpler without `async`: remove the curly braces, i.e. return the result of the service call (like in 1. Index content)

spoenemann (approved):

PR #1389, opened 2024-02-27
Fix async parser implementation

msujew:
The current implementation has a few issues in the hydrator, especially related to CST transmission. This change improves on these issues and adds tests to ensure that the CST is correctly transmitted across worker threads.

### DIFF
--- a/packages/langium/src/serializer/hydrator.ts
+++ b/packages/langium/src/serializer/hydrator.ts
@@ -12,25 +12,26 @@ import { isAbstractElement, type AbstractElement, type Grammar } from '../langua
 import type { Linker } from '../references/linker.js';
 import type { Lexer } from '../parser/lexer.js';
 import type { LangiumCoreServices } from '../services.js';
-import type { Reference, AstNode, CstNode, LeafCstNode, GenericAstNode, Mutable } from '../syntax-tree.js';
+import type { ParseResult } from '../parser/langium-parser.js';
+import type { Reference, AstNode, CstNode, LeafCstNode, GenericAstNode, Mutable, RootCstNode } from '../syntax-tree.js';
 import { isRootCstNode, isCompositeCstNode, isLeafCstNode, isAstNode, isReference } from '../syntax-tree.js';
 import { streamAst } from '../utils/ast-utils.js';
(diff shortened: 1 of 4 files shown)


packages/langium/src/serializer/hydrator.ts
```diff
@@ -12,25 +12,26 @@ import { isAbstractElement, type AbstractElement, type Grammar } from '../langua
...
-    dehydrate(node: AstNode): object;
+    dehydrate(result: ParseResult<AstNode>): object;
     /**
-     * Converts a plain object to an AST node. The resulting AST node can be used in the main thread.
-     * Calling this method on non-plain objects will result in undefined behavior.
+     * Converts a plain object to a parse result. The resulting AST node can be used in the main thread.
+     * Calling this method on objects that have not been dehydrated first will result in undefined behavior.
      */
-    hydrate(node: object): AstNode;
+    hydrate<T extends AstNode = AstNode>(result: object): ParseResult<T>;
```
Lotes, line 34: I like types. You are expecting a certain structure down below. This structure should be reflected by a type definition and not by `object`.

```ts
interface {
  lexerErrors: LexerError[],
  parserErrors: parserError[],
  value: object //for this guy you are not able to assume a type
};
```
WDYT?

msujew: Yep, thanks. That's indeed better 👍

packages/langium/src/serializer/hydrator.ts
```diff
@@ -198,15 +214,15 @@ export class DefaultHydrator implements Hydrator {
                 astNode[name] = arr;
                 for (const item of value) {
```
Lotes, line 215: The code between Line 212 until 229 looks duplicated. You may find a solution in the shape of a helper function.

montymxb (approved):
Alright, this looks pretty good to me, tests look nice as well.

sailingKieler (approved):
Just wording improvement, and I agree with @Lotes, otherwise it's great! 🚀

packages/langium/src/serializer/hydrator.ts
msujew: We can perform the cleanup in another PR.

PR #1390, opened 2024-02-28
Update version to v3

msujew:
Preparation for the v3 release.


sailingKieler (approved):
Tested the current state on Gitpod:
 * an example language
 * grammar language incl. railroad diagrams
 * langium-cli

Issue #1309, opened 2023-11-30
Parser rule used as type for cross-reference is marked as "unused"


JohannesMeierSE:
The following grammar uses the (unassigned) parser rule `Person` as type in the cross-reference of `Greeting` in order to link to `Neighbor`s or `Friend`s:

```langium
grammar HelloWorld

entry Model:
    (persons+=Neighbor | friends+=Friend | greetings+=Greeting)*;

Neighbor:
    'neighbor' name=ID;
Friend:
    'friend' name=ID;

Person: Neighbor | Friend;

Greeting:
    'Hello' person=[Person:ID] '!';

hidden terminal WS: /\s+/;
terminal ID: /[_a-zA-Z][\w_]*/;
```

## The current behavior

Nevertheless, the parser rule `Person` is marked as "unused", e.g. in the [Playground](https://langium.org/playground?grammar=OYJwhgthYgBAEgUwDbIPYHU0mQEwFD6IB2ALiAJ6wCyauKAXPrC7ABQAOiIAzmsTwDUAXgByiAJbAAFgCNssAD6wAZiAklcQ4QDF1mpbFCJEpCcWDaA4iBNmLASgBUAbkLipc7E1awA5MSSMvIgfrDEkIjCAJIAIm56GsS4Pqx%2Bakm4YREQUXFu%2BAAK3HzEDLAewQrKiZoFNnbmwKksfkioaGFcvPzCANrFPWVxALr%2BAIR%2BBdISuPTEsKTcEOZgyLAYAMrlAPQAOkI7bksgKxHrcbt9APpgALQAXgCCdwBaI317AO7XI05H%2BCAA&content=HYUwlg5gFgRg9gJwAQCk5WAKAGYLCYAE1QENRNMAJEAGxrlXWAEIrb7TRWg):

<img width="586" alt="Screenshot 2023-11-30 at 10 54 04" src="https://github.com/eclipse-langium/langium/assets/23192240/7bd08197-16b9-44a6-9643-ec277fdae02b">

## The expected behavior

I would expect, that `Person` is not marked as "unused", since it is used as type in the cross-reference.

JohannesMeierSE:
I am willing to work on this issue.

PR #1391, opened 2024-02-29
Support users to replace parser rules which are used only as type for cross-references by type declarations

JohannesMeierSE:
This PR contributes the following things:
- At the moment parser rules which are used not for parsing but whose type is used in cross-references are marked as "unused" for users. This message is not accurate and could confuse inexperienced users. Therefore, the message of the validation is improved to "This parser rule is not used for parsing, but referenced by cross-references. Consider to replace this rule by a type declaration.". See #1309 for details.
- Additionally, a quick-fix is provided for users to replace such a parser rule by a type declaration.
- While the two contributions above target a quite special case, this PR contributes also a generic helper functionality to check quick-fixes (`testQuickFix` in `packages/langium/src/test/langium-test.ts`). This function is tested by applying it to the mentioned new quick-fix for the Langium grammar.

### DIFF
--- a/packages/langium/src/grammar/lsp/grammar-code-actions.ts
+++ b/packages/langium/src/grammar/lsp/grammar-code-actions.ts
@@ -184,8 +184,10 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
     }
 
     private isRuleReplaceable(rule: ast.ParserRule): boolean {
-        // OK are: definition use only Alternatives! infers!
-        // "returns" is not relevant, since cross-references would not refer to the parser rule, but to its "return type" instead
+        /** at the moment, only "pure" parser rules are supported:
+         * - supported are only Alternatives (recursively) and "infers"
+         * - "returns" is not relevant, since cross-references would not refer to the parser rule, but to its "return type" instead
+         */
         return !rule.fragment && !rule.entry && rule.parameters.length === 0 && !rule.definesHiddenTokens && !rule.wildcard && !rule.returnType && !rule.dataType;
     }
     private replaceRule(rule: ast.ParserRule): string {
@@ -199,9 +201,6 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
         if (ast.isAlternatives(node)) {
             return node.elements.every(child => this.isDefinitionReplaceable(child));
         }
-        if (ast.isUnorderedGroup(node)) {
-            return node.elements.every(child => this.isDefinitionReplaceable(child));
-        }
         return false;
     }
     private replaceDefinition(node: ast.AbstractElement, requiresBraces: boolean): string {
@@ -210,10 +209,6 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
         }
         if (ast.isAlternatives(node)) {
             const result = node.elements.map(child => this.replaceDefinition(child, true)).join(' | ');
-            return requiresBraces && node.elements.length >= 2 ? `(${result})` : result;
-        }
-        if (ast.isUnorderedGroup(node)) {
-            const result = node.elements.map(child => this.replaceDefinition(child, true)).join(' & '); // TODO
             return requiresBraces && node.elements.length >= 2 ? `(${result})` : result;
         }
         throw new Error('missing code for ' + node);
--- a/packages/langium/src/grammar/lsp/grammar-code-actions.ts
+++ b/packages/langium/src/grammar/lsp/grammar-code-actions.ts
@@ -5,25 +5,25 @@
  ******************************************************************************/
 
 import type { Diagnostic } from 'vscode-languageserver';
+import { CodeActionKind } from 'vscode-languageserver';
 import type { CodeActionParams } from 'vscode-languageserver-protocol';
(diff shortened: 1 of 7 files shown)


JohannesMeierSE:
FYI: I just resolved conflicts with `main`. This PR is ready to review.

packages/langium/src/grammar/lsp/grammar-code-actions.ts
```diff
@@ -180,6 +183,67 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
...
+        if (rootCst) {
+            const offset = document.textDocument.offsetAt(diagnostic.range.start);
+            const cstNode = findLeafNodeAtOffset(rootCst, offset);
+            const rule = getContainerOfType(cstNode?.astNode, ast.isParserRule);
+            if (rule && rule.$cstNode) {
+                const isRuleReplaceable = this.isRuleReplaceable(rule) && this.isDefinitionReplaceable(rule.definition);
+                if (isRuleReplaceable) {
+                    const newText = `type ${this.replaceRule(rule)} = ${this.replaceDefinition(rule.definition, false)};`;
+                    return {
+                        title: 'Replace parser rule by type declaration',
```
msujew, line 228: Suggestion: This is a fairly long title for a code action
```suggestion
                        title: 'Replace with type declaration',
```

packages/langium/src/grammar/lsp/grammar-code-actions.ts
```diff
@@ -180,6 +183,67 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
...
+    private isDefinitionReplaceable(node: ast.AbstractElement): boolean {
+        if (ast.isRuleCall(node)) {
+            return node.arguments.length === 0 && ast.isParserRule(node.rule.ref) && this.isRuleReplaceable(node.rule.ref);
+        }
+        if (ast.isAlternatives(node)) {
+            return node.elements.every(child => this.isDefinitionReplaceable(child));
+        }
+        return false;
+    }
+    private replaceDefinition(node: ast.AbstractElement, requiresBraces: boolean): string {
```
msujew, line 206: Suggestion: As for the `requiresBraces`; I don't think this is relevant and we can simply place all types in one single union.

packages/langium/src/test/langium-test.ts
```diff
@@ -769,3 +768,103 @@ export function expectSemanticToken(tokensWithRanges: DecodedSemanticTokensWithR
...
+
+/**
+ * This is a helper function to easily test quick-fixes for validation problems.
+ * @param services the Langium services for the language with quick fixes
+ * @returns A function to easily test a single quick-fix on the given invalid 'input'.
+ * This function expects, that 'input' contains exactly one validation problem with the given 'diagnosticCode'.
+ * If 'outputAfterFix' is specified, this functions checks, that the diagnostic comes with a single quick-fix for this validation problem.
+ * After applying this quick-fix, 'input' is transformed to 'outputAfterFix'.
+ */
+export function testQuickFix<T extends AstNode = AstNode>(services: LangiumServices): (input: string, diagnosticCode: string, outputAfterFix: string | undefined, options?: ParseHelperOptions) => Promise<QuickFixResult<T>> {
```
msujew, line 791: Suggestion: I would keep the `CodeAction` naming convention from the LSP instead of using `QuickFix` (which we don't use anywhere else in the code base).

packages/langium/src/grammar/lsp/grammar-code-actions.ts
```diff
@@ -180,6 +183,67 @@ export class LangiumGrammarCodeActionProvider implements CodeActionProvider {
         return undefined;
     }
 
+    private isRuleReplaceable(rule: ast.ParserRule): boolean {
+        /** at the moment, only "pure" parser rules are supported:
```
msujew, line 187: Question: Do you plan to support more parser rules? I.e. those that should be transformed into interfaces?

packages/langium/src/grammar/lsp/grammar-code-actions.ts
JohannesMeierSE: Changed

packages/langium/src/grammar/lsp/grammar-code-actions.ts
JohannesMeierSE: At the moment, I don't plan to support more parser rules. In practice, I saw only the handled "simple" cases. I am not sure whether it is worth the effort to support more cases. Do you have some examples we should handle?

packages/langium/src/grammar/lsp/grammar-code-actions.ts
JohannesMeierSE: Yes, for the current cases it is not required: I removed it to make the code simpler. If we need it in the future, we can re-introduce it.

JohannesMeierSE:
Thanks @msujew for your review! I just pushed the corresponding fixes.

spoenemann:
Could this make it into v3.5.0, or rather defer to v4.0.0?

msujew (approved):
Thank you, looks good to me 👍

Issue #1392, opened 2024-02-29
`createDefaultCoreModule` isn't compatible with `ExampleGeneratedModule` (`langium` v3.0.0)


Yokozuna59:
Langium version: 3.0.0
Package name: `langium`

## Steps To Reproduce

1. Create an example language using `langium` v3.0.0. (Or use an existing one)
2. Modify the language service `inject` function parameter to use `createDefaultCoreModule` instead of `createDefaultModule`.

https://github.com/eclipse-langium/langium/blob/be0407ef03397fe04bf2dfbab387e13034d92279/examples/arithmetics/src/language-server/arithmetics-module.ts#L65-L69

Link to code example: You could use the existing examples in the repo.

Just modify those lines of code to this:

```diff
 const arithmetics = inject( 
-    createDefaultModule({ shared }), 
+    createDefaultCoreModule({ shared }), 
     ArithmeticsGeneratedModule, 
     ArithmeticsModule 
 );
```

## The current behavior

Errors are thrown.

<details>
<summary>The errors</summary>

```
error TS2345: Argument of type 'Module<LangiumServices, LangiumGeneratedCoreServices>' is not assignable to parameter of type 'Module<LangiumDefaultCoreServices & LangiumGeneratedCoreServices & { Grammar?: { readonly $type?: "Grammar" | undefined; ... 13 more ...; readonly $document?: LangiumDocument<...> | undefined; } | undefined; ... 7 more ...; shared?: { ...; } | undefined; } & InfoAddedServices, LangiumGeneratedCoreServices>'.
  Types of property 'Grammar' are incompatible.
    Type 'Module<LangiumServices, Grammar> | ((injector: LangiumServices) => Grammar)' is not assignable to type 'Module<LangiumDefaultCoreServices & LangiumGeneratedCoreServices & { Grammar?: { readonly $type?: "Grammar" | undefined; ... 13 more ...; readonly $document?: LangiumDocument<...> | undefined; } | undefined; ... 7 more ...; shared?: { ...; } | undefined; } & InfoAddedServices, Grammar> | ((injector: LangiumDefaultCo...'.
      Type 'Module<LangiumServices, Grammar>' is not assignable to type 'Module<LangiumDefaultCoreServices & LangiumGeneratedCoreServices & { Grammar?: { readonly $type?: "Grammar" | undefined; ... 13 more ...; readonly $document?: LangiumDocument<...> | undefined; } | 
[...]

PR #1393, opened 2024-02-29
Use correct module type for generated module

msujew:
Related to https://github.com/eclipse-langium/langium/issues/1392

Instead of being part of `LangiumServices`, the module should instead be part of `LangiumCoreServices`. Afaik, the current solution is working fine, but using `LangiumCoreServices` is more appropriate.


msujew:
> The actual change is just in the template in langium-cli, isn't it?

Yes, the rest is just regenerated code 👍

PR #1394, opened 2024-02-29
Correctly pass parser error message

msujew:
For some reason, copying an error via the spread operator doesn't copy its `message` property. This change ensures that the message property is being copied.


spoenemann (approved):
The reason is that the `RecognitionException` class extends the standard `Error` class. The problem shouldn't occur with `ILexingError`, though: Chevrotain uses plain objects for that.

PR #1402, opened 2024-03-01
Initialize configuration provider when creating services outside a language server

### LINKED ISSUE #1397: `DefaultConfigurationProvider.getConfiguration` stalls in tests
After upgrading from langium 3.0.0-next.e78aeba to langium 3.0.0, many of the tests in [Safe-DS](https://github.com/Safe-DS/DSL) time out. In particular, it's all tests that validate documents. This happens because some of our validation rules can be turned off using VS Code settings, so they contain code like

```ts
if (!(await settingsProvider.shouldValidateNameConvention())) {
    return;
}
```

The `settingsProvider` is just a thin wrapper around the `DefaultConfigurationProvider` and calls its `getConfiguration` method. 

➡️ #1388 is likely related to this.

Langium version: 3.0.0
Package name: langium

## Steps To Reproduce

Substitute `SafeDs` with some other language in the following code and run the tests with `vitest`:

```ts
import { describe, it } from 'vitest';
import { createSafeDsServices, SafeDsLanguageMetaData } from '../../../src/language/index.js';
import { NodeFileSystem } from 'langium/node';

const services = createSafeDsServices(NodeFileSystem).SafeDs;

describe('time out', () => {
    it('should not time out', async () => {
        await services.shared.workspace.ConfigurationProvider.getConfiguration(
            SafeDsLanguageMetaData.languageId,
            'validation',
        );
    });
});
```

Output:

```txt
Error: Test timed out in 5000ms.
If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".
    at Timeout.<anonymous> (file:///E:/Repositories/safe-ds/DSL/node_modules/@vitest/runner/dist/index.js:63:16)
    at listOnTimeout (node:internal/timers:573:17)
    at processTimers (node:internal/timers:514:7)
```

Link to code example:

* [Relevant PR](https://github.com/Safe-DS/DSL/pull/942)
* [Relevant workflow run](https://github.com/Safe-DS/DSL/actions/runs/8110154872/job/22166806456?pr=942)

## The current behavior

Tests time out because the service never readies up.

## The expected behavior

Tests should pass.

lars-reimann:
Closes #1397 

The function `create<Language>Services` that gets created from the Yeoman template now initializes the configuration provider immediately if called outside a language server. Previously, each caller (CLI/tests) had to take care of this before working with the configuration provider.

Existing projects must be updated manually.

### DIFF
--- a/examples/arithmetics/src/language-server/arithmetics-module.ts
+++ b/examples/arithmetics/src/language-server/arithmetics-module.ts
@@ -69,5 +69,10 @@ export function createArithmeticsServices(context: DefaultSharedModuleContext): 
     );
     shared.ServiceRegistry.register(arithmetics);
     registerValidationChecks(arithmetics);
+    if (!context.connection) {
+        // We don't run inside a language server
+        // Therefore, initialize the configuration provider instantly
+        shared.workspace.ConfigurationProvider.initialized({});
+    }
     return { shared, arithmetics };
 }
--- a/examples/domainmodel/src/language-server/domain-model-module.ts
+++ b/examples/domainmodel/src/language-server/domain-model-module.ts
@@ -54,5 +54,10 @@ export function createDomainModelServices(context: DefaultSharedModuleContext): 
     );
     shared.ServiceRegistry.register(domainmodel);
     registerValidationChecks(domainmodel);
+    if (!context.connection) {
+        // We don't run inside a language server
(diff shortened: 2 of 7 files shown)


msujew (approved):
Thanks, looks good to me 👍

PR #1407, opened 2024-03-07
Implemented option for parsing text with an alternative entry rule

spoenemann:
See discussion #1404. This PR adds an option to `LangiumParser` so one can change the entry rule used for parsing a given text.

### DIFF
--- a/packages/langium/src/parser/langium-parser.ts
+++ b/packages/langium/src/parser/langium-parser.ts
@@ -53,6 +53,7 @@ interface AssignmentElement {
 
 export interface BaseParser {
     rule(rule: ParserRule, impl: RuleImpl): RuleResult;
+    getRule(name: string): RuleResult | undefined;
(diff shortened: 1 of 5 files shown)


cdietrich:
nice

Lotes:
Really nice! One concern I have found is scoping. When there is a cross-reference that needs to be resolved, the scope provider writer cannot assume anymore that there are "natural" parents given. Like `getContainerOfType` would just return undefined, where actually a node is always expected, due to the nature of a translation unit (Like every statement is in a unit. When we only parse a statement we cannot go up to the outer scope). I am wondering how we should reflect this in the documentation...

spoenemann:
If you use this new feature, you should be aware of the potentially added complexity. Scoping may be completely broken. If you want to support resolving references with a partial model, you need to consider that explicitly in your custom scope provider. The generic one shouldn't have any issues.

msujew (approved):
Looks good to me 👍

PR #1401, opened 2024-03-01
feat: validate assignments when calling rules with multiplicity

bjthehun:
Closes #1246:

Adds a new validation rule for rule calls. It looks up an enclosing container with multiplicity, and then demands that an assignment occurs for said call.

PR #1412, opened 2024-03-12
Validation for assignments with `=` instead of `+=`

JohannesMeierSE:
This PR adds validations for assignments using the `=` operator, while they got multiple values assigned. This validation gives the user the hint to use `+=` as assignment operator instead.

Among others, the following details are considered during the validation and complemented with test cases:
- assignments in neighbored alternatives
- array multiplicities of outer groups
- assignments in fragments which are called multiple times
- actions creating new objects, which changes the location of stored values (that counts for looped actions as well!)
- the assignment of a rewriting action

### DIFF
--- a/packages/langium/src/grammar/validation/validator.ts
+++ b/packages/langium/src/grammar/validation/validator.ts
@@ -856,10 +856,11 @@ export class LangiumGrammarValidator {
         }
 
         // look for assignments to the same feature within groups
-        if (ast.isGroup(currentContainer) || ast.isUnorderedGroup(currentContainer)) {
+        if (ast.isGroup(currentContainer) || ast.isUnorderedGroup(currentContainer) || ast.isAlternatives(currentContainer)) {
             let countGroup = 0;
             let foundRelevantChild = false;
             for (const child of currentContainer.elements) {
+                // special case: Actions
                 if (child === relevantChild) {
                     foundRelevantChild = true;
                 }
@@ -878,37 +879,18 @@ export class LangiumGrammarValidator {
                         break;
                     }
                 }
-                countGroup += this.searchAssignmentsRecursivelyDown(child, undefined, featureName);
+
+                // count the relevant child assignments
+                const countCurrent = this.searchAssignmentsRecursivelyDown(child, undefined, featureName);
+                if (ast.isAlternatives(currentContainer)) {
+                    // for alternatives, only a single alternative is used => assume the worst case and take the maximum number of assignments
+                    countGroup = Math.max(countGroup, countCurrent);
+                } else {
+                    // all members of the group are relavant => count them all
+                    countGroup += countCurrent;
+                }
             }
             countResult += countGroup;
-        }
-        // look for assignments to the same feature within alternatives
-        if (ast.isAlternatives(currentContainer)) {
-            // TODO 2x if vereinheitlichen!
-            let countAlternatives = 0;
-            let foundRelevantChild = false;
-            for (const child of currentContainer.elements) {
-                if (child === relevantChild) {
-                    foundRelevantChild = true;
-                }
-                if (ast.isAction(child)) {
-                    // a new object is created => following assignments are put into the new object
-                    if (relevantChild) {
-                        if (foundRelevantChild) {
-                            // the assignment to check
-                            break; //  => break the for-loop here, since all following assignments are put into the new object
-                        } else {
-                            containerMultiplicityMatters = false;
-                            countAlternatives = 0;
-                        }
-                    } else {
-                        // 
-                        break;
-                    }
-                }
-                countAlternatives = Math.max(countAlternatives, this.searchAssignmentsRecursivelyDown(child, undefined, featureName));
-            }
-            countResult += countAlternatives;
         }
 
         // the current element can occur multiple times => its assignments occur multiple times as well
--- a/packages/langium/src/grammar/validation/validator.ts
+++ b/packages/langium/src/grammar/validation/validator.ts
@@ -816,10 +816,11 @@ export class LangiumGrammarValidator {
     }
 
     private searchAssignmentsRecursivelyUp(node: AstNode, assignment: ast.Assignment, accept: ValidationAcceptor): boolean {
-        let currentElement: AstNode | undefined = node;
-        while (currentElement) {
+        let currentContainer: AstNode | undefined = node;
+        let previousChild: AstNode | undefined = undefined;
+        while (currentContainer) {
             // check neighbored and nested assignments
-            const countAssignments = this.searchAssignmentsRecursivelyDown(currentElement, assignment.feature);
+            const countAssignments = this.searchAssignmentsRecursivelyDown(currentContainer, previousChild, assignment.feature);
             if (countAssignments >= 2) {
                 accept(
                     'warning',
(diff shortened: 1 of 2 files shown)


packages/langium/test/grammar/grammar-validator.test.ts
```diff
@@ -698,6 +698,233 @@ describe('Property type is not a mix of cross-ref and non-cross-ref types.', ()
...
+    });
+
+    test('two assignments with single cardinality', async () => {
+        const validation = await validate(getGrammar(`
+            entry Model:
+                persons=Person ',' persons=Person;
+            Person: 'person' name=ID ;
+        `));
+        expect(validation.diagnostics.length).toBe(2);
+        expect(validation.diagnostics[0].message).toBe(getMessage('persons'));
```
Lotes, line 740: Here you expect two errors, but only check the first. At other lines you check both.

packages/langium/test/grammar/grammar-validator.test.ts
```diff
@@ -698,6 +698,233 @@ describe('Property type is not a mix of cross-ref and non-cross-ref types.', ()
...
+        expect(validation.diagnostics[1].message).toBe(getMessage('persons'));
+    });
+
+    test('fragments: all assignments are located in a fragment', async () => {
+        const validation = await validate(getGrammar(`
+            entry Model:
+                Assign ';' Assign;
```
Lotes, line 854: Very interesting case! I have never thought of this

packages/langium/test/grammar/grammar-validator.test.ts
```diff
@@ -698,6 +698,233 @@ describe('Property type is not a mix of cross-ref and non-cross-ref types.', ()
...
+                Assign*;
+            fragment Assign:
+                ',' persons=Person;
+            Person: 'person' name=ID ;
+        `));
+        expect(validation.diagnostics.length).toBe(1);
+        expect(validation.diagnostics[0].message).toBe(getMessage('persons'));
+    });
+
+    test('no problem: fragments in alternatives', async () => {
```
Lotes, line 875: What are about fragments X that are twice in one rule alternative and only once in another :)? Then the user is forced to use a list for the second case.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -4,38 +4,41 @@
...
 import { flattenPlainType, isPlainReferenceType } from '../type-system/type-collector/plain-types.js';
+import type { AstNodeLocator } from '../../workspace/ast-node-locator.js';
 
 export function registerValidationChecks(services: LangiumGrammarServices): void {
     const registry = services.validation.ValidationRegistry;
     const validator = services.validation.LangiumGrammarValidator;
     const checks: ValidationChecks<ast.LangiumGrammarAstType> = {
         Action: [
             validator.checkAssignmentReservedName,
+            validator.checkActionOperator,
```
Lotes, line 34: The names for the check function can be more precise. It is not checking the operator, it is about the right operator multiplicity.
Theoretically we could do the same for the other direction: If `+=` was used, but only one occurrence is given, we could downgrade to `=`

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -785,6 +790,160 @@ export class LangiumGrammarValidator {
         }
     }
 
+    /** This validation is specific for assignments with '=' as assignment operator and checks,
+     * whether the operator should be '+=' instead. */
+    checkAssignmentOperator(assignment: ast.Assignment, accept: ValidationAcceptor): void {
```
Lotes, line 795: I think you should check instead the parser rule and always traverse down.
With this approach it seems that you are always moving up to the parser rule and then go down to get the entire context. That could result in redundant work if there is no cache in-between.
I also think for the sake of maintainability, would this not be easier to go always downwards within the checker, starting at the parser rule root?

Lotes: In my imagination you just need a map from property names to a counter per rule alternative. Fragments need to be resolved of course, but even with this, I think, the code becomes a lot simpler.

packages/langium/test/grammar/grammar-validator.test.ts
JohannesMeierSE: Fixed, thanks!

packages/langium/test/grammar/grammar-validator.test.ts
JohannesMeierSE: Thanks!

packages/langium/test/grammar/grammar-validator.test.ts
JohannesMeierSE: That is an interesting case and you are completely right!

In my eyes, using a list in the 2nd case as well is better than loosing a value in the 1st case.
Therefore, I would stay with the current validation.

Nevertheless, I added this as an additional test case.

What do you think?

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: You are right, I renamed it to `checkActionOperatorMultiplicity`.

Yes, we could do the same for the other direction: Do you recommend to do it now?

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: Yes, mixing upwards and downwards feels a bit strange for me as well.
(The reason for that is historic, I started in that way and the challenge grew and grew 🙂)

I like your idea to start at the parser rules and only going downwards together with a map for the found feature names. That would make it cleaner.

However, I am not sure about these details:
- Actions interrupt the straight-forward recursive search: Is it enough to handle it as "a new parser rule"? I assume yes ...
- What about fragments which are called in other Langium files? Is that no special case anymore or does it require special treatments, which is higher than the simplification before?

I will continue to think about this ... thanks for the suggestion!

JohannesMeierSE: > map from property names to a counter per rule alternative

We need to store the `Assignment` nodes as well, otherwise, we cannot annotate the warning at the correct position.

What do you mean with "per rule alternative"?

Lotes: For a parser rule P ::= A|B|C you have 3 alternatives A, B and C.

```ts
//foreach alternative do
const map = new Map<string, Assignment[]>();
foreach(const assignment in alternative…) {
  if(!map.has(assignemnt.name)) { map.set(assignment.name, []) }
  map.get(assignment.name)!.push(assignment)
}
//check if any entry of map has more then 1 assignment
```

packages/langium/src/grammar/validation/validator.ts
Lotes: No, not now...it would also conflict with rule fragments that are both used once and used twice/more

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: I just pushed a big refactoring, most of the special cases vanished and the code should be more readable.

WDYT?

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -785,6 +789,108 @@ export class LangiumGrammarValidator {
...
+
+        // assignment
+        if (ast.isAssignment(currentNode)) {
+            storeAssignmentUse(map, currentNode.feature, 1 * currentMultiplicity, currentNode);
+        }
+
+        // Search for assignments in used fragments as well, since their property values are stored in the current object.
+        // But do not search in calls of regular parser rules, since parser rules create new objects.
+        if (ast.isRuleCall(currentNode) && ast.isParserRule(currentNode.rule.ref) && currentNode.rule.ref.fragment) {
+            this.checkNodeRegardingAssignmentNumbers(currentNode.rule.ref.definition, currentMultiplicity, map, accept); // TODO fragment rules are evaluated multiple times!
```
Lotes, line 842: > // TODO fragment rules are evaluated multiple times!

Why is this a problem? I see you are using a Set<Assignment|Action>. If you make it a list, this should be not a problem anymore.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -1013,3 +1119,42 @@ function findLookAheadGroup(rule: AstNode | undefined): ast.TerminalGroup | unde
...
+            counter: 0,
+        };
+        map.set(feature, entry);
+    }
+    assignments.forEach(a => entry!.assignments.add(a)); // a Set is necessary, since assignments in Fragements might be used multiple times by different parser rules, but they should be marked only once!
+    entry.counter += increment;
+}
+
+function mergeAssignmentUse(mapSoure: Map<string, AssignmentUse>, mapTarget: Map<string, AssignmentUse>, counterOperation: (s: number, t: number) => number = (s, t) => s + t): void {
+    for (const sourceEntry of mapSoure.entries()) {
```
Lotes, line 1148: Instead of...

> `for (const sourceEntry of mapSoure.entries())`

try in-place deconstructing

`for (const [key, source] of mapSoure.entries())`

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -785,6 +789,108 @@ export class LangiumGrammarValidator {
...
+        }
+    }
+
+    private checkAssignmentOperatorMultiplicitiesLogic(startNodes: AstNode[], accept: ValidationAcceptor): void {
+        // new map to store usage information of the assignments
+        const map: Map<string, AssignmentUse> = new Map();
+
+        // top-down traversal
+        for (const node of startNodes) {
+            // TODO dürfen mehr als 1 Action auftauchen? ansonsten funktioniert das hier nicht so ganz wie gedacht!
```
Lotes, line 807: > dürfen mehr als 1 Action auftauchen? ansonsten funktioniert das hier nicht so ganz wie gedacht!

Try it out. Not sure if this is wanted to have multiple actions in series. If garbage comes out of some experiments it might be good to forbid it using a separate PR (?)...

Lotes: Actually we have an in-series case already with binary operators:

`Expression: AddTerm ({XYZ.left=current} right=AddTerm)*;`

Lotes:
Thanks @JohannesMeierSE ! Nice refactoring. No it is going down only and we can concentrate better on the special cases. Especially the one about multiple actions. Is it really a bad thing? Same we actually have for binary operators. Of course you could add some concrete example with two explicit actions and see what happens.

Apart from that, just two additional comments :).

packages/langium/src/grammar/validation/validator.ts
msujew: Yes, multiple actions are fine. Please don't comment in German though :)

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: Thanks for the hint, I forgot to remove the comment

- The "problem" I thought about was a little performance improvement, but it is not worth it, since using a map for already visited fragments makes the code less readable.
- I am using the Set in order not show multiple errors for the same assignment in this case!

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: Thanks!

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: Thanks @Lotes for the hint, I again forgot the internal comment after fixing the issue.

Thanks @msujew for the confirmation!

I improved the comments and added a test case with a second action within the same parser rule.

JohannesMeierSE:
> Thanks @JohannesMeierSE ! Nice refactoring. No it is going down only and we can concentrate better on the special cases. Especially the one about multiple actions. Is it really a bad thing? Same we actually have for binary operators. Of course you could add some concrete example with two explicit actions and see what happens.
> 
> Apart from that, just two additional comments :).

Thanks @Lotes for the second review!

Yes, I like the new structure as well! In the end, it was easier than I expected before 🙂 

The handling of two or more explicit actions within the same parser rule was already covered, I just forgot to remove my internal hint. The case with a single action with a loop around it was already covered and tested.

Additionally, I fixed this issue in another test case in the Langium code base and rebased this branch.

Lotes (approved):
Looks good to me :). Thanks.

It is funny, there is a second `checkRuleCallMultiplicity`... but it has a different scope, it is not entirely what you are checking. WDYT about it?

JohannesMeierSE:
> It is funny, there is a second `checkRuleCallMultiplicity`... but it has a different scope, it is not entirely what you are checking. WDYT about it?

Thanks @Lotes, yes, I saw https://github.com/eclipse-langium/langium/pull/1401, and you are right with the bad names: I just renamed the functions for these two validations and added a comment.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -809,6 +814,108 @@ export class LangiumGrammarValidator {
...
+
+        // create the warnings
+        for (const entry of map.entries()) {
+            if (entry[1].counter >= 2) {
+                for (const assignment of entry[1].assignments) {
+                    if (assignment.operator !== '+=') {
+                        accept(
+                            'warning',
+                            `It seems, that you are assigning multiple values to the feature '${assignment.feature}', while you are using '${assignment.operator}' as assignment operator. Consider to use '+=' instead in order not to loose some of the assigned values.`,
+                            { node: assignment, property: 'operator' }
```
msujew, line 843: Suggestion: I would rephrase that validation message to make it a bit more imperative. Also the info marker on the `operator` keyword is pretty hard to see. I would use the `feature` property instead.
```suggestion
                            `Found multiple assignments to '${assignment.feature}' with the '${assignment.operator}' assignment operator. Consider using '+=' instead to prevent data loss.`,
                            { node: assignment, property: 'feature' }
```

msujew (commented):
Looks pretty good to me. Just a few minor improvements, see below.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -809,6 +814,108 @@ export class LangiumGrammarValidator {
...
+        // look for assignments to the same feature nested within groups
+        if (ast.isGroup(currentNode) || ast.isUnorderedGroup(currentNode) || ast.isAlternatives(currentNode)) {
+            const mapAllAlternatives: Map<string, AssignmentUse> = new Map(); // store assignments for Alternatives separately
+            let nodesForNewObject: AstNode[] = [];
+            // check all elements inside the current group
+            for (const child of currentNode.elements) {
+                if (ast.isAction(child)) {
+                    // Actions are a special case: a new object is created => following assignments are put into the new object
+                    // (This counts for rewriting actions as well as for unassigned actions, i.e. actions without feature name)
+                    if (nodesForNewObject.length >= 1) {
```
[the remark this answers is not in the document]
JohannesMeierSE: I prefer `>= 1`, since it is more readable in my eyes, but I changed it to have consistent code, thanks for the hint!

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -809,6 +814,108 @@ export class LangiumGrammarValidator {
...
+                        // all collected nodes are put into the new object => check their assignments independently
+                        this.checkOperatorMultiplicitiesForMultiAssignmentsLogic(nodesForNewObject, accept);
+                        // is it possible to have two or more Actions within the same parser rule? the grammar allows that ...
+                        nodesForNewObject = [];
+                    }
+                    // push the current node into a new object
+                    nodesForNewObject.push(child);
+                } else {
+                    // for non-Actions
+                    if (nodesForNewObject.length >= 1) {
```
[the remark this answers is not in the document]
JohannesMeierSE: changed, see above

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: Thank makes sense, changed

JohannesMeierSE:
Thanks @msujew for your review! In my eyes, this feature is ready to merge.

PR #1434, opened 2024-04-02
Export grammar-related generators from CLI

Lotes:
For the [Langium playground](https://github.com/eclipse-langium/langium-website/pull/220) I need to import the Textmate generator.
Therefore I am exporting a couple of generators from the CLI.
My current alternative is to copy the code from this repo, which is actually not that nice.

Speak up if you want a different export strategy. Currently I am exporting each generator in its own export alias.

---

This PR adds a new ESM export `/textmate` to the `langium-cli` package. It contains the highlighting generator for Textmate syntax highlighters (Monaco editor).


dhuebner:
@Lotes 
In general I think it is good to export useful components to users. One thing to keep in mind, when exporting code we make it a public API, hence need to care that we do not break it in the future and we should document it. 
If you just need the **textmate** generator, maybe we should first only export just this generator, WDYT?

Lotes:
@dhuebner Thanks for the argument. That is a very valid point. Then I will only export the Textmate generator. But even there I am unsure: Shall I export it under "/textmate" or better under something more general like "/syntax-highlighting"...? Maybe start with something specific and deprecate when the time has come and it becomes more general?

montymxb:
I agree with @dhuebner here. Exporting just what we need makes sense for now. And as for the name, `textmate` seems fine, but we should document that new endpoint to make it clear what it provide (later on, before the next release).

msujew (commented):
Since the `textmate-generator.js` file imports from `../util.js`, it also attempts to import node libraries such as `node:path` or `node:url`. We should split the node related utilities from the Langium related utilities, as the node imports make the `./textmate` import difficult to bundle in a browser environment. It would require a lot of shims/polyfills in order to successfully bundle/run.

Lotes:
@montymxb As far as I can see there is no `next` change log file or something similar. So I will document the endpoint in the main description of this PR.
@msujew I have done the split. Nice catch. Please have another look.

Lotes:
I need to wait for #1437 since I want to increment the Langium CLI version and trigger Langium generator for the examples.

msujew (commented):
Looks pretty good. Just two minor suggestions, see below.

msujew (approved):
Great, looks good to me 👍

PR #1437, opened 2024-04-03
fix: generate errors cause langium-cli to fail with exit code 2

### LINKED ISSUE #1436: CI/CLI: Generator errors in linting job do not cause pipeline to fail
While investigating the bugs I introduced with pull request #1401, I found out that language generation errors do not cause the lint job to fail in the build pipeline (see pipeline run 2238):

LangiumGenerateCIErrors

Even though language generation fails due to wrongly applied validations, the generator and build step pass, causing me to miss the aforementioned bugs. Therefore, this problem needs to be fixed.

bjthehun:
Resolves #1436:

When the `generate` command of `langium-cli` fails because of an error in a grammar/config file, it will return with exit code 2 and not report a "successful" finish.

This commit _is expected_ to cause the CI pipeline to fail, because of the errors I introduced with #1401, which went unseen in the merge request.


msujew:
Hey @bjthehun, thanks for the fix!

Can you also fix the original validation issue in a secondary commit in this PR? I don't really feel comfortable merging a failing pipeline into `main`.

bjthehun:
> Can you also fix the original validation issue in a secondary commit in this PR? I don't really feel comfortable merging a failing pipeline into `main`.

Secondary commit has been appended for further review.

msujew (commented):
This doesn't actually address the whole issue:

image

The `hasAssignment` needs to be improved to include `AstUtils.getContainerOfType(call, ast.isAssignment) !== undefined`. This makes the `callInCrossReference` part obsolete, since cross references can only appear as part of an assignment.

bjthehun (commented):
Thanks for pointing that problem out. I've addressed it in my last commit.

msujew (approved):
Thanks, looks good to me 👍

Issue #1440, opened 2024-04-05
Diagnostics are stuck on deleted files


msujew:
Langium version: 3.0
Package name: `langium`

## Steps To Reproduce

1. Open a file with diagnostic markers
2. Delete or rename the file
3. The diagnostic markers stay on the non-existent file

See also https://github.com/eclipse-langium/langium/discussions/1439

## The current behavior

Diagnostic markers are still visible in the "Problem" widget of vscode

## The expected behavior

Diagnostic markers are removed as expected

PR #1441, opened 2024-04-05
Update diagnostics for deleted files

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1440

Simply replaces the diagnostics of all deleted files with an empty array, thereby unpublishing them. This effects both deleted and renamed files.


montymxb (approved):
:octocat: Looks good

PR #1442, opened 2024-04-06
Fix treatment of word transitions in fuzzy matcher

lars-reimann:
Previously, the first character of a query had to match the first character of some `camelCase` text in the fuzzy matcher. 

Now, it has to match either the first character of the text or the first character after a word transition.


msujew (approved):
Thanks, looks good to me 👍

PR #1445, opened 2024-04-07
Emit an event after a configuration section was updated

### LINKED ISSUE #1444: Watch configuration provider for updates
We have an external tool to run compiled DSL code. It is part of our language server, since it is also needed for validation $^1$. The tool gets started with some CLI command. 

When used from VS Code, this command is configured via a setting. If the tool could not be started and the setting is updated, we want to try again with the updated setting. 

➡️ For this, it would be nice if the `ConfigurationProvider` would emit an event after an update has happened.

We're currently overriding the type and the implementation of the `ConfigurationProvider` in our shared module, but this seems like functionality that could be useful elsewhere.

-----

$^1$ In particular, we want to compute the schema of tabular data when we first load it. Schemata of derived data can then get computed statically.

lars-reimann:
Closes #1444

The `DefaultConfigurationProvider` now emits an event after a configuration section was updated.

### DIFF
--- a/packages/langium/src/workspace/configuration.ts
+++ b/packages/langium/src/workspace/configuration.ts
@@ -4,7 +4,7 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { Emitter } from 'vscode-languageserver-protocol';
+import { Emitter } from '../utils/event.js';
 import type {
     ConfigurationItem,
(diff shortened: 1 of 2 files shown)


msujew (commented):
Looks pretty good, just one comment, see below.

msujew (approved):

Issue #1408, opened 2024-03-07
3.0: Multi Language cannot handle file renames


cdietrich:
having more than one language leads to service registry use the map and not the singleton.
if you rename a file from as dsl file extension to a non dsl file extension (`.hello` to `.hello_bak`) this will
lead to the language server crashing cause the update call is called with the new file extension,
which is not in the registry (of course)
cause it is not a dsl extension.

<img width="1457" alt="grafik" src="https://github.com/eclipse-langium/langium/assets/238736/269dafde-7313-49f2-8e22-68c34a2c0184">

PR #1455, opened 2024-04-16
Improve unknown file type handling in language server

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1408

Adds a few safeguards for cases in which the language client sends document updates/LSP requests for unknown file types. This is pretty simple to reproduce by renaming a file to another file that cannot be handled by the language server. For the requirements example, from `*.tst` to `*.tstx`.

We now filter those updates out before they arrive at the document builder and handle errors for other LSP requests better by returning a `ResponseError` instead of throwing an error in the code. This massively improves the UX for any user that hits this behavior by accident.

Also fixes a few exports from the `./lsp` folder to enable overrides of existing services.

### DIFF
--- a/packages/langium/src/lsp/document-update-handler.ts
+++ b/packages/langium/src/lsp/document-update-handler.ts
@@ -78,7 +78,7 @@ export class DefaultDocumentUpdateHandler implements DocumentUpdateHandler {
     protected fireDocumentUpdate(changed: URI[], deleted: URI[]): void {
         // Filter out URIs that do not have a service in the registry
         // Running the document builder update will fail for those URIs
-        changed = changed.filter(uri => this.hasService(uri));
+        changed = changed.filter(uri => this.serviceRegistry.hasServices(uri));
         // Only fire the document update when the workspace manager is ready
         // Otherwise, we might miss the initial indexing of the workspace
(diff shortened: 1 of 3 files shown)


msujew:
cc @cdietrich the behavior implemented in `DocumentUpdateHandler` can also be implemented in the `DocumentBuilder#update` method directly in the existing version of 3.0.

packages/langium/src/lsp/document-update-handler.ts
```diff
@@ -83,6 +89,23 @@ export class DefaultDocumentUpdateHandler implements DocumentUpdateHandler {
...
+     *
+     * In case we only have a single language in our registry, we can safely use that language for all URIs.
+     */
+    protected hasService(uri: URI): boolean {
+        try {
+            this.serviceRegistry.getServices(uri);
+            return true;
+        } catch {
+            return false;
+        }
```
spoenemann, line 106: Shall we introduce a `hasServices` method in ServiceRegistry? This try-catch works, but looks like a workaround for a missing service feature.

msujew: Yeah, makes sense 👍

spoenemann (approved):
👍

PR #1459, opened 2024-04-18
Mark completely optional parser rules as errors

### LINKED ISSUE #1400: Completely optional parser rules should be marked as errors
Related to https://github.com/eclipse-langium/langium/issues/1398

A parser rule that potentially consumes no input from the text is not valid, as the parser can potentially not attempt to create it:

```ts
Persons:
    ((elements+=ID) (elements+=ID)*)?;
```

This leads to unexpected behavior, and generally serves no purpose. We have a similar validation for terminals already, and should have one for parser rules as well.

bjthehun:
Closes #1400. Notes:

- Fragments are not marked as optional.
- Rules with actions only are not marked as optional.


packages/langium/src/grammar/validation/validator.ts
```diff
@@ -411,6 +413,45 @@ export class LangiumGrammarValidator {
...
+            if (element.cardinality === '?' || element.cardinality === '*') {
+                return true;
+            }
+            // Actions themselves count as non-optional.
+            if (ast.isAction(element)) {
+                return false;
+            }
+            // Assignments with ?= consume nothing, groups without + too.
+            if (ast.isAssignment(element)) {
+                return element.operator === '?=';
```
msujew, line 435: Suggestion: Every assignment always consumes something (as long as its cardinality isn't `?` or `*`):

```
A: value?='test';
```

Consumes the `'test'` keyword input. You can just remove the case completely, since it should always return `false` for assignments.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -411,6 +413,45 @@ export class LangiumGrammarValidator {
...
+            // and alternative groups, too.
+            if (ast.isGroup(element) || ast.isAlternatives(element)) {
+                return element.elements.every(mayConsumeEmpty);
+            }
+            // Else, assert that we consume _something_.
+            return false;
+
+        };
+        if (mayConsumeEmpty(parserRule.definition)) {
+            accept('error', 'This parser rule would succeed without consuming input.', { node: parserRule, property: 'definition', code: IssueCodes.ParsingRuleEmpty });
```
msujew, line 451: Suggestion: I would use the `name` property to display the message. Diagnostics should rarely cover the whole definition of a node, but rather a very specific part of it. For example, if I forgot to implement a method of an interface, I don't get an error on the whole class body, but on its name instead.

packages/langium/test/grammar/grammar-validator.test.ts
```diff
@@ -912,3 +912,128 @@ describe('Cross-reference to type union is only valid if all alternatives are AS
         expectNoIssues(validationResult);
     });
 });
+
+function detectEmptyRule(validationResult: ValidationResult, line1: number, col1: number, line2: number, col2: number) {
```
msujew, line 916: Suggestion: With the suggestion above, instead of `line1, col1...` you can just pass the name of the rule as a string and use that to find the node that contains the error. That makes it easier to write and read all tests that use this function.

msujew (commented):
Thanks for taking this on @bjthehun!

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -411,6 +413,45 @@ export class LangiumGrammarValidator {
...
+
+        // Process definition; report problem for the rule itself.
+        const mayConsumeEmpty = (element: ast.AbstractElement): boolean => {
+            // First, check cardinality of the element.
+            if (element.cardinality === '?' || element.cardinality === '*') {
+                return true;
+            }
+            // Actions themselves count as non-optional.
+            if (ast.isAction(element)) {
+                return false;
```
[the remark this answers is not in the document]
bjthehun: Actions themselves count as optional (again).

packages/langium/src/grammar/validation/validator.ts
bjthehun: Now using the 'name' property.

packages/langium/src/grammar/validation/validator.ts
bjthehun: Case removed.

packages/langium/test/grammar/grammar-validator.test.ts
bjthehun: Simplified tests to look for empty rules with a given name.

msujew (approved):
LGTM, thanks 👍

PR #1466, opened 2024-04-23
`langium-cli/textmate`: Move overlooked `process` node dependency to the `node-util.ts`

Lotes:
Unfortunately, I overlooked one dependency to the node object `process`, which does not need any import to be used in the Node context. So the `/textmate` export is only working for Node right now.

I actually tested the import with `npm link` the last time and it was working. But looking back it could be that I looked at the state before the link, because I may forgot to build.

This time I tested it properly against this [PR](https://github.com/eclipse-langium/langium-website/pull/220) (with `npm build` at the right moment).


kaisalmen (approved):
This looks good to me. 👍

Issue #1256, opened 2023-11-02
Misleading type compatibility validation error


dhuebner:
Langium version: 2.0.1
Package name: langium

## Steps To Reproduce

```js
grammar Foo

interface RuleType { prop : string };

entry Rule returns RuleType: 
     // Error at prop: The assigned type 'MyDataType' is not compatible with the declared property 'prop' of type 'string'.
    prop = MyDataType 
;

MyDataType returns string:
    INT
;

terminal INT returns number: /[0-9]+/;
```

## The current behavior
In the example above an error is reported because `MyDataType` uses a number typed terminal rule. Even thought the rule defines the return type `string`.

## The expected behavior
Because the datatype rule defines its return type as `string` the error should not be created.

PR #1478, opened 2024-04-29
Respect declared type of data type rules

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1256

In the issue, the `MyDataType` rule infers a `number` type (due to its content) instead of the declared `string` type. This change first checks whether the declared type can be assigned directly to the expected type. Only if that fails will the validation attempt to use the computed type for checking.

### DIFF
--- a/packages/langium/src/grammar/type-system/type-collector/types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/types.ts
@@ -304,7 +304,18 @@ function isTypeAssignableInternal(from: PropertyType | undefined, to: PropertyTy
         result = isArrayType(to) && isTypeAssignableInternal(from.elementType, to.elementType, visited);
     } else if (isValueType(from)) {
         if (isUnionType(from.value)) {
-            result = isTypeAssignableInternal(from.value.type, to, visited);
+            if (from.value.dataType) {
+                // We can test the primitive data type directly
+                // This potentially skips a expensive recursive call
+                // This also helps in case the computed internal data type does not fit the declared data type
+                const primitiveType: PrimitiveType = {
+                    primitive: from.value.dataType
+                };
+                result = isTypeAssignableInternal(primitiveType, to, visited);
+            }
+            if (!result) {
+                result = isTypeAssignableInternal(from.value.type, to, visited);
+            }
         } else if (!isValueType(to)) {
             result = false;
         } else if (isUnionType(to.value)) {
--- a/packages/langium/test/grammar/type-system/type-validator.test.ts
+++ b/packages/langium/test/grammar/type-system/type-validator.test.ts
@@ -281,7 +281,7 @@ describe('validate declared types', () => {
         });
     });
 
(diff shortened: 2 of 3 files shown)


packages/langium/src/grammar/type-system/type-collector/types.ts
```diff
@@ -304,7 +304,18 @@ function isTypeAssignableInternal(from: PropertyType | undefined, to: PropertyTy
...
         if (isUnionType(from.value)) {
-            result = isTypeAssignableInternal(from.value.type, to, visited);
+            if (from.value.dataType) {
+                // We can test the primitive data type directly
+                // This potentially skips a expensive recursive call
+                // This also helps in case the computed internal data type does not fit the declared data type
+                const primitiveType: PrimitiveType = {
+                    primitive: from.value.dataType
+                };
+                result = isTypeAssignableInternal(primitiveType, to, visited);
```
montymxb, line 314: Overall this looks okay for the number -> string case. But the other way around doesn't feel right. If I flip around the test below so it converts from string -> number, it also passes, but I wouldn't expect that we can assign an arbitrary string to a number type.

I figured we're doing a type check on the Datatype to make sure it's computed type is reasonable given the declared type, but it seems like this is a gap somehow.

```antlr
interface RuleType { prop : number };
Rule returns RuleType: prop = MyDataType;
MyDataType returns number: ID;
terminal ID returns string: /[_a-zA-Z][\\w_]*/;
```

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -280,6 +280,18 @@ describe('validate declared types', () => {
             property: 'name'
         });
     });
+
+    test('Can assign a data type rule to a property with its base type', async () => {
+        const validationResult = await validate(`
```
montymxb, line 285: Based on the prior note, I would also consider the inverse case as a test too.

montymxb (commented):
The checking logic doesn't seem right for the inverse case. I would double check that, as it allows coercing strings to numbers. In this case it looks like the parsed result is still a string, despite the indicated number type of the DataType (see example in comments).

packages/langium/src/grammar/type-system/type-collector/types.ts
msujew: Eventually, this simply runs through the value converter and attempts to do `Number(input)`. This will likely result in `NaN`, but it's a "valid" conversion.

I agree that it doesn't make a lot of sense, but if users want to shoot themselves in the foot, who are we to stop them? :D

montymxb: Yeah I saw that too. I still feel like maybe we can provide some advice about avoiding this at the least, maybe in a quite write-up or something later on. But that's a separate thing then.

montymxb (approved):
Cool looks good!

PR #1492, opened 2024-05-10
Track unknown documents via language id

msujew:
This is a second iteration on the fix of https://github.com/eclipse-langium/langium/pull/1455. This idea came up while helping with https://github.com/eclipse-langium/langium/discussions/1488: The `TextDocument` is always aware of the language ID used in the language client, even if the file extension doesn't fit. We can use that info (assuming a `TextDocument` instance is available for the given URI) to make the `ServiceRegistry` more accurate.

This also fixes issues experienced when creating an untitled file and setting the language to a DSL language (in multi-language projects).

### DIFF
--- a/packages/langium/src/service-registry.ts
+++ b/packages/langium/src/service-registry.ts
@@ -45,6 +45,13 @@ export class DefaultServiceRegistry implements ServiceRegistry {
     protected idMap = new Map<string, LangiumCoreServices>();
     protected extMap = new Map<string, LangiumCoreServices>();
 
+    /**
+     * @deprecated Use the new `extMap` property instead.
+     */
+    protected get map(): Map<string, LangiumCoreServices> | undefined {
+        return this.extMap;
+    }
+
     protected readonly textDocuments?: TextDocumentProvider;
 
     constructor(services?: LangiumSharedCoreServices) {
@@ -63,7 +70,6 @@ export class DefaultServiceRegistry implements ServiceRegistry {
         this.idMap.set(data.languageId, language);
         if (this.idMap.size === 1) {
             this.singleton = language;
-            return;
         } else {
             this.singleton = undefined;
         }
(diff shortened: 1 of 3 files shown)


packages/langium/src/service-registry.ts
```diff
@@ -41,44 +42,55 @@ export interface ServiceRegistry {
 export class DefaultServiceRegistry implements ServiceRegistry {
 
     protected singleton?: LangiumCoreServices;
-    protected map?: Record<string, LangiumCoreServices>;
+    protected idMap = new Map<string, LangiumCoreServices>();
+    protected extMap = new Map<string, LangiumCoreServices>();
```
spoenemann, line 46: It's unfortunate that the previous `map` was protected. I'd apply this change in the 3.x version stream anyway because it brings important fixes that we shouldn't delay. Shall we create a protected getter and setter for `map` which redirects to the new property, and mark that deprecated?

Proposal: use longer names for the new properties that are more understandable for those creating custom subclasses: `languageIdMap`, `fileExtensionMap`

packages/langium/src/service-registry.ts
```diff
@@ -41,44 +42,55 @@ export interface ServiceRegistry {
...
-        // Store the language services in the map.
-        for (const ext of language.LanguageMetaData.fileExtensions) {
-            if (this.map[ext] !== undefined && this.map[ext] !== language) {
-                console.warn(`The file extension ${ext} is used by multiple languages. It is now assigned to '${language.LanguageMetaData.languageId}'.`);
-            }
-            this.map[ext] = language;
+        this.idMap.set(data.languageId, language);
+        if (this.idMap.size === 1) {
+            this.singleton = language;
+            return;
```
spoenemann, line 66: The `return` isn't necessary here?

spoenemann (commented):
Good improvement!

packages/langium/src/service-registry.ts
spoenemann: WDYT about the naming proposal? And shall we make the properties `readonly`?

msujew: Good idea, I've renamed the fields and made them readonly.

spoenemann (approved):
👍

PR #1498, opened 2024-05-15
Fix intermittent testing failure

msujew:
In some of our CI runs (see [here](https://github.com/eclipse-langium/langium/actions/runs/9081381517/job/24955100656?pr=1474)), the workspace lock test fails due to minor timing issues in the runtime. This change makes the test a bit more lenient regarding minor inaccuracies in the `setTimeout` mechanism.


msujew:
This has been resolved by https://github.com/eclipse-langium/langium/pull/1701.

PR #1499, opened 2024-05-15
add option to calculate CompletionItemKind for keywords

cdietrich:
we use that one to assign "Operator" to some keywords


msujew (approved):
That makes sense, thanks 👍

Issue #1505, opened 2024-05-21
Incorrect array type inference


msujew:
Langium version: 3.0
Package name: `langium`

## Steps To Reproduce

Create a parser rule like:

```
A: items+=ID items+=NUMBER;
terminal ID returns string: ...;
terminal NUMBER returns number: ...;
```

This generates an unexpected type, see below.

The type below is both unexpected as well as difficult to work with. A unioned array element type would be preferable here.

## The current behavior

The parser rule above will be generated as:

```ts
interface A extends AstNode {
  $type: 'A';
  items: Array<number> | Array<string>
}
```

## The expected behavior

The expected type would be

```diff
interface A extends AstNode {
  $type: 'A';
-  items: Array<number> | Array<string>
+  items: Array<number | string>
}
```

Note that the type union has been pulled into the array.

PR #1506, opened 2024-05-21
Merge inferred array types

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1505

Note that this approach has a very minor issue in regards to alternatives of lists:

```
A: (values+=ID* | values+=NUMBER*);
```

The expected type would be `values: Array<string> | Array<number>`, but this PR adjusts this to generate `values: Array<string | number>`. I think this is "good enough" as it fixes the much more common issue of `values+=ID values+=NUMBER`. See also https://github.com/eclipse-langium/langium-website/pull/132#issuecomment-2117565838.

### DIFF
--- a/packages/langium/src/grammar/type-system/type-collector/plain-types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/plain-types.ts
@@ -210,18 +210,30 @@ function plainToPropertyType(type: PlainPropertyType, union: UnionType | undefin
 }
 
 export function mergePropertyTypes(first: PlainPropertyType, second: PlainPropertyType): PlainPropertyType {
-    const flattenedFirst = flattenPlainType(first);
-    const flattenedSecond = flattenPlainType(second);
-    for (const second of flattenedSecond) {
-        if (!includesType(flattenedFirst, second)) {
-            flattenedFirst.push(second);
+    const { union: flattenedFirstUnion, array: flattenedFirstArray } = flattenPlainType(first);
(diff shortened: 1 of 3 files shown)


packages/langium/src/grammar/type-system/type-collector/plain-types.ts
```diff
@@ -210,18 +210,30 @@ function plainToPropertyType(type: PlainPropertyType, union: UnionType | undefin
...
-    const flattenedFirst = flattenPlainType(first);
-    const flattenedSecond = flattenPlainType(second);
-    for (const second of flattenedSecond) {
-        if (!includesType(flattenedFirst, second)) {
-            flattenedFirst.push(second);
+    const { union: flattenedFirstUnion, array: flattenedFirstArray } = flattenPlainType(first);
+    const { union: flattenedSecondUnion, array: flattenedSecondArray } = flattenPlainType(second);
+    for (const second of flattenedSecondUnion) {
+        if (!includesType(flattenedFirstUnion, second)) {
+            flattenedFirstUnion.push(second);
```
Lotes, line 217: These two for-loops are almost the same. You could create some helper function 'unionTypes' or so.

Lotes: Plus: I prefer immutable operations, it also makes the code easier to read.

Lotes (changes requested):
Thanks for the change :-). Two remarks about code duplication and immutable operations.

Lotes (approved):
LGTM

Issue #1184, opened 2023-09-05
Export property types for checking at compile time


guiliannewiger:
When implementing a custom scopeProvider, you might want to compare the `context.container.$type` against a specific type like so:
``` typescript
if (context.container.$type === "Person")
```
But what happens when the grammar is adjusted so that the name of the "Person"-interface changes to "Customer"? Suddenly, this if-condition always returns false! Since the `context.container.$type` is being compared with a hard coded String, the compiler doesn't notice that there is a mismatch between the interface name and the usage of it.
Luckily, interface names are exported in the ast.ts: 
``` typescript
export const Person = 'Person';
```
The ast.ts is derived from the grammar and is therefore always up-to-date after running `langium:generate`.
Now, we can exchange the hard coded String for a reference to the exported type:
``` typescript
if (context.container.$type === Person)
```
Changing the interface name now results in a compiler error, because there is no "Person"-interface anymore. The grammar can be adjusted safely and you don't have to worry about having forgotten to update a reference. This helps immensely with maintainability.

Now let's consider the same example, but with `context.property`.
Again, you might want to compare it against a specific type like so:
``` typescript
if (context.property === "name")
```
The same problem applies here as well: changing the name of the property in the grammar to e.g. "surname" updates the generated ast.ts, but not the hard coded String reference to it.
I suggest exporting all property names of all generated interfaces like so:
``` typescript
export interface Person extends AstNode {
    readonly $container: Model;
    readonly $type: 'Person';
    age: number
    name: string
}
export const Person = 'Person';
export const PROPERTY_PERSON_AGE = "age";
export const PROPERTY_PERSON_NAME = "name";
```
The naming of these exported constants needs to be thought through diligently as always. Simply naming the constant like the property might result in naming collisions with properties of other interfaces or even interface names themselves.

This will allow for checking references to the property names at compile time, buffing maintainability for the reasons mentioned above.
hello-world-scoping.ts:
``` typescript
export class HelloWorldScopeProvider extends DefaultScopeProvider {
    override getScope(context: ReferenceInfo): Scope {
      const defaultScope = super.getScope(context);

      if (context.property === PROPERTY_PERSON_NAME)
```

spoenemann:
Would it be too strange to add additional properties to the string constants?
```typescript
export const Person: string & { [param in keyof Person]: string } = 'Person' as any;
Person.age = 'age';
Person.name = 'name';
```

msujew:
[...]

PR #1509, opened 2024-05-23
Support multi-target references

msujew:
Adds support for multi-target references. These are references that have the possibility to target multiple elements at once. See following illustration for an example:

image

Note that by default, only elements that are defined on the same scope can be targeted. I.e. a multi reference as in the following will only target one element:

image

This behavior can be freely overwritten by adopters of Langium. However, due to the inherent complexity of aligning the `References` service implementation and the scoping wrt their behavior, this feature is intended for proficient Langium developers.

The intended use cases for this feature are stuff like [declaration merging in TypeScript](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) or [partial classes in C#](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods).

Contains a bunch of breaking changes:

* The whole `References` interface now returns arrays instead of `T | undefined`.
* Every service/interface/class that works with `Reference` in a generic manner (linker mostly) now also works with `MultiReference`.
* `LinkingError` no longer inherits `ReferenceInfo` (which didn't make a lot of sense in the first place), but contains a `info` property.


Lotes (commented):
Thanks, looks good, I like especially that one can now ask for an array and do not need to check for undefined when getting references :D.
All in all I added some code improvement suggestions.
I would suggest to wait for more feedback, since I am not so deep in this area of the code, yet.

packages/langium/src/references/references.ts
```diff
@@ -67,28 +64,30 @@ export class DefaultReferences implements References {
...
+    findDeclarations(sourceCstNode: CstNode): AstNode[] {
         if (sourceCstNode) {
             const assignment = findAssignment(sourceCstNode);
             const nodeElem = sourceCstNode.astNode;
             if (assignment && nodeElem) {
                 const reference = (nodeElem as GenericAstNode)[assignment.feature];
 
-                if (isReference(reference)) {
-                    return reference.ref;
+                if (isReference(reference) || isMultiReference(reference)) {
```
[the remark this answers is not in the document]
Lotes: Or breaking the API by introducing isReference (for both; or isAnyReference), isSingleReference and isMultiReference

packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -47,7 +47,7 @@ ArrayType infers TypeDefinition:
 
 ReferenceType infers TypeDefinition:
     SimpleType |
-    {infer ReferenceType} '@' referenceType=SimpleType;
+    {infer ReferenceType} '@' referenceType=SimpleType (isMulti?='*')?;
```
spoenemann, line 50: Shouldn't we use the array notation `[]` here?

packages/langium/src/references/linker.ts
```diff
@@ -133,13 +173,20 @@ export class DefaultLinker implements Linker {
         return description ?? this.createLinkingError(refInfo);
     }
 
+    getCandidates(refInfo: ReferenceInfo): AstNodeDescription[] | LinkingError {
+        const scope = this.scopeProvider.getScope(refInfo);
+        const descriptions = scope.getElements(refInfo.reference.$refText).distinct(desc => `${desc.documentUri}#${desc.path}`).toArray();
+        return descriptions.length > 0 ? descriptions : this.createLinkingError(refInfo);
```
spoenemann, line 179: The multiplicity symbol `*` in the current grammar proposal could imply that zero link targets is okay and not a linking error. Shall we make a difference between `*` and `+` here?

spoenemann (commented):
I think it should be possible to realize this in a way that most adopters of Langium won't notice a lot of changes. The actual core changes are not too large, and are mostly adding, but not changing API (we can try to add instead of change in a few more places). I'll have a more detailed look at this within the next two months.

packages/langium/src/references/linker.ts
msujew: I think you're right. I'll change this to a `+` instead of using `*`. I think it looks nicer anyway.

spoenemann: What do you think of supporting both?

packages/langium/src/references/references.ts
```diff
@@ -97,29 +96,54 @@ export class DefaultReferences implements References {
...
+            const name = this.nameProvider.getName(node);
+            const container = node.$container;
+            // We need the name to find the siblings
+            // If the name is not available, we just return the specified node
+            // Similarly, we cannot find siblings in case the container is not available
+            if (!name || !container) {
+                return [node];
+            }
+            const siblings = streamContents(container).filter(n => this.nameProvider.getName(n) === name);
+            return siblings.toArray();
```
spoenemann: There's an overlap between this code and the scope provider service. Can we invoke the scope provider with a synthetic reference to avoid duplication?

Another thought is to determine more precisely which elements can be included in a multi-reference. This information could be added to the AST metadata (see #1184).

packages/langium/src/references/scope.ts
```diff
@@ -70,28 +74,40 @@ export class StreamScope implements Scope {
...
         return undefined;
     }
+
+    getElements(name: string): Stream<AstNodeDescription> {
+        const lowerCaseName = this.caseInsensitive ? name.toLowerCase() : name;
+        const local = this.caseInsensitive
+            ? this.elements.filter(e => e.name.toLowerCase() === lowerCaseName)
+            : this.elements.filter(e => e.name === name);
+        if (!local.isEmpty()) {
+            return local;
```
spoenemann: Shouldn't we concatenate the local scope with the outer scope?

spoenemann: Better: control this with an option when creating the scope. But IMO the default should be concatenate.

packages/langium/src/references/references.ts
spoenemann: Update: even better, take the information about multiple symbols with the same name _from one of the references to this symbol_, if any.

msujew: Done 👍

packages/langium/src/references/scope.ts
msujew: Done 👍

packages/langium/src/references/scope.ts
```diff
@@ -70,28 +90,42 @@ export class StreamScope implements Scope {
...
+            return local.concat(this.outerScope.getElements(name));
+        } else {
+            return local;
+        }
+    }
 }
 
 export class MapScope implements Scope {
-    readonly elements: Map<string, AstNodeDescription>;
+    readonly elements: MultiMap<string, AstNodeDescription>;
```
spoenemann, line 160: Shall we offer two different Scope implementations for Map and MultiMap? When performance is critical and there's no multi-reference, a Map might be better.

spoenemann (commented):
Thanks! This validation should still be added: https://github.com/eclipse-langium/langium/pull/1509#pullrequestreview-2896114275

packages/langium/src/references/scope.ts
msujew: I did some benchmarks, the `getElement` and `getElements` is roughly 5% slower, construction of the `MultiMap` roughly 20%. I kept `MapScope` and added `MultiMapScope` as a new implementation of `Scope`.

spoenemann (approved):

packages/langium/src/references/linker.ts
```diff
@@ -74,6 +85,11 @@ interface DefaultReference extends Reference {
     _nodeDescription?: AstNodeDescription;
 }
 
+export interface DefaultMultiReference extends MultiReference {
```
cdietrich, line 88: @msujew what was the reasoning behind exporting this one but not the one above

msujew: No particular reason, I thought it makes sense, and forgot that the other one isn't exported yet. Would you like to have both exported?

cdietrich: yes we are looking into writing linking traps and currently (3.5) we need to c&p everything

Issue #1513, opened 2024-05-27
Got "unterminated string literal" in generated grammar.ts error when grammar contains "${"


codehz:
Langium version: 3.0.0
Package name: langium-cli

## Steps To Reproduce

step:
1. init project, paste those code
2. npm run langium:generate
3. check the generated grammar.ts

Link to code example:

```
grammar HelloWorld

entry Model:
   persons+=Person*;

Person:
    'person' name=TemplateLiteral;

hidden terminal WS: /\s+/;

hidden terminal ML_COMMENT: /\/\*[\s\S]*?\*\//;
hidden terminal SL_COMMENT: /\/\/[^\n\r]*/;

TemplateLiteral:
    content+=TEMPLATE_LITERAL_FULL |
    // Or template literal parts with expressions in between
    (
        content+=TEMPLATE_LITERAL_START 
        content+="Expr"
        (
            content+=TEMPLATE_LITERAL_MIDDLE
            content+="Expr"
        )*
        content+=TEMPLATE_LITERAL_END
    );

terminal TEMPLATE_LITERAL_FULL:
    '`' IN_TEMPLATE_LITERAL* '`';
terminal TEMPLATE_LITERAL_START:
    '`' IN_TEMPLATE_LITERAL* '${';
terminal TEMPLATE_LITERAL_MIDDLE:
    '}' IN_TEMPLATE_LITERAL* '${';
terminal TEMPLATE_LITERAL_END:
    '}' IN_TEMPLATE_LITERAL* '`';
terminal fragment IN_TEMPLATE_LITERAL:
    /[^$`]|\\\$|\\`/;
```

## The current behavior

<img width="375" alt="image" src="https://github.com/eclipse-langium/langium/assets/13158903/72f08cfc-34f0-4582-ab72-a6d8f92256b1">

## The expected behavior

no error in output

msujew:
@codehz Thanks for the report. As a quick workaround you can adjust the mode in your `langium-config.json` file:

```json5
{
  ...
  "mode": "production"
}
```

PR #1514, opened 2024-05-27
Escape `${` in template string

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1513

Whenever `${` was included in a keyword, the generated template literal expected an expression, resulting in a parser error. This change escapes the `${` as expected.

### DIFF
--- a/packages/langium-cli/src/generator/grammar-serializer.ts
+++ b/packages/langium-cli/src/generator/grammar-serializer.ts
@@ -41,9 +41,13 @@ export function serializeGrammar(services: LangiumCoreServices, grammars: Gramma
                 });
                 // The json serializer returns strings with \n line delimiter by default
                 // We need to translate these line endings to the OS specific line ending
-                const json = normalizeEOL(serializedGrammar
+                let json = normalizeEOL(serializedGrammar
                     .replace(/\\/g, '\\\\')
                     .replace(new RegExp(delimiter, 'g'), '\\' + delimiter));
+                if (!production) {
+                    // Escape ${ in template strings
+                    json = json.replace(/\${/g, '\\${');
+                }
                 return expandToNode`
 
                     let loaded${grammar.name}Grammar: Grammar | undefined;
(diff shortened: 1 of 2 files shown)


packages/langium-cli/src/generator/grammar-serializer.ts
```diff
@@ -41,9 +41,13 @@ export function serializeGrammar(services: LangiumCoreServices, grammars: Gramma
                 });
                 // The json serializer returns strings with \n line delimiter by default
                 // We need to translate these line endings to the OS specific line ending
-                const json = normalizeEOL(serializedGrammar
+                let json = normalizeEOL(serializedGrammar
                     .replace(/\\/g, '\\\\')
                     .replace(new RegExp(delimiter, 'g'), '\\' + delimiter));
+                if (!production) {
```
Lotes, line 47: WTF? Why not in production?

packages/langium-cli/test/generator/grammar-serializer.test.ts
```diff
@@ -45,4 +51,29 @@ describe('Grammar serializer', () => {
         expect(moduleString).toMatch('"$comment": "/**\\\\n * This is the entry rule\\\\n */"');
     });
 
+    test('should escape template strings in development mode', async () => {
+        // arrange
+        const config: LangiumConfig = {
+            projectName: 'Magic',
+            languages: [],
```
Lotes, line 58: May add non-production explicitly. And add a test for production?

Lotes (commented):
Thanks for the changes, but I still need some explanation. I read the original problem. But how does minifying (mode=production) avoids this problem?

packages/langium-cli/src/generator/grammar-serializer.ts
montymxb: @Lotes The issue isn't present in production mode

msujew: Because the minifier uses `'` instead of template strings to embed the grammar in code. The `${` characters only result in issues in template strings.

PR #1516, opened 2024-05-28
Show documentation on completion item

msujew:
See discussion in https://github.com/eclipse-langium/langium/discussions/1501#discussioncomment-9579811.

This change aligns Langium's default behavior to most programming languages, where the documentation of an element is shown on the corresponding completion item.

### DIFF
--- a/packages/langium/test/lsp/completion-provider.test.ts
+++ b/packages/langium/test/lsp/completion-provider.test.ts
@@ -12,6 +12,7 @@ import { DefaultCompletionProvider } from 'langium/lsp';
 import type { LangiumServices, PartialLangiumServices } from 'langium/lsp';
 import { clearDocuments, expectCompletion, parseHelper } from 'langium/test';
 import { MarkupContent } from 'vscode-languageserver';
+import * as assert from 'assert';
 
 describe('Langium completion provider', () => {
 
@@ -206,7 +207,7 @@ describe('Completion within alternatives', () => {
                 if (MarkupContent.is(item.documentation)) {
                     return item.documentation.value;
                 } else {
-                    return item.documentation ?? '';
+                    assert.fail('Completion item should be of type `MarkupContent`.');
                 }
             }
         });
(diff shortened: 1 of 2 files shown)


packages/langium/test/lsp/completion-provider.test.ts
```diff
@@ -178,6 +179,39 @@ describe('Completion within alternatives', () => {
...
+        person A
+        hello <|>
+        `;
+
+        await completion({
+            text,
+            index: 0,
+            expectedItems: ['Hello this is A'],
+            itemToString: item => {
+                if (MarkupContent.is(item.documentation)) {
```
Lotes, line 206: For a test it confuses me that you have an if-else here... can it be something different in 2 runs?

Lotes: You could use an assertion instead. A branch that is never called, feels also weird.

msujew: Right, I've updated the code to use an assertion 👍

Lotes: I meant to assert First without an If. Can show on Monday

Lotes: Something like this here... it does not need an `if`. You just assume that you are in the right branch. If you are not you get an assertion error.

```ts
assert(MarkupContent.is(item.documentation));
return item.documentation.value;
```

In doubt, you can always create your own assert function.

```ts
function myAssert(condition: boolean) : asserts condition {
  if(!condition) throw Error();
}
```

I am not sure if you already know this, so I wrote it down. It makes testing a bit shorter and helps also to make some intrinsic assumptions visible.

msujew: Thanks, that looks way better 👍

Lotes (approved):
LGTM.

PR #1525, opened 2024-06-03
Fix range comparison

jindong-zhannng:
The start of ranges in LSP is zero-based.

I don't know is this function still in use, just fix it while reading source codes.

### DIFF
--- a/packages/langium/src/utils/cst-utils.ts
+++ b/packages/langium/src/utils/cst-utils.ts
@@ -81,13 +81,14 @@ export enum RangeComparison {
     After = 1,
     OverlapFront = 2,
     OverlapBack = 3,
-    Inside = 4
+    Inside = 4,
+    Outside = 5,
 }
 
 export function compareRange(range: Range, to: Range): RangeComparison {
-    if (range.end.line < to.start.line || (range.end.line === to.start.line && range.end.character < range.start.character)) {
+    if (range.end.line < to.start.line || (range.end.line === to.start.line && range.end.character <= to.start.character)) {
         return RangeComparison.Before;
-    } else if (range.start.line > to.end.line || (range.start.line === to.end.line && range.start.character > to.end.character)) {
+    } else if (range.start.line > to.end.line || (range.start.line === to.end.line && range.start.character >= to.end.character)) {
         return RangeComparison.After;
     }
     const startInside = range.start.line > to.start.line || (range.start.line === to.start.line && range.start.character >= to.start.character);
@@ -96,8 +97,10 @@ export function compareRange(range: Range, to: Range): RangeComparison {
         return RangeComparison.Inside;
     } else if (startInside) {
         return RangeComparison.OverlapBack;
-    } else {
+    } else if (endInside) {
         return RangeComparison.OverlapFront;
+    } else {
+        return RangeComparison.Outside;
     }
 }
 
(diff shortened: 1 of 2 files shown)


msujew (commented):
I'm not sure what this PR is supposed to fix (except for the typo in line 88 `range` -> `to`). Is there an actual issue associated with this?

Note that the code is designed to use `<` and `>`, since the `<=` and `>=` operators are used below (lines 93-94) to check whether a range is inside of another.

Also, in order to accept your contribution, please sign the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php) with the email address that was used to create the commit.

jindong-zhannng:
I added test cases to cover all my changes. For the range comparison, here is a reference implementation from vscode: https://github.com/microsoft/vscode/blob/addd445017c9c5531b89851dc8dc9bec4f1dd0f8/src/vs/editor/common/core/range.ts#L420C16-L420C31

packages/langium/test/utils/cst-utils.test.ts
```diff
@@ -60,3 +61,82 @@ describe('findLeafNode', () => {
...
+
+    test.each([
+        // Multiple lines
+        [{ start: { line: 1, character: 1 }, end: { line: 3, character: 10 } }, { start: { line: 2, character: 1 }, end: { line: 4, character: 10 } }],
+        // Same line
+        [{ start: { line: 1, character: 1 }, end: { line: 1, character: 10 } }, { start: { line: 1, character: 5 }, end: { line: 1, character: 15 } }],
+    ])('OverlapFront', (range, to) => {
+        const result = compareRange(range, to);
+        expect(result).toEqual(RangeComparison.OverlapFront);
+    });
```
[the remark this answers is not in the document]
msujew: `OverlapFront` looks right to me. IMO the function should return how the first range relates to the second (i.e. the first overlaps in front of the second one).

packages/langium/src/utils/cst-utils.ts
```diff
@@ -81,13 +81,14 @@ export enum RangeComparison {
...
     OverlapFront = 2,
     OverlapBack = 3,
-    Inside = 4
+    Inside = 4,
+    Outside = 5,
 }
 
 export function compareRange(range: Range, to: Range): RangeComparison {
-    if (range.end.line < to.start.line || (range.end.line === to.start.line && range.end.character < range.start.character)) {
+    if (range.end.line < to.start.line || (range.end.line === to.start.line && range.end.character <= to.start.character)) {
```
msujew, line 89: There is an IMO unexpected edge case in case the `range` is a zero-width range at the start of `to`. The expected result would be `Inside`, but it currently results in `Before`:

```ts
// This currently results in `Before`
[{ start: { line: 1, character: 1 }, end: { line: 1, character: 1 } }, { start: { line: 1, character: 1 }, end: { line: 1, character: 5 } }]
```

msujew (commented):
Thanks for the tests!

Have you already signed the ECA with the email that was used to create the commit? The automated checking process still says it's not signed, see [here](https://api.eclipse.org/git/eca/status/gh/eclipse-langium/langium/1525).

packages/langium/src/utils/cst-utils.ts
msujew: > I signed it just now, may need some time for syncing.

Seems to have gone through, thanks!

> Well it's actually expected for me. Zero-width range is a hard-to-imagine concept. I don't think a zero-width range is a real range, it's more like an insertable position.

Yeah, I guess. Especially in the context that we use it, it will probably never appear anyway.

msujew (approved):
LGTM, thanks 👍

Issue #1536, opened 2024-06-08
Single quotes in property types are not correctly escaped


msujew:
Langium version: 3.0
Package name: langium-cli

## Steps To Reproduce

1. Use a rule like `A: value="'test'"`;
2. The generated `ast.ts` does not compile

## The current behavior

It generates the string type as `''test''`.

## The expected behavior

It's supposed to generate the type as `'\'test\''`.

PR #1537, opened 2024-06-08
Escape strings in generated types

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1536

Correctly escapes all quotes inside of string values in the generated `ast.ts` file.

Also removes a few quotes in case they weren't needed (like for type references inside the ast reflection). This should improve bundle size.

### DIFF
--- a/packages/langium-cli/src/generator/ast-generator.ts
+++ b/packages/langium-cli/src/generator/ast-generator.ts
@@ -8,7 +8,7 @@ import { type Generated, expandToNode, joinToNode, toString } from 'langium/gene
 import type { AstTypes, Property, PropertyDefaultValue } from 'langium/grammar';
 import type { LangiumConfig } from '../package-types.js';
 import { AstUtils, MultiMap, GrammarAST } from 'langium';
-import { collectAst, collectTypeHierarchy, findReferenceTypes, isAstType, mergeTypesAndInterfaces } from 'langium/grammar';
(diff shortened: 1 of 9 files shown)


msujew:
cc @AlexandreLachanceGit I've seen that you've been encountering this issue in [this parser rule](https://github.com/AlexandreLachanceGit/langium-protobuf/blob/7c22d28701df960fe984df18c6a46a8308c6bcf7/src/language/protobuf.langium#L20-L21). Note that you can work around it by giving the rule a return type:

```
interface Syntax {
    syntax: string;
}

Syntax returns Syntax: 
    'syntax' '=' syntax=('"proto3"' | "'proto3'") ';';
```

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -147,7 +148,7 @@ function buildReferenceTypeMethod(crossReferenceTypes: CrossReferenceType[]): Ge
                 joinToNode(
                     buckets.entriesGroupedByKey(),
                     ([target, refs]) => expandToNode`
-                        ${joinToNode(refs, ref => `case '${ref}':`, { appendNewLineIfNotEmpty: true, skipNewLineAfterLastItem: true})} {
+                        ${joinToNode(refs, ref => `case "${escapeQuotes(ref)}":`, { appendNewLineIfNotEmpty: true, skipNewLineAfterLastItem: true})} {
```
spoenemann, line 151: Why is `"` the default now? We could also escape `'` instead?

spoenemann (approved):

Issue #1543, opened 2024-06-13
Concatenated stream elements can't be accessed multiple times


joao-silveira:
When accessing the elements of a concatenated stream, the concatenated elements are missing when accessing the stream a second time. In any case the elements passed in the `stream(...)` call are always available.

Langium version: 3.0.0
Package name: Langium

Code example:

```
const countTest = stream([1]).concat([2])
countTest.count()
countTest.toArray() // returns [ 1 ] but expected was [1, 2]
```

```
const toStringTest = stream([1]).concat([2])
toStringTest.toString()
toStringTest.toArray() // returns [ 1 ] but expected was [1 , 2]
```

```
const findTest = stream([1]).concat([2])
findTest.some((it: number) => it === 2) //returns true
findTest.some((it: number) => it === 2) //returns false but expected was true
```

This problem really seems to be specific to streams concatenations because i tried other variants of the code above with `stream([1], [2])` or `stream([1], stream([2]))` and all of those work.

msujew:
Thanks for catching, see https://github.com/eclipse-langium/langium/pull/1545 for a fix.

PR #1545, opened 2024-06-13
Improve idempotency of stream operations

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1543

Moves the external state of some of the stream operations to the internal state.


spoenemann (approved):
👍

Issue #1361, opened 2024-01-29
Broken cstNode calculation


cdietrich:
given the grammar

```
grammar HelloWorld

entry Expression:
    HasValue;

HasValue infers Expression:
    DotExpression({infer HasValue.left=current} 'has-value')*;

DotExpression infers Expression:
    RefExpression ({infer DotExpression.base = current} '.' target=DotTarget)*;

RefExpression:
    target=ID;

DotTarget:
   {infer DotTargetRef} ref=ID;

ID returns string: RawId | EscapedId;

hidden terminal WS: /\s+/;
terminal RawId: /[_a-zA-Z][\w_-]*/;
terminal EscapedId: /`[^`]*`/;

hidden terminal ML_COMMENT: /\/\*[\s\S]*?\*\//;
hidden terminal SL_COMMENT: /\/\/[^\n\r]*/;
```

and sample model

```
bs.cs.xx has-value
```

the cst-node for the dot expression

`bs.cs.xx`

is `.xx` and as not expected `bs.cs.xx`

msujew:
Might be related to https://github.com/eclipse-langium/langium/pull/932.

lars-reimann:
It also seems to be related to #1218.

cdietrich:
thx for fixing

Issue #1218, opened 2023-10-07
Range of CST nodes after repeated tree rewriting


lars-reimann:
Langium version: 2.0.2
Package name: langium

## Steps To Reproduce

1. Clone the repo linked below. It contains a slightly altered hello-world project that allows writing chained member accesses like `A.B` or `A.B.C`. For the member access directly below the root, a warning is issued.
2. Generate, build, and run the extension.
3. Create a new file with the `.hello` extension.
4. Type `A.B` into the file.
5. Type `A.B.C` into the file.

Link to code example: https://github.com/lars-reimann/langium-cst-range

## The current behavior

In step 4, the entire expression gets highlighted correctly:

![image](https://github.com/eclipse-langium/langium/assets/2501322/b7f4b621-8a91-4e44-9122-1863757cd663)

In step 5, however, only the part starting at the dot gets highlighted. The computed range of the CST node is also too short:

![image](https://github.com/eclipse-langium/langium/assets/2501322/1ee2b666-32cc-4a47-8803-82752174b3a1)

The [playground](https://langium.org/playground?grammar=OYJwhgthYgBAEgUwDbIPYHU0mQEwFD4CWAdgC6IgBmYAxorALJq4CuyDA3vrL7IgA8ADiEQBnMUTQkAXLACiw0RKkl8AX0KJyIAJ5MW7BqLKsQJMQbYcZPPoJHjJ0gLyNEEAEaUAgrXoS%2BADchKQU1HQMio4q0rCcmsTklDT0TB7eIH4BloIUJLiW0crOJPF2vKL0RABulHLFTqoVsBAZ9bAASohUlNr0GoTuXr7%2BTrAmZhYKSk3StnxdPX0kA4sAFC2LnMOZ2U4AdFWItZQutGai5JqLiwDkB3dbfG0jIC7dvVdrfACUAFTBULJCJpT4rNJ5bSFGYxUrlRYkSCIORiMggUjAQb4cHfYyIUzmSy4-oolqcEmrRDqWBItouACSABEgfgABZEXC4bSwcIQUhgZCwDAAZTkAHoADpiADU4pCfIFQuZEoA2gB9MAAWgAXj4tQAtAC6qslAHd1Ub-vL8EA&content=IIOgQiDCQ) (or a debugging session) shows that the AST is built properly:

```json5
{
  $type: "Module", 
  expression: 
  {
    $type: "MemberAccess", 
    receiver: 
    {
      $type: "MemberAccess", 
[...]

PR #1547, opened 2024-06-15
Fix CST related parser issues

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1218
Closes https://github.com/eclipse-langium/langium/issues/1361
Supersedes https://github.com/eclipse-langium/langium/pull/932

Related discussion https://github.com/eclipse-langium/langium/discussions/1539

Mainly fixes 3 issues related to how we process CST/AST nodes during parsing:

1. When performing an **unassigned action** (i.e. an action that just rewrites the type) we only adjust the type without touching anything else (see also https://github.com/eclipse-langium/langium/discussions/1539).
2. When performing an **assigned action**, we need to rewrite the CST order (see https://github.com/eclipse-langium/langium/issues/1218 and https://github.com/eclipse-langium/langium/issues/1361).
3. When performing an **unassigned subrule** call, we need to unassign the new node as the owner of the CST. Only after completing the CST construction does the current AST node become the owner of an CST node. Otherwise, we keep stale AST references in the CST (see #932).

### DIFF
--- a/packages/langium/src/parser/cst-node-builder.ts
+++ b/packages/langium/src/parser/cst-node-builder.ts
@@ -133,7 +133,7 @@ export abstract class AbstractCstNode implements CstNode {
         return node;
     }
 
-    set astNode(value: AstNode) {
+    set astNode(value: AstNode | undefined) {
         this._astNode = value;
     }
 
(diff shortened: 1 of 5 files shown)


spoenemann (commented):
Thanks, looks good in general. I think this would be a good opportunity to add a few more comments to langium-parser.ts – could you document the BaseParser methods and maybe add code comments in those places where specific assumptions are made?

spoenemann (approved):
👍

PR #1552, opened 2024-06-19
Fix data type rule type computation

msujew:
Fixes a regression introduced in https://github.com/eclipse-langium/langium/pull/1547 while attempting to perform a parser startup optimization. Reverts the optimization and adds regression tests for that case.


JohannesMeierSE:
The test cases look very nice for me, thanks @msujew!

spoenemann (approved):

Issue #1558, opened 2024-06-21
The `langium@3.1.0` changes broke existing grammars


Yokozuna59:
Langium version: `3.1.0`
Package name: `langium`

## Steps To Reproduce

1. Create a Hello World example using the generator.
2. Modify the grammar to has a rule with multiplicity without assigning to any value:

	```diff
	grammar HelloWorld

	entry Model:
	+  	ID*
    	(persons+=Person | greetings+=Greeting)*;
	```

Link to code example:

## The current behavior

The changes after #1401 broke the existing working grammars.

![image](https://github.com/eclipse-langium/langium/assets/87622592/6dd60c94-21e1-44d8-927b-f4bcd319bd7a)

![image](https://github.com/eclipse-langium/langium/assets/87622592/8b278e36-b45d-4957-b484-d892ccdd92f2)

## The expected behavior

No errors.

spoenemann:
The change is as expected. What is your use case for a multiplicity without an assignment?

I your second example, I'd recommend to create a data type rule:
```
Newlines returns string: NEWLINE+;
```
and then use `Newlines?` instead of `NEWLINE*`.

We could consider making an exception for terminal rules, though.

msujew:
@spoenemann I'd argue that we should make an exception for terminal and data type rules.

msujew:
The latest release (3.1.1) includes a fix for this.

spoenemann:
What should we do with the milestone assignment in such a case? Assign to v3.1.0 or v3.2.0? No milestone?

I think assigning to v3.1.0 would be most helpful.

msujew:
I agree, I've added it to the v3.1.0 milestone.

PR #1559, opened 2024-06-21
Fix false positive multiplicity validation

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1558

Adds exception in case the targeted rule is a data type rule or a terminal rule.

Also already pulls up the versions for the core package and vscode extension.


spoenemann (approved):
👍

PR #1563, opened 2024-06-27
BugFix: Correct isAstTypeInternal checking logic

montymxb:
This corrects a recently found bug where the internal `isAstTypeInternal` logic would return `false` for property types that should otherwise return true. The issue has to do with the fact we always return false when encountering a previously visited prop type, instead of returning it's previously computed result when we visited. To correct, we just make a change from `Set` -> `Map` and adjust the logic accordingly to save results as we go.

I would _really_ like to add a test case to this, but the affected grammar is unable to be shared, and I haven't been able to reliably factor out a small test grammar; at least so far. I'll keep this in draft until we make a decision in this regard.


montymxb:
Nice! Thanks @msujew , that should work.

msujew (approved):
Alright, looks good to me 👍

PR #1565, opened 2024-06-28
make document symbol provider able to handle empty names

cdietrich:


cdietrich:
@msujew this is a corner case is our grammar where empty names are possible with string literals
wdyt?
alternative would be to move 

```
name: name ?? nameNode.text,
```

or

```
{
            return [{
                kind: this.nodeKindProvider.getSymbolKind(astNode),
                kind: this.nodeKindProvider.getSymbolKind(astNode),
                name: name ?? nameNode.text,
                name: name === undefined || name.length === 0 ? nameNode.text : name,
                range: node.range,
                range: node.range,
                selectionRange: nameNode.range,
                selectionRange: nameNode.range,
                children: this.getChildSymbols(document, astNode)
                children: this.getChildSymbols(document, astNode)
            }
```

to a helper

packages/langium/src/lsp/document-symbol-provider.ts
```diff
@@ -48,7 +48,7 @@ export class DefaultDocumentSymbolProvider implements DocumentSymbolProvider {
             const name = this.nameProvider.getName(astNode);
             return [{
                 kind: this.nodeKindProvider.getSymbolKind(astNode),
-                name: name ?? nameNode.text,
+                name: name === undefined || name.length === 0 ? nameNode.text : name,
```
msujew, line 51: IMO we can just replace this with `||` since that catches `undefined` and `length === 0`, due to how falsyness operates on strings.
```suggestion
                name: name || nameNode.text,
```

msujew (approved):
LGTM

PR #1566, opened 2024-07-01
Ensure diagnostics are published after cancellation

msujew:
Fixes an issue discovered in https://github.com/eclipse-langium/langium/discussions/1562.

This is currently a draft, because I want to add some tests to confirm the behavior.

### DIFF
--- a/packages/langium/src/workspace/document-builder.ts
+++ b/packages/langium/src/workspace/document-builder.ts
@@ -15,7 +15,7 @@ import type { ValidationOptions } from '../validation/document-validator.js';
 import type { IndexManager } from '../workspace/index-manager.js';
 import type { LangiumDocument, LangiumDocuments, LangiumDocumentFactory } from './documents.js';
 import { MultiMap } from '../utils/collections.js';
-import { OperationCancelled, interruptAndCheck } from '../utils/promise-utils.js';
+import { OperationCancelled, interruptAndCheck, isOperationCancelled } from '../utils/promise-utils.js';
 import { stream } from '../utils/stream.js';
 import type { URI } from '../utils/uri-utils.js';
 import { ValidationCategory } from '../validation/validation-registry.js';
@@ -78,9 +78,21 @@ export interface DocumentBuilder {
     onUpdate(callback: DocumentUpdateListener): Disposable;
 
     /**
-     * Notify the given callback when a set of documents has been built reaching a desired target state.
+     * Notify the given callback when a set of documents has been built reaching the specified target state.
      */
     onBuildPhase(targetState: DocumentState, callback: DocumentBuildListener): Disposable;
+
+    /**
+     * Notify the specified callback when a document has been built reaching the specified target state.
+     * Unlike {@link onBuildPhase} the listener is called for every single document.
+     *
+     * There are two main advantages compared to {@link onBuildPhase}:
+     * 1. If the build is cancelled, {@link onDocumentPhase} will still fire for documents that have reached a specific state.
+     *    Meanwhile, {@link onBuildPhase} won't fire for that state.
+     * 2. The {@link DocumentBuilder} ensures that all {@link DocumentPhaseListener} instances are called for a built document.
+     *    Even if the build is cancelled before those listeners were called.
+     */
+    onDocumentPhase(targetState: DocumentState, callback: DocumentPhaseListener): Disposable;
 
     /**
      * Wait until the workspace has reached the specified state for all documents.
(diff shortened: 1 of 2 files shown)


msujew:
cc @cdietrich Can you try including this change into your project? It should now ensure that diagnostics are published as soon as **any document** reaches the `Validated` state.

cdietrich:
@msujew could you provide an alpha, this would make adoption the easiest

msujew:
@cdietrich Published as `3.2.0-next.2191297`

cdietrich:
thx. am gonna do some test rounds

cdietrich:
it looks much better now. but i very seldomly still see issues.
am not sure if other phases might have similar side effects
need also to figure out if this is a server side or client side problem

cdietrich:
i indeed can see

- all documents are in state 6 (validated)
- there are references that cannot be resolved.

it happens very very rarely. i dont have deeper insights yet.
i guess i need to reestablish more extensive logging ontop of your changes again

cdietrich:
the two effects i see.
- doLink skips some references cause they are in state linking error
- sometimes elements are not in the index at all.

cdietrich:
@msujew

cdietrich:
@msujew 

`get ref() {` this one may turn references into errors, but wont push them to document.references. (also for foreign documents)
thus i wonder if then, after a cancellation.
the unlink on the forein document, that was not linked yet,
will really unlink it. if i check the code, it wont

cdietrich:
grafik

cdietrich:
another qustion: should changed files always be unlinked?

cdietrich:
hmmmmmmm

have found this one:

```
// Some of these documents can be pretty large, so parsing them again can be quite expensive.
        // Therefore, we only parse if the text has actually changed.
        if (oldText !== text) {
        ```
        
        => the current way of invalidateDocument might not be sufficicent?

**Comment by cdietrich** [comment]
patched the code and the trap hits

```
// 0. Parse content
    await this.runCancelable(documents, DocumentState.Parsed, cancelToken, async doc => {
      await this.langiumDocumentFactory.update(doc, cancelToken)
      for (const node of AstUtils.streamAst(doc.parseResult.value)) {
        AstUtils.streamReferences(node).forEach(refInfo => {
          const ref = refInfo.reference as DefaultReference
          if (ref._ref !== undefined) {
            if (isLinkingError(ref._ref)) {
              console.log(`mimimi langiumDocumentFactory.update failure ${ref._ref.message} ${doc.uri.toString()}`)
            }
          }
        })
      }
    })
```

**Comment by cdietrich** [comment]
created https://github.com/eclipse-langium/langium/issues/1567 for further discussion

**Comment by msujew** [comment]
@cdietrich Thanks, I was able to reproduce. Newest version available at `3.2.0-next.9134e27`.

**Comment by cdietrich** [comment]
@msujew i still think the problem is also that unlink/shouldRelink only evaluates document.references, but not the references that were attempted to link with the `get ref` call before, which does not maintain document.referencs.

**Comment by msujew** [comment]
@cdietrich Yep, noticed that as well. Currently working on that :)

**Comment by cdietrich** [comment]
i have tried my unlink changes on top of your invalidateDocument change.
but i still see broken links after parsing step. need to check

**Comment by cdietrich** [comment]
problem is the js you published. it has
   if (langiumDoc) {
            // const linker = this.serviceRegistry.getServices(uri).references.Linker;
            // linker.unlink(langiumDoc);

**Comment by msujew** [comment]
@cdietrich After uncommenting the change, everything works as expected? I probably forgot to recompile.

**Comment by cdietrich** [comment]
will check.

**Comment by cdietrich** [comment]
your invalidate with my unlink look promising.

**Comment by cdietrich** [comment]
tests look good.
will continue the analyze the deleted case.

maybe the didChangeContent and the didChangeWatchedFiles reace with each other.

**Comment by cdietrich** [comment]
delete case seems to be a client side problem. so here we are looking good

File: packages/langium/src/workspace/document-builder.ts [hunk 1]
```diff
@@ -209,22 +225,49 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+     * Sort the given documents by priority. By default, documents with an open text document are prioritized.
+     * This is useful to ensure that visible documents show their diagnostics before all other documents.
+     *
+     * This improves the responsiveness in large workspaces as users usually don't care about diagnostics
+     * in files that are currently not opened in the editor.
+     */
+    protected sortDocuments(documents: LangiumDocument[]): LangiumDocument[] {
+        const hasTextDocument = new Map<LangiumDocument, boolean>();
+        for (const doc of documents) {
+            hasTextDocument.set(doc, !!this.textDocuments?.get(doc.uri.toString()));
```
spoenemann, line 256: Detail: I'd prefer `Boolean(...)` over `!!`

spoenemann (approved):
Great improvements! 👍

spoenemann:
Shall we add this to the 3.1.0 milestone and make another patch release?

cdietrich:
a patch release would be welcome 🙏

PR #1577, opened 2024-07-12
Detect cyclic reference resolution

spoenemann:
When a custom scope provider (directly or indirectly) tries to access the reference it's supposed to create a scope for, we get an infinite recursive loop. This change detects such a case early and generates a proper error message.

Further changes:
- When an error is caught during linking (usually caused by a custom scope provider), we print it to the error log.
- Don't strictly require a containing document for resolving references.


kaisalmen (commented):
Changes look good to me, but @msujew should review as well and approve.
A new release with this change will be appreciated. 🙂

PR #1578, opened 2024-07-14
Add indentation-aware TokenBuilder and Lexer

### LINKED ISSUE #1016: Indentation tokenization based on terminal annotations
*Inspired by https://github.com/langium/langium/discussions/1015*

We could offer to build `INDENT`/`DEDENT` tokens in the `DefaultTokenProvider` based on comment annotations on terminals:

```antlr
/**
 * @token.indent
 */
terminal INDENT: 'indent';
/**
 * @token.dedent
 */
terminal DEDENT: 'dedent';
```

aabounegm:
This is useful for languages where indentation is used as a block delimiter, such as Python.
I will try to add a recipe for it in the documentation (and maybe some unit tests) later on.

The code was heavily inspired by the [Python indentation chevrotain example](https://github.com/chevrotain/chevrotain/blob/master/examples/lexer/python_indentation/python_indentation.js), but modified to be more extensible and hopefully fitting for Langium.

Usage example (just a snippet, not full grammar):

```langium
fragment Body:
  INDENT
  statements+=Statement
  DEDENT
;

If: 'if' cond=Expr ':' Body;

hidden terminal NEW_LINE: /\r?\n/;
terminal INDENT: 'synthetic:indent'; // The content doesn't really matter.
terminal DEDENT: 'synthetic:dedent'; // It will be overridden anyway
hidden terminal WS: /[ \t]+/; // whitespace and newlines now have to be separated
```

The main drawback is that the whole language now becomes indentation-sensitive, even areas which would normally not be white-space sensitive, such as the following Python example:
```python
numbers = [
  1,  // unexpected indentation
  2,
]
```
[Multi-mode lexing](https://github.com/eclipse-langium/langium-website/pull/132) can be leveraged to overcome this, and if I find a generic way to integrate it with the code from this PR, I will follow it up with another one that does so.

Closes #1016.
Related to #608, #663, #782, and #1085.


aabounegm:
Whoops, looks like I messed up a bit with the imports there. That's what I get for using github.dev without cloning locally, I guess 😅. I'll fix it in a sec

aabounegm:
@msujew Should hopefully be fine now 😄

packages/langium/src/parser/lexer.ts
```diff
@@ -66,6 +67,45 @@ export class DefaultLexer implements Lexer {
...
+ * In your module, you can override the default lexer with this one as such:
+ * ```ts
+ * parser: {
+ *    TokenBuilder: () => new IndentationAwareTokenBuilder(),
+ *    Lexer: (services) => new IndentationAwareLexer(services),
+ * }
+ * ```
+ */
+export class IndentationAwareLexer extends DefaultLexer {
+    private indentationTokenBuilder?: IndentationAwareTokenBuilder;
```
msujew, line 84: Suggestion: With the way our service implementations are designed (i.e. being capable of overriding everything), we don't keep `private` methods/fields. Please mark them all as `protected`. This also applies to the other file. Also, make this field `readonly`.

packages/langium/src/parser/lexer.ts
```diff
@@ -66,6 +67,45 @@ export class DefaultLexer implements Lexer {
...
+
+    constructor(services: LangiumCoreServices) {
+        super(services);
+        if (services.parser.TokenBuilder instanceof IndentationAwareTokenBuilder) {
+            this.indentationTokenBuilder = services.parser.TokenBuilder;
```
msujew, line 89: Suggestion: I believe we should make it very clear that this lexer is supposed to be used in conjunction with the `IndentationAwareTokenBuilder` (or an extension of it). Given that every other combination is pretty nonsensical, we should throw an error if we find that the token builder is not of type `IndentationAwareTokenBuilder`.

This also allows us to make the `indentationTokenBuilder` field non-optional.

packages/langium/src/parser/token-builder.ts
```diff
@@ -118,3 +119,261 @@ export class DefaultTokenBuilder implements TokenBuilder {
...
+     * ```langium
+     * hidden terminal WS: /[ \t]+/;
+     * ```
+     *
+     * @default 'WS'
+     */
+    whitespaceTokenName: string;
+}
+
+const indetationBuilderDefaultOptions: IndentationTokenBuilderOptions = {
```
msujew, line 156: Suggestion: Adopters might want to reuse the value. Also, typo:
```suggestion
export const IndentationBuilderDefaultOptions: IndentationTokenBuilderOptions = {
```

packages/langium/src/parser/token-builder.ts
```diff
@@ -118,3 +119,261 @@ export class DefaultTokenBuilder implements TokenBuilder {
...
+        return tokenType;
+    }
+
+    /**
+     * Resets the indentation stack between different runs of the lexer
+     *
+     * @param text Full text that was tokenized
+     * @returns Remaining dedent tokens to match all previous indents at the end of the file
+     */
+    public popRemainingDedents(text: string) {
```
msujew, line 367:
```suggestion
    popRemainingDedents(text: string): IToken[] {
```

packages/langium/src/parser/token-builder.ts
```diff
@@ -118,3 +119,261 @@ export class DefaultTokenBuilder implements TokenBuilder {
...
+        const dedent = tokenTypes.find(tok => tok.name === dedentTokenName);
+        const indent = tokenTypes.find(tok => tok.name === indentTokenName);
+        const ws = tokenTypes.find(tok => tok.name === whitespaceTokenName);
+        if (!dedent || !indent || !ws) {
+            throw new Error('Some indentation/whitespace tokens not found!');
+        }
+
+        const spaceTokens = [dedent, indent, ws];
+        const otherTokens = tokenTypes.filter(tok => !spaceTokens.includes(tok));
+        return [...spaceTokens, ...otherTokens];
```
msujew, line 227: Suggestion: This iterates over all tokens 4 times. While it's probably pretty minor, we can make this a bit more efficient:
```suggestion
        let dedent: TokenType | undefined;
        let indent: TokenType | undefined;
        let ws: TokenType | undefined;
        const otherTokens: TokenType[] = [];
        for (const tokenType of tokenTypes) {
            if (tokenType.name === dedentTokenName) {
                dedent = tokenType;
            } else if (tokenType.name === indentTokenName) {
                indent = tokenType;
            } else if (tokenType.name === whitespaceTokenName) {
                ws = tokenType;
            } else {
                otherTokens.push(tokenType);
            }
        }
        if (!dedent || !indent || !ws) {
            throw new Error('Some indentation/whitespace tokens not found!');
        }
        return [dedent, indent, ws, ...otherTokens];
```

msujew (commented):
Thanks for the contribution @aabounegm!

Do you mind creating some tests cases for the new lexer?

aabounegm:
Sure, no problem, but that might take me some time (not sure if I will be able to manage it today)

msujew:
@aabounegm All good, take your time. I appreciate it!

spoenemann (commented):
Thanks for the contribution! From a code organization standpoint, I'd prefer the two new service classes to be in a new file `indentation-aware.ts`.

aabounegm:
Thanks for your comments, I have now added unit tests and moved the new classes to a new file as suggested.
Please feel free to nitpick, or to edit directly if you so prefer

msujew (approved):
Thanks @aabounegm, looks pretty good to me 👍

packages/langium/test/parser/indentation-aware.test.ts
```diff
@@ -181,69 +194,100 @@ describe('indentationAwareLexer', () => {
...
 
 });
+
+type Statement = If | Return;
+
+interface If extends AstNode {
+    $type: 'If';
+    condition: boolean;
+    thenBlock: Statement[];
+    elseBlock?: Statement[];
```
aabounegm, line 287: Why is this field optional? Since it uses `+=`, shouldn't there always be an array (even if empty) by default?

aabounegm (commented):
Thank you for your fixes!

packages/langium/test/parser/indentation-aware.test.ts
msujew: GitHub Copilot strikes again. Yes, you're right :D

spoenemann (approved):
👍

PR #1587, opened 2024-07-18
Hotfix: undefined token on indentation-aware empty input

aabounegm:
If the input was empty, the `IndentationAwareLexer` would previously return `[ undefined ]` for the list of tokens.


msujew (approved):
Ah, thanks. Looks like my optimization didn't work out as expected. Thanks for catching that 👍

PR #1589, opened 2024-07-19
Detect \n and \t as whitespace

aabounegm:
Previously, splitting a terminal rule into separate "spaces/tabs" and "newlines" terminals would cause the "newlines" one to be detected as a comment when generating TextMate grammar.

This would generate a TextMate pattern that captures the whole line as such:
```json
{
  "begin": "",
  "beginCaptures": {
    "1": {
      "name": "punctuation.whitespace.comment.leading.mylang"
    }
  },
  "end": "(?=$)",
  "name": "comment.line.mylang"
}
```

To test current behavior, go to the playground and replace the rule `hidden terminal WS: /s+/` with:
```langium
hidden terminal WS: /[ \t]+/;
hidden terminal NL: /[\r\n]+/;
```
and all the text will be highlighted as a comment.

P.S.: The main reason to split `\s+` into two terminals like so is for use with the new `IndentationAwareTokenBuilder`

### DIFF
--- a/packages/langium/src/utils/regexp-utils.ts
+++ b/packages/langium/src/utils/regexp-utils.ts
@@ -138,9 +138,17 @@ export function isMultilineComment(regexp: RegExp | string): boolean {
     }
 }
 
+/**
+ * A set of all characters that are considered whitespace by the '\s' RegExp character class.
+ * Taken from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Character_classes).
+ */
+export const whitespaceCharacters = (
+    '\f\n\r\t\v\u0020\u00a0\u1680\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007' +
+    '\u2008\u2009\u200a\u2028\u2029\u202f\u205f\u3000\ufeff').split('');
+
 export function isWhitespace(value: RegExp | string): boolean {
     const regexp = typeof value === 'string' ? new RegExp(value) : value;
-    return [' ', '\n', '\t'].some((ws) => regexp.test(ws));
+    return whitespaceCharacters.some((ws) => regexp.test(ws));
 }
 
 export function escapeRegExp(value: string): string {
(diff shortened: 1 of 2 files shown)


packages/langium/src/utils/regexp-utils.ts
```diff
@@ -140,7 +140,7 @@ export function isMultilineComment(regexp: RegExp | string): boolean {
 
 export function isWhitespace(value: RegExp | string): boolean {
     const regexp = typeof value === 'string' ? new RegExp(value) : value;
-    return regexp.test(' ');
+    return [' ', '\n', '\t'].some((ws) => regexp.test(ws));
```
msujew, line 143: Suggestion: There are a few more (but more rarely used) line feed characters that should probably be included in there
```suggestion
    return [' ', '\n', '\t', '\v', '\r'].some((ws) => regexp.test(ws));
```

msujew (commented):
Thanks, that looks reasonable, just a minor comment, see below.

packages/langium/src/utils/regexp-utils.ts
aabounegm: I considered adding them but thought they are most likely not used by anyone. I am not against including those, but then again `\s` also accepts other characters, such as form feed (`\f`), non-breaking space (`U+00A0`), and a few others. Based on what do we decide which to include here 😅?

> For reference, [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Character_classes) specifies that `\s` is equivalent to `[\f\n\r\t\v\u0020\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]`

msujew: It's unlikely that people use them, but I think it still makes sense to include all of them. Using the `split('')` approach sounds reasonable. Can you write a comment explaining that that this is supposed to be `\s` with a link to MDN?

aabounegm: Sure thing

msujew (approved):
Looks good to me, thanks 👍

PR #1590, opened 2024-07-19
Fix language server in standalone mode

borisbrodski:
Running language server without VS Code and without build configuration resulted in an exception. This is fixed now.

ECL signed.


msujew (approved):
Thanks for the contribution @borisbrodski,

I've adjusted the method to ensure that we actually get the expected type from the `getConfiguration` call. The way we previously did it was pretty unsafe anyway.

PR #1600, opened 2024-07-25
Add support for overriding semantic token types

### LINKED ISSUE #1515: Add the possibility of easily define custom Semantic Tokens
I have a proposal for better define custom semantic tokens, as this was a feature that I needed for my language

# Current Behavior
When using SemanticTokenAcceptor, I can only use the current defined semantic tokens. This happens because the `highlightToken` function maps the string token id to a int, and without the newer token in the `SemanticTokenTypes` list, it does not work.

# Desired Behavior
As a User, I want to be able to easily define custom Semantic Tokens for my language using Langium methods.

# Background
I am using Langium since its 0.1 and have a language project that was developed during a scientific initiation and my graduation project in Computer Science at University. Now, I am doing a M.Sc in Computer Science and I'm still developing my language called Tonto. I recently got my first paper accepted showcasing it! if you want to take a look, here is the [website](https://matheuslenke.github.io/tonto-docs/) and here is the [github project](https://github.com/matheuslenke/Tonto).

The use case that I needed was that the language needed to use different colors for different ontological categories, making it easier to visualize in a model:
image

I can take care of this issue. Just reporting here to ask if anyone have any suggestions on how to implement it and if including this feature is approved.

aabounegm:
Closes #1515 

Example usage:
```ts
export class MySemanticTokenProvider extends AbstractSemanticTokenProvider {
    override get tokenTypes() {
        return {
            ...super.tokenTypes,
            myTokenType: 23,
        };
    }
    // ...
}
```

To be honest, I don't like the manual numbering and would instead prefer to just use an array of strings (and maybe something like `Object.fromEntries(tokenTypes.map((tokenType, idx) => [tokenType, idx]))` to reconstruct the object if needed), but I tried to stick with the current interface as much as possible. Is there a use case in which manually assigning the indexes is preferred?

### DIFF
--- a/packages/langium/src/lsp/language-server.ts
+++ b/packages/langium/src/lsp/language-server.ts
@@ -37,7 +37,7 @@ import type { ConfigurationInitializedParams } from '../workspace/configuration.
 import { DocumentState, type LangiumDocument } from '../workspace/documents.js';
 import { mergeCompletionProviderOptions } from './completion/completion-provider.js';
 import type { LangiumSharedServices, PartialLangiumLSPServices } from './lsp-services.js';
-import { mergeSemanticProviderOptions } from './semantic-token-provider.js';
+import { mergeSemanticTokenProviderOptions } from './semantic-token-provider.js';
 import { mergeSignatureHelpOptions } from './signature-help-provider.js';
 
 export interface LanguageServer {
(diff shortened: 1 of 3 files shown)


msujew (commented):
I went with manual assignments in the original implementation, since I didn't expect adopters to actually add new semantic token types (the LSP documentation indicates that this is generally pretty difficult - however this has proven to be inaccurate).

I think we can change the manual assignments to automatic ones. Otherwise, using multiple languages with different semantic token types will quickly become very difficult.

aabounegm:
I hope the latest commit is somewhat in line with what you had in mind

msujew:
@aabounegm thanks for getting back on this. However, I'm not sure this will work well: Let's say we have two languages with different semantic token providers, using different token types. For simplicity, let's say we have language A with `[SemanticTokenTypes.class]` and language B with `[SemanticTokenTypes.variable]`. After merging those types, the language server will respond that it supports `[SemanticTokenTypes.class, SemanticTokenTypes.variable]` at indices `[0, 1]`.

Now for the problem. Internally, each token provider will internally do `this.tokenTypes.indexOf(type);`, which **always yields** `0` for both `class` and `variable`. Therefore, all tokens will be highlighted as `class`. We need some way to feed the merged information back into the semantic token providers to resolve this.

aabounegm:
Hmm, that's a good point. Do you have any suggestions on how to best do that? If not, maybe I can just revert the last commit and we can think about it in a separate issue because it already feels out of scope of the original issue

msujew:
@aabounegm Not right now, no. I think it's fine to revert the last commit, yes :)

msujew (approved):
I've adjusted the options merging algorithm a bit - it should now work as expected (and will throw errors if it cannot merge correctly).

The rest looks good to me 👍

PR #1604, opened 2024-07-27
Alignment of Default Class Implementations to Interface definitions

### LINKED ISSUE #1584: Default class implementations do not fully implement interfaces
Langium version: 3.1.x
Package name: Langium

## Steps To Reproduce

1. See issue #1579 

Link to code example:
```typescript
export interface DocumentSymbolProvider {
    ...
    getSymbols(document: LangiumDocument, params: DocumentSymbolParams, cancelToken?: CancellationToken): MaybePromise<DocumentSymbol[]>;
}


export class DefaultDocumentSymbolProvider implements DocumentSymbolProvider {
    ...
    getSymbols(document: LangiumDocument): MaybePromise<DocumentSymbol[]> {
        return this.getSymbol(document, document.parseResult.value);
    }
    ...
}
```

## The current behavior
Some `DefaultXYZProvider` classes do not fully implement their corresponding interface. Due to TypeScript's structural typing, this is allowable but may prevent proper subclassing and/or overriding of default methods.

## The expected behavior
Where appropriate, `DefaultXYZ` classes should fully implement their corresponding interfaces and/or provide documentation regarding the implementation.

## Notes
image

snarkipus:
Resolves #1584 

### Summary:
- Added cancellation token support to various language server protocol (LSP) methods, allowing operations to be interrupted.
- Updated method signatures across multiple LSP providers (e.g. completion, definition, document highlight) with updated documentation and optional parameters.
- Implemented default no-op implementations for several DocumentUpdateHandler methods to reduce boilerplate in subclasses.
- Refactored several methods to use optional parameters with default values instead of explicitly passing default values, improving code readability.

### DIFF
--- a/packages/langium/src/lsp/document-update-handler.ts
+++ b/packages/langium/src/lsp/document-update-handler.ts
@@ -40,20 +40,20 @@ export interface DocumentUpdateHandler {
     willSaveDocument?(event: TextDocumentWillSaveEvent<TextDocument>): void;
 
     /**
-     * A docuemnt save event (initiated) was triggered by the `TextDocuments` service.
+     * A document save event (initiated) was triggered by the `TextDocuments` service.
      * @param event The document change event.
      * @returns An array of text edits which will be applied to the document before it is saved.
      */
     willSaveDocumentWaitUntil?(event: TextDocumentWillSaveEvent<TextDocument>): MaybePromise<TextEdit[]>;
 
     /**
-     * A docuemnt save event (completed) was triggered by the `TextDocuments` service.
+     * A document save event (completed) was triggered by the `TextDocuments` service.
      * @param event The document change event.
      */
     didSaveDocument?(event: TextDocumentChangeEvent<TextDocument>): void;
 
     /**
-     * A docuemnt close event was triggered by the `TextDocuments` service.
+     * A document close event was triggered by the `TextDocuments` service.
      * @param event The document change event.
      */
     didCloseDocument?(event: TextDocumentChangeEvent<TextDocument>): void;
--- a/packages/langium/src/lsp/document-update-handler.ts
+++ b/packages/langium/src/lsp/document-update-handler.ts
@@ -141,26 +141,4 @@ export class DefaultDocumentUpdateHandler implements DocumentUpdateHandler {
             .toArray();
         this.fireDocumentUpdate(changedUris, deletedUris);
     }
-
-    didOpenDocument(_event: TextDocumentChangeEvent<TextDocument>): void {
-        // To be implemented by subclasses if needed
-    }
-
-    willSaveDocument(_event: TextDocumentWillSaveEvent<TextDocument>): void {
-        // To be implemented by subclasses if needed
-    }
-
-    willSaveWaitUntilDocument(_event: TextDocumentWillSaveEvent<TextDocument>): MaybePromise<TextEdit[]> {
-        // To be implemented by subclasses if needed
-        return [];
-    }
-
-    didSaveDocument(_event: TextDocumentChangeEvent<TextDocument>): void {
-        // To be implemented by subclasses if needed
-    }
-
-    didCloseDocument(_event: TextDocumentChangeEvent<TextDocument>): void {
-        // To be implemented by subclasses if needed
-    }
-
 }
--- a/packages/langium/src/lsp/document-update-handler.ts
+++ b/packages/langium/src/lsp/document-update-handler.ts
@@ -21,23 +21,46 @@ import type { MaybePromise } from '../utils/promise-utils.js';
  */
 export interface DocumentUpdateHandler {
 
+    /**
+     * A document open event was triggered by the `TextDocuments` service.
+     * @param event The document change event.
+     */
     didOpenDocument?(event: TextDocumentChangeEvent<TextDocument>): void;
 
     /**
      * A content change event was triggered by the `TextDocuments` service.
+     * @param event The document change event.
      */
     didChangeContent?(event: TextDocumentChangeEvent<TextDocument>): void;
 
+    /**
+     * A document save event (initiated) was triggered by the `TextDocuments` service.
+     * @param event The document change event.
+     */
     willSaveDocument?(event: TextDocumentWillSaveEvent<TextDocument>): void;
 
+    /**
+     * A docuemnt save event (initiated) was triggered by the `TextDocuments` service.
+     * @param event The document change event.
+     * @returns An array of text edits which will be applied to the document before it is saved.
+     */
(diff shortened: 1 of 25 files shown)


packages/langium/src/lsp/document-update-handler.ts
```diff
@@ -119,4 +142,25 @@ export class DefaultDocumentUpdateHandler implements DocumentUpdateHandler {
         this.fireDocumentUpdate(changedUris, deletedUris);
     }
 
+    didOpenDocument(_event: TextDocumentChangeEvent<TextDocument>): void {
```
spoenemann, line 145: The interface methods are optional, so there's no need to provide empty implementations here.

packages/langium/src/lsp/workspace-symbol-provider.ts
```diff
@@ -46,11 +54,12 @@ export class DefaultWorkspaceSymbolProvider implements WorkspaceSymbolProvider {
         this.fuzzyMatcher = services.lsp.FuzzyMatcher;
     }
 
-    async getSymbols(params: WorkspaceSymbolParams, cancelToken = CancellationToken.None): Promise<WorkspaceSymbol[]> {
+    async getSymbols(params: WorkspaceSymbolParams, cancelToken?: CancellationToken): Promise<WorkspaceSymbol[]> {
+        const token = cancelToken ?? CancellationToken.None;
```
spoenemann, line 58: See also https://github.com/eclipse-langium/langium/pull/1605#discussion_r1696957672. Maybe it's a matter of taste, but I don't see this as an improvement w.r.t. readability.

My argument for keeping the original syntax: By specifying the default value as part of the method signature, it's like a part of the API and thus more transparent for those using it.

packages/langium/src/lsp/workspace-symbol-provider.ts
```diff
@@ -76,4 +85,9 @@ export class DefaultWorkspaceSymbolProvider implements WorkspaceSymbolProvider {
             return undefined;
         }
     }
+
+    async resolveSymbol(symbol: WorkspaceSymbol, _cancelToken?: CancellationToken): Promise<WorkspaceSymbol> {
```
spoenemann, line 89: The interface method is optional, so there's no need to provide a default value here.

packages/langium/src/workspace/configuration.ts
```diff
@@ -174,7 +174,11 @@ export class DefaultConfigurationProvider implements ConfigurationProvider {
         return `${languageId}`;
     }
 
-    get onConfigurationSectionUpdate(): Event<ConfigurationSectionUpdate> {
+    onConfigurationSectionUpdate(callback: ConfigurationSectionUpdateListener): Disposable {
+        return this.onConfigurationSectionUpdateEmitter.event(callback);
+    }
+
+    protected get onConfigurationSectionUpdateEvent(): Event<ConfigurationSectionUpdate> {
```
spoenemann, line 181: What's the reason for splitting this in two getters?

spoenemann (commented):
Thank you! Most of the changes look good to me.

However, I'm not convinced of these:

> - Implemented default no-op implementations for several DocumentUpdateHandler methods to reduce boilerplate in subclasses.
> - Refactored several methods to use optional parameters with default values instead of explicitly passing default values, improving code readability.

packages/langium/src/lsp/workspace-symbol-provider.ts
spoenemann: The modified solution also introduces an additional symbol in the scope so now there's `token` and `cancelToken`. This is a potential source of error, as the difference is easy to overlook. It can be mitigated by overriding the parameter value instead of adding another const variable. But even then, it's an additional line of code that needs to be understood, so I prefer the more concise form.

packages/langium/src/lsp/document-update-handler.ts
snarkipus: Rremoved.

packages/langium/src/lsp/workspace-symbol-provider.ts
snarkipus: Reverted default parameters.

packages/langium/src/lsp/workspace-symbol-provider.ts
snarkipus: Removed no-op method implementations.

packages/langium/src/workspace/configuration.ts
snarkipus: Semantics - reverted to original. 👍

packages/langium/src/lsp/document-update-handler.ts
snarkipus: Removed no-op implementations.

packages/langium/src/lsp/document-update-handler.ts
```diff
@@ -21,23 +21,46 @@ import type { MaybePromise } from '../utils/promise-utils.js';
...
 
+    /**
+     * A docuemnt save event (initiated) was triggered by the `TextDocuments` service.
+     * @param event The document change event.
+     * @returns An array of text edits which will be applied to the document before it is saved.
+     */
     willSaveDocumentWaitUntil?(event: TextDocumentWillSaveEvent<TextDocument>): MaybePromise<TextEdit[]>;
 
+    /**
+     * A docuemnt save event (completed) was triggered by the `TextDocuments` service.
```
msujew, line 50: Typo:
```suggestion
     * A document save event (completed) was triggered by the `TextDocuments` service.
```

packages/langium/src/lsp/document-update-handler.ts
```diff
@@ -21,23 +21,46 @@ import type { MaybePromise } from '../utils/promise-utils.js';
...
+     * @param event The document change event.
+     */
     didSaveDocument?(event: TextDocumentChangeEvent<TextDocument>): void;
 
+    /**
+     * A docuemnt close event was triggered by the `TextDocuments` service.
```
msujew, line 56: Typo:
```suggestion
     * A document close event was triggered by the `TextDocuments` service.
```

packages/langium/src/workspace/configuration.ts
```diff
@@ -174,7 +174,7 @@ export class DefaultConfigurationProvider implements ConfigurationProvider {
         return `${languageId}`;
     }
 
-    get onConfigurationSectionUpdate(): Event<ConfigurationSectionUpdate> {
+    get onConfigurationSectionUpdate():  Event<ConfigurationSectionUpdate> {
```
msujew, line 177: Formatting:
```suggestion
    get onConfigurationSectionUpdate(): Event<ConfigurationSectionUpdate> {
```

msujew (approved):
Looks good to me, thanks 👍

PR #1605, opened 2024-07-27
optimize document sorting and update documentation

snarkipus:
- Optimized the `sortDocuments` method in `DefaultDocumentBuilder` class, replacing the previous sort algorithm with a more efficient in-place partitioning approach (non-stable) to separate documents with and without associated text documents
- Updated documentation

### DIFF
--- a/packages/langium/src/references/scope-computation.ts
+++ b/packages/langium/src/references/scope-computation.ts
@@ -73,8 +73,9 @@ export class DefaultScopeComputation implements ScopeComputation {
         this.descriptions = services.workspace.AstNodeDescriptionProvider;
     }
 
-    async computeExports(document: LangiumDocument, cancelToken = CancellationToken.None): Promise<AstNodeDescription[]> {
-        return this.computeExportsForNode(document.parseResult.value, document, undefined, cancelToken);
+    async computeExports(document: LangiumDocument, cancelToken?: CancellationToken): Promise<AstNodeDescription[]> {
+        const token = cancelToken ?? CancellationToken.None;
+        return this.computeExportsForNode(document.parseResult.value, document, undefined, token);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


packages/langium/src/references/scope-computation.ts
```diff
@@ -73,8 +73,9 @@ export class DefaultScopeComputation implements ScopeComputation {
         this.descriptions = services.workspace.AstNodeDescriptionProvider;
     }
 
-    async computeExports(document: LangiumDocument, cancelToken = CancellationToken.None): Promise<AstNodeDescription[]> {
-        return this.computeExportsForNode(document.parseResult.value, document, undefined, cancelToken);
+    async computeExports(document: LangiumDocument, cancelToken?: CancellationToken): Promise<AstNodeDescription[]> {
+        const token = cancelToken ?? CancellationToken.None;
```
spoenemann, line 77: What's the issue with the default parameter value syntax?

packages/langium/src/workspace/document-builder.ts
```diff
@@ -251,21 +251,28 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
-        });
+
+            while (right >= 0 && !this.hasTextDocument(documents[right])) {
+                right--;
+            }
+
+            if (left < right) {
+                [documents[left], documents[right]] = [documents[right], documents[left]];
+            }
+        }
```
spoenemann, line 269: Could you provide a unit test for this algorithm by mocking the `textDocuments` service?

spoenemann (commented):
Thanks! The change looks good, just two remarks below.

packages/langium/src/references/scope-computation.ts
snarkipus: No issue per se - I think I just had interface alignment in mind after the previous exercise (#1604 ).

packages/langium/src/workspace/document-builder.ts
snarkipus: Sure - is a naive method extraction acceptable? If not, it will take me a minute to figure out how to wield the dark arts of dependency injection to conjure up a modified `createServicesForGrammar()` function (if that's even feasible).

spoenemann: IMO it's ok to override TypeScript's `protected` method visibility by casting the service to `any` in the test.

packages/langium/src/references/scope-computation.ts
spoenemann: See my comment https://github.com/eclipse-langium/langium/pull/1604#discussion_r1696977473 – I prefer the previous parameter initialization syntax over this modified variant.

snarkipus: Reverted the change. 👍

packages/langium/src/workspace/document-builder.ts
snarkipus: Thanks for the hint - test code isn't perfect, but I think it shows that it works.

spoenemann (approved):
👍

PR #1614, opened 2024-07-30
Extend the API of the validation registry for more performant custom validations

JohannesMeierSE:
When writing custom validations, often it is required to stream the (whole or parts of the) AST inside the registered checks, e.g. for checking unique names or for identifying circles/loops in relationships between the AST nodes (calls, inheritance). For huge ASTs, this approach has the drawback, that the AST is traversed multiple times, which decreases performance.

This PR proposes to register additional "checks" which are executed once before or after validating an AST/Langium document. The idea is to exploit the AST traversal already done by the DocumentBuilder and to remember only relevant information during AstNode-specific checks. The collected information is evaluated once after all checks are done.
The expected benefit are faster validations by less AST traversals inside custom checks for big ASTs.

In order to demonstrate the new API, I wrote simple test cases to check unique fully-qualified names and implemented the validation twice, once with the existing API (and with `streamAllContents`) and once with the new API (and without `streamAllContents`).

Aspects to consider during the review:
- Did I found the best names for the new API?
- Drawback is that the validations are not state-less anymore.
- It is no breaking change, since the existing checks are unchanged and only some additional checks can be registered.

### DIFF
--- a/packages/langium/src/validation/validation-registry.ts
+++ b/packages/langium/src/validation/validation-registry.ts
@@ -5,15 +5,16 @@
  ******************************************************************************/
 
 import type { CodeDescription, DiagnosticRelatedInformation, DiagnosticTag, integer, Range } from 'vscode-languageserver-types';
-import type { CancellationToken } from '../utils/cancellation.js';
+import { assertUnreachable } from '../index.js';
 import type { LangiumCoreServices } from '../services.js';
 import type { AstNode, AstReflection, Properties } from '../syntax-tree.js';
-import type { MaybePromise } from '../utils/promise-utils.js';
-import type { Stream } from '../utils/stream.js';
-import type { DocumentSegment } from '../workspace/documents.js';
+import type { CancellationToken } from '../utils/cancellation.js';
 import { MultiMap } from '../utils/collections.js';
+import type { MaybePromise } from '../utils/promise-utils.js';
 import { isOperationCancelled } from '../utils/promise-utils.js';
+import type { Stream } from '../utils/stream.js';
 import { stream } from '../utils/stream.js';
+import type { DocumentSegment } from '../workspace/documents.js';
 
 export type DiagnosticInfo<N extends AstNode, P extends string = Properties<N>> = {
     /** The AST node to which the diagnostic is attached. */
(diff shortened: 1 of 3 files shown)


packages/langium/src/validation/validation-registry.ts
```diff
@@ -181,4 +205,45 @@ export class ValidationRegistry {
...
+        };
+    }
+
+    getChecksBefore(): ValidationPreparation[] {
+        return this.entriesBefore;
+    }
+
+    getChecksAfter(): ValidationPreparation[] {
+        return this.entriesAfter;
+    }
```
spoenemann, line 248: I'd declare the properties as `public readonly` rather than using such getter methods.

packages/langium/test/validation/document-validator.test.ts
```diff
@@ -110,3 +111,142 @@ describe('Generic `AstNode` validation applies to all nodes', () => {
...
+            list = [];
+            fqnMap.set(fqn, list);
+        }
+        list.push(child);
+    }
+
+    function registerValidationForRootElement() {
+        const validationChecks: ValidationChecks<object> = {
+            // Note that there are no generated types like Model and NamedElement here
+            AstNode: (node, accept) => {
```
spoenemann, line 156: Even though there are no generated types for the in-memory test language, it's possible to write the corresponding types manually in order to improve the readability and type safety of the tests.

spoenemann (commented):
Thanks, the changes look good to me! Just two remarks below.

packages/langium/test/validation/document-validator.test.ts
JohannesMeierSE: Done, it improves the readability a lot, thanks!

packages/langium/src/validation/validation-registry.ts
JohannesMeierSE: The problem with `public` properties is, that users could register checks directly in this property and without calling `registerBeforeDocument(...)`, which bypasses our internal handling of exceptions.
Therefore I decided to keep the "simple getters".

JohannesMeierSE:
Thanks for the discussions in the Langium dev meeting and @spoenemann for your review! I just pushed the improvements.

packages/langium/src/validation/validation-registry.ts
spoenemann: The getter does not change that: someone could still push elements to the returned array. If you really want to prevent that, you'd need to clone the array. But an easier and faster solution is just to document the property/method saying you shouldn't manipulate the array.

If you want to keep a getter, I suggest using a `get` property:
```typescript
    get checksAfter(): ValidationPreparation[] {
        return this.entriesAfter;
    }
```

spoenemann (approved):
Looks good, thanks. See comment thread above about code style of getter methods.

JohannesMeierSE:
@spoenemann I pushed the fix and rebased to solve the conflicts with `main`.

JohannesMeierSE:
FYI: We need this feature in Typir. Therefore we are looking forward to a Langium release with this feature!

PR #1617, opened 2024-08-05
Use languageId as key in prism generator

cdietrich:
changed to `languageId` to harmonize with textmate/vscode behaviour

````
```language-id
Hello World!
```
````


msujew:
@cdietrich Running `npm run langium:generate` in the root folder will run the generator for the langium language and all example languages.

cdietrich:
thx it works ;)

msujew (approved):

PR #1520, opened 2024-05-29
generator-langium: Transform generated project to npm workspace

kaisalmen:
TODO:
- [x] Align template structure with target structure
- [x] Integrate a global tsconfig.build.json in target project
- [x] Unify how build and compile is done in target project
- [x] Update to the latest version of the wrapper lib stack
- [x] Fix Build issues: Create index files and export code to other packages
- [x] Fix regular `generator-langium` unit tests
- [x] Add new test that check the complete structure
- [x] Option test not yet included even if selected
- [x] `tsconfig.build.json` needs to be dynamically created. Sources not contained should not be included as it leads to build failures otherwise.
- [x] Align dependencies and update vitest
- [x] Remove the need to use `http-server`

Fixes #1495 

TODO: from review:
- [x] Update langium-quickstart.md
- [x] Add clean script and `build:clean` to all packages

Issue #999, opened 2023-03-27
IntelliJ example


ctron:
On the homepage it is mentioned that Langium can be used for IntelliJ editors. However, I wasn't able to find any information on how this could be achieved.

I think it makes sense to provide an example project, or at least outline the steps necessary.

ben5311:
> On the homepage it is mentioned that Langium can be used for IntelliJ editors. However, I wasn't able to find any information on how this could be achieved.
> 
> I think it makes sense to provide an example project, or at least outline the steps necessary.

I have tried both the IntelliJ Marketplace Plugin https://plugins.jetbrains.com/plugin/10209-lsp-support and the IntelliJ Plugin dependency https://github.com/ballerina-platform/lsp4intellij.

The Marketplace Plugin is actually working: Validations, Quick Fixes, Formatting, etc.
But it crashes randomly and is not actively developed anymore.

With the lsp4intellij solution you need to setup you own IntelliJ extension project and integrate the Langium LSP server into it. It did not work for me, even though the project seems to be still alive. 
It does connect to the Langium LSP Server and I can even see that it receives notfications from it (e.g. on Hover) but it does not display it in the IntelliJ UI.

For me the IntelliJ LSP integration looks not really stable at the moment.
Also note that Syntax Highlighting usually does not work via LSP. Meaning you need to add Syntax Highlighting manually for IntelliJ / Eclipse / Other LSP editor.

nadavhames:
Instead of IDEA, Jetbrains is coming out with Fleet soon which uses LSP, so support should be much easier for that product once it supports editor plugins.

angelozerr:
Please try https://github.com/redhat-developer/lsp4ij which is a free lsp support for intellij

Yokozuna59:
Fellows at [zenstack](https://github.com/zenstackhq/zenstack) were able to create the JetBrains extension:

- Original PR where the extension was created: https://github.com/zenstackhq/zenstack/pull/904
- The latest commit for the extension: https://github.com/zenstackhq/zenstack/tree/29962e0b48a73ae6d42f43f2575048ba9cf6a953/packages/ide/jetbrains

spoenemann:
In the meantime, JetBrains has started offering LSP support for their paid IDEs: https://blog.jetbrains.com/platform/2023/07/lsp-for-plugin-developers/

As I understand this, the solution is not available for the open source IntelliJ Platform.

spoenemann:
[...]

PR #1618, opened 2024-08-06
Intellij template

aabounegm:
This PR adds the option to generate a plugin for IntelliJ-based IDEs (including WebStorm, Android Studio, and many others listed [here](https://plugins.jetbrains.com/docs/intellij/intellij-platform.html#ides-based-on-the-intellij-platform)). It is largely based on [JetBrains/intellij-platform-plugin-template](https://github.com/JetBrains/intellij-platform-plugin-template).
It depends on the [LSP4IJ](http://github.com/redhat-developer/lsp4ij) community plugin by RedHat to add LSP support to non-paid versions of the IDEs.

While LSP4IJ currently supports IntelliJ starting with version 2023.2, the plugin generated by this template requires at least 2024.1 for TextMate highlighting support. Ideally, we should instead generate IntelliJ's own [syntax highlighting factory](https://plugins.jetbrains.com/docs/intellij/syntax-highlighter-and-color-settings-page.html), but I don't think I will have the time to do that anytime soon since TextMate fits my needs (I already had a comprehensive TextMate grammar defined long ago and wanted to reuse it).

I should also note that I am totally not an expert in Java/Kotlin, so the code there may not be the best Kotlin code, but it does the job and I was guided by IntelliJ IDEA 😁.


packages/generator-langium/src/index.ts
```diff
@@ -397,19 +438,23 @@ export * from './generated/module.js';
         return this.destinationPath(USER_DIR, this.answers.extensionName, ...path);
     }
 
-    _replaceTemplateWords(fileExtensionGlob: string, languageName: string, languageId: string, content: string | Buffer): string {
+    _replaceTemplateWords(fileExtensionGlob: string, fileExtensionDefault: string, fileExtensionsSemi: string, languageName: string, languageId: string, content: string | Buffer): string {
```
aabounegm, line 441: I had a strong urge to replace this with a single object parameter to look better (like named parameters), but I didn't want to modify existing code as much as possible.

packages/generator-langium/src/index.ts
```diff
@@ -397,19 +438,23 @@ export * from './generated/module.js';
...
+            .replace(FILE_EXTENSION_DEFAULT, fileExtensionDefault)
+            .replace(FILE_EXTENSIONS_SEMI, fileExtensionsSemi)
             .replace(LANGUAGE_NAME, languageName)
             .replace(LANGUAGE_ID, languageId)
             .replace(NEWLINES, EOL);
     }
 
-    _replaceTemplateNames(languageId: string, path: string): string {
-        return path.replace(LANGUAGE_PATH_ID, languageId);
+    _replaceTemplateNames(languageId: string, languageName: string, path: string): string {
```
aabounegm, line 454: Same thing here

packages/generator-langium/templates/packages/intellij-plugin/build.gradle.kts
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+    }
+}
+
+tasks {
+    prepareSandbox {
+        // Inspired by Prettier plugin (https://github.com/JetBrains/intellij-plugins/blob/master/prettierJS/build.gradle.kts)
+        from("../extension/out/language") {
+            include("main.cjs")
```
aabounegm, line 133: This here is the main reason this package depends on the VS Code extension. If this dependence is undesirable, we can add a minimal JS package to this (`intellij-plugin`) subpackage instead.
For my purposes, I need a VS Code extension anyway, and I assumed it would be rare for someone to need only IntelliJ, so I disabled the IntelliJ prompt if VS Code wasn't included.

packages/generator-langium/templates/packages/intellij-plugin/build.gradle.kts
```diff
@@ -0,0 +1,153 @@
...
+            into("${intellijPlatform.pluginConfiguration.name.get()}/<%= LanguageName %>LanguageService")
+        }
+
+        // Copy TextMate files from VS Code extension
+        from("../extension") {
+            include("package.json")
+            include("syntaxes/*.tmLanguage.json")
+            include("language-configuration.json")
```
aabounegm, line 141: This is the other reason it depends on the VS Code extension. Technically, it should be easy to copy the TM syntax from `language` directly, but I'm not sure about `language-configuration.json` (actually, I'm not even sure if IntelliJ needs/understands it, I forgot to test for that 😅)

packages/generator-langium/src/index.ts
```diff
@@ -135,6 +140,17 @@ export class LangiumGenerator extends Generator {
...
+            {
+                type: 'confirm',
+                name: 'includeIntelliJ',
+                // Currently, the IntelliJ plugin depends on files generated for the VS Code extension.
+                when: (answers: Answers): boolean => answers.includeVSCode,
+                prefix: description(
+                    'Your language can be run inside of a JetBrains IntelliJ-based IDE plugin.'
+                ),
+                message: 'Include IntelliJ integration?',
+                default: 'yes',
```
aabounegm, line 152: Should I keep this default as `yes` or does it make more sense to not include it by default?

aabounegm (commented):
Just some comments and questions

aabounegm:
Hi @msujew, I understand your point, especially about the maintenance, but after spending several days trying to get everything right for an IntelliJ plugin, I thought it would be helpful to others if I add it to the template, especially that the documentation about it is not so great and I had to search through multiple forums and deal with some bugs that had no logs and so on, so I just wanted to share it to save other people's time (especially those not proficient in Java/Kotlin, like me).
About the Langium-specific part being minimal, isn't that also the case with the `web` package, for example?

We could technically make it into a guide, but it will mostly consist of "add this file with that content" blocks, so I don't see why maintaining the guide would be any less effort than maintaining the template (which should hopefully not require that much maintenance anyway).

kaisalmen:
@aabounegm and @msujew maybe after #1520 is merged, we could think about making the generator extendable. If the npm package exposes callable functions/extendable methods to process an additional package, then any further code extending the generator must not be contained here. Then even for the web package could be moved to an external location. Let me think about that, but only after #1520 is merged (it is already too big).

aabounegm:
Yes, I definitely agree that this one should only be considered after the main one is merged, I just wanted to get it out of the way before I forget the context of how I made the plugin 😁.
Perhaps it would be better if I convert it to a draft PR for now.

spoenemann:
> Hey @aabounegm, thanks for looking into this. We've had discussions around this in the past (see https://github.com/eclipse-langium/langium/issues/999 and https://github.com/eclipse-langium/langium-website/issues/243) and I believe this is better served as a guide on our website rather than included in the yeoman generator.
> I'm kind of torn on this issue right now. While this increases our reach beyond the VS Code ecosystem in an easy-to-use way, it's also quite a bit of maintenance effort and we don't have anyone (in the active Langium committers) working on IntelliJ based plugins right now at all. Since the langium-specific part of this is rather minimal (most of this PR is IntelliJ boilerplate), I would rather expect a yeoman generator or similar from the RedHat people for this, instead of us offering something.
> As mentioned, I'd rather see this as a page on our website, mostly just linking to the [developer guide for LSP4IJ](https://github.com/redhat-developer/lsp4ij/blob/main/docs/DeveloperGuide.md) with some langium-specific additions on top. @spoenemann @kaisalmen your thoughts on this?

I agree. The increased maintenance effort would become a problem with time.

What's the normal way to create projects in IntelliJ? Eclipse has a concept of "Project Wizards" that do the same as Yeoman, but with a GUI.

aabounegm:
> I agree. The increased maintenance effort would become a problem with time.

Alright, if you prefer, I can simply make a template repo on my personal account so it doesn't become a maintenance burden on you, because I believe a template would be much more useful than a guide. Perhaps a dedicated README covering what the template does (and how it changed from the original IntelliJ plugin template) could serve as a guide for how to keep it up to date as well.

> What's the normal way to create projects in IntelliJ?

I'm not sure, to be frank. I'm not an IntelliJ user myself, but some of the users of the language I'm working on are. This particular project was based on [JetBrains/intellij-platform-plugin-template](https://github.com/JetBrains/intellij-platform-plugin-template).

aabounegm:
I finally made a template repo with the contents of this PR so I can finally close it: [aabounegm/langium-intellij-template](https://github.com/aabounegm/langium-intellij-template)

spoenemann:
@aabounegm that repo is archived, is that on purpose?

aabounegm:
@spoenemann yeah, it seemed to me that the base PR (#1520) kinda stalled, and I don't think I intend to maintain the IntelliJ any further, so I archived to avoid annoying dependabot alerts. If there is a good enough reason (at least some demand), I can unarchive.

Issue #1362, opened 2024-01-29
Formatter shows cyclic behaviour


insafuhrmann:
I am implementing a code formatter for a language that has a number of constructs with code regions enclosed in keywords, in a nested fashion, for example a minimal program could be something like:
```
BEGIN_NAMESPACE
    myNamespace
    BEGIN_STRUCT
        myStruct
    END_STRUCT
END_NAMESPACE
```
I would like to have a free line before the opening BEGIN_STRUCT keyword and if there already is one or more than one, I wish to keep those lines, but when I implement it according to the documentation (at least to my understanding of it) and apply the auto format multiple times, the formatting changes back and forth, adding and removing the desired line.

Langium version: current development (3.0 candidate)

## Steps To Reproduce

For example just add this to the domain-model.langium grammar:
```
Structure:
    'BEGIN_STRUCT' 
    name=QualifiedName
    'END_STRUCT';

NameSpace:
    'BEGIN_NAMESPACE'
    name=QualifiedName
    (elements+=AbstractElement)*
    'END_NAMESPACE';
```

and register the two as AbstractElement:
```
AbstractElement:
    PackageDeclaration | Type | Structure | NameSpace;
```
then add this to the format function of the DomainModelFormatter:

```
} else if (ast.isStructure(node)) {
            const formatter = this.getNodeFormatter(node);
            const structOpen = formatter.keyword('BEGIN_STRUCT');
            const structClose = formatter.keyword('END_STRUCT');
            formatter.interior(structOpen, structClose).prepend(Formatting.indent());
            structOpen.prepend(Formatting.newLines(2, {allowMore: true}));
            structClose.prepend(Formatting.newLine());
  } else if (ast.isNameSpace(node)) {
            const formatter = this.getNodeFormatter(node);
            const nsOpen = formatter.keyword('BEGIN_NAMESPACE');
            const nsClose = formatter.keyword('END_NAMESPACE');
            formatter.interior(nsOpen, nsClose).prepend(Formatting.indent());
            nsClose.prepend(Formatting.newLine());
  }
```
Then apply `Format Document` repeatedly to a document with the minimal program I wrote above.
  
## The current behavior
The line before BEGIN_STRUCT alternatingly appears and is removed. More than one line is initially removed, after that same behaviour: alternate between empty line and no empty line before BEGIN_STRUCT.

## The expected behavior
[...]

PR #1628, opened 2024-08-13
Fix formatter

ydaveluy:
Hello,

This PR addresses several issues in the current formatter implementation:

- Return only the necessary edits: This ensures that only the edits which actually change the document are returned, rather than all possible edits.

- Fix `avoidOverlappingEdits`: This change improves the removal of overlapping edits, ensuring they are handled correctly across multiple levels.

- Prevent formatting on the same line: This update prevents formatting actions from being applied to the same line if the previous node is hidden.

These improvements aim to enhance the accuracy and efficiency of the formatter. Please let me know if you have any questions or need further clarifications.

### DIFF
--- a/packages/langium/src/lsp/formatter.ts
+++ b/packages/langium/src/lsp/formatter.ts
@@ -206,7 +206,7 @@ export abstract class AbstractFormatter implements Formatter {
 
 
     protected isNecessary(edit: TextEdit, document: TextDocument): boolean {
-        return edit.newText !== document.getText(edit.range);
+        return edit.newText !== document.getText(edit.range).replace(/\r/g, '');
     }
 
     protected iterateCstFormatting(document: LangiumDocument, formattings: Map<string, FormattingAction>, options: FormattingOptions, range?: Range): TextEdit[] {
--- a/packages/langium/src/lsp/formatter.ts
+++ b/packages/langium/src/lsp/formatter.ts
@@ -150,12 +150,16 @@ export abstract class AbstractFormatter implements Formatter {
     protected avoidOverlappingEdits(textDocument: TextDocument, textEdits: TextEdit[]): TextEdit[] {
         const edits: TextEdit[] = [];
         for (const edit of textEdits) {
-            const last = edits[edits.length - 1];
-            if (last) {
+            let last = edits[edits.length - 1];
+            while (last) {
                 const currentStart = textDocument.offsetAt(edit.range.start);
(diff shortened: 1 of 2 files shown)


cdietrich:
Can we double check if this won’t reintroduce https://github.com/eclipse-langium/langium/pull/1550/files

packages/langium/src/lsp/formatter.ts
```diff
@@ -200,11 +204,9 @@ export abstract class AbstractFormatter implements Formatter {
...
     }
 
-    /**
-     * @deprecated This method has been deprecated with 3.1. It now always returns `true` and is no longer used by the default formatter implementation.
-     */
-    protected isNecessary(_edit: TextEdit, _document: TextDocument): boolean {
-        return true;
+
+    protected isNecessary(edit: TextEdit, document: TextDocument): boolean {
+        return edit.newText !== document.getText(edit.range);
```
msujew, line 209: Suggestion: I've already noticed in https://github.com/eclipse-langium/langium/pull/1550 that this comparison yields faulty behavior on Windows. `document.getText()` will return line endings with `\r`, while `edit.newText` will not contain `\r`.
```suggestion
        return edit.newText !== document.getText(edit.range).replace(/\r/g, '');
```

msujew (commented):
Looks mostly good to me. Just one issue (on Windows systems, see below). Can you also please create a test case for that: You will need to manually invoke the `Formatter` service for that, and assert that no text edits are returned for a correctly formatted document (ideally with formattings that are doing newline or indentation).

msujew:
@cdietrich I can't imagine this reintroducing https://github.com/eclipse-langium/langium/issues/1362, but I also don't have a test case for this anymore. Do you still have one lying around and try it against this branch?

packages/langium/src/lsp/formatter.ts
ydaveluy: Test case added and fixed `\r` on Windows.

msujew:
@cdietrich You could add the test repo as a temporary part the `examples`. That way, npm should do the linking for you.

@ydaveluy There is a linting error, see https://github.com/eclipse-langium/langium/actions/runs/10387365742/job/28760687114?pr=1628

cdietrich:
@msujew thx. looks good. formatter stays idempotent

msujew (approved):
@cdietrich Thanks for checking!

Issue #1571, opened 2024-07-08
Formatter removes first option in alternation rules


samir-abis-everest:
Langium version: 3.1.1 (extension)
Package name: langium

## Steps To Reproduce

1. Create a .langium file with an alternation rule like:

```langium
NumberTypeSpecialization:
    {infer Int} 'Int' |
    {infer Float} 'Float' |
    {infer Decimal} 'Decimal';
```

2. Activate the format on Langium files

## The current behavior

The formatter removes the first option in alternation rules, resulting in:

```langium
NumberTypeSpecialization:
    
     |
    {infer Float} 'Float' |
    {infer Decimal} 'Decimal';
```

This occurs for any rule that starts with an alternation (|).

## The expected behavior

The formatter should preserve all options in alternation rules, maintaining the original structure:

```langium
NumberTypeSpecialization:
    {infer Int} 'Int' |
    {infer Float} 'Float' |
    {infer Decimal} 'Decimal';
```

ydaveluy:
Hello,

I found where is the issue in the formatter code: 
Sometime the between range has a start position after the end position that lead to incorrect edits :https://github.com/eclipse-langium/langium/blob/b2c40e3a8f06becfa7b18b04cfbef7b7931f164a/packages/langium/src/lsp/formatter.ts#L362

Returning an empty edit in such case fix the issue. 
```
// Ignore the edit if the previous node ends after the current node starts
if (a && (a.range.end.line > b.range.start.line ||
    (a.range.end.line === b.range.start.line && a.range.end.character > b.range.start.character))) {
    return [];
}
const betweenRange: Range = {
    start: a?.range.end ?? {
        character: 0,
        line: 0
    },
    end: b.range.start
};
```
@msujew I can provide a PR if you are interested. It can be included in #1628 or in a separate PR if you prefer.

msujew:
@ydaveluy Thanks for looking into this! A separate PR would be appreciated :)

ydaveluy:
ok see PR #1629

PR #1629, opened 2024-08-14
Take CST node boundaries into account for formatting

ydaveluy:
Closes https://github.com/eclipse-langium/langium/issues/1571

Do not perform formatting if the previous node ends after the current node starts.


ydaveluy:
This PR  fix issue  #1571

msujew (approved):
Looks good to me, thanks 👍

PR #1631, opened 2024-08-14
Fix false positive on regex flag validation

msujew:
Just a minor fix for a warning that would incorrectly appear when mix & matching non-regex terminal elements with regex terminal elements.

### DIFF
--- a/packages/langium/test/grammar/grammar-validator.test.ts
+++ b/packages/langium/test/grammar/grammar-validator.test.ts
@@ -200,6 +200,24 @@ describe('Langium grammar validation', () => {
         const validationResult = await validate(grammar);
         expectNoIssues(validationResult);
     });
+
+    test('Composite terminal regex flags', async () => {
+        const grammar = `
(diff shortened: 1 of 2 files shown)


packages/langium/test/grammar/grammar-validator.test.ts
```diff
@@ -200,6 +200,24 @@ describe('Langium grammar validation', () => {
         const validationResult = await validate(grammar);
         expectNoIssues(validationResult);
     });
+
```
dhuebner, line 203: It would be nice to have an additional test that test a false positive when no regex flags are used 
```suggestion
    test('Composite terminal no regex flags', async () => {
        const grammar = `
        terminal Test: 'Test' /test/;
        `;
        const validationResult = await validate(grammar);
        expectNoIssues(validationResult);
    });

```

dhuebner (approved):
Nice fix!
Please consider to add the suggested additional test case.

Issue #1633, opened 2024-08-15
Language Server Crash: 


ballcoach12:
When I add the following grammar rule, the Langium language server crashes in VSCode (also in the Langium Playground);

```
terminal D_BYTE_CHAR_STR returns string:
    '"' D_BYTE_CHAR_VALUE* '"';
```

Langium version: 3.1.1
Package name:

## Steps To Reproduce

1. Given this grammar:
```
grammar StructuredText

entry StructuredTextModel:
    (functionBlock=FunctionBlock | 
    (
        'BUILDIN'
        'FUNCTIONS' ':' buildin+=Intr_Function*
        'BUILDIN_END'
    ))
    ;

FunctionBlock:
    'FUNCTION_BLOCK' name=ID
      inputs=Inputs_List
      outputs=Outputs_List
      (localVariables=InternalVars_List)?
      ((statements+=Stmt)?';')*
    'END_FUNCTION_BLOCK'
    ;

Inputs_List:
    'VAR_INPUT'
    ((items+=Var_Decl_Init))*
    'END_VAR'
    ;

Outputs_List:
    'VAR_OUTPUT'
    ((items+=Output_Var_Decl_Init))+ 
    'END_VAR'
    ;

InternalVars_List:
    'VAR'
    ((items+=Local_Var_Decl_Init))*
    'END_VAR'
    ;

Var_Decl_Init:
    variables+=Variable_Primary (',' variables+=Variable_Primary)* ':'
    typeName=Native_Type_Name
    (':=' initialValue=Constant)? ';'
    ;

Output_Var_Decl_Init:
    variables+=Variable_Primary (',' variables+=Variable_Primary)* ':'
    typeName=Native_Type_Name 
    (':=' initialValue=Constant)? ';'
    ;

Local_Var_Decl_Init:
    (param ?='PARAM')?
    variables+=Variable_Primary (',' variables+=Variable_Primary)* ':'
    typeName=Native_Type_Name
    (':=' initialValue=Constant)? ';'
;

Variable_Primary infers PrimaryVariable:
    name=Variable_Name
    ;

Variable_Name returns string:
    ID | 'E' | 'e'
    ;

Native_Type_Name returns string: // enum in xtext, but not in langium
    'NUMERIC' | 'BOOL' | 'STRING' | 'ENUM' | 'ACSOURCE'
    ;

Constant:
    Numeric_Literal | String_Literal | Bool_Literal
    ;

Numeric_Literal infers NumericLiteral:
[...]

PR #1634, opened 2024-08-15
Make Langium language server more robust

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1633

Makes the type inference methods more robust against errors due to missing elements. In some cases, this could make some validations "disappear", but only in the presence of parser errors - this is alright, since the file doesn't compile anyway.


msujew:
@ballcoach12 Do you mind testing this PR against your grammar? You can run the `Run Grammar Extension` config to test that this works as expected.

dhuebner (approved):
Looks good! 
I've added a test case that was failing before and passing after this fix.

PR #1635, opened 2024-08-17
Make service containers readonly

msujew:
Fixes something that was encountered in https://github.com/eclipse-langium/langium/discussions/1621#discussioncomment-10362924.

Essentially, our API (and runtime as well - to some degree) allows to override the values of services simply by performing assignments. However, this almost certainly breaks runtime behavior in case the service that is being assigned is used as a service dependency somewhere.

This change adds readonly modifiers where necessary and adds a `set` method to the service proxies that ensure that no service can be assigned.

### DIFF
--- a/packages/langium/src/dependency-injection.ts
+++ b/packages/langium/src/dependency-injection.ts
@@ -74,6 +74,9 @@ export function eagerLoad<T>(item: T): T {
 function _inject<I, T>(module: Module<I, T>, injector?: any): T {
     const proxy: any = new Proxy({} as any, {
         deleteProperty: () => false,
+        set: () => {
+            throw new Error('Cannot set values on service container');
+        },
         get: (obj, prop) => _resolve(obj, prop, module, injector || proxy),
         getOwnPropertyDescriptor: (obj, prop) => (_resolve(obj, prop, module, injector || proxy), Object.getOwnPropertyDescriptor(obj, prop)), // used by for..in
         has: (_, prop) => prop in module, // used by ..in..
--- a/packages/langium/src/dependency-injection.ts
+++ b/packages/langium/src/dependency-injection.ts
@@ -75,14 +75,19 @@ function _inject<I, T>(module: Module<I, T>, injector?: any): T {
     const proxy: any = new Proxy({} as any, {
         deleteProperty: () => false,
         set: () => {
-            throw new Error('Cannot set values on service container');
+            throw new Error('Cannot set property on injected service container');
         },
-        get: (obj, prop) => _resolve(obj, prop, module, injector || proxy),
+        get: (obj, prop) => {
+            if (prop === isProxy) {
+                return true;
+            } else {
+                return _resolve(obj, prop, module, injector || proxy);
+            }
+        },
         getOwnPropertyDescriptor: (obj, prop) => (_resolve(obj, prop, module, injector || proxy), Object.getOwnPropertyDescriptor(obj, prop)), // used by for..in
         has: (_, prop) => prop in module, // used by ..in..
-        ownKeys: () => [...Reflect.ownKeys(module), isProxy] // used by for..in
+        ownKeys: () => [...Object.getOwnPropertyNames(module)] // used by for..in
     });
-    proxy[isProxy] = true;
     return proxy;
 }
 
--- a/packages/langium/src/lsp/default-lsp-module.ts
+++ b/packages/langium/src/lsp/default-lsp-module.ts
@@ -28,7 +28,7 @@ import { DefaultWorkspaceSymbolProvider } from './workspace-symbol-provider.js';
  * Context required for creating the default language-specific dependency injection module.
  */
 export interface DefaultModuleContext extends DefaultCoreModuleContext {
-    shared: LangiumSharedServices;
+    readonly shared: LangiumSharedServices;
 }
 
 /**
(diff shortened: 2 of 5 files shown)


dhuebner (approved):
It was not clear to me that it was possible. Or I just never had an idea to do this... 
Even thought it's a bad idea to override the services, it is as breaking change because of changing public API. We need a dedicated entry in the change log.

packages/langium/test/dependency-injection.test.ts
```diff
@@ -352,18 +352,9 @@ describe('The inject result', () => {
...
-        expect(obj.a).toBeUndefined();
-        expect(() => obj.a = 1).not.toThrow();
-        expect(obj.a).toBe(1);
-    });
-
-    test('should be sealable', () => {
-        const obj: any = Object.seal(inject({}));
-        expect(Object.isExtensible(obj)).toBe(false);
-        expect(() => (obj.a = 1)).toThrowError('Cannot define property a, object is not extensible');
+        expect(() => (obj.a = 1)).toThrowError('Cannot set property on injected service container');
```
Lotes, line 357: Looks a bit weird because `a` was not initially set to a value.
I would assume that any property that is not defined on this object is inaccessiable.

packages/langium/src/dependency-injection.ts
```diff
@@ -74,12 +74,20 @@ export function eagerLoad<T>(item: T): T {
...
+            if (prop === isProxy) {
+                return true;
+            } else {
+                return _resolve(obj, prop, module, injector || proxy);
+            }
+        },
         getOwnPropertyDescriptor: (obj, prop) => (_resolve(obj, prop, module, injector || proxy), Object.getOwnPropertyDescriptor(obj, prop)), // used by for..in
         has: (_, prop) => prop in module, // used by ..in..
-        ownKeys: () => [...Reflect.ownKeys(module), isProxy] // used by for..in
+        ownKeys: () => [...Object.getOwnPropertyNames(module)] // used by for..in
```
Lotes, line 89: What is the difference here? The documentation says it is the same, except that symbols are not listed with `Object.getOwnPropertyNames`.
What is the idea behind this change? On first sight it looks more complicated, because you need to handle `isProxy` more.

packages/langium/src/lsp/default-lsp-module.ts
```diff
@@ -28,7 +28,7 @@ import { DefaultWorkspaceSymbolProvider } from './workspace-symbol-provider.js';
  * Context required for creating the default language-specific dependency injection module.
  */
 export interface DefaultModuleContext extends DefaultCoreModuleContext {
-    shared: LangiumSharedServices;
+    readonly shared: LangiumSharedServices;
```
Lotes, line 31: Since you are throwing an Error on the setter of the proxy, would it be a good idea to require that all properties of the Module are readonly?

Lotes (commented):
Some small findings/understanding questions :)...

packages/langium/test/dependency-injection.test.ts
msujew: Without the `set` override on the proxy it was previously completely valid to set the value. This test ensures that it's not possible.

packages/langium/src/dependency-injection.ts
msujew: `isProxy` is a very special variable that should not be exposed from the service container using reflection. We only need it internally.

packages/langium/src/lsp/default-lsp-module.ts
msujew: I'm not sure it's possible to enforce this on the type level in TypeScript.

pbrostean:
I just saw that my code is no longer working and saw that this change caused it. I used to create my own `FoldingRangeProvicer` and `CommandHandler `and overwrote  `lsp.FoldingRAngeProvider` and `lsp.ExecuteCommandHandler` with them. Is there any possibility of changing the behaviour without overriding them in this version?

msujew:
@pbrostean You're supposed to use the [`Module` API](https://langium.org/docs/reference/configuration-services/#overriding-and-extending-services) to change the implementation used for a specific service. Everything else will likely result in runtime errors (see the referenced discussion).

pbrostean:
@msujew I've alreadly tried using the `Module API` but never managed to change the implementation of the FoldingRangeProvider this way. Would you mind sharing an example?

cdietrich:
simply 

```
export const YourDslModule ....
lsp: {
 ....
 FoldingRangeProvider: services => new YourFoldingRangeProvider(services),
}
```

does not work?
see LangiumGrammarModule in the codebase of this repo

pbrostean:
It does, thank you! I don't know why I had it in the declaration of custom services. Is there an equal implementation for a custom `ExecuteCommandHandler`?

msujew:
@pbrostean Yes, but the command handler needs to go into the shared services:

```ts
export const YourDslSharedModule = {
  lsp: {
    CommandHandler: services => new YourDslCommandHandler(services)
  }
}
```

See also [here](https://github.com/eclipse-langium/langium/blob/21b105164186dfff6a4763c21d450931e2200e5b/packages/langium/test/lsp/execute-command-handler.test.ts#L57-L61) and [here](https://github.com/eclipse-langium/langium/blob/21b105164186dfff6a4763c21d450931e2200e5b/packages/langium-vscode/src/language-server/main.ts#L17-L21) for some examples.

pbrostean:
@msujew @cdietrich you two are the best, thank's for the super fast replies!

PR #1637, opened 2024-08-18
Fix start column of indentation token instances

aabounegm:
I was not previously aware that chevrotain tokens are 1-indexed and started the indentation token from column 0 instead of 1. This caused an `Error: Illegal argument: character must be non-negative` when sending diagnostics, causing parse errors to not be shown to the user.


msujew (approved):
Makes sense, thanks!

PR #1640, opened 2024-08-21
Perform read lock on LSP requests

msujew:
I've noticed that some long running requests (i.e. completion in large files, sometimes semantic highlighting) can break pretty heavily and lead to unresolved references if the specified document gets a new build triggered. This change adds a new type of read to the `WorkspaceLock` service that allows to instantly serve a read request - all other requests are queued up until this read request has finished (or has been aborted).

### DIFF
--- a/packages/langium/src/lsp/language-server.ts
+++ b/packages/langium/src/lsp/language-server.ts
@@ -552,12 +552,14 @@ export function addCodeLensHandler(connection: Connection, services: LangiumShar
 
 export function addWorkspaceSymbolHandler(connection: Connection, services: LangiumSharedServices): void {
     const workspaceSymbolProvider = services.lsp.WorkspaceSymbolProvider;
+    const lock = services.workspace.WorkspaceLock;
     if (workspaceSymbolProvider) {
         const documentBuilder = services.workspace.DocumentBuilder;
         connection.onWorkspaceSymbol(async (params, token) => {
             try {
                 await documentBuilder.waitUntil(DocumentState.IndexedContent, token);
-                return await workspaceSymbolProvider.getSymbols(params, token);
+                const result = await lock.read(() => workspaceSymbolProvider.getSymbols(params, token), true);
+                return result;
             } catch (err) {
                 return responseError(err);
             }
@@ -567,7 +569,8 @@ export function addWorkspaceSymbolHandler(connection: Connection, services: Lang
             connection.onWorkspaceSymbolResolve(async (workspaceSymbol, token) => {
                 try {
                     await documentBuilder.waitUntil(DocumentState.IndexedContent, token);
-                    return await resolveWorkspaceSymbol(workspaceSymbol, token);
+                    const result = await lock.read(() => resolveWorkspaceSymbol(workspaceSymbol, token), true);
+                    return result;
                 } catch (err) {
                     return responseError(err);
                 }
@@ -654,6 +657,7 @@ export function createHierarchyRequestHandler<P extends TypeHierarchySupertypesP
     serviceCall: (services: LangiumCoreAndPartialLSPServices, params: P, cancelToken: CancellationToken) => HandlerResult<R, E>,
     sharedServices: LangiumSharedServices,
 ): ServerRequestHandler<P, R, PR, E> {
+    const lock = sharedServices.workspace.WorkspaceLock;
     const serviceRegistry = sharedServices.ServiceRegistry;
     return async (params: P, cancelToken: CancellationToken) => {
         const uri = URI.parse(params.item.uri);
@@ -668,7 +672,10 @@ export function createHierarchyRequestHandler<P extends TypeHierarchySupertypesP
         }
         const language = serviceRegistry.getServices(uri);
         try {
-            return await serviceCall(language, params, cancelToken);
+            const result = await lock.read(async () => {
+                return await serviceCall(language, params, cancelToken);
+            }, true); // Give this priority, since we already waited until the target state
+            return result;
         } catch (err) {
             return responseError<E>(err);
         }
@@ -681,6 +688,7 @@ export function createServerRequestHandler<P extends { textDocument: TextDocumen
     targetState?: DocumentState
 ): ServerRequestHandler<P, R, PR, E> {
     const documents = sharedServices.workspace.LangiumDocuments;
+    const lock = sharedServices.workspace.WorkspaceLock;
     const serviceRegistry = sharedServices.ServiceRegistry;
     return async (params: P, cancelToken: CancellationToken) => {
         const uri = URI.parse(params.textDocument.uri);
@@ -694,9 +702,17 @@ export function createServerRequestHandler<P extends { textDocument: TextDocumen
             return responseError<E>(new Error(errorText));
         }
         const language = serviceRegistry.getServices(uri);
+        const document = documents.getDocument(uri);
+        if (!document) {
+            const errorText = `Could not find document for uri: '${uri}'`;
+            console.debug(errorText);
+            return responseError<E>(new Error(errorText));
+        }
         try {
-            const document = await documents.getOrCreateDocument(uri);
-            return await serviceCall(language, document, params, cancelToken);
+            const result = await lock.read(async () => {
+                return await serviceCall(language, document, params, cancelToken);
+            }, true); // Give this priority, since we already waited until the target state
+            return result;
         } catch (err) {
             return responseError<E>(err);
         }
@@ -709,6 +725,7 @@ export function createRequestHandler<P extends { textDocument: TextDocumentIdent
     targetState?: DocumentState
 ): RequestHandler<P, R | null, E> {
     const documents = sharedServices.workspace.LangiumDocuments;
+    const lock = sharedServices.workspace.WorkspaceLock;
     const serviceRegistry = sharedServices.ServiceRegistry;
     return async (params: P, cancelToken: CancellationToken) => {
         const uri = URI.parse(params.textDocument.uri);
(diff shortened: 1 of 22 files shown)


dhuebner:
As what I understood from the discussion in the weekly dev meeting, the problem is the following:
We have a running `read` operation (e.g. semantic highlighting) and a new `write` operation request (e.g. document changed) arrives.  What happens, is that the `write` operation resets the document where the `read` operation still operates on this document reference. This seems to break the document and also the running `read` operation may fail..?
Introducing a `priority read` doesn't feels right to me as some? read operations become a priority over the write operations. It also makes the locking logic more complicated and difficult to understand. 

Maybe the solution in this PR is the only one, but I still want to share some other ideas (not sure it works well with Langium):
- Would creating a new Document instead of reseting an existing one work? Doing so, the running `read` operation can finish its work on the old outdated document and nothing should break.
- What one can also consider is actively trigger canceling on all running `read` services when a `write` arrives and wait until they are canceled.  After that apply the write operation.

packages/langium/src/workspace/workspace-lock.ts
```diff
@@ -28,8 +28,13 @@ export interface WorkspaceLock {
...
      *
      * Note that read actions are not allowed to modify anything in the workspace. Please use {@link write} instead.
+     *
+     * @param action the action to perform.
+     * @param priority
+     * If set to true, the action will receive immediate attention and will be instantly executed. All other queued read/write actions will be postponed.
+     * This should only be used for actions that require a specific document state and have already awaited the necessary write actions.
      */
-    read<T>(action: () => MaybePromise<T>): Promise<T>;
+    read<T>(action: () => MaybePromise<T>, priority?: boolean): Promise<T>;
```
Lotes, line 37: I am associating `priority` with an ordering/number. I understand the usage here as `hasPriority`. Maybe there is a better name, like: `isUrgent`, `important`, `handleFirst`...

packages/langium/src/workspace/workspace-lock.ts
```diff
@@ -59,8 +64,25 @@ export class DefaultWorkspaceLock implements WorkspaceLock {
         return this.enqueue(this.writeQueue, action, tokenSource.token);
     }
 
-    read<T>(action: () => MaybePromise<T>): Promise<T> {
-        return this.enqueue(this.readQueue, action);
+    read<T>(action: () => MaybePromise<T>, priority?: boolean): Promise<T> {
```
Lotes, line 67: Just thinking loud: Is this the first time we have a "priority" queue?
You could refactor a common component/function if not.

Lotes (commented):
Some small things I noticed

packages/langium/src/workspace/workspace-lock.ts
msujew: The workspace lock is a mutex, not a queue. It features some very specific semantics about read/read with prio/write actions that are likely not relevant for a common component.

spoenemann (commented):
I share @dhuebner's concerns. Is it right to push through an LSP request even though the text has changed in the meantime?

Reading the specification section [Implementation Considerations](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#implementationConsiderations), it looks like the best solution would be to return an error code `ContentModified` when a change has happened.

msujew:
@spoenemann The link you've posted literally says this about changed content:

> servers should therefore not decide by themselves to cancel requests simply due to that fact that a state change notification is detected in the queue. As said the result could still be useful for the client.

We do exactly this: Prevent the response from failing by blocking the workspace lock and returning the result. In the meantime, the client is free to cancel the pending request which we completely respect and unblock the workspace lock

packages/langium/src/lsp/language-server.ts
```diff
@@ -721,9 +738,17 @@ export function createRequestHandler<P extends { textDocument: TextDocumentIdent
...
+            const errorText = `Could not find document for uri: '${uri}'`;
+            console.debug(errorText);
+            return responseError<E>(new Error(errorText));
+        }
         try {
-            const document = await documents.getOrCreateDocument(uri);
-            return await serviceCall(language, document, params, cancelToken);
+            const result = await lock.read(async () => {
+                return await serviceCall(language, document, params, cancelToken);
+            }, true); // Give this priority, since we already waited until the target state
```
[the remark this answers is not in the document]
msujew: All of these are getting cancelled by the language client - they shouldn't be blocking the workspace in any way for longer than our timeout (5ms).

spoenemann:
But that section also says:
> if a server detects an internal state change (for example, a project context changed) that invalidates the result of a request in execution the server can error these requests with `ContentModified`
This PR is about long running requests; when the state of a document is changed so that we can no longer finish the request, we should return an error. Note that this is not the same as canceling the request (which should be done only by the client).

spoenemann:
Could it help to use a utility function like this in LSP request processing?
```typescript
export async function interruptAndCheckDocument(document: LangiumDocument, token: CancellationToken): Promise<void> {
    const previousState = document.state;
    await interruptAndCheck(token);
    if (document.state < previousState) {
        throw new ResponseError(LSPErrorCodes.ContentModified, 'Document content has been modified.');
    }
}
```

msujew:
> This PR is about long running requests; when the state of a document is changed so that we can no longer finish the request

The issue is: We cannot really know? For example, for the document highlight:

1. User edits text on line 2 and the language client requests a highlighting delta
2. Server is computing delta request for line 2
3. Meanwhile, user edits text on line 10 and the language client requests a highlighting delta for that line
4. ???

Aborting the initial operation for line 2 when receiving a document update is not the correct move here - the data for that highlighting is still valid and aborting the request might leave the user with incorrectly highlighted text.

This is the case for most LSP requests. The internal state change the documentation is talking about (like project context switches) are much more fundamental changes than the "normal" document changes.

packages/langium/src/workspace/workspace-lock.ts
```diff
@@ -59,8 +64,25 @@ export class DefaultWorkspaceLock implements WorkspaceLock {
...
+                run();
+                // Ensure that in any case the counter is decremented and the next operation is performed
+                this.counter--;
+                this.performNextOperation();
+            };
+            // Instanly run the action and resolve/reject the promise afterwards
+            Promise.resolve(action()).then(
+                result => end(() => deferred.resolve(result)),
+                err => end(() => deferred.reject(err))
+            );
```
spoenemann, line 81: I see two potential issues with this code:
- The solution with the local `end` function is a more complicated way of expressing a `finally` block.
- I suspect that `Promise.resolve(action())` does not behave as we want when the `action` throws an error – the error would just be propagated instead of rejecting the resulting promise.

I suggest making this whole method `async` so we can `await` the action and wrap that in a `try`-`finally` block.

packages/langium/src/lsp/language-server.ts
```diff
@@ -668,7 +672,10 @@ export function createHierarchyRequestHandler<P extends TypeHierarchySupertypesP
         }
         const language = serviceRegistry.getServices(uri);
         try {
-            return await serviceCall(language, params, cancelToken);
+            const result = await lock.read(async () => {
+                return await serviceCall(language, params, cancelToken);
+            }, true); // Give this priority, since we already waited until the target state
```
spoenemann, line 677: I think this can be simplified to
```typescript
await lock.read(() => serviceCall(language, params, cancelToken), true)
```
(same in `createServerRequestHandler` and `createRequestHandler`).

spoenemann (commented):
Yes, I see the problem with semantic highlighting (Semantic Tokens) (you said document highlighting, but that's a different service). The situation is different for other services. But it's true that it's the client that should decide whether to use a potentially outdated result or not.

We could apply this and then check how it affects the editing experience in larger projects. I'm particularly curious how it performs when completion is continuously triggered while typing.

packages/langium/src/lsp/language-server.ts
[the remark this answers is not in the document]
spoenemann: Where does that 5 ms timeout come from?

spoenemann:
Question: with this change, does it make any sense for LSP services to use `interruptAndCheck`? I see it currently used by inlay-hint-provider, semantic-token-provider and workspace-symbol-provider.

Should we make it clear in the function documentation that it should be used only in the document building process and the associated services?

packages/langium/src/workspace/workspace-lock.ts
```diff
@@ -59,8 +72,26 @@ export class DefaultWorkspaceLock implements WorkspaceLock {
...
+                    const result = await action();
+                    deferred.resolve(result);
+                } catch (err) {
+                    deferred.reject(err);
+                } finally {
+                    this.counter--;
+                    this.performNextOperation();
+                }
+            })();
+            return deferred.promise;
```
spoenemann, line 91: I think you can get rid of the `Deferred` if you `return` the result in the inline function and either `return` the inline function's result or eliminate it by making the whole method `async`.

packages/langium/src/utils/index.ts
```diff
@@ -15,8 +16,7 @@ export * from './stream.js';
 export * from './uri-utils.js';
 
 import * as AstUtils from './ast-utils.js';
-import * as Cancellation from './cancellation.js';
 import * as CstUtils from './cst-utils.js';
 import * as GrammarUtils from './grammar-utils.js';
 import * as RegExpUtils from './regexp-utils.js';
-export { AstUtils, Cancellation, CstUtils, GrammarUtils, RegExpUtils };
+export { AstUtils, CstUtils, GrammarUtils, RegExpUtils };
```
spoenemann, line 22: We should keep the `Cancellation` namespace to avoid a braking change. Can we mark it deprecated?

spoenemann:
This is a good change in general, but I'm hesitant to make it effective for all LSP requests given that the scope of the initial problem is rather small (it affects you only in specific circumstances, e.g. very large files and a certain combination of editor actions).

Proposal: Reduce the scope of these changes so the new read synchronization is applied only to the semantic tokens service.

cdietrich-everest:
@msujew any plans for this one. still seeing stuff like

Request textDocument/semanticTokens/range failed. Message: Document state of
  file:///Users/dietrich/tmp/localvscodethingy-new2/xxxx/somefile.someextension
  is Parsed, requiring Linked, but workspace state is already Validated. Returning
  undefined. Code: -32803
  
  from time to time

msujew:
@cdietrich-everest I plan to get back to this. I haven't encountered this issue anymore recently, but I'll probably get back to this next week, when I have some more time. Thanks for the reminder.

msujew:
I'm getting back onto this, but first I'll close this PR. Creating a fresh PR for this change.

Issue #1642, opened 2024-08-23
`langium` package cannot be tree-shaken well


jindong-zhannng:
**Reproduction**: https://github.com/jindong-zhannng/langium-bundle-test
**Bundler**: vite

There is only one import in `main.ts` (to simulate the case of `generated/ast.ts`): 

```ts
import { AbstractAstReflection } from 'langium'
```

And class `AbstractAstReflection` also has **zero** dependency.

**Expect**: Bundle should only include `AbstractAstReflection` relevant code.
**Actual**:

![image](https://github.com/user-attachments/assets/4b6d66cf-641d-4f05-8ec1-29e80a36e7b3)

*(run `npx vite-bundle-visualizer --open -o ./stats.html` to generate report)*

Obviously it includes lots of dead code like parser, validator and their dependencies. It brings the bundle size up to 174.67 kB (minified).

**Idea**

I tried to add ` "sideEffects": false` to `package.json`. It significantly reduces the size to 1.27 kB and has a perfect report:

![image](https://github.com/user-attachments/assets/61a582c5-3df8-4e29-ad42-44801859450f)

I'm not sure if it's suitable for langium and I also don't have a very good understanding of this setting. Just put it here for your reference.

msujew:
Hey @jindong-zhannng,

thanks for the investigation. Since we actually don't feature any side effects, I've filed https://github.com/eclipse-langium/langium/pull/1643. It will be released as part of 3.2 (hopefully next week).

PR #1643, opened 2024-08-23
Add `sideEffects` flag to library packages

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1642

Adds the `sideEffects: false` flag to our library package.json files. Since we only export methods/constants/classes and don't perform any polyfills, our libraries don't have any side effects. This results in smaller bundles when only a subset of Langium's features are used.

### DIFF
--- a/packages/langium-railroad/package.json
+++ b/packages/langium-railroad/package.json
@@ -18,6 +18,7 @@
     "lib",
     "src"
   ],
+  "sideEffects": false,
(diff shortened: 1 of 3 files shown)


dhuebner (approved):
I tested our code (hope I did it right) with Rollup to check for side effects and it seems like we don't have them, at least nothing was reported.

PR #1659, opened 2024-09-02
Enable cache eviction on specific document state

msujew:
Right now, our caches only evict for documents that have actually changed. However, those documents with direct references to those files (see `isAffected` computation) aren't evicted from cache. This change allows to perform more fine-grained eviction behavior on caches. This came to my mind when helping with https://github.com/eclipse-langium/langium/discussions/1652.

### DIFF
--- a/packages/langium/src/utils/caching.ts
+++ b/packages/langium/src/utils/caching.ts
@@ -7,6 +7,7 @@
 import type { Disposable } from './disposable.js';
 import type { URI } from './uri-utils.js';
 import type { LangiumSharedCoreServices } from '../services.js';
+import type { DocumentState } from '../workspace/documents.js';
 
 export abstract class DisposableCache implements Disposable {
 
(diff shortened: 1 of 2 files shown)


packages/langium/src/utils/caching.ts
```diff
@@ -157,10 +177,24 @@ export class DocumentCache<K, V> extends ContextCache<URI | string, K, V, string
...
-            this.clear();
-        }));
+        if (state) {
+            this.onDispose(sharedServices.workspace.DocumentBuilder.onBuildPhase(state, () => {
+                this.clear();
+            }));
+        } else {
+            this.onDispose(sharedServices.workspace.DocumentBuilder.onUpdate(() => {
+                this.clear();
+            }));
```
spoenemann, line 197: I know this form of registering disposables is used a lot in other frameworks like Theia, but IMO it's a bit unfortunate: if you don't read carefully, `this.onDispose(doSomething())` looks like `doSomething` happens only on dispose. It can be clarified by splitting it in separate statements:
```typescript
const cleanup = doSomething();
this.onDispose(cleanup);
```

spoenemann (approved):
👍

JohannesMeierSE:
FYI: We need this feature in Typir. Therefore we are looking forward to a Langium release with this feature!

PR #1660, opened 2024-09-02
Normalize all LSP URIs

msujew:
Addresses an issue discussed in https://github.com/eclipse-langium/langium/discussions/1619.

The main issue is that VS Code itself uses a non spec-compliant URI format in which they encode the `C:` to `c%A3`. This behavior is also mirrored in the `vscode-uri` package, which we use to store `LangiumDocument` objects.

However, language clients like LSP4E correctly send `C:` - which leads us down a very deep rabbit hole. In the end, the best solution to this is to write our own `TextDocuments` service that normalizes the incoming URIs to the same schema used by VS Code. This way, any incoming URI can find its respective document in the internal map.

Note that it seems like LSP4E can deal with our outgoing URIs just fine - it's the incoming URIs that are problematic.

### DIFF
--- a/packages/langium/src/lsp/normalized-text-documents.ts
+++ b/packages/langium/src/lsp/normalized-text-documents.ts
@@ -11,6 +11,11 @@ import type {
 import { TextDocumentSyncKind, Disposable, Emitter } from 'vscode-languageserver';
 import { UriUtils } from '../utils/uri-utils.js';
 
+/**
+ * A manager service that keeps track of all currently opened text documents.
+ *
+ * Designed to be compatible with the `TextDocuments` class in the `vscode-languageserver` package.
+ */
 export interface TextDocuments<T extends { uri: string }> {
     /**
      * An event that fires when a text document managed by this manager
--- a/packages/langium/src/lsp/normalized-text-documents.ts
+++ b/packages/langium/src/lsp/normalized-text-documents.ts
@@ -1,0 +1,217 @@
+/******************************************************************************
+ * Copyright 2024 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import type {
+    Connection, DidOpenTextDocumentParams, DidChangeTextDocumentParams, DidCloseTextDocumentParams, TextDocumentsConfiguration, TextDocumentChangeEvent,
+    TextDocumentWillSaveEvent, RequestHandler, TextEdit, Event, WillSaveTextDocumentParams, CancellationToken, DidSaveTextDocumentParams
+} from 'vscode-languageserver';
+import { TextDocumentSyncKind, Disposable, Emitter } from 'vscode-languageserver';
+import { UriUtils } from '../utils/uri-utils.js';
+
+export interface TextDocuments<T extends { uri: string }> {
+    /**
+     * An event that fires when a text document managed by this manager
+     * has been opened.
+     */
+    readonly onDidOpen: Event<TextDocumentChangeEvent<T>>;
+    /**
+     * An event that fires when a text document managed by this manager
+     * has been opened or the content changes.
+     */
+    readonly onDidChangeContent: Event<TextDocumentChangeEvent<T>>;
+    /**
+     * An event that fires when a text document managed by this manager
+     * will be saved.
+     */
+    readonly onWillSave: Event<TextDocumentWillSaveEvent<T>>;
+    /**
+     * Sets a handler that will be called if a participant wants to provide
+     * edits during a text document save.
+     */
+    onWillSaveWaitUntil(handler: RequestHandler<TextDocumentWillSaveEvent<T>, TextEdit[], void>): void;
+    /**
+     * An event that fires when a text document managed by this manager
+     * has been saved.
+     */
+    readonly onDidSave: Event<TextDocumentChangeEvent<T>>;
+    /**
+     * An event that fires when a text document managed by this manager
+     * has been closed.
+     */
+    readonly onDidClose: Event<TextDocumentChangeEvent<T>>;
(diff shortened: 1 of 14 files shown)


borisbrodski (commented):
Tested on Windows with Eclipse 2024-06-R. Works great!

packages/langium/src/lsp/normalized-text-documents.ts
```diff
@@ -0,0 +1,217 @@
...
+ ******************************************************************************/
+
+import type {
+    Connection, DidOpenTextDocumentParams, DidChangeTextDocumentParams, DidCloseTextDocumentParams, TextDocumentsConfiguration, TextDocumentChangeEvent,
+    TextDocumentWillSaveEvent, RequestHandler, TextEdit, Event, WillSaveTextDocumentParams, CancellationToken, DidSaveTextDocumentParams
+} from 'vscode-languageserver';
+import { TextDocumentSyncKind, Disposable, Emitter } from 'vscode-languageserver';
+import { UriUtils } from '../utils/uri-utils.js';
+
+export interface TextDocuments<T extends { uri: string }> {
```
sailingKieler, line 14: I think it makes sense to leave a hint on the origin of this definition, which is the class declaration in `vscode-languageserver/lib/common/textDocuments.d.ts`, right?

sailingKieler (approved):
Subclassing the existing `TextDocuments` service and, e.g., wrapping the Map `_syncedDocuments` doesn't work (or wouldn't be more elegant) as everything is private or buried within inline function expressions, correct? So I agree to copy-paste the `TextDocuments` service.

I assume this thing is quite stable in the vscode repo and would only have to be touched in case new document event types would be introduced by the vscode folks, right?

msujew:
> Subclassing the existing TextDocuments service and, e.g., wrapping the Map _syncedDocuments doesn't work (or wouldn't be more elegant) as everything is private or buried within inline function expressions, correct?

Yes, It'd be great if we could have something like a `normalizeUri` method in the `TextDocumentsConfiguration` interface, but right now, we need to create our own class due to the inline functions.

sailingKieler:
I'm however wondering what about `uri` references that are included when notifying the LC about diagnostics and the like, see https://github.com/eclipse-langium/langium/blob/99fb880e08361ddb511ff4a03f003c67e73eb9a0/packages/langium/src/lsp/language-server.ts#L331-L339

msujew:
@sailingKieler I've mentioned this in the initial post:

> Note that it seems like LSP4E can deal with our outgoing URIs just fine - it's the incoming URIs that are problematic.

Outgoing URIs are no problems, as the language client should parse the malformed URIs correctly - it's just that we cannot deal with "correctly" formatted URIs on the incoming side right now.

sailingKieler:
>> Note that it seems like LSP4E can deal with our outgoing URIs just fine - it's the incoming URIs that are problematic.
>
> Outgoing URIs are no problems, as the language client should parse the malformed URIs correctly

That's an interesting assumption. 😅

I assume that VS Code implementations sitting at the beginning of the message pipeline of our LS (like `Connection`) are in charge of this specific encoding, I believe you that LSP4E and others are happy now.
I'm just wondering whether a Langium-based LS in VS Code on Windows still works as expected, as we now don't provide "other" uris in notifications than we got earlier.

msujew:
> I assume that VS Code implementations sitting at the beginning of the message pipeline of our LS (like Connection) are in charge of this specific encoding

They're not, the they just pass in whatever the language client gives them. They don't perform any manipulation on the LSP data. We do :)

sailingKieler:
I think I got it now.
> They're not, the they just pass in whatever the language client gives them. They don't perform any manipulation on the LSP data. We do :)

You're basically saying: Everything that happens in https://github.com/microsoft/vscode-languageserver-node/blob/f58f4dff16ad2760028bb1cb95e882de30a1000f/server/src/common/textDocuments.ts#L189-L244 relies on the plain uri strings provided by the language client, while we're doing https://github.com/eclipse-langium/langium/blob/99fb880e08361ddb511ff4a03f003c67e73eb9a0/packages/langium/src/lsp/document-update-handler.ts#L128 on first contact with a new document, and later on `document.uri.toString()`, e.g. in https://github.com/eclipse-langium/langium/blob/99fb880e08361ddb511ff4a03f003c67e73eb9a0/packages/langium/src/workspace/documents.ts#L409-L417

With your changes the book keeping in `TextDocuments` is in sync with what `LangiumDocuments` & friends do. 👍

msujew:
@sailingKieler Exactly - and the main issue is essentially this: https://github.com/eclipse-langium/langium/blob/99fb880e08361ddb511ff4a03f003c67e73eb9a0/packages/langium/src/workspace/documents.ts#L271

We try to get the text document based on the langium URI - however, these don't match on Windows for a bunch of language clients. That's why we need to keep every URI formatted in the same way.

Note that the latest commit adds a few more utility methods to removes a few hacks inside our test framework.

packages/langium/src/lsp/normalized-text-documents.ts
```diff
@@ -142,8 +153,28 @@ export class NormalizedTextDocuments<T extends { uri: string }> implements TextD
         return this._onDidClose.event;
     }
 
-    public get(uri: string): T | undefined {
-        return this._syncedDocuments.get(UriUtils.normalize(uri));
+    public get(uri: string | URI): T | undefined {
+        return this._syncedDocuments.get(UriUtils.normalize(uri.toString()));
```
sailingKieler, line 157: It's just a detail: the `toString()` is not required here considering that `normalize` accepts `URI|string`... 😉.

sailingKieler:
> Note that the latest commit adds a few more utility methods to removes a few hacks inside our test framework.

Had a look. Great that we can get rid of those workarounds at this occasion. 👌

PR #1663, opened 2024-09-04
Rename "delimeter" to "delimiter"

aabounegm:
Funny how I seem to have originally thought it's from `"deli" + "meter"` rather than `"de" + "limit" + "er"` :D


msujew (approved):
Great, thanks 👍

PR #1664, opened 2024-09-04
Change fields to methods to support overriding

aabounegm:
`indentMatcher` and `dedentMatcher` were defined as normal fields whose values were functions, which prevented calling them using `super` in an override. This converts them to normal methods to support overriding them.


msujew (approved):
LGTM 👍

PR #1665, opened 2024-09-05
Export 'createParser' function to ease completion parser customization

martin-fleck-at:
When creating a custom completion parser with the `createCompletionParser` function, the class `LangiumCompletionParser` is fixed. Doing customizations is therefore difficult and would be easier if the `createParser` function would be exposed.


msujew (approved):
Fine with me 👍

PR #1668, opened 2024-09-06
Extend Lexer interface to expose diagnostics and data as a lexing report

martin-fleck-at:
This is based on a discussion that started in https://github.com/eclipse-langium/langium/discussions/1653 and was already marked as a TODO in-code:
```
// TODO: find a way to report error diagnostics message
``` 

- Add support to report diagnostics during lexing process
- Properly map diagnostic severities
- Mark method and report as optional for backwards-compatibility

For indentation:
- Add dedent tokens to report until consumed by lexer

Please note this PR is based on https://github.com/eclipse-langium/langium/pull/1664 which was already approved but has yet to be merged.

### DIFF
--- a/packages/langium/src/parser/indentation-aware.ts
+++ b/packages/langium/src/parser/indentation-aware.ts
@@ -6,7 +6,7 @@
 
 import type { CustomPatternMatcherFunc, TokenType, IToken, IMultiModeLexerDefinition } from 'chevrotain';
 import type { Grammar, TerminalRule } from '../languages/generated/ast.js';
-import type { TokenBuilderOptions } from './token-builder.js';
+import type { ILexingReport, TokenBuilderOptions } from './token-builder.js';
 import type { LexerResult } from './lexer.js';
 import type { LangiumCoreServices } from '../services.js';
 import { createToken, createTokenInstance, Lexer } from 'chevrotain';
@@ -69,22 +69,28 @@ export enum LexingMode {
     IGNORE_INDENTATION = 'ignore-indentation',
 }
 
+export interface IndentationLexingReport extends ILexingReport {
+    /** Dedent tokens that are necessary to close the remaining indents. */
+    remainingDedents: IToken[];
+}
+
 /**
  * A token builder that is sensitive to indentation in the input text.
  * It will generate tokens for indentation and dedentation based on the indentation level.
  *
  * The first generic parameter corresponds to the names of terminal tokens,
- * while the second one corresonds to the names of keyword tokens.
+ * while the second one corresponds to the names of keyword tokens.
  * Both parameters are optional and can be imported from `./generated/ast.js`.
  *
  * Inspired by https://github.com/chevrotain/chevrotain/blob/master/examples/lexer/python_indentation/python_indentation.js
  */
 export class IndentationAwareTokenBuilder<Terminals extends string = string, KeywordName extends string = string> extends DefaultTokenBuilder {
     /**
-     * The stack in which all the previous matched indentation levels are stored
-     * to understand how deep a the next tokens are nested.
+     * The stack stores all the previously matched indentation levels to understand how deeply the next tokens are nested.
+     * The stack is valid for lexing
      */
     protected indentationStack: number[] = [0];
+
     readonly options: IndentationTokenBuilderOptions<Terminals, KeywordName>;
 
     /**
@@ -123,7 +129,7 @@ export class IndentationAwareTokenBuilder<Terminals extends string = string, Key
         });
     }
 
-    override buildTokens(grammar: Grammar, options?: TokenBuilderOptions | undefined) {
+    override buildTokens(grammar: Grammar, options?: TokenBuilderOptions | undefined): TokenType[] | IMultiModeLexerDefinition {
         const tokenTypes = super.buildTokens(grammar, options);
         if (!isTokenTypeArray(tokenTypes)) {
             throw new Error('Invalid tokens built by default builder');
@@ -173,6 +179,14 @@ export class IndentationAwareTokenBuilder<Terminals extends string = string, Key
         }
     }
 
+    override popLexingReport(text: string): IndentationLexingReport {
+        const result = super.popLexingReport(text);
+        return {
+            ...result,
+            remainingDedents: this.popRemainingDedents(text),
+        };
+    }
+
     /**
      * Helper function to check if the current position is the start of a new line.
      *
@@ -191,7 +205,7 @@ export class IndentationAwareTokenBuilder<Terminals extends string = string, Key
      * @param offset The current position at which to attempt a match
      * @returns The current and previous indentation levels and the matched whitespace
      */
-    protected matchWhitespace(text: string, offset: number) {
+    protected matchWhitespace(text: string, offset: number, _tokens: IToken[], _groups: Record<string, IToken[]>): { currIndentLevel: number, prevIndentLevel: number, match: RegExpExecArray | null } {
         this.whitespaceRegExp.lastIndex = offset;
         const match = this.whitespaceRegExp.exec(text);
         return {
@@ -210,8 +224,8 @@ export class IndentationAwareTokenBuilder<Terminals extends string = string, Key
      * @param offset Current position in the input string
      * @returns The indentation token instance
      */
-    protected createIndentationTokenInstance(tokenType: TokenType, text: string, image: string, offset: number) {
-        const lineNumber = text.substring(0, offset).split(/\r\n|\r|\n/).length;
+    protected createIndentationTokenInstance(tokenType: TokenType, text: string, image: string, offset: number): IToken {
+        const lineNumber = this.getLineNumber(text, offset);
         return createTokenInstance(
             tokenType,
             image,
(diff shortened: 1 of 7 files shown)


martin-fleck-at:
@aabounegm @msujew FYI

msujew (commented):
Can you cherry pick my latest commit from https://github.com/eclipse-langium/langium/tree/msujew/indentation-diagnostics? Since you opened this PR from an organization repository, I can't push to it myself.

msujew:
I'm not happy with the API by the way, but we kind of boxed ourselves into this, by simply reusing all the types that chevrotain provides us. I'll probably perform a refactoring of the whole lexer/token builder infrastructure for 4.0, since we also have a project that struggles with the way the existing API works (i.e. no known URI during the lexing phase, etc.).

martin-fleck-at:
@msujew I rebased the commit and cherry-picked yours on top of it. Unfortunately, I couldn't open a branch in this repository directly.

I agree with what you said about the API. I aimed to stay backwards-compatible but having the `lexerErrors` and a separate but optional `lexerReport` is indeed a bit unexpected. But I'm happy to leave the API breakages to you where also the overall terminology (e.g., is "report" really the best term?) could be re-worked.

Thanks again for the very fast turnaround on this!

msujew (approved):
Looks good to me, thanks 👍 

Using `report` is fine for now, I might revisit this decision at a later point.

packages/langium/src/parser/indentation-aware.ts
```diff
@@ -191,7 +205,7 @@ export class IndentationAwareTokenBuilder<Terminals extends string = string, Key
      * @param offset The current position at which to attempt a match
      * @returns The current and previous indentation levels and the matched whitespace
      */
-    protected matchWhitespace(text: string, offset: number) {
+    protected matchWhitespace(text: string, offset: number, _tokens: IToken[], _groups: Record<string, IToken[]>): { currIndentLevel: number, prevIndentLevel: number, match: RegExpExecArray | null } {
```
aabounegm, line 208: Can you also update the JSDoc with the new parameters?

packages/langium/src/parser/token-builder.ts
```diff
@@ -20,9 +20,31 @@ export interface TokenBuilderOptions {
...
+/**
+ * A custom lexing report that can be produced by the token builder during the lexing process.
+ * Adopters need to ensure that the any custom fields are serializable so they can be sent across worker threads.
+ */
+export interface LexingReport {
+    diagnostics: LexingDiagnostic[];
+}
+
+export interface LexingDiagnostic extends ILexingError {
+    severity?: 'error' | 'warning' | 'info' | 'hint';
```
aabounegm, line 40: We should probably have some global `type DiagnosticSeverity = 'error' | 'warning' | 'info' | 'hint';` since it's used a lot

packages/langium/src/parser/token-builder.ts
```diff
@@ -42,6 +64,16 @@ export class DefaultTokenBuilder implements TokenBuilder {
         return tokens;
     }
 
+    popLexingReport(_text: string): LexingReport {
```
aabounegm, line 67: I think a name like `finalize` or `finalizeLexing` would probably be more user-friendly

packages/langium/src/parser/lexer.ts
```diff
@@ -21,6 +22,7 @@ export interface LexerResult {
      */
     hidden: IToken[];
     errors: ILexingError[];
+    report?: LexingReport;
```
aabounegm, line 25: Personally, I would mark `errors` as `@deprecated` and include them in the report. They can point to the same array for backwards compatibility, only to be remove in the next major version.
Also, it would be great if you can add JSDoc

packages/langium/src/parser/lexer.ts
```diff
@@ -31,10 +33,12 @@ export interface Lexer {
 export class DefaultLexer implements Lexer {
 
     protected chevrotainLexer: ChevrotainLexer;
+    protected tokenBuilder: TokenBuilder;
     protected tokenTypes: TokenTypeDictionary;
 
-    constructor(services: LangiumCoreServices) {
-        const tokens = services.parser.TokenBuilder.buildTokens(services.Grammar, {
+    constructor( services: LangiumCoreServices) {
```
aabounegm, line 39: Why the extra space after `(`?

packages/langium/src/parser/langium-parser.ts
```diff
@@ -19,11 +19,13 @@ import { isAssignment, isCrossReference, isKeyword } from '../languages/generate
...
 import { assignMandatoryProperties, getContainerOfType, linkContentToContainer } from '../utils/ast-utils.js';
 import { CstNodeBuilder } from './cst-node-builder.js';
+import type { LexingReport } from './token-builder.js';
 
 export type ParseResult<T = AstNode> = {
     value: T,
     parserErrors: IRecognitionException[],
-    lexerErrors: ILexingError[]
+    lexerErrors: ILexingError[],
+    lexerReport?: LexingReport
```
aabounegm, line 28: These 2 can also be merged, with one being marked as deprecated, as mentioned in `LexingReport`

packages/langium/src/validation/document-validator.ts
```diff
@@ -97,21 +98,23 @@ export class DefaultDocumentValidator implements DocumentValidator {
     }
 
     protected processLexingErrors(parseResult: ParseResult, diagnostics: Diagnostic[], _options: ValidationOptions): void {
-        for (const lexerError of parseResult.lexerErrors) {
+        const lexerDiagnostics = [...parseResult.lexerErrors, ...parseResult.lexerReport?.diagnostics ?? []] as LexingDiagnostic[];
+        for (const lexerDiagnostic of lexerDiagnostics) {
+            const severity = lexerDiagnostic?.severity ?? 'error';
```
aabounegm, line 103: I don't think the diagnostic element itself can be undefined
```suggestion
            const severity = lexerDiagnostic.severity ?? 'error';
```

packages/langium/src/validation/document-validator.ts
```diff
@@ -245,8 +248,26 @@ export function toDiagnosticSeverity(severity: 'error' | 'warning' | 'info' | 'h
...
+export function toDiagnosticData(severity: 'error' | 'warning' | 'info' | 'hint'): DiagnosticData {
+    switch (severity) {
+        case 'error':
+            return diagnosticData(DocumentValidator.LexingError);
+        case 'warning':
+            return diagnosticData(DocumentValidator.LexingWarning);
+        case 'info':
+            return diagnosticData(DocumentValidator.LexingInfo);
+        case 'hint':
+            return diagnosticData(DocumentValidator.LexingHint);
```
aabounegm, line 260: I personally prefer a `Record<DiagnosticSeverity, DocumentValidator>` over a `switch`, and also using an enum instead of a namespace (to make this type possible)

aabounegm (commented):
Sorry for the delay. Nothing critical, just mostly opinions and questions regarding the code style. Nothing important

packages/langium/src/validation/document-validator.ts
martin-fleck-at: I aligned this with the `toDiagnosticSeverity` function in the same file so I'd say we should at least be consistent even though I do not have any hard feelings regarding either way.

packages/langium/src/validation/document-validator.ts
martin-fleck-at: Absolutely correct!

packages/langium/src/parser/langium-parser.ts
martin-fleck-at: Yeah, I didn't wanna mark parts of the API as deprecated and @msujew already mentioned that he may re-work those parts when he does a 4.x release. But I fully agree that this is an odd API now.

packages/langium/src/parser/lexer.ts
martin-fleck-at: Ah, this was a pure oversight ;-) Thanks for cacthing that!

packages/langium/src/parser/token-builder.ts
martin-fleck-at: Ah, the difficulty of good naming. I was debating over this and I previously had `flushLexingReport` to indicate that afterwards the state is reset. My issue with `finalizeX` or `endX` or `stopX` was the asymmetry because we do not have the opposing `startX` or `beginX`. In the end, I chose `popLexingReport` as this is what was used in the indendation-based token provider with the `popRemainingDedents`. But as usual, no hard opinions.

packages/langium/src/parser/token-builder.ts
martin-fleck-at: Yeah, that might be a good idea, I'll look into it.

packages/langium/src/parser/indentation-aware.ts
martin-fleck-at: Good point!

martin-fleck-at:
@aabounegm Thank you for your feedback! I am finalizing https://github.com/eclipse-langium/langium/pull/1669 right now and I'll add the non-controversial changes there. For anything else (like marking API as deprecated or particular types) I'd suggest to open a new PR since this may require some discussion.

packages/langium/src/parser/token-builder.ts
aabounegm: Yeah that makes sense, but `pop` also implies a corresponding `push` function. I like the `flush` variant more 👍

PR #1669, opened 2024-09-06
Introduce tokenizing options for full and partial mode

martin-fleck-at:
This is based on a discussion that started in https://github.com/eclipse-langium/langium/discussions/1653 where it was discussed that auto-completing the dedent tokens in the indentation-aware lexer messes with the auto-completion in the editor.

Add tokenizing mode to tokenizing method
- Full: We get the full text to tokenize
- Partial: We get only a portion of the text to tokenize

In indentation lexing, we do not auto-complete dedents for partial mode

Please note this PR is based on https://github.com/eclipse-langium/langium/pull/1664 which was already approved but has yet to be merged.

### DIFF
--- a/packages/langium/src/parser/lexer.ts
+++ b/packages/langium/src/parser/lexer.ts
@@ -28,7 +28,7 @@ export interface LexerResult {
 export type TokenizeMode = 'full' | 'partial';
 
 export interface TokenizeOptions {
-    mode: TokenizeMode;
+    mode?: TokenizeMode;
 }
 
 export const DEFAULT_TOKENIZE_OPTIONS: TokenizeOptions = { mode: 'full' };
--- a/packages/langium/src/parser/lexer.ts
+++ b/packages/langium/src/parser/lexer.ts
@@ -25,9 +25,15 @@ export interface LexerResult {
     report?: LexingReport;
 }
 
+export interface TokenizeOptions {
+    mode: 'full' | 'partial';
+}
+
+export const DEFAULT_TOKENIZE_OPTIONS: TokenizeOptions = { mode: 'full' };
+
 export interface Lexer {
     readonly definition: TokenTypeDictionary;
-    tokenize(text: string): LexerResult;
+    tokenize(text: string, options?: TokenizeOptions): LexerResult;
 }
 
 export class DefaultLexer implements Lexer {
@@ -52,7 +58,7 @@ export class DefaultLexer implements Lexer {
         return this.tokenTypes;
     }
 
-    tokenize(text: string): LexerResult {
+    tokenize(text: string, _options: TokenizeOptions = DEFAULT_TOKENIZE_OPTIONS): LexerResult {
         const chevrotainResult = this.chevrotainLexer.tokenize(text);
         return {
             tokens: chevrotainResult.tokens,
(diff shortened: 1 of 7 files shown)


martin-fleck-at:
@aabounegm @msujew FYI

msujew (commented):
Can you add a completion test to for this as well? Just to confirm that it really fixes the completion issue :)

msujew:
@martin-fleck-at By the way, after this change do you consider this feature to be "complete", at least for most common use cases? I would like to perform the release soon (ideally today, since I'm on a team event next week).

martin-fleck-at:
@msujew Yes, if that is working, I think we can consider this complete. I'd still love to try it out on our use case first though but I'll try to get this done today. Unfortunately, I am a bit struggling with creating the completion test case. Somehow, the tokens do not appear as I hoped they would but maybe it has something to do with the test function.

martin-fleck-at:
@msujew I pushed an update with the failing test cases and some other minor comments I received on https://github.com/eclipse-langium/langium/pull/1668. I'll quickly see if I can try the changes locally on my use case, i.e., whether it is really just a testing issue or the issue is not properly resolved.

packages/langium/src/parser/lexer.ts
```diff
@@ -25,9 +25,17 @@ export interface LexerResult {
     report?: LexingReport;
 }
 
+export type TokenizeMode = 'full' | 'partial';
+
+export interface TokenizeOptions {
+    mode: TokenizeMode;
```
msujew, line 31: Suggestion: With these kinds of option interfaces (especially if they're optional), their properties should be optional as well.
```suggestion
    mode?: TokenizeMode;
```

msujew:
@martin-fleck-at Looking into this, this seems like a more complicated issue. Essentially, in the position with index `1` the completion provider identifies that we're "within" an `INDENT` token and attempts to "complete" it. This of course doesn't yield anything and the test fails. The completion provider was written in the assumption that whitespace tokens are all hidden.

martin-fleck-at:
@msujew Yeah, I also had a feeling that something was off. Curiously, when I try that code (I published it locally to a verdaccio npm registry to consume it correctly without copying code), the completion provider I have in my project works. Without this code, it does not. That is why I was not sure whether it was something in the test code or not.

packages/langium/test/parser/indentation-aware.test.ts
```diff
@@ -389,6 +401,27 @@ describe('IndentationAware parsing', () => {
         expect(return2.value).toBe(true);
     });
 
+    test('should offer correct auto-completion parsing', async () => {
```
msujew, line 404:
```suggestion
    test.fails('should offer correct auto-completion parsing', async () => {
```

msujew (approved):
We can probably fix the outstanding completion issue later. See below. I think we can merge this.

martin-fleck-at:
Perfect, will push an update in a second. Actually, found another minor issue in the indentation token builder where I used a 0-based column index instead of a 1-based (chevrotain).

msujew:
> the completion provider I have in my project works

Note that in our case, the completion provider fails because it's being triggered right after a `INDENT` or `DEDENT` token. If you trigger it after a different token, the change you did makes it behave as expected. That's why I'm approving this in the first place. See also https://github.com/eclipse-langium/langium/discussions/1653#discussioncomment-10562312.

Issue #1645, opened 2024-08-25
Comment not extracted from first alternative


ydaveluy:
When defining a grammar with some alternatives, the first alternative comment is not properly extracted

Langium version: 3.1
Package name: langium

## Steps To Reproduce

1. Define this rule in the grammar:
```
Test returns string:
    /** comment 1 */ 'A' | /** comment 2 */  'B' | /** comment 3 */ 'C';
```
2. Check the generated grammar.ts
```
...
    {
      "$type": "ParserRule",
      "name": "Test",
      "dataType": "string",
      "definition": {
        "$type": "Alternatives",
        "elements": [
          {
            "$type": "Keyword",
            "value": "A"
          },
          {
            "$type": "Keyword",
            "value": "B",
            "$comment": "/**comment 2 */"
          },
          {
            "$type": "Keyword",
            "value": "C",
            "$comment": "/** comment 3 */"
          }
        ]
      },
      "definesHiddenTokens": false,
      "entry": false,
      "fragment": false,
      "hiddenTokens": [],
      "parameters": [],
      "wildcard": false
    },
...
```

## The current behavior
comment 1 is not extracted from the grammar.
comment 2 and comment 3 are properly extracted. 

## The expected behavior
comment 1 should be extracted from the grammar.

```
...
    {
      "$type": "ParserRule",
      "name": "Test",
      "dataType": "string",
      "definition": {
        "$type": "Alternatives",
        "elements": [
          {
            "$type": "Keyword",
            "value": "A",
            "$comment": "/**comment 1 */"
          },
          {
            "$type": "Keyword",
            "value": "B",
            "$comment": "/**comment 2 */"
          },
          {
            "$type": "Keyword",
            "value": "C",
            "$comment": "/** comment 3 */"
          }
        ]
      },
      "definesHiddenTokens": false,
      "entry": false,
      "fragment": false,
      "hiddenTokens": [],
      "parameters": [],
      "wildcard": false
    },
...
```

PR #1674, opened 2024-09-11
Fix the CST after performing assigned actions

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1645

We accidentally never actually removed the CST node that was added before executing the action. This led to CST nodes being available in the CST multiple times and accidentally shadowing some other nodes. In the case of the linked issue, the erroneous CST node shadowed a comment node.


ydaveluy:
Looks great !

spoenemann (approved):

PR #1679, opened 2024-09-15
Expose workspace folders via a readonly getter

aabounegm:
I could not figure out how to get information about the current workspace roots (important in a multi-root workspace), so I added a getter for it.

If we want to enforce that it is only accessed after `ready` is resolved, we could have `ready` resolve to it and remove the getter.

### DIFF
--- a/packages/langium/src/workspace/workspace-manager.ts
+++ b/packages/langium/src/workspace/workspace-manager.ts
@@ -36,6 +36,12 @@ export interface WorkspaceManager {
     readonly ready: Promise<void>;
 
     /**
+     * The workspace folders of the current workspace.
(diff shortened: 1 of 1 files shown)


aabounegm:
Specifically, the reason I needed this is these commits: https://github.com/SnakeskinTpl/language-tools/pull/8/files/7f54af2eaffee9085209975609521fe0db86b9b9..98e2571ba34e1dcbf1ad30412f8752cd5c160a67

If there is a better way to implement multi-root workspace support (with separate config files in each) using existing APIs, please let me know.

msujew (approved):
Yep, makes sense to expose the workspace folders on the manager. Thanks 👍

PR #1685, opened 2024-09-17
Specify `LINE_BREAKS` on tokens with custom patterns

aabounegm:
A warning was printed by Chevrotain for keyword tokens using a custom matcher function for the pattern.
While the default implementation does not use a custom matcher function, it is possible to override `buildKeywordPattern` to return one.

### DIFF
--- a/packages/langium/src/parser/token-builder.ts
+++ b/packages/langium/src/parser/token-builder.ts
@@ -131,13 +131,17 @@ export class DefaultTokenBuilder implements TokenBuilder {
 
     protected buildKeywordToken(keyword: Keyword, terminalTokens: TokenType[], caseInsensitive: boolean): TokenType {
         const keywordPattern = this.buildKeywordPattern(keyword, caseInsensitive);
-
-        return {
+        const tokenType: TokenType = {
             name: keyword.value,
             PATTERN: keywordPattern,
-            LINE_BREAKS: typeof keywordPattern === 'function',
             LONGER_ALT: this.findLongerAlt(keyword, terminalTokens)
         };
+
+        if (typeof keywordPattern === 'function') {
+            tokenType.LINE_BREAKS = true;
+        }
+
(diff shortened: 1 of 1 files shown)


packages/langium/src/parser/token-builder.ts
```diff
@@ -130,9 +130,12 @@ export class DefaultTokenBuilder implements TokenBuilder {
     }
 
     protected buildKeywordToken(keyword: Keyword, terminalTokens: TokenType[], caseInsensitive: boolean): TokenType {
+        const keywordPattern = this.buildKeywordPattern(keyword, caseInsensitive);
+
         return {
             name: keyword.value,
-            PATTERN: this.buildKeywordPattern(keyword, caseInsensitive),
+            PATTERN: keywordPattern,
+            LINE_BREAKS: typeof keywordPattern === 'function',
```
msujew, line 138: Suggestion: This doesn't really work out. By default, Chevrotain will attempt to identify whether the Regex/string passed into the `PATTERN` is capable of matching line break characters. This obviously cannot be done for functions. However, by automatically passing false for any non-function token pattern, we prevent this automatic identification from executing.

The suggestion would be to only set `LINE_BREAKS` to `true` if it is a function, otherwise keep the field undefined. And I don't mean setting it to `undefined` explicitly, see:

https://github.com/eclipse-langium/langium/blob/fa9f405fd7574e518481f74b22f736e01397967f/packages/langium/src/parser/token-builder.ts#L93-L96

aabounegm: Fixed 👍

msujew (approved):
Thanks 👍

While we're at it, do you think we can do the same for the terminal tokens?

aabounegm:
Sure thing

msujew (approved):
Perfect, thank you :)

Issue #1687, opened 2024-09-19
Add an option to suppress Chevrotain warnings


aabounegm:
Sometimes, when the grammar is too complicated, Chevrotain would print to the console warnings like this:
```
Ambiguous Alternatives Detected: <0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1> in <OPTION2> inside <Tag​> Rule
```
Which are not always very informative, but also mainly they flood the user's console. While they might signal real issues with the grammar, sometimes these ambiguities are unavoidable.
It would be great if there is an option to turn these warnings off, at least in production mode.

aabounegm:
@dhuebner I remember you said you had a project in which you were able to suppress these warnings. Can you please share the workaround (temporarily, until it is implemented in Langium)?

dhuebner:
@aabounegm
Sure. In the [BBj LS project](https://github.com/BBx-Kitchen/bbj-language-server/) we suppress all this ambiguity warnings and just report it once in debug log. The solution is a bit hacky, so I think we should have an easier nook to customize the behavior.

[Here is the parser adjustment.](https://github.com/BBx-Kitchen/bbj-language-server/blob/b7d1dd6f140dcd4428cdcbc73d248e9b7c8529a4/bbj-vscode/src/language/bbj-module.ts#L92)

msujew:
See https://github.com/eclipse-langium/langium/pull/1688 for a fix for this.

aabounegm:
@dhuebner Thanks a lot! That should help at least until the PR above is merged and released

PR #1688, opened 2024-09-19
Perform parser optimizations in production mode

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1687

Improves the `production` mode setting for the langium CLI. It's now generated into the language meta data and is used to skip validations and ignore ambiguity issues in the parser if activated.

This should improve the startup time of all grammars and is quite noticeable on larger grammars.

### DIFF
--- a/packages/langium-cli/src/generator/module-generator.ts
+++ b/packages/langium-cli/src/generator/module-generator.ts
@@ -5,7 +5,7 @@
  ******************************************************************************/
 
 import type { Grammar, IParserConfig } from 'langium';
-import { EOL, type Generated, expandToNode, joinToNode, toString } from 'langium/generate';
+import { type Generated, expandToNode, joinToNode, toString } from 'langium/generate';
 import type { LangiumConfig, LangiumLanguageConfig } from '../package-types.js';
 import { generatedHeader } from './node-util.js';
 
@@ -42,13 +42,14 @@ export function generateModule(grammars: Grammar[], config: LangiumConfig, gramm
             grammarsWithName,
             grammar => {
                 const config = grammarConfigMap.get(grammar)!;
-                const modeValue = mode ? `,${EOL}    mode: '${mode}'` : '';
+                const modeValue = mode === 'production' ? mode : 'development';
                 return expandToNode`
 
                     export const ${ grammar.name }LanguageMetaData = {
                         languageId: '${config.id}',
                         fileExtensions: [${config.fileExtensions && joinToNode(config.fileExtensions, e => appendQuotesAndDot(e), { separator: ', ' })}],
-                        caseInsensitive: ${Boolean(config.caseInsensitive)}${modeValue}
+                        caseInsensitive: ${Boolean(config.caseInsensitive)},
+                        mode: '${modeValue}'
                     } as const satisfies LanguageMetaData;
                 `;
             },
--- a/packages/langium-cli/src/generator/module-generator.ts
+++ b/packages/langium-cli/src/generator/module-generator.ts
@@ -5,13 +5,14 @@
  ******************************************************************************/
 
 import type { Grammar, IParserConfig } from 'langium';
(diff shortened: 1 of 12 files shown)


packages/langium-cli/src/generator/module-generator.ts
```diff
@@ -41,12 +42,13 @@ export function generateModule(grammars: Grammar[], config: LangiumConfig, gramm
...
             grammar => {
                 const config = grammarConfigMap.get(grammar)!;
+                const modeValue = mode ? `,${EOL}    mode: '${mode}'` : '';
                 return expandToNode`
 
                     export const ${ grammar.name }LanguageMetaData = {
                         languageId: '${config.id}',
                         fileExtensions: [${config.fileExtensions && joinToNode(config.fileExtensions, e => appendQuotesAndDot(e), { separator: ', ' })}],
-                        caseInsensitive: ${Boolean(config.caseInsensitive)}
+                        caseInsensitive: ${Boolean(config.caseInsensitive)}${modeValue}
```
aabounegm, line 51: This and the definition of `modeValue` are a bit unreadable. How about something like this:
```suggestion
                        caseInsensitive: ${Boolean(config.caseInsensitive)},
                        mode: '${mode ?? 'development'}',
```
or factor the `?? 'development'` part out to `modeValue`

packages/langium/src/languages/language-meta-data.ts
```diff
@@ -4,8 +4,17 @@
...
+ * Metadata of a language.
+ */
 export interface LanguageMetaData {
     languageId: string;
     fileExtensions: readonly string[];
     caseInsensitive: boolean;
+    /**
+     * Mode used to optimize code for development or production environments.
+     * 
+     * In production mode, all lexer/parser validations are disabled.
```
aabounegm, line 17: Maybe clarify that this refers specifically to the Chevrotain lexer/parser validations?
As I understand, this doesn't apply to the validations introduced in #1668

aabounegm (commented):
Looks good, but as I understand, this needs to be explicitly configured in `langium-config.json`, right?
Perhaps it is possible to add a default value in `loadConfig` like so?
```ts
config.mode ??= process.env.NODE_ENV ?? 'development';
```

Such that warning would be present in dev mode (running with `npm run langium:watch`), but not in the production build

msujew:
@aabounegm Thanks for taking a look!

> Looks good, but as I understand, this needs to be explicitly configured in langium-config.json, right?

You can also use the `--mode=production` CLI argument to switch between development and production mode. I'm not a big fan of using env variables for that.

aabounegm:
Ok that makes sense, I wasn't aware of the CLI option. In that case, maybe there should be an explicit `--mode=production` in the `langium:generate` scripts?

msujew:
There already is:

https://github.com/eclipse-langium/langium/blob/335f9a99b1568829cd1c0f5d3bbe33ca754daf90/packages/langium/package.json#L62

It's invoked before publishing the Langium artifacts.

aabounegm:
Sorry, I meant in the template so that new projects would make use of it

spoenemann:
How would we handle this when we publish the npm packages and vscode extension? Shall we re-generate the grammar language with production mode before publishing?

spoenemann (approved):
Oh I see, that's what we already do: https://github.com/eclipse-langium/langium/blob/335f9a99b1568829cd1c0f5d3bbe33ca754daf90/.github/workflows/publish.yml#L28

It's a bit strange to change a generated file that is also in version control during the release process, but this is likely the most pragmatic solution for now.

cdietrich-everest:
@msujew will there a preview/alpha available for this feature

cdietrich:
@msujew will there a preview/alpha available for this feature

msujew:
@cdietrich I can release one after tomorrow's dev meeting. I'll let you know.

PR #1689, opened 2024-09-20
Fix missing group proposal

### LINKED ISSUE #1677: Proposal missing from the content assist
Hey !

I probably found a minor issue in the content assist in case of multiple keyword alternatives. It seems that if the previous token is of a keyword kind, this keyword is not proposed by the content assist.

Langium version: 3.1
Package name: Langium

## Steps To Reproduce

1. Open this  [playground](https://langium.org/playground?grammar=OYJwhgthYgBAEgUwDbIPYHU0mQEwFD4AWAlrrogHawAuiIEJlYysGAygFywD0AOgGcA1DwDc%2BOgyYtYASQAi3HgG0A%2BmAC0ALwCCGgFoBdPgHcAVGMLEyFapMbNWAWQAyqgMIB5J04CiAOQAVJT5%2BM2VBPnZDMwB%2BPjNQy1JyKlp6Bxl2Ny8fAODeUNDlAD0%2BSj4QGMt8KhoQAE9YX0p6hs58WC7YAApO7oGAcgEaAFdbGkHYEfG64QBeBVgAH36BrsG6MABjInoprd36BYU12ABKM3xRIA&content=FAZwLgrgJgpgdmABAIWGGBDAxgCxgJ0QEFgg) 
2. Ctrl+space on line 4

## The current behavior
The content assist propose to create a new 'student' but not an other 'teacher'. After a 'student', the content assist propose to create a 'teacher' but not an other 'student'.

## The expected behavior
The content assist should propose a 'student' and a 'teacher' in all cases.

ydaveluy:
This PR fix #1677.

Add an additional test case and implement the fix.

### DIFF
--- a/packages/langium/src/lsp/completion/follow-element-computation.ts
+++ b/packages/langium/src/lsp/completion/follow-element-computation.ts
@@ -52,8 +52,6 @@ function findNextFeaturesInternal(options: { next: NextFeature, cardinalities: M
     const feature = next.feature;
     if (visited.has(feature)) {
         return [];
-    } else {
-        visited.add(feature);
     }
     let parent: ast.Group | undefined;
     let item = feature;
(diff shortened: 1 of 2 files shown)


msujew (commented):
Thanks for taking a look at this @ydaveluy. I can verify that the fix works, but I'll need a bit more time to investigate whether it breaks anything - adding elements to the set was supposed to break infinite loops in the completion calculation, but I'm not able to reproduce those anymore.

PR #1697, opened 2024-09-23
API docs

aabounegm:
Adds automatic documentation generation using [TypeDoc](https://typedoc.org/), which gets published to GitHub Pages on release.
I did not implement version management yet. Currently, only the latest version of the docs is available.

A preview of the website can be found here for now: https://aabounegm.github.io/langium/

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,15 +6,12 @@ permissions:
 on:
   release:
     types: [published]
-  # To manually trigger a rebuild of the API docs if needed
-  workflow_dispatch:
 
 jobs:
   publish:
     name: Langium Publish
     runs-on: ubuntu-latest
     timeout-minutes: 20
-    if: ${{ github.event_name == 'release' }}
     steps:
     - name: Checkout
       uses: actions/checkout@v3
@@ -59,31 +56,3 @@ jobs:
         PACKAGE_VERSION=`npm pkg get version --workspaces=false | tr -d \"`
         vsce package
         ovsx publish langium-vscode-$PACKAGE_VERSION.vsix -p ${{ secrets.OVSX_TOKEN }}
-
-  docs:
-    name: Build and publish API docs
-    runs-on: ubuntu-latest
-    permissions:
-      pages: write      # to deploy to Pages
-      id-token: write   # to verify the deployment originates from an appropriate source
-    environment:
-      name: github-pages
-      url: ${{ steps.deployment.outputs.page_url }}
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-      - name: Use Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: '18'
-      - name: Build API docs
-        run: |
-          npm ci
-          npm run docs -w langium
-      - name: Upload docs artifact
-        uses: actions/upload-pages-artifact@v3
-        with:
-          path: packages/langium/docs
-      - name: Deploy to GitHub Pages
-        id: deployment
-        uses: actions/deploy-pages@v4
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,12 +6,15 @@ permissions:
 on:
   release:
     types: [published]
+  # To manually trigger a rebuild of the API docs if needed
+  workflow_dispatch:
 
 jobs:
   publish:
     name: Langium Publish
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    if: ${{ github.event_name == 'release' }}
     steps:
     - name: Checkout
       uses: actions/checkout@v3
@@ -56,3 +59,31 @@ jobs:
         PACKAGE_VERSION=`npm pkg get version --workspaces=false | tr -d \"`
         vsce package
         ovsx publish langium-vscode-$PACKAGE_VERSION.vsix -p ${{ secrets.OVSX_TOKEN }}
+
+  docs:
+    name: Build and publish API docs
+    runs-on: ubuntu-latest
+    permissions:
+      pages: write      # to deploy to Pages
+      id-token: write   # to verify the deployment originates from an appropriate source
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18'
+      - name: Build API docs
+        run: |
+          npm ci
+          npm run docs -w langium
+      - name: Upload docs artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: packages/langium/docs
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
--- a/packages/langium/src/generate/generator-node.ts
+++ b/packages/langium/src/generate/generator-node.ts
@@ -14,7 +14,7 @@ export const EOL = (typeof process === 'undefined') ? '\n' : (process.platform =
 
 /**
  * Common type of expected results of functions contributing to code generation.
- * Includes `undefined` for allowing contributing functions to explicitely contribute
+ * Includes `undefined` for allowing contributing functions to explicitly contribute
  * nothing, if required, in contrast to contributing empty strings,
  * which facilitates better formatting of the desired output, for example.
  */
--- a/packages/langium/src/index.ts
+++ b/packages/langium/src/index.ts
@@ -2,7 +2,9 @@
  * Copyright 2021 TypeFox GmbH
  * This program and the accompanying materials are made available under the
  * terms of the MIT License, which is available in the project root.
- ******************************************************************************/
+ *
+ * @module langium
+ */
 
 export * from './default-module.js';
 export * from './dependency-injection.js';
--- a/packages/langium/src/parser/async-parser.ts
+++ b/packages/langium/src/parser/async-parser.ts
@@ -16,8 +16,8 @@ import { Emitter } from '../utils/event.js';
 /**
  * Async parser that allows cancellation of the current parsing process.
  *
- * @remark The sync parser implementation is blocking the event loop, which can become quite problematic for large files.
- * @remark The default implementation is not actually async. It just wraps the sync parser in a promise. A real implementation would create worker threads or web workers to offload the parsing work.
+ * The sync parser implementation is blocking the event loop, which can become quite problematic for large files.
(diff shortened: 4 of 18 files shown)


packages/langium/src/index.ts
```diff
@@ -2,7 +2,9 @@
  * Copyright 2021 TypeFox GmbH
  * This program and the accompanying materials are made available under the
  * terms of the MIT License, which is available in the project root.
- ******************************************************************************/
```
aabounegm, line 5: It seems that TypeDoc was not able to parse this comment ending properly, so I reduced it to the normal `*/`. If this is a big deal, I can create an issue and wait until it gets resolved.

aabounegm: FYI, the `@module` tag is used to rename modules in this tree
Screenshot 2024-09-25 at 5 23 09 PM

By default, it would use the module folder name directly (so without the `langium/` prefix), and `index` for the main entry point. I thought this could be a little confusing and renamed them to correspond to the package exports.

borisbrodski:
Looks great!

.github/workflows/publish.yml
```diff
@@ -6,12 +6,15 @@ permissions:
...
     types: [published]
+  # To manually trigger a rebuild of the API docs if needed
+  workflow_dispatch:
 
 jobs:
   publish:
     name: Langium Publish
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    if: ${{ github.event_name == 'release' }}
```
montymxb, line 17: I don't know if we want to constrain the publish job like this. We didn't need it before, and I think we may want to run this under different circumstances than this constraint would allow.

packages/langium/src/generate/generator-node.ts
```diff
@@ -14,7 +14,7 @@ export const EOL = (typeof process === 'undefined') ? '\n' : (process.platform =
 
 /**
  * Common type of expected results of functions contributing to code generation.
- * Includes `undefined` for allowing contributing functions to explicitely contribute
+ * Includes `undefined` for allowing contributing functions to explicitly contribute
```
montymxb, line 17: 👍

packages/langium/src/index.ts
montymxb: Yeah both sound fine to me, interesting that typedoc couldn't pick up the longer closing block comment.

packages/langium/src/parser/async-parser.ts
```diff
@@ -16,8 +16,8 @@ import { Emitter } from '../utils/event.js';
 /**
  * Async parser that allows cancellation of the current parsing process.
  *
- * @remark The sync parser implementation is blocking the event loop, which can become quite problematic for large files.
- * @remark The default implementation is not actually async. It just wraps the sync parser in a promise. A real implementation would create worker threads or web workers to offload the parsing work.
```
montymxb, line 20: Was there a reason for removing the remark annotations?

packages/langium/src/parser/async-parser.ts
```diff
@@ -27,15 +27,15 @@ export interface AsyncParser {
...
-     * @throw `OperationCancelled` if the parsing process is cancelled.
+     * @throws `OperationCancelled` if the parsing process is cancelled.
      */
     parse<T extends AstNode>(text: string, cancelToken: CancellationToken): Promise<ParseResult<T>>;
 }
 
 /**
  * Default implementation of the async parser which simply wraps the sync parser in a promise.
  *
- * @remark A real implementation would create worker threads or web workers to offload the parsing work.
```
montymxb, line 38: Same here as above. If it's not an issue with TypeDoc it would be good to retain the remarks as they were.

packages/langium/src/validation/document-validator.ts
```diff
@@ -236,13 +236,13 @@ export function getDiagnosticRange<N extends AstNode>(info: DiagnosticInfo<N, st
...
+            return 1 satisfies typeof DiagnosticSeverity.Error;
         case 'warning':
-            return 2; // according to vscode-languageserver-types/lib/esm/main.js#DiagnosticSeverity.Warning
+            return 2 satisfies typeof DiagnosticSeverity.Warning;
         case 'info':
-            return 3; // according to vscode-languageserver-types/lib/esm/main.js#DiagnosticSeverity.Information
+            return 3 satisfies typeof DiagnosticSeverity.Information;
         case 'hint':
-            return 4; // according to vscode-languageserver-types/lib/esm/main.js#DiagnosticSeverity.Hint
+            return 4 satisfies typeof DiagnosticSeverity.Hint;
```
montymxb, line 245: Nice, at first I read this and wasn't sure about being so specific, but honestly it looks pretty good. My only thought would be to retain at least one comment (somewhere) denoting `according to vscode-languageserver-types/lib/esm/main.js#DiagnosticSeverity`. It may be obvious via the type, but retaining the information might help for others are looking at it later.

montymxb (commented):
Nice! I just have a few questions in here, but overall it's looking really good (and the preview by extension as well).

In particular I wanted to say thanks for cleaning up some of the old doc comments, especially with respect to params that are no longer present.

packages/langium/src/index.ts
aabounegm: Yup, perhaps I should report it as a bug

.github/workflows/publish.yml
aabounegm: Because we didn't have any other trigger before, but now I added the `workflow_dispatch` trigger to allow updating the docs manually even before a release. Now that I think about it, I'm not sure if this would actually ever be used

packages/langium/src/parser/async-parser.ts
aabounegm: They are not recognized by JSDoc, so TypeDoc was complaining about them

packages/langium/src/parser/async-parser.ts
aabounegm: It's just a warning, so we can ignore it, but I just didn't see a reason for it since it is not a valid JSDoc tag

packages/langium/src/validation/document-validator.ts
aabounegm: Specific is good :D
But it's mainly about type safety and readability, to guarantee that copy-pasting (and forgetting to change the number) errors won't happen.

> retain at least one comment

I don't understand why such a comment would be needed. `DiagnosticSeverity` comes directly from that package, which can be jumped to using a simple "go to definition" (<kbd>CTRL/⌘</kbd> + click</kbd>)

aabounegm:
> In particular I wanted to say thanks for cleaning up some of the old doc comments, especially with respect to params that are no longer present.

My pleasure, and that's another benefit of using TypeDoc, it emits warnings whenever there is a discrepancy between the docs and the implementation.

.github/workflows/publish.yml
montymxb: Hmm, yeah I'm still not sure about this one. @msujew thoughts here?

packages/langium/src/parser/async-parser.ts
montymxb: Interesting, hmm, could you tweak it a bit to be `remark: ` instead? It would be nice to retain the intent.

packages/langium/src/validation/document-validator.ts
montymxb: Good comments can be surprisingly helpful, even when they appear redundant. The types are also excellent, but I would always error on the side of being verbose here. To be more specific, the comment _reinforces_ the developer's intent, and the type _agrees_ with that intent. This can make it easier to check later on if something changes (or needs to change).

packages/langium/src/parser/async-parser.ts
aabounegm: Alright, fair enough

packages/langium/src/validation/document-validator.ts
aabounegm: Sounds good. The function was even missing a JSDoc comment, so it seems like a good place to add it

.github/workflows/publish.yml
msujew: I think it makes more sense to have a separate `workflow_dispatch` job for the API docs publishing - right now, the release publishing process is broken due to expired credentials.

aabounegm: By "separate job", do you mean an entirely separate .yml file or just the additional trigger here?

msujew: A separate yml file :)

aabounegm: done

packages/langium/src/validation/document-validator.ts
montymxb: Perfect, thank you 😄

montymxb (approved):
:octocat: nice this looks great. Thanks a ton @aabounegm ! I have no problems on my end with bringing this in now.

The other question is when we should run the docs, ex. waiting for the next release or the current state? But that's less of an issue, either way we get hosted docs 😎 .

aabounegm:
My pleasure, hopefully this will help newcomers (including me 😄) explore Langium, and also make linking to relevant functionality easier.

> The other question is when we should run the docs, ex. waiting for the next release or the current state?

The workflow has both triggers for this reason precisely 😁. I think it makes sense to publish it right away for v3.2, and after that it should be automatically updated with each new release.

> either way we get hosted docs 😎

Yes we do! 😎

packages/langium/src/parser/async-parser.ts
Yokozuna59: It's possible to define your own tags, see https://typedoc.org/guides/tags/#defining-tags.
Here is [an example](https://typedoc.org/guides/tags/#block-tags) that introduce `@remarks` tag.

aabounegm: Oh, I didn't notice that part of the docs. In that case, we can either define a custom tag or simply use `@remarks` which is already recognized by TypeDoc (though there can only be one per comment). I personally vote for the latter, what do you think?

aabounegm: Actually, I just noticed now that `@remarks` is already used in `lsp/node-kind-provider.ts`, so I'm guessing `@remark` in async parser was just a typo?

montymxb: I would be on board with the latter, even if it's just a single `@remarks` 👍

aabounegm: Done ✅

msujew:
TypeDocs of 3.2.0 (ish) are now online at https://eclipse-langium.github.io/langium/ 🎉 

Thanks a lot for your work @aabounegm!

PR #1699, opened 2024-09-25
added failOnParseErrors option to validationHelper

cdietrich:


cdietrich:
question: does anybody know why we put the code here

```
data: {
      code: DocumentValidator.ParsingError,
    }
 ```   
    and not into
    
`code` at top level

File: packages/langium/src/test/langium-test.ts [hunk 1]
```diff
@@ -550,18 +551,29 @@ export interface ValidationResult<T extends AstNode = AstNode> extends AsyncDisp
     document: LangiumDocument<T>;
 }
 
-export function validationHelper<T extends AstNode = AstNode>(services: LangiumCoreServices): (input: string, options?: ParseHelperOptions) => Promise<ValidationResult<T>> {
+export type ValidationHelperOptions = ParseHelperOptions & { failOnParseErrors?: true };
```
spoenemann, line 554: Suggestion: Generalize the property type to `boolean` to enable forwarding the option from some other source (so we don't know whether it's set to true or false).

Other suggestion: rename to `failOnParsingErrors` to be consistent with the constant from DocumentValidator.

spoenemann (approved):

cdietrich:
my committer paperwork is still stuck somewhere ....

PR #1701, opened 2024-09-26
Improvements for cancelable operations

spoenemann:
- Use `performance.now()` for time measurements (see [Performance.now vs. Date.now](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#performance.now_vs._date.now))
- Use the `startCancelableOperation` function to reset the `lastTick` variable: in the current state, `interruptAndCheck` leads to unnecessary delays at the beginning of an operation because the internal variable used to compare the elapsed time is outdated. This change eliminates those unnecessary delays on document changes.

### DIFF
--- a/packages/langium/test/workspace/workspace-lock.test.ts
+++ b/packages/langium/test/workspace/workspace-lock.test.ts
@@ -59,10 +59,10 @@ describe('WorkspaceLock', () => {
 
     test('Write action results can be awaited', async () => {
         const mutex = new DefaultWorkspaceLock();
-        const now = Date.now();
+        const now = performance.now();
         const magicalNumber = await mutex.read(() => new Promise(resolve => setTimeout(() => resolve(42), 10)));
         // Confirm that at least 10ms have elapsed
-        expect(Date.now() - now).toBeGreaterThanOrEqual(10);
+        expect(performance.now() - now).toBeGreaterThanOrEqual(10);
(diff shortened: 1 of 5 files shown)


msujew (commented):
Shouldn't we use `performance.now()` here as well?

https://github.com/eclipse-langium/langium/blob/5be87adae9bc918b590dfdce479bd962f38e09cb/packages/langium/src/utils/promise-utils.ts#L34

packages/langium/test/workspace/workspace-lock.test.ts
```diff
@@ -59,10 +59,10 @@ describe('WorkspaceLock', () => {
 
     test('Write action results can be awaited', async () => {
         const mutex = new DefaultWorkspaceLock();
-        const now = Date.now();
+        const now = performance.now();
         const magicalNumber = await mutex.read(() => new Promise(resolve => setTimeout(() => resolve(42), 10)));
         // Confirm that at least 10ms have elapsed
-        expect(Date.now() - now).toBeGreaterThanOrEqual(10);
+        expect(performance.now() - now).toBeGreaterThanOrEqual(10);
```
msujew, line 65: The CI is failing here due to the higher precision:

image

We should decrease the value from 10 to something like 5 (which should be good enough).

msujew (approved):
LGTM 👍

PR #1708, opened 2024-10-02
Fix indentation tokens lengths

aabounegm:
Removed the duplication of indent tokens as whitespace, and ensured that dedent tokens are 0-width (empty string).
The dedent token could previously contain the matched whitespace of the following indentation (if it matched a previous indentation and didn't introduce a new indent token).

For example, in this snippet:
```python
if True:
    if False:
        print('impossible')
    else:
        print('makes sense')
```
the dedent token after the first `print` statement contained the whitespace before the `else` keyword

Additionally, fixed a minor issue (mostly inconvenience with folding) where the dedent tokens were placed after all newline tokens. For example, in the snippet:
```python
if True:
    print('yes')


else:
    print('no')
```
the dedent was placed on the 5th line, right before the `else` keyword. This PR moves it such that it is on the 3rd line.

### DIFF
--- a/packages/langium/src/parser/indentation-aware.ts
+++ b/packages/langium/src/parser/indentation-aware.ts
@@ -272,16 +272,7 @@ export class IndentationAwareTokenBuilder<Terminals extends string = string, Key
 
         this.indentationStack.push(currIndentLevel);
 
-        const indentToken = this.createIndentationTokenInstance(
-            this.indentTokenType,
-            text,
-            match?.[0] ?? '',
-            offset,
-        );
-        tokens.push(indentToken);
-
-        // Token already added, let the indentation now be consumed as whitespace and ignored
-        return null;
+        return match;
     }
 
     /**
--- a/packages/langium/src/parser/indentation-aware.ts
+++ b/packages/langium/src/parser/indentation-aware.ts
@@ -312,13 +312,14 @@ export class IndentationAwareTokenBuilder<Terminals extends string = string, Key
         }
 
         const numberOfDedents = this.indentationStack.length - matchIndentIndex - 1;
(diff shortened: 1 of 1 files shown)


packages/langium/src/parser/indentation-aware.ts
```diff
@@ -321,19 +312,20 @@ export class IndentationAwareTokenBuilder<Terminals extends string = string, Key
         }
 
         const numberOfDedents = this.indentationStack.length - matchIndentIndex - 1;
+        const newlinesBeforeDedent = text.substring(0, offset).match(/[\r\n]+$/)?.[0].length ?? 1;
```
msujew, line 315: I think this should not only backtrack through all the newline characters, but also whitespace. For example, if we dedent later:

```
{
    value
    <-- whitespace
    <-- whitespace
} <-- Dedent token appears at start of this line
```

This seems pretty weird, as if we remove the whitespace, the dedent already appears in the 3rd line. Since this PR is there to fix behavior such as the folding, I would've thought it should keep the `DEDENT` token in the 3rd line.

aabounegm: But if emit the `DEDENT` token on the beginning of the 3rd line, then it means that the whitespace after it will be detected as `INDENT`, and naturally followed by another `DEDENT` on the next line (and once more, for this example). These extra `INDENT`/`DEDENT` tokens are unexpected and will cause parsing errors.

I generally think that trailing whitespace is problematic and should be marked as a parsing error, especially in indentation-sensitive languages.

msujew: I see. I guess that makes sense. Thanks for the explanation 👍

msujew (approved):
Makes sense to me, thanks 👍

PR #1711, opened 2024-10-09
Remove undefined from joinTracedToNode signature

aabounegm:
Since it calls the function returned by `traceToNode`, which always returns `CompositeGeneratorNode`, it can never return `undefined`

### DIFF
--- a/packages/langium/src/generate/node-joiner.ts
+++ b/packages/langium/src/generate/node-joiner.ts
@@ -183,7 +183,7 @@ export function joinTracedToNode<T extends AstNode>(astNode: T, property?: Prope
  *   `.appendNewLine()
  */
 export function joinTracedToNode(sourceRegion: SourceRegion | undefined): // eslint-disable-next-line @typescript-eslint/indent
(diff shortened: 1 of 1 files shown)


packages/langium/src/generate/node-joiner.ts
```diff
@@ -167,7 +167,7 @@ export function joinTracedToNode<T extends AstNode>(astNode: T, property?: Prope
  *  arguments as expected by {@link joinToNode}, i.e. an `iterable`, a function `toGenerated`
  *  converting each element into a `Generated`, as well as some `options`.
  *
- * That function than joins the elements of `iterable` by delegating to {@link joinToNode}.
+ * That function then joins the elements of `iterable` by delegating to {@link joinToNode}.
```
msujew, line 170: Suggestion: The signature before this change also needs to be adjusted.

aabounegm: Right, seems I forgot it earlier

msujew (approved):
Thanks, looks good to me 👍

PR #1716, opened 2024-10-15
Add lexer error message provider

msujew:
Addresses an issue exposed in https://github.com/eclipse-langium/langium/discussions/1715. Currently, there's no easy way to override the lexer error message provider for chevrotain.

This change adds a service that will be used by the lexer for that.

### DIFF
--- a/packages/langium/src/parser/lexer.ts
+++ b/packages/langium/src/parser/lexer.ts
@@ -4,10 +4,21 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { ILexingError, IMultiModeLexerDefinition, IToken, TokenType, TokenTypeDictionary, TokenVocabulary } from 'chevrotain';
+import type { ILexerErrorMessageProvider, ILexingError, IMultiModeLexerDefinition, IToken, TokenType, TokenTypeDictionary, TokenVocabulary } from 'chevrotain';
 import type { LangiumCoreServices } from '../services.js';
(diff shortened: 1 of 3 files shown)


packages/langium/src/parser/lexer.ts
```diff
@@ -4,11 +4,22 @@
...
  ******************************************************************************/
 
-import type { ILexingError, IMultiModeLexerDefinition, IToken, TokenType, TokenTypeDictionary, TokenVocabulary } from 'chevrotain';
+import type { ILexerErrorMessageProvider, ILexingError, IMultiModeLexerDefinition, IToken, TokenType, TokenTypeDictionary, TokenVocabulary } from 'chevrotain';
 import type { LangiumCoreServices } from '../services.js';
-import { Lexer as ChevrotainLexer } from 'chevrotain';
+import { Lexer as ChevrotainLexer, defaultLexerErrorProvider } from 'chevrotain';
 import type { LexingReport, TokenBuilder } from './token-builder.js';
 
+export abstract class AbstractLexerErrorMessageProvider implements ILexerErrorMessageProvider {
```
aabounegm, line 12: Wouldn't it be more in line with the rest of the implementation (especially in the parser module) if this was instead:
```suggestion
export class DefaultLexerErrorMessageProvider implements ILexerErrorMessageProvider {
```
and then this was the class used in `src/default-module.ts` instead of using `defaultLexerErrorProvider` directly?

aabounegm (approved):

PR #1733, opened 2024-10-29
Support discarding the CST to preserve memory

msujew:
This (surprisingly small) change enables us to discard the CST on closed documents (in LSP use cases) to reduce our memory footprint. I've seen several crashes of Langium based language servers on large (> 100mb) workspaces, and this change should help alleviate those issues.

This change includes:
1. A new `Environment` service that identifies whether we're currently running as a language server. Running in a CLI environment usually requires full access to the CST.
2. A new `CstParserMode` that allows to discard parsed CSTs. The default is `Retain`, but `Discard` will be used when we encounter a closed document while running as a language server. This also goes in tandem with the new `discardCst` function that removes CSTs from ASTs that were parsed in `Retain` mode. Useful for documents that have been closed.
3. A new `$segments` property has been added for caching of CST data. In particular, it contains the full range of any AST node and the ranges for all properties. Ranges for keywords are not stored (yet).
4. This new property is used extensively in the LSP related services to find the correct positions of properties within closed files. Therefore, most services that attempt to find some specific position/range now use the `$segments` property.

Specific tests for the discard behavior will follow soon. This PR just shows the feasibility and backwards compatibility of the changes at hand so far.


packages/langium/src/workspace/ast-descriptions.ts
```diff
@@ -131,8 +130,8 @@ export class DefaultReferenceDescriptionProvider implements ReferenceDescription
 
     protected createDescription(refInfo: ReferenceInfo): ReferenceDescription | undefined {
         const targetNodeDescr = refInfo.reference.$nodeDescription;
-        const refCstNode = refInfo.reference.$refNode;
-        if (!targetNodeDescr || !refCstNode) {
+        const refSegment = refInfo.container.$segments?.properties[refInfo.property]?.[refInfo.index ?? 0];
```
danieldietrich, line 133: is the length of the refSegments array (last in the chain) always > 0 such that the fallback index 0 will succeed?

packages/langium/src/lsp/call-hierarchy-provider.ts
```diff
@@ -54,12 +54,12 @@ export abstract class AbstractCallHierarchyProvider implements CallHierarchyProv
             return undefined;
         }
 
-        const declarationNode = this.references.findDeclarationNode(targetNode);
+        const declarationNode = this.references.findDeclaration(targetNode);
```
danieldietrich, line 57: will this introduce a breaking change of public API?

packages/langium/src/references/name-provider.ts
```diff
@@ -21,9 +21,16 @@ export function isNamed(node: AstNode): node is NamedAstNode {
...
-     * @param node Specified `AstNode` whose name node shall be retrieved.
+     * @param node Specified `AstNode` whose name shall be retrieved.
      */
     getName(node: AstNode): string | undefined;
+
+    /**
+     * Returns the property name that is used to store the name of an AstNode.
+     * @param node The AstNode for which the name property shall be retrieved.
+     */
+    getNameProperty(node: AstNode): string | undefined;
```
danieldietrich, line 32: asking myself: why is the property name for storing the ast node name dynamic?

danieldietrich (commented):
Hi Mark,
I just did a quick 'scan' of the code and asked questions.
When I read "cache" my reflex is "and what about invalidation?", i.e. how do you ensure that the closed file contents don't deviate from the cached regions?
Daily reminder (and note to myself): It would be easier to have code formatting in a separate commit ;)
Thanks
Daniel

packages/langium/src/workspace/ast-descriptions.ts
msujew: Luckily, JS doesn't do bounds checking and simply returns `undefined` if the array is empty. Since we need to check for undefined anyway (see next line), this shouldn't lead to any issues.

packages/langium/src/references/name-provider.ts
msujew: Because some developers prefer not using `name` as the main property for giving an element a name. For example, some devs use `id` or something similar. The `NameProvider` in general is just there as a way for adopters to override how you retrieve the name property from an AST node.

packages/langium/src/lsp/call-hierarchy-provider.ts
msujew: We don't follow semantic versioning, since every change in an internal class (which we need to expose due to the override mechanism) would require a new major version.

msujew:
> When I read "cache" my reflex is "and what about invalidation?", i.e. how do you ensure that the closed file contents don't deviate from the cached regions?

As soon as the document is changed (either on disk, or via the `textDocument/open` notification), the language server will be notified by the language client and reparses the document. Reparsing the document automatically invalidates the cache, so we don't need to worry about desync in that regard.

packages/langium/src/workspace/ast-descriptions.ts
danieldietrich: Right! `[][-1]` also works in JS 🙃

danieldietrich:
> > When I read "cache" my reflex is "and what about invalidation?", i.e. how do you ensure that the closed file contents don't deviate from the cached regions?
> 
> As soon as the document is changed (either on disk, or via the `textDocument/open` notification), the language server will be notified by the language client and reparses the document. Reparsing the document automatically invalidates the cache, so we don't need to worry about desync in that regard.

Sounds all good! Thx

danieldietrich (approved):
(see my last review)

packages/langium/src/lsp/definition-provider.ts
```diff
@@ -67,21 +67,27 @@ export class DefaultDefinitionProvider implements DefinitionProvider {
...
-        if (goToLink) {
-            return [LocationLink.create(
-                goToLink.targetDocument.textDocument.uri,
-                (goToLink.target.astNode.$cstNode ?? goToLink.target).range,
-                goToLink.target.range,
-                goToLink.source.range
-            )];
+        if (goToLink && goToLink.target.$segments) {
+            const name = this.nameProvider.getNameProperty(goToLink.target);
+            if (name) {
```
spoenemann, line 72: Shall we use the whole composite node as target in case a name is not available?

packages/langium/src/lsp/document-update-handler.ts
```diff
@@ -141,4 +145,18 @@ export class DefaultDocumentUpdateHandler implements DocumentUpdateHandler {
...
     }
+
+    didCloseDocument(event: TextDocumentChangeEvent<TextDocument>): void {
+        const document = this.documents.getDocument(URI.parse(event.document.uri));
+        if (document) {
+            // Preserve memory by discarding the CST of the document
+            // Whenever the user reopens the document, the CST will be rebuilt
+            discardCst(document.parseResult.value);
+        }
+        // Discard the diagnostics for the closed document
```
spoenemann, line 156: Shall we make this configurable (without method overriding)? It could be simply a public property on the service class to turn this behavior on or off?

I imagine there are users/language implementors who would like to keep all problem markers even for closed documents.

packages/langium/src/parser/async-parser.ts
```diff
@@ -30,7 +30,7 @@ export interface AsyncParser {
      *
      * @throws `OperationCancelled` if the parsing process is cancelled.
      */
-    parse<T extends AstNode>(text: string, cancelToken: CancellationToken): Promise<ParseResult<T>>;
+    parse<T extends AstNode>(text: string, options: ParserOptions | undefined, cancelToken: CancellationToken): Promise<ParseResult<T>>;
```
spoenemann, line 33: The PR contains multiple "minor" breaking changes, but in sum I'd say the changes are significant enough to move this to v4.0. Shall we start planning the new major version more concretely?

packages/langium/src/syntax-tree.ts
```diff
@@ -23,6 +23,7 @@ export interface AstNode {
     readonly $containerIndex?: number;
     /** The Concrete Syntax Tree (CST) node of the text range from which this node was parsed. */
     readonly $cstNode?: CstNode;
+    readonly $segments?: AstNodeSegments;
```
spoenemann, line 26: Please add a comment explaining this new property.

packages/langium/src/validation/document-validator.ts
```diff
@@ -61,6 +61,9 @@ export class DefaultDocumentValidator implements DocumentValidator {
 
     async validateDocument(document: LangiumDocument, options: ValidationOptions = {}, cancelToken = CancellationToken.None): Promise<Diagnostic[]> {
         const parseResult = document.parseResult;
+        if (!parseResult.value.$cstNode) {
```
spoenemann, line 64: Why the early-exit here?

packages/langium/src/workspace/environment.ts
```diff
@@ -0,0 +1,51 @@
...
+ ******************************************************************************/
+
+import type { InitializeParams, InitializedParams } from 'vscode-languageserver-protocol';
+
+export interface EnvironmentInfo {
+    readonly isLanguageServer: boolean;
+    readonly locale: string;
+}
+
+export interface Environment extends EnvironmentInfo {
```
spoenemann, line 14: Do we really need this as a separate service? AFAICT the information is only used by the WorkspaceManager.

packages/langium/src/validation/document-validator.ts
msujew: Because the validator requires the CST to identify where to place the validation markers. If I want to place a diagnostic on a keyword, the validator doesn't know where to place it, because all information about that keyword was lost when we discarded the CST.

packages/langium/src/workspace/environment.ts
msujew: I feel like this could benefit other services as well and is well suited as a stand-alone service.

msujew:
@spoenemann I agree, this is something for a v4.0 of Langium. This PR isn't finished anyway - the formatter needs a refactoring as the new way of doing comments in the CST broke the formatter and I still haven't gotten to writing any tests for this.

packages/langium/src/validation/document-validator.ts
spoenemann: But at least the lexer and parser errors don't need the CST, so we should at least report those.

packages/langium/src/workspace/ast-descriptions.ts
```diff
@@ -53,15 +52,15 @@ export class DefaultAstNodeDescriptionProvider implements AstNodeDescriptionProv
         if (!name) {
             throw new Error(`Node at path ${path} has no name.`);
         }
-        let nameNodeSegment: DocumentSegment | undefined;
-        const nameSegmentGetter = () => nameNodeSegment ??= toDocumentSegment(this.nameProvider.getNameNode(node) ?? node.$cstNode);
+        const nameProperty = this.nameProvider.getNameProperty(node);
+        const nameSegment: DocumentSegment | undefined = nameProperty
+            ? node.$segments?.properties.get(nameProperty)[0]
+            : undefined;
```
spoenemann, line 58: The description providers are rather central services. I suggest to keep the CST-based segment resolution as fallback.

spoenemann (commented):
This change improves the memory usage for large workspaces. But it actually worsens the situation when you open a large file because the parser attaches both `$cstNode` and `$segments` to the AST.

Could we change that so the parser attaches either `$cstNode` or `$segments`, but not both? Then the `discardCst` function can take care of gathering the necessary information from the CST to create `$segments`. Of course that's more effort then, but I think it's acceptable given that it's done only when a document is closed by the user, so it can happen in the background without the user noticing it.

The LSP services would have to support both ways to obtain range information, but that might be extracted into utility functions to keep it simple.

msujew:
I don't think we'll move forward with this change. There are better ways to preserve memory in the workspace - which we will likely implement soon. Also, the general idea of generating a CST and then immediately throwing it away is simply just not good. See [Fastbelt](https://github.com/Typefox/fastbelt) for a better revised approach to problem.

Issue #1496, opened 2024-05-14
Support notebook document messages


msujew:
As of LSP version 3.17, [notebook messages](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notebookDocument_synchronization) have been added to the protocol. In a similar vein, the `vscode-languageserver` library has added a `NotebookDocuments` collection that corresponds to the `TextDocuments` (but for notebook cells). We should support notebook cells natively.

Currently all cell documents are handled via `TextDocuments`, but that isn't optimal.

PR #1743, opened 2024-11-06
Add normalizing notebook documents service

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1496.

This is a continuation of https://github.com/eclipse-langium/langium/pull/1660. Since the `NotebookDocuments` class exported by `vscode-languageserver` is quite specific in what it accepts as a `TextDocuments` instance (only instances of the class exported by `vscode-languageserver`), adopters of Langium, can no longer plug Langium's `TextDocuments` service in there.

However, the `NotebookDocuments` implementation of `vscode-languageserver` had the same URI encoding issues as `TextDocuments` anyway, and so we need to roll our own. This new `NormalizingNotebookDocuments` is now a default service of the shared LSP services and can be accessed by all Langium adopters.

### DIFF
--- a/packages/langium/src/lsp/default-lsp-module.ts
+++ b/packages/langium/src/lsp/default-lsp-module.ts
@@ -23,7 +23,7 @@ import { DefaultNodeKindProvider } from './node-kind-provider.js';
 import { DefaultReferencesProvider } from './references-provider.js';
 import { DefaultRenameProvider } from './rename-provider.js';
 import { DefaultWorkspaceSymbolProvider } from './workspace-symbol-provider.js';
-import { NormalizedTextDocuments } from './normalized-text-documents.js';
+import { NormalizedNotebookDocuments, NormalizedTextDocuments } from './normalized-text-documents.js';
 
 /**
  * Context required for creating the default language-specific dependency injection module.
@@ -96,7 +96,8 @@ export function createDefaultSharedLSPModule(context: DefaultSharedModuleContext
             FuzzyMatcher: () => new DefaultFuzzyMatcher(),
         },
         workspace: {
-            TextDocuments: () => new NormalizedTextDocuments(TextDocument)
+            TextDocuments: () => new NormalizedTextDocuments(TextDocument),
+            NotebookDocuments: () => new NormalizedNotebookDocuments(TextDocument)
         }
     };
 }
--- a/packages/langium/src/lsp/default-lsp-module.ts
+++ b/packages/langium/src/lsp/default-lsp-module.ts
@@ -97,7 +97,7 @@ export function createDefaultSharedLSPModule(context: DefaultSharedModuleContext
         },
         workspace: {
             TextDocuments: () => new NormalizedTextDocuments(TextDocument),
-            NotebookDocuments: () => new NormalizedNotebookDocuments(TextDocument)
+            NotebookDocuments: (services) => new NormalizedNotebookDocuments(services.workspace.TextDocuments)
         }
     };
 }
--- a/packages/langium/src/lsp/normalized-text-documents.ts
+++ b/packages/langium/src/lsp/normalized-text-documents.ts
@@ -6,11 +6,26 @@
 
 import type {
     Connection, DidOpenTextDocumentParams, DidChangeTextDocumentParams, DidCloseTextDocumentParams, TextDocumentsConfiguration, TextDocumentChangeEvent,
-    TextDocumentWillSaveEvent, RequestHandler, TextEdit, Event, WillSaveTextDocumentParams, CancellationToken, DidSaveTextDocumentParams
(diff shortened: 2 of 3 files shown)


cdietrich:
@msujew please let me know when we have an alpha for this. would like to test.

cdietrich:
thx will give it a test

packages/langium/src/lsp/normalized-text-documents.ts
```diff
@@ -253,3 +268,279 @@ export class NormalizedTextDocuments<T extends { uri: string }> implements TextD
...
+
+    private readonly notebookDocuments = new Map<DocumentUri, NotebookDocument>();
+    private readonly notebookCellMap = new Map<DocumentUri, [NotebookCell, NotebookDocument]>();
+
+    private readonly _onDidOpen = new Emitter<NotebookDocument>();
+    private readonly _onDidSave = new Emitter<NotebookDocument>();
+    private readonly _onDidChange = new Emitter<NotebookDocumentChangeEvent>();
+    private readonly _onDidClose = new Emitter<NotebookDocument>();
+
+    private readonly _cellTextDocuments: TextDocuments<T>;
```
cdietrich, line 306: the default impl seems to forward this to the normal textdocuments

packages/langium/src/lsp/normalized-text-documents.ts
```diff
@@ -253,3 +268,279 @@ export class NormalizedTextDocuments<T extends { uri: string }> implements TextD
...
+    listen(connection: Connection): Disposable {
+        const cellTextDocumentConnection = new CellTextDocumentConnection();
+        const disposables: Disposable[] = [];
+
+        disposables.push(this.cellTextDocuments.listen(cellTextDocumentConnection));
+        disposables.push(connection.notebooks.synchronization.onDidOpenNotebookDocument((params) => {
+            const uri = UriUtils.normalize(params.notebookDocument.uri);
+            this.notebookDocuments.set(uri, params.notebookDocument);
+            for (const cellTextDocument of params.cellTextDocuments) {
+                cellTextDocumentConnection.openTextDocument({ textDocument: cellTextDocument });
```
cdietrich, line 375: hmmmm. shoulnt this CellTextDocumentConnection work on _cellTextDocuments

cdietrich:
we do this in our service registry
     const textDocument = this.sharedServices.workspace.TextDocuments.get(uri.toString())

but 

this.sharedServices.workspace.TextDocuments.all()

is empty

msujew:
@cdietrich Thanks for looking into it. I'll take another look to see what's wrong. I probably missed something when adapting the Microsoft implementation.

cdietrich:
Then I ask for cell and document by cell I get one step more 
But then services crash with document already present 

need to figure out why

cdietrich:
hmmm no. when i use 

```
  let languageId: string | undefined = undefined
      const cell = this.sharedServices.workspace.NotebookDocuments.getNotebookCell(uri.toString())
      if (cell !== undefined) {
        languageId = this.sharedServices.workspace.NotebookDocuments.getCellTextDocument(cell)?.languageId
      } else {
        console.log("nooooo ceeeellll")
      }
```

in the service registry the cell map is still empty in the early stage (guard in fireDocumentUpdate)

cdietrich:
hmmm somehow not all docs end up in
_syncedDocuments

this is cause vscode sends only 4 of 9 cells. need to digg why

cdietrich:
looks like i somehow end up with multiple NormalizedTextDocs. need to figure out why.

cdietrich:
hmmmm
NotebookDocuments: () => new NormalizedNotebookDocuments(TextDocument)

packages/langium/src/lsp/default-lsp-module.ts
```diff
@@ -96,7 +96,8 @@ export function createDefaultSharedLSPModule(context: DefaultSharedModuleContext
             FuzzyMatcher: () => new DefaultFuzzyMatcher(),
         },
         workspace: {
-            TextDocuments: () => new NormalizedTextDocuments(TextDocument)
+            TextDocuments: () => new NormalizedTextDocuments(TextDocument),
+            NotebookDocuments: () => new NormalizedNotebookDocuments(TextDocument)
```
cdietrich, line 100: NotebookDocuments: (services) => new NormalizedNotebookDocuments(services.workspace.TextDocuments)

cdietrich: alternatively it seems to work to call
const textDocument = this.sharedServices.workspace.NotebookDocuments.cellTextDocuments.get(uri.toString())
in our service reg.

cdietrich: nope. the latter does not work neither

cdietrich: cool will try

cdietrich:
looks better now. we will some thorough testing tomorrow

cdietrich (approved):

PR #1760, opened 2024-11-26
Add optional AST node type

msujew:
The proposed `OptionalAstNode<T>` is a generic type that sets all properties, that could be potentially missing due to parser errors, to be optional. However, this excludes specificially:

* Properties inherited from `AstNode` (mainly the `$type` property which is always set).
* All `boolean` properties. Those will be set to `false` by the parser, even if the property wasn't parsed.
* All array property. Those will be set to `[]` by the parser.

This change adjusts the arithmetics language validator to use that new type to ensure that all properties are properly checked for truthyness before they are accessed. Inspired by the discussion in https://github.com/eclipse-langium/langium/discussions/1758.

### DIFF
--- a/packages/langium/src/syntax-tree.ts
+++ b/packages/langium/src/syntax-tree.ts
@@ -35,6 +35,11 @@ export interface GenericAstNode extends AstNode {
     [key: string]: unknown
 }
 
+export type OptionalAstNode<T extends AstNode> = {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    [P in keyof T]: P extends keyof AstNode ? T[P] : T[P] extends any[] | boolean ? T[P] : T[P] | undefined;
+};
+
 type SpecificNodeProperties<N extends AstNode> = keyof Omit<N, keyof AstNode | number | symbol>;
 
 /**
--- a/examples/arithmetics/src/language-server/arithmetics-util.ts
+++ b/examples/arithmetics/src/language-server/arithmetics-util.ts
@@ -4,7 +4,7 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { ResolvedReference } from 'langium';
+import type { OptionalAstNode, ResolvedReference } from 'langium';
 import { isDefinition, type BinaryExpression, type Definition, type FunctionCall } from './generated/ast.js';
 
 export function applyOp(op: BinaryExpression['operator']): (x: number, y: number) => number {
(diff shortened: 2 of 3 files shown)


packages/langium/src/syntax-tree.ts
```diff
@@ -35,6 +35,11 @@ export interface GenericAstNode extends AstNode {
     [key: string]: unknown
 }
 
+export type OptionalAstNode<T extends AstNode> = {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    [P in keyof T]: P extends keyof AstNode ? T[P] : T[P] extends any[] | boolean ? T[P] : T[P] | undefined;
```
sorawee, line 40: Thanks!

Would it make sense to make this recursive too? I believe a child node that we want to recur download is either `AstNode[]` or `AstNode | undefined` or `AstNode`. Do I miss anything else?

Here's my attempt to take a stab at it. (This is the first time in my life that I do type-level programming actually! Apologies if anything is off)

```
export type OptionalAstNode<T extends AstNode> = {
  [P in keyof T]: 
    P extends keyof AstNode ? 
      T[P] : 
      T[P] extends AstNode[] ? 
        OptionalAstNode<T[P][number]>[] : 
        T[P] extends any[] | boolean ? 
          T[P] : 
          T[P] extends AstNode | undefined ? 
            OptionalAstNode<Extract<T[P], AstNode>> | undefined :  
            (T[P] extends AstNode ? OptionalAstNode<T[P]> : T[P]) | undefined;
};
```

sailingKieler, line 40:
```suggestion
    [P in keyof T]: P extends keyof AstNode ? T[P] : T[P] extends unknown[] | boolean ? T[P] : T[P] | undefined;
```
... should do the trick, too, without any need for the linter exclusion.

sailingKieler: > Thanks!
> 
> Would it make sense to make this recursive too?

I think it does! 😀

msujew:
@sailingKieler @sorawee Making AST nodes deep optional, makes them very difficult to deal with. While technically correct on a type level, you cannot pass them into any method expecting a "normal" `AstNode`, due to type mismatch. This makes validations/scoping suddenly very difficult to write. I'll probably get back to this in the new year.

msujew:
I am not absolutely happy with this change. Adopters are free to use this type definition for themselves, but I don't think it's good enough to offer to all adopters as it is. Closing this for now.

PR #1766, opened 2024-12-03
Validation for multiple assignments takes rewrite actions into account now

### LINKED ISSUE #1756: False positive in grammar
Langium version: 3.3

given the grammar

```
entry Model:
    expr=Expression;

Expression: Equality;

Equality infers Expression:
    Literal (({infer Equals.left=current} '==' | {infer NotEquals.left=current} '!=')  right=Literal)*;

Literal: value=INT;
```

the validator complains about

```
src/language/hello-world.langium:9:88 - Found multiple assignments to 'right' with the '=' assignment operator. Consider using '+=' instead to prevent data loss.
```
imho this is a false positive with the ast rewrite. maybe it is not properly detected if happing on all branches of an alternative.

JohannesMeierSE:
Fixes #1756.

This PR contributes the following improvements for the validation for multiple assignments with `=` instead of `+=` (see #1412):

* unassigned actions don't create new objects, only rewrite actions create new objects
* respect looped rewrite actions, even when they are deeply nested or occur as top-level elements of rules
* handle asymmetric use of rewrite actions in alternatives

Beyond that, this PR ...

* simplified the validation logic
* improved error logging in test utils

### DIFF
--- a/packages/langium/src/grammar/validation/validator.ts
+++ b/packages/langium/src/grammar/validation/validator.ts
@@ -877,18 +877,13 @@ export class LangiumGrammarValidator {
     checkOperatorMultiplicitiesForMultiAssignments(rule: ast.ParserRule, accept: ValidationAcceptor): void {
         // for usual parser rules AND for fragments, but not for data type rules!
         if (!rule.dataType) {
-            this.checkOperatorMultiplicitiesForMultiAssignmentsLogic([rule.definition], accept);
+            this.checkOperatorMultiplicitiesForMultiAssignmentsIndependent([rule.definition], new Map(), accept);
         }
     }
 
-    private checkOperatorMultiplicitiesForMultiAssignmentsLogic(startNodes: AstNode[], accept: ValidationAcceptor): void {
-        // new map to store usage information of the assignments
-        const map: Map<string, AssignmentUse> = new Map();
-
-        // top-down traversal for all starting nodes
-        for (const node of startNodes) {
-            this.checkAssignmentNumbersForNode(node, 1, map, accept);
-        }
+    private checkOperatorMultiplicitiesForMultiAssignmentsIndependent(startNodes: AstNode[], map: Map<string, AssignmentUse>, accept: ValidationAcceptor): void {
+        // check all starting nodes
+        this.checkOperatorMultiplicitiesForMultiAssignmentsNested(startNodes, 1, map, accept);
 
         // create the warnings
         for (const entry of map.values()) {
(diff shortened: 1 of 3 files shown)


cdietrich:
checked with my usecase in the arithmetics grammar + in the extension host and it seems to work

packages/langium/test/grammar/grammar-validator.test.ts
```diff
@@ -1031,6 +1031,104 @@ describe('Assignments with = instead of +=', () => {
...
+    test('rewrite actions inside loop #1756 (only a single alternative creates a new object)', async () => {
+        // Since we assume the worst case and since 'right' is assigned twice, if the alternative without the rewrite action is used, we expect a warning here as well!
+        const validation = await validate(getGrammar(`
+            entry Model:
+                expr=Expression;
+
+            Expression: Equality;
+
+            Equality infers Expression:
+                Literal ( ( {infer Equals.left=current} '==' | '!=' )  right=Literal)*;
```
sailingKieler, line 1104: Very nice case, fully agree to bring up the warning here, as that indicates that the grammar design is probably sub optimal.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -877,18 +877,13 @@ export class LangiumGrammarValidator {
...
 
-    private checkOperatorMultiplicitiesForMultiAssignmentsLogic(startNodes: AstNode[], accept: ValidationAcceptor): void {
-        // new map to store usage information of the assignments
-        const map: Map<string, AssignmentUse> = new Map();
-
-        // top-down traversal for all starting nodes
-        for (const node of startNodes) {
-            this.checkAssignmentNumbersForNode(node, 1, map, accept);
-        }
+    private checkOperatorMultiplicitiesForMultiAssignmentsIndependent(startNodes: AstNode[], map: Map<string, AssignmentUse>, accept: ValidationAcceptor): void {
```
sailingKieler, line 884: Just a minor thing: You could flip the last two arguments and initialize `map` via a default value, like
```
..., accept: ValidationAcceptor, map: Map<string, AssignmentUse> = new Map()): void {
```
and skip the map creation in the call above.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -906,72 +901,79 @@ export class LangiumGrammarValidator {
...
 
-        // Search for assignments in used fragments as well, since their property values are stored in the current object.
-        // But do not search in calls of regular parser rules, since parser rules create new objects.
-        if (ast.isRuleCall(currentNode) && ast.isParserRule(currentNode.rule.ref) && currentNode.rule.ref.fragment) {
-            this.checkAssignmentNumbersForNode(currentNode.rule.ref.definition, currentMultiplicity, map, accept);
-        }
+            // the current element can occur multiple times => its assignments can occur multiple times as well
+            let currentMultiplicity = parentMultiplicity;
+            if (ast.isAbstractElement(currentNode) && isArrayCardinality(currentNode.cardinality)) {
+                currentMultiplicity *= 2; // note that the result is not exact (but it is sufficient for the current use case)!
```
sailingKieler, line 926: cool trick 😉

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -906,72 +901,79 @@ export class LangiumGrammarValidator {
...
+            }
 
-        // rewriting actions are a special case for assignments
-        if (ast.isAction(currentNode) && currentNode.feature) {
-            storeAssignmentUse(map, currentNode.feature, 1 * currentMultiplicity, currentNode);
-        }
+            // assignment
+            if (ast.isAssignment(currentNode)) {
+                storeAssignmentUse(map, currentNode.feature, 1 * currentMultiplicity, currentNode);
```
sailingKieler, line 931: what's the rational of this `1 * ...`, which has no actual effect?

sailingKieler (approved):
LVGTM! Just a few remarks and questions, nothing to be changed urgently. 🚀

sailingKieler:
And explicit kudos for your elaborate test cases!

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: done

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: It is a relict, in former versions I used `currentMultiplicity` multiple times with different multipliers, where `1 * ` was used to make this multiplier explicit. I removed it now.

JohannesMeierSE:
Thanks for your review @sailingKieler! In my eyes, this PR is ready to merge.

PR #1771, opened 2024-12-05
langium: follow-up on #972 contributing convenience type defs for cross-ref property distinction

sailingKieler:
* fixed a nasty bug in internal type 'ExtractKeysOfValueType' by adding '-?' to the result object key list
* improved documentation
* contributed a bunch of tests in 'syntax-tree.test.ts' which is effectively tested during the TypeScript compilation: test failures occur as compilation errors

### DIFF
--- a/packages/langium/test/syntax-tree.test.ts
+++ b/packages/langium/test/syntax-tree.test.ts
@@ -1,0 +1,260 @@
+/******************************************************************************
+ * Copyright 2024 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { describe, expectTypeOf, test } from 'vitest';
+import type { AstNode, AstNodeTypesWithCrossReferences, AstTypeList, CrossReferencesOfAstNodeType, Reference } from '../src/syntax-tree.js';
+
+describe('Utility types revealing cross-reference properties', () => {
+
+    /**
+     * The tests listed below don't check correctness of 'CrossReferencesOfAstNodeType'
+     *  and 'AstNodeTypesWithCrossReferences' by executing code and comparing results.
+     * Instead, they check if the types are correctly inferred by TypeScript.
+     *
+     * Hence, test failure are indicated by TypeScript errors, while the absence of any compile errors denotes success.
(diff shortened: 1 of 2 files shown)


packages/langium/test/syntax-tree.test.ts
```diff
@@ -0,0 +1,260 @@
...
+        OptionalMultiRef: OptionalMultiRef
+    }
+
+    const getAnyInstanceOfType = <T extends AstTypeList<T>>()  => <AstNodeTypesWithCrossReferences<T>>{};
+    const getCrossRefProps =     <T extends AstNode>(_type: T) => <CrossReferencesOfAstNodeType<T>>'';
+    const checkNever =           (props: never)                => props;
+
+    test('Should not reveal cross-ref properties for NoRefs', () => {
+        const props = getCrossRefProps(<NoRefs>{});
+        checkNever(props);
```
msujew, line 65: Suggestion: Instead using a custom `checkNever` function, you can use vitest's type testing utilities. See [their type testing guide](https://vitest.dev/guide/testing-types).
```suggestion
        expectTypeOf(props).toBeNever();
```

packages/langium/test/syntax-tree.test.ts
```diff
@@ -0,0 +1,260 @@
...
+                    }
+                }
+            }
+            default: {
+                checkNever(instance);
+                return checkNever(getCrossRefProps(instance));
+            }
+        }
+    });
+});
```
msujew, line 260: Formatting: Missing newline at EOF.

msujew (commented):
@sailingKieler As discussed yesterday during the dev-meeting, there actually is a way to explicitly test types by invoking the TypeScript compiler. This is done by vitest (see [type testing guide](https://vitest.dev/guide/testing-types)) when renaming the file to `*.test-d.ts`.

packages/langium/test/syntax-tree.test.ts
sailingKieler: Fixed.

packages/langium/test/syntax-tree.test.ts
sailingKieler: Switched to that API, it's actually quite cool! 💡

PR #1784, opened 2025-01-08
Extend file system provider interface

msujew:
Extends the file system provider to have the common file system operations:

1. Reading text file
2. Reading directory
3. Reading binary file
4. Checking whether a file exists

Both of these are now available in their respective sync and async versions. However, the framework itself exclusively uses the async methods. The sync method is available for adopters.

This change required to update the `@types/node` dev dependency to 18, as the `Buffer` class only started to correctly subtypethe `Uint8Array` class during that release.

### DIFF
--- a/packages/langium/src/node/node-file-system-provider.ts
+++ b/packages/langium/src/node/node-file-system-provider.ts
@@ -33,17 +33,51 @@ export class NodeFileSystemProvider implements FileSystemProvider {
         };
     }
 
+    exists(uri: URI): Promise<boolean> {
+        return new Promise(resolve => {
+            fs.stat(uri.fsPath, err => {
+                resolve(!err);
+            });
+        });
+    }
+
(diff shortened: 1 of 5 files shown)


msujew:
cc @cdietrich I believe you were interested in sync fs methods, right?

spoenemann (commented):
Since the new interface methods are not optional, we should regard this as a breaking change.

aabounegm:
I'm a bit late, but what if instead of a breaking change, this is split into 2 interfaces/modules/providers, one for sync and the other for async?

msujew:
> I'm a bit late, but what if instead of a breaking change, this is split into 2 interfaces/modules/providers, one for sync and the other for async?

We had something similar in Xtext, and it was fairly annoying to use. If adopters don't want to implement the sync methods, they can always opt to just stub it.

PR #1786, opened 2025-01-09
Watch changes on directories

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1230

Even though I expected this to be a small fix, this was surprisingly complicated. The main issue is that we don't get notified what is contained in the directory that has changed. Therefore, we need to "manually" dig through the changed directory. Performs a few changes:

1. Adds a `UriTrie` data structure. It's a tree structure for document storage that respects directories (compared to the previous flat map). It is required to quickly figure out which documents belong to a changed directory.
2. Adjusts the document builder to find all documents that belong either to a changed directory or a deleted directory.
3. Adds a few new methods to the workspace manager so we can efficiently traverse newly added folders.
4. Adds a new testing utility `VirtualFileSystemProvider` that can be programmed for specific tests.


packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -130,10 +144,15 @@ export class DefaultWorkspaceManager implements WorkspaceManager {
...
         await Promise.all(
-            folders.map(wf => [wf, this.getRootFolder(wf)] as [WorkspaceFolder, URI])
-                .map(async entry => this.traverseFolder(...entry, fileExtensions, collector))
+            folders.map(wf => this.getRootFolder(wf))
+                .map(async entry => this.traverseFolder(entry, uris))
         );
+        await Promise.all(uris.map(async uri => {
+            const document = await this.langiumDocuments.getOrCreateDocument(uri);
+            collector(document);
+        }));
```
spoenemann, line 155: This can be simplified now: instead of using a collector function, just map the uris to the created documents.

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -160,33 +179,37 @@ export class DefaultWorkspaceManager implements WorkspaceManager {
      * Traverse the file system folder identified by the given URI and its subfolders. All
      * contained files that match the file extensions are added to the collector.
```
spoenemann, line 180: Please update the comment (-collector, +uris)

spoenemann (commented):
Great solution! Let's merge this after the v3.5.0 release so we can work towards v4.0.0

packages/langium/src/workspace/workspace-manager.ts
spoenemann: Ok let's keep it as it is.

PR #1787, opened 2025-01-09
Prevent interruptAndCheck from delaying too often when delay takes to…

thejoecode:
…o long

I ran into an issue where Firefox setTimeout(resolve, 0) was taking the majority of the time between interruptAndCheck calls.

performance.now() - current will be greater than globalInterruptionPeriod so it gets called every iteration

As a work around I can greatly increase the globalinterruptionPeriod with something like setInterruptionPeriod(150) - but then other browsers that don't have the slower setTimeout callback will have slower cancellation responses.

Other fixes I considered: have delayNextTick return performance.now() only for setTimout path (code below) or letting users provide a function that gets called back with current and they can either return current or process.now()

```ts
export function delayNextTick(current: number): Promise<number> {
    return new Promise(resolve => {
        // In case we are running in a non-node environment, `setImmediate` isn't available.
        // Using `setTimeout` of the browser API accomplishes the same result.
        if (typeof setImmediate === 'undefined') {
            // there is no way to guarantee
            setTimeout(() => resolve(performance.now()), 0);
        } else {
            setImmediate(() => resolve(current));
        }
    });
}
```

### DIFF
--- a/packages/langium/src/utils/promise-utils.ts
+++ b/packages/langium/src/utils/promise-utils.ts
@@ -76,8 +76,11 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
     }
     const current = performance.now();
     if (current - lastTick >= globalInterruptionPeriod) {
-        lastTick = current;
         await delayNextTick();
+        // prevent calling delayNextTick every iteration of loop
+        // where delayNextTick takes up the majority or all of the
+        // globalInterruptionPeriod itself
+        lastTick = performance.now();
     }
     if (token.isCancellationRequested) {
         throw OperationCancelled;
--- a/packages/langium/src/utils/promise-utils.ts
+++ b/packages/langium/src/utils/promise-utils.ts
@@ -76,6 +76,7 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
     }
     const current = performance.now();
     if (current - lastTick >= globalInterruptionPeriod) {
+        lastTick = current;
         await delayNextTick();
         // prevent calling delayNextTick every iteration of loop
         // where delayNextTick takes up the majority or all of the


packages/langium/src/utils/promise-utils.ts
```diff
@@ -76,8 +76,11 @@ export async function interruptAndCheck(token: CancellationToken): Promise<void>
     }
     const current = performance.now();
     if (current - lastTick >= globalInterruptionPeriod) {
-        lastTick = current;
```
msujew, line 79: I think we still need this line of code. If we interrupt to let other code run (i.e. how it's intended to be used, not just waiting on FireFox to do it's thing), the newly called code will then also call `interruptAndCheck` which **immediately** results in another next tick delay. We don't want that, so we would need to update the `lastTick` before and after running the delay.

msujew (commented):
I think that makes sense (aside from a comment, see below). Thanks for bringing that to our attention :)

msujew (commented):
@thejoecode Quick question before going deeper into this: When using Langium in a browser, do you run it in a dedicated worker? I'm a bit confused on why FireFox is even interrupting the event loop that long. I'm driving FireFox as well and never had any issues with Langium in the browser yet.

packages/langium/src/utils/promise-utils.ts
thejoecode: Ah, that makes sense. Will update.

msujew (approved):
I believe that's a good change. @spoenemann what do you think?

msujew:
I was actually able to reproduce this in a project and it essentially makes the language server unusable. I'll provide a patch release soon that includes this change once we merge this.

This honestly seems like a FireFox bug, but I think it makes sense anyway to include this fix anyway.

msujew:
@spoenemann Is out sick. I've reconfirmed in a different project that this change fixes the issues experienced and does not seem to lead to any noticable regression. I'll create a new release after merging this.

cdietrich:
@msujew will the release include the nb changes?

msujew:
@cdietrich https://github.com/eclipse-langium/langium/pull/1743 hasn't been approved yet, but since it adds new API, I planned it for a (soonish) 3.4 anyway.

PR #1788, opened 2025-01-13
added `getStream()` to `MultiMap` in collection.ts, adjusted evaluation of precomputed scopes in `DefaultScopeProvider`

sailingKieler:
With this additional method hook, customized implementations of `PrecomputedScopes` contributing lazily computed proposals (for the sake of memory usage optimization) can be employed, and eager in-between evaluations can be avoided by overriding the new method hook `getStream(key)`.


msujew (commented):
Are you overriding `MultiMap` in your adopter code? This change feels like the wrong abstraction layer. I would much rather have a dedicated `PrecomputedScope` interface (with a default implementation based on `MultiMap`), than extend the data type based on a specific adopter need. You can then create your own, stream based `PrecomputedScope` implementation.

sailingKieler:
I introduced an
```
CustomPrecomputedScopes extends MultiMap<AstNode, AstNodeDescription> {

    private lazyComputations = new Map<AstNode, () => Stream<AstNodeDescription>>();
    
    ...

    public getStream(key: AstNode): Stream<AstNodeDescription> {
        const local = stream(super.get(key)); // to be replaced with `super.getStream(key)`
        const additional = this.lazyComputations.get(key);
        return local && additional ? local.concat(additional()) : local ?? additional?.() ?? stream();
    }
}
```

Indeed, from an outside point-of-view the additional method looks a bit weird, with the constraint in mind that the memory footprint of the precomputed scopes might get too high with a high number of documents this is the origin of the change https://github.com/eclipse-langium/langium/blob/80786815a56d0e145fbdffda63805bd1eca51745/packages/langium/src/references/scope-provider.ts#L59-L63

Motivation: Constraint-based visibility of external symbols per document.

packages/langium/src/references/scope-computation.ts
```diff
@@ -122,6 +123,19 @@ export class DefaultScopeComputation implements ScopeComputation {
...
+     */
+    protected newPrecomputedScopes(_document: LangiumDocument): PrecomputedScopes {
+        const map = new MultiMap<AstNode, AstNodeDescription>();
+        return {
+            add:       map.add.bind(map),
+            addAll:    map.addAll.bind(map),
+            get:       map.get.bind(map),
+            getStream: (key: AstNode) => stream(map.get(key))
+        };
+    }
```
msujew, line 137: Suggestion: I don't think that the "default implementation" of `PrecomputedScopes` should be exclusively available as a method inside of the default implementation of `ScopeComputation`. Instead, I would much rather have a `MultiMapPrecomputedScopes` class, that can be used by adopters that don't want to override `DefaultScopeComputation`.

msujew (commented):
Thanks, I think the new API looks much better. I just have one comment, see below.

packages/langium/src/references/scope-provider.ts
```diff
@@ -56,11 +56,11 @@ export class DefaultScopeProvider implements ScopeProvider {
...
-                const allDescriptions = precomputed.get(currentNode);
-                if (allDescriptions.length > 0) {
-                    scopes.push(stream(allDescriptions).filter(
-                        desc => this.reflection.isSubtype(desc.type, referenceType)));
-                }
+                scopes.push(
+                    precomputed.getStream(currentNode).filter(
+                        desc => this.reflection.isSubtype(desc.type, referenceType)
+                    )
+                );
```
spoenemann, line 63: This creates empty scopes, which was avoided in the previous version. The change might have a performance impact for larger documents. I suggest keeping a check for an empty stream.

packages/langium/src/workspace/documents.ts
```diff
@@ -95,10 +94,15 @@ export enum DocumentState {
...
+ * Result of the scope pre-computation phase (`ScopeComputation` service).
  * It maps every AST node to the set of symbols that are visible in the subtree of that node.
  */
-export type PrecomputedScopes = MultiMap<AstNode, AstNodeDescription>
+export interface PrecomputedScopes {
+    add(key: AstNode, value: AstNodeDescription): void
+    addAll(key: AstNode, values: AstNodeDescription[]): void
+    get(key: AstNode): readonly AstNodeDescription[]
+    getStream(key: AstNode): Stream<AstNodeDescription>
+}
```
spoenemann, line 105: This is a breaking change; we should defer the PR to v4.0.0.

spoenemann:
Proposal: rename `PrecomputedScopes` to `SymbolTable`. I think that would clear up the relation to the `ScopeProvider`, which computes a `Scope` from a collection of symbol tables.

packages/langium/src/references/scope-computation.ts
sailingKieler: I now introduced https://github.com/eclipse-langium/langium/blob/fb4558100002b91f0fc1d7c81632a80bd8ff2340/packages/langium/src/references/scope-computation.ts#L56-L68

packages/langium/src/references/scope-provider.ts
sailingKieler: I restored the original structure and added a check of the stream being`EMPTY_STREAM`, see https://github.com/eclipse-langium/langium/blob/fb4558100002b91f0fc1d7c81632a80bd8ff2340/packages/langium/src/references/scope-provider.ts#L59-L63

spoenemann (commented):
Hmm, IMO the confusion with the LSP concept is too problematic.

What about `SymbolMap` to keep it as simple as possible?

packages/langium/src/documentation/documentation-provider.ts
```diff
@@ -69,13 +69,13 @@ export class JSDocDocumentationProvider implements DocumentationProvider {
 
     protected findNameInPrecomputedScopes(node: AstNode, name: string): AstNodeDescription | undefined {
```
spoenemann: This method could be renamed.

packages/langium/src/grammar/references/grammar-scope.ts
```diff
@@ -55,12 +55,11 @@ export class LangiumGrammarScopeProvider extends DefaultScopeProvider {
...
         let localScope: Stream<AstNodeDescription> | undefined;
-        const precomputed = getDocument(context.container).precomputedScopes;
+        const localSymbols = getDocument(context.container).localSymbols;
         const rootNode = findRootNode(context.container);
-        if (precomputed && rootNode) {
-            const allDescriptions = precomputed.get(rootNode);
-            if (allDescriptions.length > 0) {
-                localScope = stream(allDescriptions).filter(des => des.type === Interface.$type || des.type === Type.$type || des.type === InferredType.$type);
+        if (localSymbols && rootNode) {
+            if (localSymbols.has(rootNode)) {
```
spoenemann: The nested if could be merged with the outer one.

packages/langium/src/workspace/documents.ts
```diff
@@ -95,10 +94,15 @@ export enum DocumentState {
 }
 
 /**
- * Result of the scope precomputation phase (`ScopeComputation` service).
- * It maps every AST node to the set of symbols that are visible in the subtree of that node.
+ * Result of the scope pre-computation phase performed by the `ScopeComputation` service.
+ * It maps AST nodes of a document to their corresponding sets of symbols that are accessible
+ * by those nodes/subtrees, provided any symbols corresponding specifically to those nodes/subtrees exist.
+ * The sets of symbols are assumed to be un-ordered. Hence, no assumptions about the order of symbols in the sets should be made.
```
spoenemann: I'd add that usually a MultiMap is used for collecting symbols, which conforms to this interface.

spoenemann (approved):
Thanks a lot! A few detail remarks below, but otherwise I think this is a good change.

> Do we wanna keep that? Renaming that one requires lot's of documentation updates, doesn't it?

I agree, we should keep the `ComputedScopes` document state and `ScopeComputation` service name to avoid unnecessary disruption. And the name still makes sense if "scope (pre)computation" means that both the exported symbols and the local symbols have been collected. After that, lazy linking works as expected.

PR #1790, opened 2025-01-14
Improve performance of hidden node parsing

msujew:
I've noticed in a project with a large file with a lot of comments (~20 000 LoC, ~10 000 comments), that parsing started taking a huge performance hit. It seems like our current method of iterating through the CST to add hidden nodes leads to super-linear behavior in the parsing phase.

You can reproduce this by using the test I've added to `langium-parser.test.ts` and run it on the current `main` branch. While the parse takes only ~150ms on my machine with the new changes, it takes roughly ~5000ms on `main`.

This change replaces the existing hidden node placement algorithm with a different one that should perform linearly. Note that the placement of the hidden nodes has not changed with this algorithm. Existing tests (and also a few new ones) verify this behavior.

Some ideas for this new algorithm were taken from https://github.com/eclipse-langium/langium/pull/1733.

### DIFF
--- a/packages/langium/src/parser/cst-node-builder.ts
+++ b/packages/langium/src/parser/cst-node-builder.ts
@@ -9,7 +9,6 @@ import type { Range } from 'vscode-languageserver-types';
 import type { AbstractElement } from '../languages/generated/ast.js';
 import type { AstNode, CompositeCstNode, CstNode, LeafCstNode, RootCstNode } from '../syntax-tree.js';
 import { Position } from 'vscode-languageserver-types';
-import { isCompositeCstNode } from '../syntax-tree.js';
 import { tokenToRange } from '../utils/cst-utils.js';
 
 export class CstNodeBuilder {
@@ -17,7 +16,7 @@ export class CstNodeBuilder {
     private rootNode!: RootCstNodeImpl;
     private nodeStack: CompositeCstNodeImpl[] = [];
 
-    private get current(): CompositeCstNodeImpl {
+    get current(): CompositeCstNodeImpl {
         return this.nodeStack[this.nodeStack.length - 1] ?? this.rootNode;
     }
 
@@ -37,8 +36,8 @@ export class CstNodeBuilder {
         return compositeNode;
     }
 
-    buildLeafNode(token: IToken, feature: AbstractElement): LeafCstNode {
-        const leafNode = new LeafCstNodeImpl(token.startOffset, token.image.length, tokenToRange(token), token.tokenType, false);
+    buildLeafNode(token: IToken, feature?: AbstractElement): LeafCstNode {
+        const leafNode = new LeafCstNodeImpl(token.startOffset, token.image.length, tokenToRange(token), token.tokenType, !feature);
         leafNode.grammarSource = feature;
         leafNode.root = this.rootNode;
         this.current.content.push(leafNode);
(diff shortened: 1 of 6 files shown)


packages/langium/src/parser/cst-node-builder.ts
```diff
@@ -55,6 +54,33 @@ export class CstNodeBuilder {
...
+        leafNode.root = this.rootNode;
+        let current: CompositeCstNode = this.current;
+        let added = false;
+        // If we are within a composite node, we add the hidden node to the content
+        if (current.content.length > 0) {
+            current.content.push(leafNode);
+            return;
+        }
+        while (current.container) {
+            const index = current.container.content.indexOf(current);
```
Lotes, line 68: Question: I thought I saw a performance improvement, but then I saw that `$containerIndex` is only defined on `AstNode`. Is there a reason why the `CstNode` does not have a `$containerIndex` property? This would be a difference from `O(N)` to `O(1)` here, if the given `N` is the number of children.

packages/langium/src/parser/langium-parser.ts
```diff
@@ -232,15 +233,16 @@ export class LangiumParser extends AbstractLangiumParser {
...
-        const lexerResult = this.lexer.tokenize(input);
+        const lexerResult = this.lexerResult = this.lexer.tokenize(input);
         this.wrapper.input = lexerResult.tokens;
         const ruleMethod = options.rule ? this.allRules.get(options.rule) : this.mainRule;
         if (!ruleMethod) {
             throw new Error(options.rule ? `No rule found with name '${options.rule}'` : 'No main rule available.');
         }
         const result = ruleMethod.call(this.wrapper, {});
-        this.nodeBuilder.addHiddenTokens(lexerResult.hidden);
+        this.appendHiddenTokens(lexerResult.hidden);
```
Lotes, line 243: Thinking loud, I assume that these hidden tokens are sorted by position. This can be used to optimize their insertion (correct me if I am wrong, in my eyes we can be faster, or? But maybe I missed a detail that does this already?).

Having a `CstNode traverser` could be a solution. Like:

```ts
interface CstNodeTraverser {
  stack: [CompositeCstNode, ChildIndex][];
  advanceBefore(token): void;
  insert(token): void
}
```

The `advanceBefore` method traverses the tree from its last position, not from the root.
I saw that the node builder has already a stack. But I assume that everything is completed and the stack is empty, before you start to add hidden nodes.

msujew: Not necessary to make it this complicated, but I was able to get further performance improvements out of this by adding multiple tokens at the same time. Thanks.

packages/langium/src/parser/langium-parser.ts
```diff
@@ -275,12 +275,6 @@ export class LangiumParser extends AbstractLangiumParser {
         };
     }
 
-    private appendHiddenTokens(tokens: IToken[]): void {
-        for (const token of tokens) {
-            this.nodeBuilder.addHiddenNode(token);
-        }
-    }
-
     private getHiddenTokens(token: IToken): IToken[] {
```
Lotes, line 278: GET is misleading, because while reading you will maybe assume that the structure stays untouched. I suggest POP or EXTRACT.

Lotes (approved):
I was hesitating with the approval, but actually there is no need for waiting, I got it. If you want an additional approval or more comments, you could reach out for Ben. Or ask me, what I should look at specifically.

PR #1791, opened 2025-01-16
added .nvmrc

cdietrich:


msujew (commented):
I think this makes sense. As an additional change, can you make the GitHub Actions use this specific node version? You can replace the `node-version: '18'` with `node-version-file: '.nvmrc'` in all `.yml` files.

msujew (approved):
LGTM 👍

PR #1805, opened 2025-02-07
langium: additional tests of UriUtils.relative() with windows paths

sailingKieler:


packages/langium/test/utils/uri-utils.test.ts
```diff
@@ -21,18 +22,36 @@ describe('URI Utils', () => {
         expect(UriUtils.relative(from, to)).toBe('../d.txt');
     });
 
+    test.skipIf(process.platform !== 'win32')('relative path in parent directory win32', () => {
```
msujew, line 25: FYI, feel free to run these tests on Windows as well - `URI` is capable of reading windows style file paths even on non-windows systems :)

msujew: Huh, that's unexpected. I don't like that, but it seems like you're right :)

sailingKieler:
@msujew what do you think about merging these test cases, mainly for documentation purposes?

msujew (approved):
Yep, good idea. Looks good to me 👍

PR #1810, opened 2025-02-14
Langium generator: Transform  to npm workspace and remove web

kaisalmen:
This will replace #1520 but is not yet ready for review


kaisalmen:
I rebased the PR and this is ready for review.

kaisalmen:
Should be rebased after #1930 is merged and then reviewed.

packages/generator-langium/templates/packages/extension/package.json
```diff
@@ -1,4 +1,7 @@
 {
+    "name": "<%= extension-name %>-extension",
```
msujew, line 2: Minor: This name will eventually end up as the ID for the extension on the vscode marketplace (assuming the adopter uploads it there). We should align to other extensions there and rather call it `vscode-whatever`, like [`vscode-yaml`](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml).
```suggestion
    "name": "vscode-<%= extension-name %>",
```

packages/generator-langium/templates/packages/language/vitest.config.ts
msujew, line 1: Suggestion: This file is getting generated even though the `test` option is disabled. We should skip this file in that case.

msujew (approved):
Looks really good. Just a few minor comments, see below.

packages/generator-langium/templates/packages/language/vitest.config.ts
kaisalmen: Good point. I will fix that.

packages/generator-langium/templates/packages/extension/package.json
kaisalmen: Will do

kaisalmen: Done ⬇️

packages/generator-langium/templates/packages/language/vitest.config.ts
kaisalmen: Done ⬇️

PR #1814, opened 2025-02-19
langium generate: fixed a hidden bug in 'node-processor.ts' of Langium's 'generate' facility

sailingKieler:
* added some test cases
* extended 'Vitest: Run Selected File' launch config
* brought config of the Vitest VSCode extension closer to the Vitest setup in 'vite.config.mts'

The actual bug fix is this the addition of the following line, everything else is just polishing:
https://github.com/eclipse-langium/langium/blob/1df3bcfed9c50f49cc2a66d0a4beb984bbf28962/packages/langium/src/generate/node-processor.ts#L97

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -158,19 +158,25 @@
             "skipFiles": ["<node_internals>/**", "**/node_modules/**"],
             "program": "${workspaceRoot}/node_modules/vitest/vitest.mjs",
             "args": ["run", "--no-color", "--no-coverage", "--no-watch"],
-            "smartStep": true,
             "console": "integratedTerminal",
+            "smartStep": true,
         },
         {
             "name": "Vitest: Run Selected File",
             "type": "node",
             "request": "launch",
             "autoAttachChildProcesses": true,
-            "skipFiles": ["<node_internals>/**", "**/node_modules/**"],
+            "skipFiles": ["<node_internals>/**", "**/node_modules/**", "!**/node_modules/vscode-*/**"],
             "program": "${workspaceRoot}/node_modules/vitest/vitest.mjs",
             "args": ["run", "${relativeFile}"],
-            "smartStep": true,
             "console": "integratedTerminal",
+            "smartStep": true,
+            "sourceMaps": true,
+            "outFiles": [
+                // cs: surprisingly, it makes a significant difference whether the "outFiles" property is not mentioned at all or an empty array is given here;
+                //  this setup seems to work best here, cross check with 'uri-utils.test.ts', for example
+                // my assumption is that vitest now relies on it's on the fly generated source maps plus those being available in the folder of the particular js modules, like in case of external libraries
+            ]
         }
     ]
 }
--- a/packages/langium/test/generate/template-node.test.ts
+++ b/packages/langium/test/generate/template-node.test.ts
@@ -832,9 +832,47 @@ describe('Multiple nested substitution templates', () => {
                         generated text!
         `);
     });
+
+    test('Nested substitution of with indented nested template starting with an _undefined_ line', () => {
+        expect(
+            toString(n`
+                begin:
+                  ${n`
+                    ${undefined}
+                    ${nestedNode}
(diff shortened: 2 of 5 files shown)


.vscode/launch.json
```diff
@@ -171,6 +171,11 @@
             "args": ["run", "${relativeFile}"],
             "smartStep": true,
             "console": "integratedTerminal",
+            "sourceMaps": true,
+            "outFiles": [
+                "${workspaceFolder}/packages/vscode-uri/lib/**/*.js",
```
msujew, line 176: We don't have a package called `packages/vscode-uri` in our repo.

sailingKieler: Damn, True. It's supposed to be `node_modules` 🙈, but I think I'll just drop that line.

packages/langium/test/generate/template-node.test.ts
```diff
@@ -832,9 +832,47 @@ describe('Multiple nested substitution templates', () => {
                         generated text!
         `);
     });
+
+    test('Nested substitution of with indented nested template starting with an _undefined_ line', () => {
```
JohannesMeierSE, line 836: This test case reflects the critical case, nice!

packages/langium/test/generate/template-node.test.ts
```diff
@@ -832,9 +832,47 @@ describe('Multiple nested substitution templates', () => {
...
+        ).toBe(
+            s`
+                begin:
+                  More
+                  generated text!
+            `
+        );
+    });
+
+    test('Nested substitution of with indented nested template starting with an _empty string_ line', () => {
```
JohannesMeierSE, line 854: This case already worked before

JohannesMeierSE (approved):
I reviewed the changes in the generator: Thanks @sailingKieler for fixing this bug!

packages/langium/test/generate/template-node.test.ts
sailingKieler: True. It's good to have it still, especially in contrast to the above one.

.vscode/launch.json
```diff
@@ -166,11 +166,15 @@
...
-            "skipFiles": ["<node_internals>/**", "**/node_modules/**"],
+            "skipFiles": ["<node_internals>/**", "**/node_modules/**", "!**/node_modules/vscode-uri/**"],
             "program": "${workspaceRoot}/node_modules/vitest/vitest.mjs",
             "args": ["run", "${relativeFile}"],
             "smartStep": true,
             "console": "integratedTerminal",
+            "sourceMaps": true,
+            "outFiles": [
+                "${workspaceFolder}/node_modules/vscode-uri/lib/**/*.js",
+            ]
```
msujew, line 177: The `outFiles` property on here behaves as a filter not as an addition to the existing outfiles. Meaning that the only `outFiles` being considered are now those from `node_modules/vscode-uri`. This prevents debugging the Langium source code.

msujew (changes requested):

.vscode/launch.json
sailingKieler: I kept the line, and made it work.
I however dropped the line mentioning `langium`, as it doesn't bring any benefit.

That's most likely due to Vitest/Vite (executed via launch config) that one the fly translates the executed test module implemented in TS and all transitively referenced ones. However, it messes up the sourcemaps, or I wasn't able to configure it correctly.
Running the test via the Vitest VSCode extension works nicely in this regard 😐

.vscode/launch.json
sailingKieler: I cannot confirm that, see below. However, the source maps generated by Vitest/Vite on the fly are out-of-sync 😖, see also https://github.com/eclipse-langium/langium/pull/1814#discussion_r1963601132
source-maps-out-of-sync

msujew: Interesting. Yes, I can confirm that keeping it empty also works for me 👍

msujew (approved):

PR #1816, opened 2025-02-21
Improve Windows drive letter URI handling

sailingKieler:
* added additional test cases

### DIFF
--- a/packages/langium/src/utils/uri-utils.ts
+++ b/packages/langium/src/utils/uri-utils.ts
@@ -16,17 +16,17 @@ export namespace UriUtils {
     export const joinPath = Utils.joinPath;
     export const resolvePath = Utils.resolvePath;
 
-    const isWindows = process?.platform === 'win32';
+    const isWindows = typeof process === 'object' && process?.platform === 'win32';
 
     export function equals(a?: URI | string, b?: URI | string): boolean {
         return a?.toString() === b?.toString();
     }
 
     export function relative(from: URI | string, to: URI | string): string {
-        const fromPath = typeof from === 'string' ? from : from.path;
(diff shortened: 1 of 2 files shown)


packages/langium/src/utils/uri-utils.ts
```diff
@@ -16,6 +16,8 @@ export namespace UriUtils {
     export const joinPath = Utils.joinPath;
     export const resolvePath = Utils.resolvePath;
 
+    const isWindows = process?.platform === 'win32';
```
spoenemann, line 19: This is not portable. Here's a good reference for how to make it work both in Node.js and browser:

https://github.com/eclipse-theia/theia/blob/master/packages/core/src/common/os.ts

msujew: Additionally, I'd be in favor of calling `URI.parse` on the input in case the input is a string. It isn't obvious we are expecting a path, and should much rather expect a stringified URI, like we're doing in the other methods.

packages/langium/src/utils/uri-utils.ts
```diff
@@ -36,6 +36,10 @@ export namespace UriUtils {
             if (toParts[0] && upperCaseDriveLetter.test(toParts[0])) {
                 toParts[0] = toParts[0].toLowerCase();
             }
+            if (fromParts[0] !== toParts[0]) {
+                // in case of different drive letters, we cannot compute a relative path, so...
+                return toPath.toLowerCase();
```
msujew, line 41: Question: Why lowercase the whole string? We can just keep the original capitalization.

sailingKieler: True.

msujew (approved):
Alright, looks good to me 👍

PR #1822, opened 2025-03-02
Improve chevrotain parser error recovery

msujew:
While looking into the [Chevrotain documentation on fault tolerance](https://chevrotain.io/docs/tutorial/step4_fault_tolerance.html), I've noticed that we're not supposed to catch the errors thrown by Chevrotain ourselves, but let them propagate. We've been effectively starving the `catch` block in [this method](https://github.com/Chevrotain/chevrotain/blob/e1479dfe848838c26a4b5a3ae12b4c8430eaeee8/packages/chevrotain/src/parse/parser/traits/recognizer_engine.ts#L244-L245), which is in charge of actually recovering from the error.

This results in some less-than-optimal behavior from Langium when it comes to error recovery. I've been playing around with this change using our example languages and it seems to behave pretty well. Refer to the additional comments to see why some of the changes were required.

### DIFF
--- a/packages/langium/test/parser/langium-parser.test.ts
+++ b/packages/langium/test/parser/langium-parser.test.ts
@@ -105,7 +105,6 @@ describe('parser error recovery', () => {
             const closing = ')'.repeat(close);
             const result = services.parser.LangiumParser.parse(`${opening}a${closing};`);
             // Expect only one parser error independent of the number of missing closing parenthesis
-            console.log(result.parserErrors);
             expect(result.parserErrors).toHaveLength(1);
         });
     }
--- a/packages/langium/test/parser/langium-parser.test.ts
+++ b/packages/langium/test/parser/langium-parser.test.ts
@@ -89,6 +89,29 @@ describe('hidden node parsing', () => {
 
 });
 
+describe('parser error recovery', () => {
+
+    for (const [open, close] of [[1, 0], [2, 0], [3, 0], [2, 1], [3, 1], [3, 2]]) {
+        test(`recovers from lexer error with ${open} open and ${close} close parenthesis`, async () => {
+            const text = `
+                grammar Test
+                entry Model: value=Expr ';';
+                Expr: '(' Expr ')' | value=ID;
+                terminal ID: /[_a-zA-Z][\\w_]*/;
+                hidden terminal WS: /\\s+/;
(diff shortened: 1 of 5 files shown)


msujew:
@cdietrich Are you interested in testing this? I believe this should be beneficial for most adopters of Langium. I can create a next release of this change if you want to.

packages/langium/test/parser/langium-parser.test.ts
```diff
@@ -89,6 +89,29 @@ describe('hidden node parsing', () => {
...
+                Expr: '(' Expr ')' | value=ID;
+                terminal ID: /[_a-zA-Z][\\w_]*/;
+                hidden terminal WS: /\\s+/;
+            `;
+            const services = await createServicesForGrammar({ grammar: text });
+            const opening = '('.repeat(open);
+            const closing = ')'.repeat(close);
+            const result = services.parser.LangiumParser.parse(`${opening}a${closing};`);
+            // Expect only one parser error independent of the number of missing closing parenthesis
+            console.log(result.parserErrors);
```
aabounegm, line 108: I don't think you intended to leave this `console.log` here :)

msujew: Thanks, yes I didn't :D

msujew:
This PR is available as `3.4.0-next.10c0eb7` on npm.

spoenemann (commented):
I haven't seen any difference in error recovery with the example languages. Do you have an example where the recovery behaves differently?

One minor difference: in the statemachine example, when I insert `state state` between the `events` and the `initialState`, the previous version showed two parser errors

> Expecting token of type 'initialState' but found `state`.statemachine
> Expecting end of file but found `state`.

while the new version shows only one. But in both versions, the recovery does _not_ work in this case: I get no cross-reference resolution for the rest of the file.

msujew:
> I haven't seen any difference in error recovery with the example languages. Do you have an example where the recovery behaves differently?

Me neither (at least with the examples). It might be that the only difference actually is related to the improved parser error message. Thinking further about this change, it's likely that this change mainly improves the error message behavior, as we likely end up at the same recovered parser state, but with less steps (and therefore fewer parser errors).

spoenemann (approved):
👍

PR #1830, opened 2025-03-10
Resolve wait until if document already in desired state

### LINKED ISSUE #1827: DocumentBuild.waitUntil doesn't resolve if document exists and is already in desired state
Langium version: 3.3.0
Package name: langium

## Steps To Reproduce

1. Add a document to the document cache.
2. Ask the document builder to build that document.
3. Wait until the document has finished building.
4. Call `.waitUntil` with the desired state.
5. The promise will not resolve if the document is already in the desired state.

The current check requires that the document be in a more advanced state than the requested state:

https://github.com/eclipse-langium/langium/blob/c3afb247d7dcfd41733c051b8db0395d50398b33/packages/langium/src/workspace/document-builder.ts#L423-L428

On the other hand, the fallback check for the workspace only requires the same or more advanced state:

https://github.com/eclipse-langium/langium/blob/c3afb247d7dcfd41733c051b8db0395d50398b33/packages/langium/src/workspace/document-builder.ts#L429-L433

Link to code example:

## The current behavior

`waitUntil` never resolves (unless the document is rebuilt later) if the document is available in the target state.

## The expected behavior

`waitUntil` should resolve immediately if the document is already available in the desired state. Or at least that's what I think I'd expect.

colin-grant-work:
Fixes #1827

As discussed on the issue, should be `>=` rather than `>` so that the promise resolves immediately if a document is already in the desired state.

### DIFF
--- a/packages/langium/test/workspace/document-builder.test.ts
+++ b/packages/langium/test/workspace/document-builder.test.ts
@@ -417,7 +417,7 @@ describe('DefaultDocumentBuilder', () => {
         await builder.build([document], { validation: true });
     });
 
-    test.only('`waitUntil` will correctly resolve if the document is already in the target state.', async () => {
+    test('`waitUntil` will correctly resolve if the document is already in the target state.', async () => {
         const services = await createServices();
         const documentFactory = services.shared.workspace.LangiumDocumentFactory;
         const documents = services.shared.workspace.LangiumDocuments;
--- a/packages/langium/test/workspace/document-builder.test.ts
+++ b/packages/langium/test/workspace/document-builder.test.ts
@@ -417,6 +417,25 @@ describe('DefaultDocumentBuilder', () => {
         await builder.build([document], { validation: true });
     });
 
+    test.only('`waitUntil` will correctly resolve if the document is already in the target state.', async () => {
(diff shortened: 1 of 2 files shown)


msujew (commented):
Thanks, can you create a unit test for that?

msujew (approved):
LGTM, thanks 👍

packages/langium/test/workspace/document-builder.test.ts
```diff
@@ -417,6 +417,25 @@ describe('DefaultDocumentBuilder', () => {
         await builder.build([document], { validation: true });
     });
 
+    test.only('`waitUntil` will correctly resolve if the document is already in the target state.', async () => {
```
msujew, line 420: This fails the CI.

```suggestion
    test('`waitUntil` will correctly resolve if the document is already in the target state.', async () => {
```

colin-grant-work: Whoops, that's a bad habit of mine. I'm glad that the CI catches it as a bad idea.

colin-grant-work:
@msujew, I don't seem able to merge this myself, so please feel free to do so. Thanks!

Issue #1065, opened 2023-05-26
Add a special syntax for binary operators


pluralia:
Motivation is in the discussion #751.

The proposal is to add the ability to define binary operations shorter with the following syntax:
```typescript
@binary_op (left_operand=left, right_operand=right, operator=op)
BinaryExpression infers Expression:
    , infixl ("*" | "/") Primary
    , infixl ('+' | '-') Primary
    , infixl "&&" Primary
    , infixl "||" Primary
    , infixn  ".." Primary // range operator; no associativity
    , infixr '=' Primary
;
```
* `@binary_op` -- an annotation to defined a rule with the special syntax
* `left_operand`, `right_operand`, `operator` -- arguments of the annotation that reflect property names in the generated interface `BinaryExpression`
* arguments can be omitted -- then default names can be used: `left_operand`, `right_operand`, `operator`
* the rule with the special annotation `@binary_op` consists of comma-separated operator declarations; the order of these declarations reflects the operator **precedence**
* each operator declaration consists of 3 parts:
   - associativity keyword `infix(r|l|n)`
   - operators -- usual syntax with pipes
   - the name of the rule responsible for parsing the primary expression

---

The syntax above must be semantically equivalent to the current syntax:
```typescript
Expression:
    Assignment;

Assignment infers Expression:
    Range ({infer BinaryExpression.left=current} operator='=' right=Assignment)?;

Range infers Expression:
    Or ({infer BinaryExpression.left=current} operator='..' right=Or)?;

Or infers Expression:
    And ({infer BinaryExpression.left=current} operator='||' right=And)*;

And infers Expression:
    Addition ({infer BinaryExpression.left=current} operator='&&' right=Addition)*;

Addition infers Expression:
    Multiplication ({infer BinaryExpression.left=current} operator=('+' | '-') right=Multiplication)*;

Multiplication infers Expression:
    Primary ({infer BinaryExpression.left=current} operator=('*' | '/') right=Primary)*;
```

---

Unary operators must be defined using the current syntax, for example:
```typescript
Primary infers Expression:
      '(' Expression ')'
    | {infer UnaryPrefix} op=('+' | '-') value=Expression
    | {infer NumberLiteral} value=NUMBER 
;
```

---

Both syntaxes have to generate the same types for the `ast.ts`:
```typescript
export type Expression = BinaryExpression | NumberLiteral | UnaryPrefix;

export interface BinaryExpression extends AstNode {
    left: Expression
[...]

PR #1836, opened 2025-03-11
Support infix operator rules

msujew:
Closes https://github.com/eclipse-langium/langium/issues/1065

Adds support for a special infix operator syntax that aims to improve Langium on multiple fronts:
1. Improves readability/maintainability of Langium grammars
2. Simplify specifying precedence for infix binary expressions
3. Improves performance by parsing the binary expression in a flat list - the parser will automatically transform the resulting object into a tree structure.

Note that the same AST node is constructed as if you were to write the parser rules manually. Also note that this PR assumes that all binary operators are left-associative (i.e. `A + B + C` is parsed as `(A + B) + C`). We can revisit this decision in case we find a case for right-associative grammar rules. Note that this only requires changes in the tree restructuring routine.

### DIFF
--- a/examples/arithmetics/src/language-server/arithmetics.langium
+++ b/examples/arithmetics/src/language-server/arithmetics.langium
@@ -20,19 +20,9 @@ Evaluation:
     expression=Expression ';';
 
 Expression:
-    Addition;
+    BinaryExpression;
 
-Addition infers Expression:
-    Multiplication ({infer BinaryExpression.left=current} operator=('+' | '-') right=Multiplication)*;
-
-Multiplication infers Expression:
-    Exponentiation ({infer BinaryExpression.left=current} operator=('*' | '/') right=Exponentiation)*;
-
-Exponentiation infers Expression:
-    Modulo ({infer BinaryExpression.left=current} operator='^' right=Modulo)*;
-
-Modulo infers Expression:
-    PrimaryExpression ({infer BinaryExpression.left=current} operator='%' right=PrimaryExpression)*;
+infix BinaryExpression on PrimaryExpression: '%' > '^' > '*' | '/' > '+' | '-';
 
 PrimaryExpression infers Expression:
     '(' Expression ')' |
--- a/packages/langium/src/grammar/langium-grammar.langium
+++ b/packages/langium/src/grammar/langium-grammar.langium
@@ -56,13 +56,13 @@ SimpleType infers TypeDefinition:
 PrimitiveType returns string:
     'string' | 'number' | 'boolean' | 'Date' | 'bigint';
 
-type AbstractType = Interface | Type | ParserRule | InferredType;
+type AbstractType = Interface | Type | ParserRule | InfixRule | InferredType;
 
 Type:
     'type' name=ID '=' type=TypeDefinition ';'?;
 
 AbstractRule:
-    ParserRule | TerminalRule;
+    ParserRule | TerminalRule | InfixRule;
 
 GrammarImport:
     'import' path=STRING ';'?;
@@ -73,6 +73,12 @@ ParserRule:
     (wildcard?='*' | ('returns' (returnType=[AbstractType:ID] | dataType=PrimitiveType)) | inferredType=InferredType<false>)?
     (definesHiddenTokens?='hidden' '(' (hiddenTokens+=[AbstractRule:ID] (',' hiddenTokens+=[AbstractRule:ID])*)? ')')? ':'
     definition=Alternatives ';';
+
+InfixRule: 'infix' RuleNameAndParams 'on' call=RuleCall ':' operators=InfixRuleOperators ';';
+
+InfixRuleOperators: precedences+=InfixRuleOperatorList ('>' precedences+=InfixRuleOperatorList)*;
+
+InfixRuleOperatorList: operators+=Keyword ('|' operators+=Keyword)*;
 
 InferredType<imperative>:
     (<imperative> 'infer' | <!imperative> 'infers') name=ID;
--- a/packages/langium/src/parser/langium-parser.ts
+++ b/packages/langium/src/parser/langium-parser.ts
@@ -6,16 +6,17 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import type { DSLMethodOpts, ILexingError, IOrAlt, IParserErrorMessageProvider, IRecognitionException, IToken, TokenType, TokenVocabulary } from 'chevrotain';
-import type { AbstractElement, Action, Assignment, ParserRule } from '../languages/generated/ast.js';
+import type { AbstractElement, Action, Assignment, InfixRule, ParserRule } from '../languages/generated/ast.js';
+import { isInfixRule } from '../languages/generated/ast.js';
 import type { Linker } from '../references/linker.js';
 import type { LangiumCoreServices } from '../services.js';
-import type { AstNode, AstReflection, CompositeCstNode, CstNode } from '../syntax-tree.js';
+import type { AstNode, AstReflection, CompositeCstNode, CstNode, Mutable } from '../syntax-tree.js';
 import type { Lexer, LexerResult } from './lexer.js';
 import type { IParserConfig } from './parser-config.js';
 import type { ValueConverter } from './value-converter.js';
 import { defaultParserErrorProvider, EmbeddedActionsParser, LLkLookaheadStrategy } from 'chevrotain';
 import { LLStarLookaheadStrategy } from 'chevrotain-allstar';
-import { isAssignment, isCrossReference, isKeyword } from '../languages/generated/ast.js';
+import { isAssignment, isCrossReference, isKeyword, isParserRule } from '../languages/generated/ast.js';
 import { getExplicitRuleType, isDataTypeRule } from '../utils/grammar-utils.js';
 import { assignMandatoryProperties, getContainerOfType, linkContentToContainer } from '../utils/ast-utils.js';
 import { CstNodeBuilder } from './cst-node-builder.js';
@@ -36,6 +37,13 @@ interface DataTypeNode {
     $type: symbol
     /** Used as a storage for all parsed terminals, keywords and sub-datatype rules */
     value: string
+}
+
+interface InfixElement {
+    $type: string;
+    $cstNode: CompositeCstNode;
+    parts: AstNode[];
+    operators: string[];
 }
 
 function isDataTypeNode(node: { $type: string | symbol | undefined }): node is DataTypeNode {
@@ -63,7 +71,7 @@ export interface BaseParser {
     /**
      * Adds a new parser rule to the parser
      */
-    rule(rule: ParserRule, impl: RuleImpl): RuleResult;
+    rule(rule: ParserRule | InfixRule, impl: RuleImpl): RuleResult;
     /**
      * Returns the executable rule function for the specified rule name
      */
@@ -159,7 +167,7 @@ export abstract class AbstractLangiumParser implements BaseParser {
         this.wrapper.wrapAtLeastOne(idx, callback);
     }
 
-    abstract rule(rule: ParserRule, impl: RuleImpl): RuleResult;
+    abstract rule(rule: ParserRule | InfixRule, impl: RuleImpl): RuleResult;
     abstract consume(idx: number, tokenType: TokenType, feature: AbstractElement): void;
     abstract subrule(idx: number, rule: RuleResult, fragment: boolean, feature: AbstractElement, args: Args): void;
     abstract action($type: string, action: Action): void;
@@ -198,6 +206,7 @@ export class LangiumParser extends AbstractLangiumParser {
     private lexerResult?: LexerResult;
     private stack: any[] = [];
     private assignmentMap = new Map<AbstractElement, AssignmentElement | undefined>();
+    private operatorPrecedence = new Map<string, Map<string, number>>();
 
     private get current(): any {
         return this.stack[this.stack.length - 1];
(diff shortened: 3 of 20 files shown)


msujew:
@aabounegm Good question: Splitting the infix operator rule into multiple rules would result in decreased performance - we get a boost by reducing the amount of parser rules (and therefore subrule calls) we need to create.

Instead, I would propose something like this perhaps?

```
Expr: BinaryExpr;

infix BinaryExpr on Primary: {infer Equal} '==' > {infer Multiplication} '*' > {infer Addition} '+';

Primary infers Expr: '(' Expression ')' | NumberLiteral;
```

I think that should work pretty well. However, I would attempt to implement such a feature in a separate step/PR. This PR is complex enough as it is.

aabounegm:
In my grammar (linked above), I already have multiple rules since the language is for educational purposes. Do you mean the performance will be even worse or the same as the way it is currently defined in my grammar?

msujew:
> Do you mean the performance will be even worse or the same as the way it is currently defined in my grammar?

It will be the same (maybe *very slightly* worse, due to the additional transformation step done to convert the flat list into an expression tree), but you won't gain anything from it. This PR allows to eliminate multiple levels of precedence in order to massively boost performance when parsing expressions.

ydaveluy:
How does it works when types are not infered (using pre-defined interfaces)?
```
interface Expression {
}
interface BinaryOperation extends Expression {
	leftOperand: Expression
	rightOperand: Expression
	feature: string
}
interface UnaryOperation extends Expression {
	operand: Expression
	feature: string
}
// how to map leftOperand, rightOperand and feature ?
infix BinaryExpression on UnaryOperation: '%' > '^' > '*' | '/' > '+' | '-';

UnaryOperation returns Expression:
    {UnaryOperation} feature=("!" | "-" | "+" | "~")  operand=Literal
;

```

szarnekow:
> in order to massively boost performance when parsing expressions.

can you provide some numbers?

msujew:
> can you provide some numbers?

@szarnekow Sure, I just ran the branch through ~4100 lines of the arithmetics example code (parser only). On `main`, this currently takes 430ms, whereas this branch only requires 260ms. So we're talking about 40% improved performance wrt to expression parsing. I used this code and placed in the `packages/arithmetics/test` directory:

```ts
import { describe, expect, test } from 'vitest';
import { createArithmeticsServices } from '../src/language-server/arithmetics-module.js';
import { EmptyFileSystem } from 'langium';

describe('Test the arithmetics parsing', () => {

    const services = createArithmeticsServices(EmptyFileSystem);
    const parser = services.arithmetics.parser.LangiumParser;

    test('Simple expression', () => {
        let text = '1 + 2 ^ 3 * 4 - 5 / 6;';
        for (let i = 0; i < 12; i++) {
            text = text + text;
        }
        console.time('parse');
        const parseResult = parser.parse('module test ' + text);
        console.timeEnd('parse');
        expect(parseResult.parserErrors).toHaveLength(0);
    });
});
```

> I don't know if it is technically possible.

I'm not sure either. Your proposal still includes building a tree-like structure for infix operators - however, that's exactly what we want to prevent. Tranforming that into the expected flat structure we need for parsing is maybe possible, although you have so many degrees of freedom that it likely just becomes a bottomless hole. Instead I thought that something like this could work:

```
interface Expression {
}
interface BinaryExpression extends Expression {
	leftOperand: Expression
	rightOperand: Expression
	feature: string
}
infix BinaryExpression on Primary returns BinaryExpression {
  left: leftOperand,
  right: rightOperand,
  operator: feature
}: '%' > '^' > '*' | '/' > '+' | '-';
```

Or alternatively for inferred types:

```
infix BinaryExpression on Primary infers BinaryExpression {
  left: leftOperand,
  right: rightOperand,
  operator: feature
}: '%' > '^' > '*' | '/' > '+' | '-';
```

aabounegm:
@ydaveluy's suggestion had me wondering, isn't the new `infix Binary on Primary: '+';` syntax basically syntax sugar for `Binary: Primary | Binary '+' Binary;` I understand that left-recursion in general is not yet supported, but this infix rule seems to be a specific case of left-recursion. If my understanding is correct, would it also not be technically possible to just detect this specific pattern (`Rule: Base | lhs=Rule op='+' rhs=Rule;`) and perform the same transformation that this PR does to the infix rule, rather than introduce new syntax?

examples/arithmetics/src/language-server/arithmetics.langium
```diff
@@ -20,19 +20,9 @@ Evaluation:
...
-
-Multiplication infers Expression:
-    Exponentiation ({infer BinaryExpression.left=current} operator=('*' | '/') right=Exponentiation)*;
-
-Exponentiation infers Expression:
-    Modulo ({infer BinaryExpression.left=current} operator='^' right=Modulo)*;
-
-Modulo infers Expression:
-    PrimaryExpression ({infer BinaryExpression.left=current} operator='%' right=PrimaryExpression)*;
+infix BinaryExpression on PrimaryExpression: '%' > '^' > '*' | '/' > '+' | '-';
```
Lotes, line 25: Normally the precedence of > is greater than the one of |. It reads a bit unnatural.
Adding parentheses could be an option to enumerate all alternatives.
Or maybe a equal sign instead of the |.
WDYT?

packages/langium/src/parser/langium-parser.ts
```diff
@@ -63,7 +71,7 @@ export interface BaseParser {
     /**
      * Adds a new parser rule to the parser
      */
-    rule(rule: ParserRule, impl: RuleImpl): RuleResult;
+    rule(rule: ParserRule | InfixRule, impl: RuleImpl): RuleResult;
```
Lotes, line 74: ParserRule | InfixRule occur together quite often. Maybe add a common type and use that one instead.

examples/arithmetics/src/language-server/arithmetics.langium
ydaveluy: And why not using the comma operator instead of '>' ?
```
'%' , '^' , '*' | '/' , '+' | '-';
```

Lotes: ‚>‘ implies an order between the operators, that makes totally sense to me. Having a comma, you do not know what comes first: low or high precedence. But you can assume some convention, true.

spoenemann: You can mitigate this by writing each group in its own line:
```
infix BinaryExpression on PrimaryExpression:
    '%'
    > '^'
    > '*' | '/'
    > '+' | '-';
```
This is how I'd format such expressions for readability.

packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -74,6 +74,12 @@ ParserRule:
     (definesHiddenTokens?='hidden' '(' (hiddenTokens+=[AbstractRule:ID] (',' hiddenTokens+=[AbstractRule:ID])*)? ')')? ':'
     definition=Alternatives ';';
 
+InfixRule: 'infix' RuleNameAndParams 'on' call=RuleCall ':' operators=InfixRuleOperators ';';
+
+InfixRuleOperators: precedences+=InfixRuleOperatorList ('>' precedences+=InfixRuleOperatorList)*;
+
+InfixRuleOperatorList: operators+=Keyword ('|' operators+=Keyword)*;
```
spoenemann, line 81: I assume the associativity is left-to-right?

Shall we add an option to change that?

```
InfixRuleOperatorList:
    (associativity=Associativity 'assoc')?
    operators+=Keyword ('|' operators+=Keyword)*;

Associativity returns string: 'left' | 'right';
```

If omitted, the default should be `left`. With this, you can realize assignments as operators:

```
infix BinaryExpression on PrimaryExpression:
    '*' | '/'
    > '+' | '-'
    > right assoc '=' | '*=' | '/=' | '+=' | '-=';
```

Maybe also allow `no assoc` to forbid more than two operands on the same level?

packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -74,6 +74,12 @@ ParserRule:
     (definesHiddenTokens?='hidden' '(' (hiddenTokens+=[AbstractRule:ID] (',' hiddenTokens+=[AbstractRule:ID])*)? ')')? ':'
     definition=Alternatives ';';
 
+InfixRule: 'infix' RuleNameAndParams 'on' call=RuleCall ':' operators=InfixRuleOperators ';';
```
spoenemann, line 77: We should allow `infers` and `returns` as usual. If a declared type is given, it *must* have the properties `left`, `right`, and `operator` with respective types.

packages/langium/src/utils/ast-utils.ts
```diff
@@ -16,7 +16,7 @@ import { inRange } from './cst-utils.js';
  * Link the `$container` and other related properties of every AST node that is directly contained
  * in the given `node`.
  */
-export function linkContentToContainer(node: AstNode): void {
+export function linkContentToContainer(node: AstNode, deep = false): void {
```
spoenemann, line 19: I'd try to avoid boolean parameters and use an `options` object instead.

packages/langium/src/parser/langium-parser.ts
spoenemann: That would also help for `isParserRule(rule) || isInfixRule(rule)`, which I've seen at least twice.

packages/langium/src/grammar/langium-grammar.langium
spoenemann: Ah 👉 https://github.com/eclipse-langium/langium/pull/1836#issuecomment-2715732901 – 👍 for a separate PR, but let's not forget that.

spoenemann:
To this discussion: https://github.com/eclipse-langium/langium/pull/1836#issuecomment-2716688994 (too bad GitHub doesn't have threads 😞)

I actually see value in standardizing the structure of BinaryOperator nodes. That can be very useful for other features that we might add later. So I vote for not giving the option to change the property names.

If you already have custom code relying on different property names, I'd use TypeScript rename refactoring to change it.

ydaveluy:
If I understand correctly, this PR involves a post-parsing refactoring for binary expressions.

I wonder if this case could be handled with a more generic approach:

Write the BinaryExpression rule without precedence:
`BinaryExpression: lhs=BinaryExpression op=('*'|'/'|'-'|'+'|'|'|'&') rhs=BinaryExpression | PrimaryExpression;`

Provide an API to register refactoring actions for the parser (similar to what is done for validation):
```
        actions: RefactoringAction<ast.TestAstType> = {
            BinaryExpression: refactorBinaryExpression
        };
    parser.register(actions);
    refactorBinaryExpression(expr: ast.BinaryExpression): void {
       if(ast.isBinaryExpression(expr.lhs) && isBinaryExpression(expr.rhs)) {
           // do refactoring
       }
    }
```

Provide a generic action to restructure the AST for binary expressions with operator precedence.

This would be easily customizable by the user and could allow other types of refactoring.

msujew:
> If I understand correctly, this PR involves a post-parsing refactoring for binary expressions.

Right, but that's not the only advantage. I mainly created this change to speed up expression parsing performance (between 40-50% in my tests). This is done by converting the infix parser rule into a *flat list of expressions* and then constructing a parse tree. Your proposed change still constructs a parse tree, which is then getting rewritten. This would unfortunately eliminate the performance benefits of this PR.

> BinaryExpression: lhs=BinaryExpression op=('*'|'/'|'-'|'+'|'|'|'&') rhs=BinaryExpression | PrimaryExpression;

Note that this syntax is left recursive, and we don't support left recursion (due to using an LL parser).

ydaveluy:
Sorry the syntax was wrong...

```
Expression: PrimaryExpression ({BinaryExpression.lhs=current} operator=('*'|'/'|'-'|'+'|'|'|'&') rhs=PrimaryExpression)*;
```

msujew:
I see. I like the idea of making the tree-rewriting configurable, but I see relative little use beyond rewriting binary expression, since we cannot influence the shape of the generated AST that way. Also, I feel like this would hide the feature from a lot of users, even though it's supposed to be a simplification of the existing way of writing binary expressions.

msujew:
Hm, I quite like the idea. I'll iterate on this during the course of the week :)

msujew:
@spoenemann I've added support for right associativity in my latest commit.

examples/arithmetics/src/language-server/arithmetics.langium
spoenemann: @msujew shall we change the formatting here to give a better readable example?

packages/langium/src/grammar/langium-grammar.langium
```diff
@@ -219,7 +229,7 @@ CharacterRange infers AbstractElement:
     {infer CharacterRange} left=Keyword ('..' right=Keyword)?;
 
 FeatureName returns string:
-    'current' | 'entry' | 'extends' | 'false' | 'fragment' | 'grammar' | 'hidden' | 'import' | 'interface' | 'returns' | 'terminal' | 'true' | 'type' | 'infer' | 'infers' | 'with' | PrimitiveType | ID;
+    'right' | 'left' | 'assoc' | 'current' | 'entry' | 'extends' | 'false' | 'fragment' | 'grammar' | 'hidden' | 'import' | 'interface' | 'returns' | 'terminal' | 'true' | 'type' | 'infer' | 'infers' | 'with' | PrimitiveType | ID;
```
spoenemann, line 232: What about the new keywords `'infix'` and `'on'`?

spoenemann (approved):
This is almost good to go – I already give my 👍, but please check out the comments https://github.com/eclipse-langium/langium/pull/1836#discussion_r2092467582 and https://github.com/eclipse-langium/langium/pull/1836#discussion_r1997744694, and create a follow-up issue for https://github.com/eclipse-langium/langium/pull/1836#discussion_r2014351621

PR #1837, opened 2025-03-11
allow user defined validation categories

ydaveluy:
Langium provides by default 3 pre-defined validation categories: `fast`, `slow` and `built-in`.

This PR allow to register custom/user categories, see #1834.

It is the user responsibility to trigger these checks.

### DIFF
--- a/packages/langium/src/test/langium-test.ts
+++ b/packages/langium/src/test/langium-test.ts
@@ -556,10 +556,11 @@ export type ValidationHelperOptions = ParseHelperOptions & { failOnParsingErrors
 export function validationHelper<T extends AstNode = AstNode>(services: LangiumCoreServices): (input: string, options?: ValidationHelperOptions) => Promise<ValidationResult<T>> {
     const parse = parseHelper<T>(services);
     return async (input, options) => {
-        const document = await parse(input, {
+        const parseOptions: ValidationHelperOptions = {
             ...(options ?? {}),
-            validation: true
-        });
+        };
+        parseOptions.validation = parseOptions.validation ?? true;
+        const document = await parse(input, parseOptions);
         const result = {
             document,
             diagnostics: document.diagnostics ?? [],
--- a/packages/langium/src/test/langium-test.ts
+++ b/packages/langium/src/test/langium-test.ts
@@ -559,7 +559,7 @@ export function validationHelper<T extends AstNode = AstNode>(services: LangiumC
         const parseOptions: ValidationHelperOptions = {
             ...(options ?? {}),
         };
-        parseOptions.validation = parseOptions.validation ?? true;
+        parseOptions.validation ??= true;
         const document = await parse(input, parseOptions);
         const result = {
             document,
--- a/packages/langium/src/validation/validation-registry.ts
+++ b/packages/langium/src/validation/validation-registry.ts
@@ -113,7 +113,7 @@ export type ValidationChecks<T> = {
  * to you to schedule these checks: after the fast checks are done, or after saving a document,
  * or with an explicit command, etc.
  */
-export type ValidationCategory = 'fast' | 'slow' | 'built-in' | string;
+export type ValidationCategory = 'fast' | 'slow' | 'built-in' | (string & {});
 
 export namespace ValidationCategory {
     export const all: readonly ValidationCategory[] = ['fast', 'slow', 'built-in'];
--- a/packages/langium/src/validation/validation-registry.ts
+++ b/packages/langium/src/validation/validation-registry.ts
@@ -98,6 +98,8 @@ export type ValidationChecks<T> = {
 }
 
 /**
+ * There are 3 pre-defined categories: `fast`, `slow` and `built-in`.
+ *
  * `fast` checks can be executed after every document change (i.e. as the user is typing). If a check
  * is too slow it can delay the response to document changes, yielding bad user experience. By marking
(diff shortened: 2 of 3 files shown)


packages/langium/test/validation/document-validator.test.ts
```diff
@@ -259,4 +259,82 @@ describe('Register Before/AfterDocument logic for validations with state', () =>
...
+
+        test('user&fast categories', async () => {
+            const validationResult = await validate('model test', {validation:{categories:['user', 'fast']}});
+            const diagnostics = validationResult.diagnostics;
+            expect(diagnostics).toHaveLength(2);
+            expect(diagnostics.filter(d => d.severity === 1)).toHaveLength(1); // 1 error
+            expect(diagnostics.filter(d => d.severity === 2)).toHaveLength(1); // 1 warning
+        });
+
+        test('default categories', async () => {
```
ydaveluy, line 331: I'm surprised that by default all categories are enabled. I was expecting only `built-in` and `fast` checks.

I don't know if the behavior is different between the test framework and the default builder.

packages/langium/src/test/langium-test.ts
```diff
@@ -556,10 +556,11 @@ export type ValidationHelperOptions = ParseHelperOptions & { failOnParsingErrors
 export function validationHelper<T extends AstNode = AstNode>(services: LangiumCoreServices): (input: string, options?: ValidationHelperOptions) => Promise<ValidationResult<T>> {
     const parse = parseHelper<T>(services);
     return async (input, options) => {
-        const document = await parse(input, {
+        const parseOptions: ValidationHelperOptions = {
             ...(options ?? {}),
-            validation: true
-        });
+        };
+        parseOptions.validation = parseOptions.validation ?? true;
```
aabounegm, line 562: FYI, the nullish coalescing assignment operator (`??=`) is supported since TypeScript 4.0 and Node 15.0

packages/langium/src/validation/validation-registry.ts
```diff
@@ -106,8 +108,12 @@ export type ValidationChecks<T> = {
  *
  * `built-in` checks are errors produced by the lexer, the parser, or the linker. They cannot be used
  * for custom validation checks.
+ *
+ * You can also provide user-defined categories. These check will be skipped by default. Then it's up
+ * to you to schedule these checks: after the fast checks are done, or after saving a document,
+ * or with an explicit command, etc.
  */
-export type ValidationCategory = 'fast' | 'slow' | 'built-in'
+export type ValidationCategory = 'fast' | 'slow' | 'built-in' | string;
```
aabounegm, line 116: This leads to `ValidationCategory` collapsing to just `string`, hence losing IntelliSense for the built-in ones. The work-around for this is to define it as such:
```suggestion
export type ValidationCategory = 'fast' | 'slow' | 'built-in' | (string & {});
```

packages/langium/src/test/langium-test.ts
ydaveluy: Thanks for the information.
I'm not (yet) expert in typescript

ydaveluy:
@aabounegm any idea on how to solve `export type ValidationCategory = 'fast' | 'slow' | 'built-in' | (string & {});` ?
The linter does not like this syntax.

aabounegm:
@ydaveluy Technically, `Record<string, never>` should also work, but I personally don't like it. I would actually prefer to just use `// eslint-disable-next-line @typescript-eslint/ban-types`

packages/langium/test/validation/document-validator.test.ts
ydaveluy: ok,

In the test framework  `buildOptions.validation` is true by default so all validator are executed.
In the DefaultBuilder `buildOptions.validation` is a ValidationOptions with categories `built-in` and `fast`

msujew (approved):
I think that makes sense. @spoenemann care to chime in as well? You originally contributed the feature.

spoenemann (commented):
Looks good, but there's a potential problem with the `ValidationCategory.all` array, which is then no longer complete. It's currently used by `DefaultDocumentBuilder`.

Proposal: mark that value as `@deprecated` and instead define a public property `allValidationCategories` in `DefaultDocumentBuilder` so it's possible to change that (see how `updateBuildOptions` is already exposed there). WDYT?

ydaveluy:
@spoenemann I don't understand well your proposal and for me it is not clear what the code has to do.

From my understanding an initial validation is done with `ValidationCategory.all` (if no validation option is provided by the workspace manager) then on update `updateBuildOptions` is used and stored in the buildState for the next updates.

For user defined validation categories I think it would be great to provide something like

```ts
    protected getAdditionalValidationCategories(document: LangiumDocument): ValidationCategory[]  {
        // retrieve the list of category from an external configuration file, or from a data in the document or ...
        return [];
    }
```

The initial validation could be done with `ValidationCategory.all`  + `getAdditionalValidationCategories(...)` then on update only `updateBuildOptions`  + `getAdditionalValidationCategories(...)` .

Does it makes sense for you ?

spoenemann (approved):
Nevermind – the code that I referred to in DefaultDocumentBuilder is only for a super-special case. Let's keep it simple.

ydaveluy:
Hey @msujew 

I would be very interested to use this functionality in a near future, do you plan to make a new release soon ?

PR #1842, opened 2025-03-12
provide hover documentation on keywords

ydaveluy:
Few months ago I posted #1603 to provide hover on keywords.

It was not working when the user ML_COMMENT rule was different from the Langium grammar (`'/*'->'*/'`).

This PR solves this limitation and enable by default hover on keywords.

The documentation prefixing a keyword in the grammar will be available when hovering on these keywords in the user language.

in grammar:
```
entry Element: 
    /** documentation of the element keyword */ 
    'element' name=ID ;
```

in the user language:
```
element a_name
```
The hover information on `element` keyword will be `documentation of the element keyword ` .

### DIFF
--- a/packages/langium/src/lsp/hover-provider.ts
+++ b/packages/langium/src/lsp/hover-provider.ts
@@ -13,10 +13,10 @@ import type { AstNode } from '../syntax-tree.js';
 import type { MaybePromise } from '../utils/promise-utils.js';
 import type { LangiumDocument } from '../workspace/documents.js';
 import type { DocumentationProvider } from '../documentation/documentation-provider.js';
-import { findDeclarationNodeAtOffset } from '../utils/cst-utils.js';
+import { findCommentNode, findDeclarationNodeAtOffset } from '../utils/cst-utils.js';
+import type { Keyword } from '../languages/generated/ast.js';
 import { isKeyword } from '../languages/generated/ast.js';
-import { createLangiumGrammarServices } from '../grammar/langium-grammar-module.js';
-import { EmptyFileSystem } from '../workspace/file-system-provider.js';
+import { isJSDoc, parseJSDoc } from '../documentation/jsdoc.js';
 
 /**
  * Language-specific service for handling hover requests.
@@ -35,12 +35,10 @@ export abstract class AstNodeHoverProvider implements HoverProvider {
 
     protected readonly references: References;
     protected readonly grammarConfig: GrammarConfig;
-    protected readonly grammarDocumentationProvider: DocumentationProvider;
 
     constructor(services: LangiumServices) {
         this.references = services.references.References;
         this.grammarConfig = services.parser.GrammarConfig;
-        this.grammarDocumentationProvider = createLangiumGrammarServices(EmptyFileSystem).grammar.documentation.DocumentationProvider;
     }
 
     getHoverContent(document: LangiumDocument, params: HoverParams): MaybePromise<Hover | undefined> {
@@ -56,7 +54,7 @@ export abstract class AstNodeHoverProvider implements HoverProvider {
 
                 // Add support for documentation on keywords
                 if (isKeyword(cstNode.grammarSource)) {
-                    return this.getGrammarAstNodeHoverContent(cstNode.grammarSource);
+                    return this.getKeywordHoverContent(cstNode.grammarSource);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


packages/langium/src/lsp/hover-provider.ts
```diff
@@ -14,6 +14,9 @@ import type { MaybePromise } from '../utils/promise-utils.js';
 import type { LangiumDocument } from '../workspace/documents.js';
 import type { DocumentationProvider } from '../documentation/documentation-provider.js';
 import { findDeclarationNodeAtOffset } from '../utils/cst-utils.js';
+import { isKeyword } from '../languages/generated/ast.js';
+import { createLangiumGrammarServices } from '../grammar/langium-grammar-module.js';
```
msujew, line 18: Importing the `langium/grammar` code in general purpose code is not allowed (as it will always include the whole grammar language in all adopter bundles - which we definitely don't want). I'm actually a bit confused, we used to have a linting rule for that, which should fail the CI.

ydaveluy: Ok I understand. 
I just a pushed a new commit that fix this.

packages/langium/src/lsp/hover-provider.ts
```diff
@@ -48,13 +51,33 @@ export abstract class AstNodeHoverProvider implements HoverProvider {
...
+                }
             }
         }
         return undefined;
     }
 
     protected abstract getAstNodeHoverContent(node: AstNode): MaybePromise<Hover | undefined>;
 
+    protected getKeywordHoverContent(node: Keyword): MaybePromise<Hover | undefined> {
+        const comment = findCommentNode(node.$cstNode, ['ML_COMMENT'])?.text;
```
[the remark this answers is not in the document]
ydaveluy: Yes you're right!

I also have to keep `findCommentNode`, otherwise it does not works in the test environment.

praisethemoon:
Hello, I like this feature very much!

But as someone who documents his grammar frequently, I think it would be better to activate it explicitly for example when commenting with a special syntax:

```
entry Element: 
    /** A very cool element 
      * @hover: documentation of the element keyword 
      */ 
    'element' name=ID ;
```

This way it possible to avoid showing entire documentation on hover, but only what is intended.

What are your thoughts on this?

ydaveluy:
The only limitation I see in doing this is that it will not be possible to add some tags in the documentation for hover.
If the expected hover documentation is like below it will not work:
```
documentation of the element keyword
@remark this cannot be in a parent @hover tag
```

On one side it is less straightforward to write the documentation but on the other it permits to add  different kinds of metadata that can be retrieved at runtime based on the tag
```
entry Element: 
    /** A very cool element 
      * @hover documentation of the element keyword 
      * @info an info message for an element keyword
      * @test a test message for an element keyword
      */ 
    'element' name=ID ;

```

I don't have a use case in mind but it is more extensible.

@msujew can you decide on the matter ?

praisethemoon:
Agreed, 

But I am not sure if the regular comments are "attached" to the node like JSDoc, 
for instance I export the JSDoc of the langium grammar and probably wouldn't want them as hover information 😅 

I guess having an option to disable this default behavior would be also a good option.

But that's just my opinion :)

packages/langium/src/lsp/hover-provider.ts
```diff
@@ -48,13 +51,36 @@ export abstract class AstNodeHoverProvider implements HoverProvider {
...
 
     protected abstract getAstNodeHoverContent(node: AstNode): MaybePromise<Hover | undefined>;
 
+    protected getKeywordHoverContent(node: AstNode): MaybePromise<Hover | undefined> {
+        let comment = isAstNodeWithComment(node) ? node.$comment : undefined;
+        if (!comment) {
+            comment = findCommentNode(node.$cstNode, ['ML_COMMENT'])?.text;
+        }
+        if (comment && isJSDoc(comment)) {
+            const content = parseJSDoc(comment).toMarkdown();
```
ydaveluy, line 72: @praisethemoon For specific use-cases it is always possible to override this method and provide a different logic

```ts
const content = parseJSDoc(comment).getTag('hover')?.content.toMarkdown();
```

praisethemoon: I like this! Thanks for the tip :)

msujew (approved):

Issue #1591, opened 2024-07-21
Name collision in generated ast.ts


ydaveluy:
When defining in mydsl.langium file an interface named Reference, the generated ast.ts does not compile.

Langium version: 3.1
Package name: langium

## Steps To Reproduce

1.
define an interface named 'Reference' in mydsl.langium
```
interface Reference {
name:string
}
```
2.

Build the project

## The current behavior

The project fails to build due to a conflict with imported types from langium in src/language/generated/ast.ts:
`import type { AstNode, Reference, ReferenceInfo, TypeMetaData } from 'langium';`

## The expected behavior

Definition of interfaces like AstNode, Reference, ReferenceInfo or TypeMetaData in langium file should not collide with imports.

msujew:
We have a validation for protecting against name collisions with the JavaScript runtime. See [here](https://github.com/eclipse-langium/langium/blob/8bbaa1b3364d4443f3cee9c6664d55106e9e7035/packages/langium/src/grammar/validation/validator.ts#L987). We would need a separate list for reserved names of the Langium imports.

ydaveluy:
Another alternative would be to import types with a specific prefix (e.g. `_langium_`) and to prohibit rule names in the grammar from starting with this prefix ?

e.g:
```
import type { AstNode as _langium_AstNode, 
Reference as _langium_Reference, 
ReferenceInfo as _langium_ReferenceInfo, 
TypeMetaData as _langium_TypeMetaData } from 'langium';

```
I don't know if this concept would be valid for other reserved keywords.

ydaveluy:
Hey @msujew,

I can try to provide a PR for this issue.

I think that doing a single import like 

```
import * as langium from 'langium';
```

and adding the 'langium' keyword to reserved names would be more generic.

msujew:
Sure, that sounds reasonable to me 👍

PR #1843, opened 2025-03-12
avoid names collisions between ast and langium

ydaveluy:
Closes https://github.com/eclipse-langium/langium/issues/1591

Using type names such as AstNode, Reference, ReferenceInfo or TypeMetaData in the Langium grammar generate file ast.ts with names collision between the generated ast and Langium.

This PR, in the generated ast, import langium types in the `langium` namespace: `import * as langium from 'langium'`.

The `langium` name is also added to `reservedNames` in the grammar validator.

### DIFF
--- a/examples/arithmetics/src/language-server/generated/ast.ts
+++ b/examples/arithmetics/src/language-server/generated/ast.ts
@@ -4,8 +4,7 @@
  ******************************************************************************/
 
 /* eslint-disable */
-import type { AstNode, Reference, ReferenceInfo, TypeMetaData } from 'langium';
-import { AbstractAstReflection } from 'langium';
+import * as langium from 'langium';
 
 export const ArithmeticsTerminals = {
     WS: /\s+/,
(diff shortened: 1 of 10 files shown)


msujew (approved):
Looks pretty good to me 👍

aabounegm:
Can you please check if this change leads to any difference in the bundle size? I'm afraid that `import * as` can break tree-shaking in some bundlers (at least in earlier versions, this may have improved lately).

ydaveluy:
@aabounegm I'm not sure to understand exactly what you ask for but in one of my DSL I replaced in the generated ast.ts all the imports with `import * as langium from 'langium'` and when I compare the size of the generated vsix the sizes are the sames.

aabounegm:
Right, after further search, it seems the issue I was talking about is an edge case that does not apply here (like dynamically accessing a property on the namespace import, or some other bugs in some versions of some bundlers) 👍

Issue #1398, opened 2024-03-01
Nodes that consume no text are undefined


drhagen:
Langium version: 3.0.0

When a node succeeds but consumes no input, it is `undefined` rather than a well-defined `AstNode` object.

Consider this grammar:

```langium
grammar HelloWorld

entry Model:
    'header' persons=Persons;

Persons:
    ((elements+=ID) (elements+=ID)*)?;

hidden terminal WS: /\s+/;
terminal ID: /[_a-zA-Z][\w_]*/;
```

when parsing this file:

```
header
```

When there are no persons listed, I would expect the resulting AST to be something like `{persons: {elements: []}}`. Instead, it is `{persons: undefined}`.

msujew:
This is by design, the parser will never attempt to construct synthetic AST nodes (which `{elements: []}` would be) and doing so likely leads to very unexpected behavior, as there is no CST to associate to it.

drhagen:
That makes sense. If it is not a node but merely an attribute, it is still expected that the result is undefined? For example, this grammar

```langium
grammar HelloWorld

entry Model:
    'header'
    (persons+=ID ('+' persons+=ID)*)?
;

hidden terminal WS: /\s+/;
terminal ID: /[_a-zA-Z][\w_]*/;
```

when parsing 

```
header
```

also generates the AST `{persons: undefined}`.

This is unexpected because the generated interface does not say this is an optional attribute:

```typescript
export interface Model extends AstNode {
    readonly $type: 'Model';
    persons: Array<string>;
}
```

drhagen:
> On the other hand, the AST is guaranteed to align to the generated types in case the parser returns without any errors.

`header` is a valid document for the parser in my previous message. It parses without error, but `model.person` is still `undefined` during validation of `Model`. Does the validation code run before parsing is complete and the default values are assigned?

msujew:
> It parses without error, but model.person is still undefined during validation of Model.

Yes, that's exactly why I said it's supposed to be a grammar error. It produces a nonsensical AST, even on correct input, because the grammar isn't written in a correct way. My thoughts are: correct grammar + correct input = correct AST.

> Does the validation code run before parsing is complete and the default values are assigned?

No, validation is the last phase of the document lifecycle, see [here](https://langium.org/docs/document-lifecycle/).

PR #1848, opened 2025-03-14
Generate valid AST from parsing with recovered tokens

ydaveluy:
When an incomplete document is parsed the parser can generate tokens on the fly (recovered token).

In such case the generated AST does not match types definition: some mandatory attributes may be undefined.

This imply that accessing the AST in user services in unsafe and raise errors such as `Cannot read properties of undefined (reading 'ref')`.

This PR assign a default value to attribute parsed with an invalid token. It means that the generated is AST is compliant with its type definition.

### DIFF
--- a/packages/langium/src/parser/langium-parser.ts
+++ b/packages/langium/src/parser/langium-parser.ts
@@ -12,11 +12,11 @@ import type { LangiumCoreServices } from '../services.js';
 import type { AstNode, AstReflection, CompositeCstNode, CstNode } from '../syntax-tree.js';
 import type { Lexer, LexerResult } from './lexer.js';
 import type { IParserConfig } from './parser-config.js';
-import type { ValueConverter } from './value-converter.js';
+import type { ValueConverter, ValueType } from './value-converter.js';
 import { defaultParserErrorProvider, EmbeddedActionsParser, LLkLookaheadStrategy } from 'chevrotain';
 import { LLStarLookaheadStrategy } from 'chevrotain-allstar';
-import { isAssignment, isCrossReference, isKeyword } from '../languages/generated/ast.js';
-import { getExplicitRuleType, isDataTypeRule } from '../utils/grammar-utils.js';
+import { isAssignment, isCrossReference, isKeyword, isRuleCall } from '../languages/generated/ast.js';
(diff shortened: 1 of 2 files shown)


msujew (commented):
I'm not sure this is a good idea, see the discussion over in https://github.com/eclipse-langium/langium/issues/1398#issuecomment-1973861098, https://github.com/eclipse-langium/langium/pull/1760 and https://github.com/eclipse-langium/langium/discussions/1758. The main issue is that it misrepresents the missing value in the AST - for example, it becomes impossible to tell whether the string value `''` (empty string) was parsed from the input or whether it's a result of the error recovery.

Also, this only fixes half the problem - as it only answers the question on how to deal with properties that are using `CONSUME`. What is the default value supposed to be for elements that call `SUBRULE`?

msujew:
> I didn't think it was possible to have a rule that parse nothing.

It is (even though we show a warning that you shouldn't do that): `Test: name=ID?;`

Like I already said, it was extensively discussed in the linked issues. You can wrap the object into some `DeepOptional` type if you are in scoping and validation code to ensure that you don't accidentally access `undefined` properties. There's literally no way to please all adopters, as there are conflicting views on this topic:

1. The parsed AST should represent the input text (current `main`)
2. The parsed AST should always represent the "ideal" AST (this PR)
3. The generated AST interfaces should always have `undefined` properties, due to parser errors.

All three of those approaches are completely valid, but in conflict with one another. Some people want Langium only for LSP features (where (3) is actually ideal), while others need to generate code or do some other post-processing on the AST which requires (1) or (2). However, as parsing errors can appear during scoping and validation, doing (1) automatically requires to take care of (3), whereas (2) has the issue that it does not correctly represent the input text (and therefore makes it difficult to differentiate between in-recovery values and real parsed values).

ydaveluy:
Thanks for the clarification. I close this PR as it is not relevant.

msujew:
FYI @ydaveluy, I'd be open for a PR that makes all properties optional (if specified in the Langium config file). I think it's a valid perspective, but I don't want to force it on all adopters.

ydaveluy:
@msujew ok I will try to provide a PR next week. It seems that array attributes are already properly supported, do you confirm ? The PR would make only single attributes optional?

msujew:
Yes, exactly. Arrays and boolean properties get default values, but all others can be optional.

PR #1851, opened 2025-03-16
Alternative implementation of comment generation

dsogari:
Alternative implementation of #1835.

Noteworthy changes:

- add `offset` property in transformed types, to identify the original CST node
- add `getComments` and `registerComments` methods to `CommentProvider` interface
- modify `toAstTypesString` method of `UnionType` and `InterfaceType` to generate comments
- modify `DefaultCommentProvider` class:
  - implement registration of comments from tokens
  - reimplement `getComment` to use the new `getComments` method
  - (optional) the `findCommentNode` function is no longer necessary, so it can be removed
- modify `LangiumParser` class:
  - add `commentProvider` member to hold the comment provider instance
  - add `preprocessTokens` method to perform the association of comments with tokens
  - remove the `extractHiddenTokens`, as it is no longer necessary
  - update the `parse` and `consume` methods to use the new changes

Other important notes:

- The previous implementation of hidden token extraction had a time complexity of $O(n \cdot m)$, as it iterated over hidden tokens for each non-hidden token. The new implementation has linear complexity in the number of non-hidden tokens.
- Multiple block comments are preserved for each `ID` token, which makes it possible to add comments to multiple rule assignments; the only caveat is that only the last JSDoc comment is displayed in Intellisense.
- It would be interesting to test the performance of langium-cli on grammars with a large number of rules. Below is a sample test that can be inlcuded in [ast-generator.test.ts].
   <details>
   <summary>Spoiler</summary>

   ```ts
    test('should parse the grammar in a reasonable amount of time', { timeout: 100 }, async () => {
        const ruleCount = 1000;
        const grammar = `
            grammar TestGrammar
            entry Test: Test0;
            ${Array(ruleCount).fill(null).map((_, i) => `
                /** Test rule ${i} */
                Test${i}: name=ID;
            `).join('')}
            hidden terminal WS: /\\s+/;
            terminal ID: /[_a-zA-Z][\\w_]*/;
        `;
        const { value } = (await parse(grammar)).parseResult;
        expect(value.rules).toHaveLength(ruleCount + 3);
    });
   ```

   </details>

[ast-generator.test.ts]: https://github.com/eclipse-langium/langium/blob/main/packages/langium-cli/test/generator/ast-generator.test.ts

### DIFF
--- a/packages/langium/src/documentation/comment-provider.ts
+++ b/packages/langium/src/documentation/comment-provider.ts
@@ -22,22 +22,22 @@ export interface CommentProvider {
     getComment(node: AstNode): string | undefined;
 
     /**
-     * Returns the comment associated with the specified offset.
-     * @param node The AST node to get the comment for.
-     * @returns The comment associated with the specified offset or `undefined` if there is no comment.
+     * Returns the comments associated with the specified offset.
+     * @param offset The source offset get the comment for.
+     * @returns The comments associated with the specified offset or `undefined` if there is no comment.
      */
-    getCommentByOffset(offset?: number): string | undefined;
+    getComments(offset?: number): ReadonlyArray<string> | undefined;
 
     /**
-     * Registers the documentation comment of a token.
+     * Registers the comments of a token.
      * @param token The token
      */
-    registerComment(token: IToken): void;
+    registerComments(token: IToken): void;
 }
 
 export class DefaultCommentProvider implements CommentProvider {
     protected readonly grammarConfig: () => GrammarConfig;
-    protected readonly offsetToComment: Map<number, string> = new Map();
+    protected readonly commentsByOffset: Map<number, Array<string>> = new Map();
     constructor(services: LangiumCoreServices) {
         this.grammarConfig = () => services.parser.GrammarConfig;
     }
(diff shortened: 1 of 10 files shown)


dsogari:
After all, I think this issue deserves a design discussion.

packages/langium/src/parser/langium-parser.ts
```diff
@@ -231,18 +233,39 @@ export class LangiumParser extends AbstractLangiumParser {
...
+    private processHiddenTokens(lexerResult: LexerResult): Array<IToken> { // O(n) where n is the number of tokens
+        const { tokens, hidden } = lexerResult;
+        this.commentProvider.clearComments();
+        let j = 0;
+        for (let i = 0, last = 0, lastId = 0; i < tokens.length; ) {
+            if (j == hidden.length || tokens[i].startOffset < hidden[j].startOffset) {
+                const token = tokens[i++];
+                let payload = token.payload = hidden.slice(last, j);
+                last = j; // save index for immediate comments
+                if (['entry', 'ID'].includes(token.tokenType.name)) {
```
msujew, line 245: Langium can never make any assumptions about the names used by the adopter language.

packages/langium/src/documentation/comment-provider.ts
```diff
@@ -20,17 +20,52 @@ export interface CommentProvider {
...
+
+    /**
+     * Clears the comments.
+     */
+    clearComments(): void;
 }
 
 export class DefaultCommentProvider implements CommentProvider {
     protected readonly grammarConfig: () => GrammarConfig;
+    protected readonly commentsByOffset: Map<number, Array<string>> = new Map();
```
msujew, line 46: Having state in services is generally pretty undesirable.

packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
```diff
@@ -21,13 +21,15 @@ interface TypePart {
     children: TypePart[]
     end?: TypePart
     actionWithAssignment: boolean
+    offset?: number // used as identifier
```
msujew, line 24: Using the offset of the cst node (which in the end just retrieves the content of the CST node anyway) seems unnecessarily complicated to me. Why not store the comment text associated with the AST node directly?

msujew (commented):
Hey @dsogari,

thanks for looking into this, but I feel like the changes are a bit over-engineered. See some of the comments below. I think we should move forward with https://github.com/eclipse-langium/langium/pull/1835 instead.

dsogari:
@msujew That's alright. Thanks for taking the time to review this. It was fun diving deep into Langium's codebase, nonetheless.

PR #1860, opened 2025-03-19
Adds a BNF generator to generate GBNF definition from the Langium grammar

dhuebner:
Adds an optional BNF Generator to Langium Configuration:
```js
/** Enable GBNF file generation */
bnf?: {
    /** Output path for GBNF file */
    out: string,
    /** Dialect of the generated BNF file. GBNF is the default. In EBNF RegEx terminals are not supported. */
    dialect?: 'GBNF' | 'EBNF'
}
``` 
 
The option `GBNF` will produce a grammar file corresponding to this [description](https://github.com/ggml-org/llama.cpp/blob/master/grammars/README.md).

An experimental `EBNF` option will produce an EBNF like grammar definition. Note: in most cases one will need to rewrite the Regex terminal rules to the EBNF syntax.

### DIFF
--- a/packages/langium-cli/src/generate.ts
+++ b/packages/langium-cli/src/generate.ts
@@ -5,7 +5,7 @@
  ******************************************************************************/
 
 import type { AstNode, Grammar, LangiumDocument, Mutable } from 'langium';
-import type { LangiumConfig, LangiumLanguageConfig} from './package-types.js';
+import type { LangiumConfig, LangiumLanguageConfig } from './package-types.js';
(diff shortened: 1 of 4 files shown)


montymxb (commented):
Some general feedback. I'll also give this a spin on another grammar that's medium size & see how it holds up in practice.

montymxb (approved):
:octocat: Alright I think this is suitable to take in as is. Generated EBNF will likely have follow-ups to improve what's produced, which we can apply iteratively.

PR #1861, opened 2025-03-19
fix generation issue on AST getReferenceType

### LINKED ISSUE #1857: AstReflection:getReferenceType generation issue
The `AstReflection:getReferenceType` method in `ast.ts` is not properly generated in some cases and an exception with message `XXX:YYY is not a valid reference id.` is raised.

Langium version: 3.4
Package name: langium-cli

## Steps To Reproduce

1. Generate language from this grammar
```
grammar HelloWorld
interface NamedElement {
    name:string
}
interface WithRef extends NamedElement {
    ref:@NamedElement
}
interface MyElement extends NamedElement, WithRef {
}
entry MyElement returns MyElement:
    name=ID ref=[NamedElement:ID];
hidden terminal WS: /\s+/;
```

The switch case for `MyElement:ref` is not provided in `HelloWorldAstReflection:getReferenceType`. At runtime an exception with message `MyElement:ref is not a valid reference id.` is raised.

## The current behavior
 The getReferenceType method generated in the ast.ts is:
```ts
    getReferenceType(refInfo: ReferenceInfo): string {
        const referenceId = `${refInfo.container.$type}:${refInfo.property}`;
        switch (referenceId) {
            case 'WithRef:ref': {
                return NamedElement;
            }
            default: {
                throw new Error(`${referenceId} is not a valid reference id.`);
            }
        }
    }
```

## The expected behavior

 The getReferenceType method generated in the ast.ts should be:
```ts
    getReferenceType(refInfo: ReferenceInfo): string {
        const referenceId = `${refInfo.container.$type}:${refInfo.property}`;
        switch (referenceId) {
            case 'MyElement:ref':
            case 'WithRef:ref': {
                return NamedElement;
            }
            default: {
                throw new Error(`${referenceId} is not a valid reference id.`);
            }
        }
    }
```

Note:
When WithRef is defined like below it works properly:
```
interface WithRef {
    ref:@NamedElement
}
```

ydaveluy:
Closes #1857 

There was a false assumption in the implementation.

This PR iterate a second time on the interfaces to process super-types cross references.

### DIFF
--- a/packages/langium-cli/src/generator/ast-generator.ts
+++ b/packages/langium-cli/src/generator/ast-generator.ts
@@ -176,16 +176,20 @@ function buildCrossReferenceTypes(astTypes: AstTypes): CrossReferenceType[] {
                 });
             }
         }
-        // Since the types are topologically sorted we can assume
-        // that all super type properties have already been processed
+    }
+
+    for (const typeInterface of astTypes.interfaces) {
+        const superFeatures = new Set<string>();
         for (const superType of typeInterface.interfaceSuperTypes) {
-            const superTypeCrossReferences = crossReferences.get(superType.name).map(e => ({
-                ...e,
-                type: typeInterface.name
-            }));
-            crossReferences.addAll(typeInterface.name, superTypeCrossReferences);
+            for (const superTypeCrossReference of crossReferences.get(superType.name)) {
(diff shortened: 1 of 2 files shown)


msujew (approved):
Thank you, that makes sense 👍

PR #1863, opened 2025-03-20
do not require properties with default values in rules

### LINKED ISSUE #1862: Interface properties with a default value should not be required in a rule
When an interface declare a property with a defaultValue, it should not be required to define this property in a rule (like for optional properties).
Langium version: 3.4.0
Package name: langium

## Steps To Reproduce

Test the following grammar

```
grammar HelloWorld

interface Person {
    name: string = ''
}
entry Person returns Person:
    {Person} 'person' ;

hidden terminal WS: /\s+/;

```

## The current behavior

A validation error is raised on rule `Person`: `A property 'name' is expected. in a rule that returns type 'Person'.`

## The expected behavior
The grammar should be valid.

ydaveluy:
fix #1862

When a property has a default value it should not be required to assign it in a rule.


spoenemann (approved):
Thanks!

Issue #1871, opened 2025-03-26
Provides filenames support in langium-config.json


ydaveluy:
In `package.json` it is possible to define a language with extensions and filenames:

```json
{
          "id": "mylanguage",
          "aliases": [
              "MyLanguage"
          ],
          "filenames": [
              "a_file_name.myext"
          ],
          "configuration": "./language-configuration.json"
}
```
In `langium-config.json`, it is only possible to provides an array of extensions.
```json
 {
        "id": "mylanguage",
        "grammar": "src/model/mylanguage.langium",
        "fileExtensions": [ ".myext"],
        "textMate": {
            "out": "syntaxes/mylanguage.tmLanguage.json"
        }
    },
```
To handle a specific filename it is required to override some Langium services and to use a custom filter to process only `a_file_name.myext` and not `*.myext`.

Langium could provide by default filenames in `langium-config.json`

montymxb:
Hi @ydaveluy this seems like a reasonable enough request. I'll see if I can't take a look at the PR later today.

ydaveluy:
Hi @montymxb 
FYI VSCode provides support for extensions, file names, filename patterns, firstLine and mimetypes: https://code.visualstudio.com/api/references/contribution-points#contributes.languages

```
Associate file name extensions (extensions), file names (filenames), file name [glob patterns](https://code.visualstudio.com/docs/editor/glob-patterns) (filenamePatterns), files that begin with a specific line (such as hashbang) (firstLine), 
and mimetypes to that languageId.
```

In #1873, I only added support for file names, but the new API can be extended later without breaking it again to provide support for filenamePatterns, firstLine, and mimetypes. These last three elements are more complex to implement and would likely require a third-party library.

spoenemann:
Fixed with #1873

PR #1873, opened 2025-03-27
add support for languages with specific file names

ydaveluy:
Solves #1871

Add an optional fileNames entry in the langium configuration.

In VSCode contributions it is also possible to provide a filenamePatterns: https://code.visualstudio.com/api/references/contribution-points#contributes.languages but to add this to Langium it would requires to add a glob library as dependency.

It changes the API of traverseFolder and includeEntry in workspace-manager.ts so maybe for 4.0 ?

### DIFF
--- a/packages/langium/src/service-registry.ts
+++ b/packages/langium/src/service-registry.ts
@@ -103,9 +103,9 @@ export class DefaultServiceRegistry implements ServiceRegistry {
 
         if (!services) {
             if (languageId) {
-                throw new Error(`The service registry contains no services for the file '${name}' for language '${languageId}'.`);
+                throw new Error(`The service registry contains no services for the extension '${ext}' for language '${languageId}'.`);
             } else {
-                throw new Error(`The service registry contains no services for the file '${name}'.`);
+                throw new Error(`The service registry contains no services for the extension '${ext}'.`);
             }
         }
         return services;
--- a/packages/langium/src/service-registry.ts
+++ b/packages/langium/src/service-registry.ts
@@ -44,6 +44,7 @@ export class DefaultServiceRegistry implements ServiceRegistry {
     protected singleton?: LangiumCoreServices;
     protected readonly languageIdMap = new Map<string, LangiumCoreServices>();
     protected readonly fileExtensionMap = new Map<string, LangiumCoreServices>();
+    protected readonly fileNameMap = new Map<string, LangiumCoreServices>();
 
     /**
      * @deprecated Use the new `fileExtensionMap` (or `languageIdMap`) property instead.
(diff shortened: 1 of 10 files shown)


spoenemann (commented):
Shall we unify the two into `filePatterns` where an extension is a special case translated to `*.ext`?

It should be easy to implement by translating the pattern to a regexp: `*` becomes `.*` and `?` becomes `.`

packages/langium/src/service-registry.ts
```diff
@@ -89,12 +98,14 @@ export class DefaultServiceRegistry implements ServiceRegistry {
...
+        const name = UriUtils.basename(uri);
+        const services = this.fileNameMap.get(name) ?? this.fileExtensionMap.get(ext);
+
         if (!services) {
             if (languageId) {
-                throw new Error(`The service registry contains no services for the extension '${ext}' for language '${languageId}'.`);
+                throw new Error(`The service registry contains no services for the file '${name}' for language '${languageId}'.`);
             } else {
-                throw new Error(`The service registry contains no services for the extension '${ext}'.`);
+                throw new Error(`The service registry contains no services for the file '${name}'.`);
```
montymxb, line 108: We should refactor this to retain the existing message when an extension entry is retrieved vs. an entry for a file. Might be confusing otherwise when all messages are changed, especially considering that all projects have just a list of extensions right now.

packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -66,6 +66,10 @@ export interface WorkspaceManager {
     initializeWorkspace(folders: WorkspaceFolder[], cancelToken?: CancellationToken): Promise<void>;
 
 }
+export interface FileSelector {
```
montymxb, line 69: Would be nice to add a little block comment to this one too

packages/langium/test/service-registry.test.ts
```diff
@@ -41,6 +41,16 @@ describe('DefaultServiceRegistry', () => {
         expect(registry.getServices(URI.parse('file:/test.y'))).toBe(language2);
         expect(registry.all).toHaveLength(2);
     });
+    test('a file name has a highest priority than an extension', () => {
```
montymxb, line 44: Small typo:
> higher priority

montymxb (commented):
Nice! This looks pretty good so far. Just a handful of comments for a couple small things that would be nice to adjust, and one requested change for how we report errors in the service registry.

packages/langium/src/service-registry.ts
ydaveluy: Ok, it was useless to change the message.

montymxb (approved):
:octocat: approved! Plus thanks for the changes.

montymxb:
One last note before we pull this in. Per @spoenemann 's comment above, we can still consider a follow-up PR where we unify the two patterns we're now using.

PR #1875, opened 2025-03-28
add DeepPartialAstNode utility

ydaveluy:
Add an utility that convert the properties of an AstNode to optional.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,10708 +1,10708 @@
 {
-    "name": "langium-workspaces",
-    "lockfileVersion": 3,
-    "requires": true,
-    "packages": {
-      "": {
-        "name": "langium-workspaces",
-        "workspaces": [
-          "packages/langium",
-          "packages/langium-railroad",
-          "packages/langium-cli",
-          "packages/langium-sprotty",
-          "packages/langium-vscode",
-          "packages/generator-langium",
-          "examples/arithmetics",
-          "examples/domainmodel",
-          "examples/requirements",
-          "examples/statemachine"
-        ],
-        "devDependencies": {
-          "@types/node": "~16.18.41",
-          "@types/vscode": "~1.67.0",
-          "@typescript-eslint/eslint-plugin": "~6.4.1",
-          "@typescript-eslint/parser": "~6.4.1",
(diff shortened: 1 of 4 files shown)


ydaveluy:
@msujew For the unit test I added ts-expect as devDependency in Langium package.
To work I had to do `npm install` but it introduced lots of changes in `package-lock.json` and the updated versions seems to break something in the `yeoman-generator.test.ts`.

How can I solve this issue ?

spoenemann (commented):
The change to package.json is missing.

Maybe we can implement the test without the additional dependency? You could write an assignment to a variable so we use the compiler's type compatibility check as test.

ydaveluy:
@spoenemann Yes in https://github.com/eclipse-langium/langium/pull/1875/commits/18f2a4083fa62865df978d04404c4e58894d9141 I reverted the import from ts-expect and added `expectType` for the test.

```ts
export const expectType = <Type>(_: Type): void => void 0;
```
I forget to revert the changes in `package-lock.json` but it is ok now.

spoenemann (approved):
👍

sorawee:
Nice to see this implemented :) Should we also fix various mistakes in the example directory? For example, there's a mistake reported in https://github.com/eclipse-langium/langium/discussions/1758, which got fixed in https://github.com/eclipse-langium/langium/pull/1760, but the PR was abandoned.

I think fixing these mistakes would also be a good opportunity to show case this utility type.

-----

Off-topic: I'm curious why this PR doesn't need to check against `AstNode | undefined`. My earlier proposal in #1760  does this:

```
export type OptionalAstNode<T extends AstNode> = {
  [P in keyof T]: 
    P extends keyof AstNode ? 
      T[P] : 
      T[P] extends AstNode[] ? 
        OptionalAstNode<T[P][number]>[] : 
        T[P] extends any[] | boolean ? 
          T[P] : 
          T[P] extends AstNode | undefined ? 
            OptionalAstNode<Extract<T[P], AstNode>> | undefined :  
            (T[P] extends AstNode ? OptionalAstNode<T[P]> : T[P]) | undefined;
};
```

PR #1878, opened 2025-03-28
Improve case insensitive Regex generation

msujew:
Our current way of representing case insensitive tokens in the token builder and textmate grammars is rather unusual. Instead of using `[xX]` for every possible character `x`, we can make use of the `i` or `(?i)` regex options instead. This massively improves readability of the generated textmate grammars.


spoenemann (approved):
👍

PR #1889, opened 2025-04-02
Completed the descriptions in the example README.md how to execute the examples

JohannesMeierSE:
This issue was discussed in #1768

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ node_modules/
 *.vsix
 *.tsbuildinfo
 hello-world
+generated/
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -9,3 +9,4 @@ node_modules/
 *.vsix
 *.tsbuildinfo
 hello-world
+generated/
```
msujew, line 12: We don't want to exclude changes to the `generated` files in this repo, as that would make (eventually) it impossible to bootstrap the Langium grammar. It also serves as a regression test for changes in the `langium-cli` package.

JohannesMeierSE: You are completely right, thanks for the hint! I just reverted the change and rebased the PR.

PR #1895, opened 2025-04-04
Value converters can generate errors on invalid input

spoenemann:
See https://github.com/eclipse-langium/langium/discussions/1887#discussioncomment-12725628


spoenemann:
We need a better solution.

PR #1908, opened 2025-04-15
eslint 9 update

### LINKED ISSUE #1904: Transform code and packages to ESM with exception of langium-vscode
The packages should ship only ESM code and be transformed to `"type"="module"`. With the exception of `langium-vscode` packages should no longer ship CommonJS code.

The TypeScript compiler configuration should move away from "module": "CommonJS" (`"target": "ES2022"`, `"module": "NodeNext"` and `"moduleResolution": "nodenext"`) . The exception will be the VSCode extension code where the bundled code must be CommonJS and and the package format must stay the same.

Node 18 will be EOL in April 2025 therefore it is a good time to move the minimum engine requirements at least to node 20.

### LINKED ISSUE #1905: Migrate to eslint 9
We should migrate to eslint 9. We have to transform to the configuration and adapt some of the rules (https://github.com/eclipse-oct/open-collaboration-tools/pull/118 serves as a very good template).

kaisalmen:
The [eslint configuration](https://github.com/eclipse-oct/open-collaboration-tools/blob/main/eslint.config.mjs) of open-collaboration-tools has been the baseline for this implementation. Specific adjustment from the old configuration have been added.

The TypeScript configuration has been clean and aligned with [OCT](https://github.com/eclipse-oct/open-collaboration-tools/blob/main/tsconfig.json) as well.
I needed to lower the compiler target to `ES2020` because otherwise almost half of the unit test using the chevrotain parser failed. The lib reference was lowered from `ESNEXT` to `ES2020` as well to ensure lib and target are in-line. This required some code changes, because `string.replaceAll` is only available after `ES2021`.
This PR also bumps the minimum engines up to node 20. `ES2022`/`ES2023` should be valid TS/node20 target according to this [list from TypeScript](https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping), but it doesn't help if our dependency stacks doesn't work with (potential chevrotain issue/investigation required?).

All dependencies but yeoman-generator were updated to current versions, because of incompatible API changes. This should be tracked in a new issue.

One hack remains: I can't set `'no-useless-escape': 'off'` inherited rule from `eslint:recommended`, although it should it work. 🤷‍♂️  It is taken into account, because once I configure an illegal string eslint does not start. I excluded the generated **syntaxes/*.monarch.ts** files from processing. The real cause `/[/\*]/` (backslah not required) in the monarch generator should be fixed at one point, but still the config override must work. **Update:** I just saw those files were contained in the deleted eslintignore as well, so the problem is not new.

Fixes #1904
Fixes #1905


kaisalmen:
Next time I will separate compiler update, eslint update and required dependencies into separate PRs. From the recent upgrade experience in OCT I thought this is an straightforward task, but larger code bases bring more complexity. I know that and should not fall into the trap. 🙂 Implementation becomes harder and the review as well. My apologies for making this review more complex than needed.

kaisalmen:
Is replaced by #1930 and other PRs following.

PR #1921, opened 2025-05-14
Await language client start in generated extension

msujew:
See discussion in https://github.com/eclipse-langium/langium/discussions/1801#discussioncomment-13121766. In some cases, the language server didn't pick up on dirty files in the editor, since we didn't await the startup sequence of the language client. This change adjusts this for our examples & yeoman generator.


spoenemann (approved):

spoenemann:
Shall we plan a 3.6.0 release? Then we could create a respective milestone and cherry-pick this change to the `maintenance-3.x` branch.

msujew:
@spoenemann I don't think this change has any immediate need to be released. We could release it as a 3.5.1 for the generator package, if needed.

PR #1930, opened 2025-05-17
Updated tsc and tsconfig. Node raised to v20

kaisalmen:
This is a follow-up from #1908 It does the same smaller steps.

The TypeScript configuration has been clean and aligned with [OCT](https://github.com/eclipse-oct/open-collaboration-tools/blob/main/tsconfig.json).
I needed to lower the compiler target to ES2020 because otherwise almost half of the unit test using the chevrotain parser failed. The lib reference was lowered from ESNEXT to ES2020 as well to ensure lib and target are in-line. This required some code changes, because `string.replaceAll` is only available after ES2021.
This PR also bumps the minimum engines up to node 20. ES2022/ES2023 should be valid TS/node20 target according to this [list from TypeScript](https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping).

Additionally, yeoman-generator was updated already due to problems with the new tsconfig. Some files forgotten in #1810 are also removed. `shx` version is aligned with the generator.

eslint and full dependencies update will follow in a separate PR.


package.json
```diff
@@ -36,15 +37,15 @@
     "eslint": "~8.56.0",
     "eslint-plugin-header": "~3.1.1",
     "shx": "~0.3.4",
```
msujew, line 39: Suggestion: We should also update `shx` to `0.4.0`. This version is using an outdated version of `glob`.

kaisalmen: Yes, that sounds like a good idea and I will drop eslint as we discussed offline.

kaisalmen: Done ⬇️

msujew (approved):
Looks good to me 👍

PR #1937, opened 2025-05-20
Improve range calculation for linking errors

msujew:
In some rare cases (grammar specific), the error would appear on the incorrect range. Instead of searching through the CST for the range of the reference, we should simply use `$refNode.range`, which is already available at this point and can be accessed without expensive search.


spoenemann (approved):
👍

PR #1939, opened 2025-05-22
langium DI: fixed erroneous source module modification in 'Module.merge'

sailingKieler:
* added corresponding tests

### DIFF
--- a/packages/langium/src/dependency-injection.ts
+++ b/packages/langium/src/dependency-injection.ts
@@ -20,6 +20,16 @@ export type Module<I, T = I> = {
 }
 
 export namespace Module {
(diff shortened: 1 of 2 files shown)


JohannesMeierSE (changes requested):
Thanks @sailingKieler for fixing this nasty bug!

The implementation and the test cases look very nice! I only would like to see some more comments.

I suggest to add a documention for `Module.merge`, including the explicit hint, that both `m1` and `m2` are unchanged / read-only and a *new* module is returned.

JohannesMeierSE (approved):
Thanks for the comments!

Issue #1755, opened 2024-11-18
Provide a Builder/Functions for creating ASTs without a CST


nametable:
I find that a builder to create an AST programmatically would be helpful for generating code targeting a language built using Langium. My current approach is to use the `CompositeGeneratorNode` syntax to manually craft text which should be compliant with my language's syntax, however this can be cumbersome and prone to error.

My thought is that if a person implements a formatter for an AST, it should be able to produce the CST and thus spit out valid syntax. The construction of the AST node types however currently requires setting the parent `$container` of each AST node.

Is there a recommended way to do something like this already, or would adding AST node generator functions/utilities to Langium potentially make sense?

msujew:
Hey @nametable,

> My thought is that if a person implements a formatter for an AST, it should be able to produce the CST and thus spit out valid syntax.

Note that the current formatter implementation operates on the existing CST, and does not produce a CST (i.e. it simply shifts the existing text nodes, and does not create any nodes).

> The construction of the AST node types however currently requires setting the parent $container of each AST node.

You can just set the `$container` property to `undefined!` and then run the [`linkContentToContainer`](https://eclipse-langium.github.io/langium/functions/langium.AstUtils.linkContentToContainer.html) function on everything.

> Is there a recommended way to do something like this already, or would adding AST node generator functions/utilities to Langium potentially make sense?

The recommendation is to create a manual serializer right now (as you're already doing). What you're looking for is an automatic serializer. However, building a completely generic, grammar agnostic serializer is *really* complicated, and while it can be done (see Xtext's serializer), we currently don't have any plans to work on that.

spoenemann:
See this article about developing a generic serializer: https://www.typefox.io/blog/lionweb-langium/

nametable:
Interesting read, is there a current issue for tracking the design and implementation of such a serializer. I'm just trying to think of possible solutions, as this is an interesting problem.

For the comments, I wonder if "full fidelity parsing" as described [here](https://langdev.stackexchange.com/a/2356) could be a solution to that. In his case, they used leading and trailing "trivia".

Also maybe its philosophical, but if the grammar rules define overlapping alternatives, does it matter which one is chosen during serialization, if parsing the produced text would produce the same AST again? :sweat_smile:

spoenemann:
There's no issue or plan yet. Maybe @JohannesMeierSE can say more.

In any case, it's a huge endeavor. Xtext has this feature, and it's very complex:
[...]

PR #1942, opened 2025-06-06
Refactored AST reflection

### LINKED ISSUE #1184: Export property types for checking at compile time

spoenemann:
Closes #1184.

The new implementation of AstReflection no longer needs overriding any of the generic methods. All the necessary metadata is in the `types` field.

In addition, the new structure is useful for type-safe access to AST property names, e.g. `reflection.types.MyType._myProperty`. The underscore prefix is necessary to separate the property name constants from the metadata properties.

### DIFF
--- a/packages/langium/src/grammar/type-system/type-collector/types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/types.ts
@@ -453,15 +453,13 @@ export function isMandatoryPropertyType(propertyType: PropertyType): boolean {
 
 function addReflectionInfo(name: string, properties: Property[] = []): Generated {
     return expandToNode`
-        /** @deprecated Use \`$${name}.$type\` instead. */
-        export const ${name} = '${name}';
-        export const $${name} = {
+        export const ${name} = {
             $type: '${name}'${properties.length > 0 ? ',' : ''}
             ${joinToNode(properties.sort((a, b) => a.name.localeCompare(b.name)), prop => `${prop.name}: '${escapeQuotes(prop.name, "'")}'`, { separator: ',', appendNewLineIfNotEmpty: true })}
         } as const;
 
         export function is${name}(item: unknown): item is ${name} {
-            return reflection.isInstance(item, $${name}.$type);
+            return reflection.isInstance(item, ${name}.$type);
         }
     `.appendNewLine();
 }
--- a/examples/arithmetics/src/language-server/generated/ast.ts
+++ b/examples/arithmetics/src/language-server/generated/ast.ts
@@ -35,38 +35,32 @@ export type ArithmeticsTokenNames = ArithmeticsTerminalNames | ArithmeticsKeywor
 
 export type AbstractDefinition = DeclaredParameter | Definition;
 
-/** @deprecated Use `$AbstractDefinition.$type` instead. */
-export const AbstractDefinition = 'AbstractDefinition';
-export const $AbstractDefinition = {
+export const AbstractDefinition = {
     $type: 'AbstractDefinition'
 } as const;
 
 export function isAbstractDefinition(item: unknown): item is AbstractDefinition {
-    return reflection.isInstance(item, $AbstractDefinition.$type);
+    return reflection.isInstance(item, AbstractDefinition.$type);
 }
 
 export type Expression = BinaryExpression | FunctionCall | NumberLiteral;
 
-/** @deprecated Use `$Expression.$type` instead. */
-export const Expression = 'Expression';
-export const $Expression = {
+export const Expression = {
     $type: 'Expression'
 } as const;
 
 export function isExpression(item: unknown): item is Expression {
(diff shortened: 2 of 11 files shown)


ydaveluy:
Hey @spoenemann,

And what about prefixing the metadata properties  `name`, `properties` and `superTypes` with `$` to avoid the underscore prefix for the property name constants ?

Something like:
```ts
 BinaryExpression: {
            $name: BinaryExpression,
            $properties: {
                left: {
                    name: 'left'
                },
                operator: {
                    name: 'operator'
                },
                right: {
                    name: 'right'
                }
            },
            $superTypes: ['Expression'],
            // Property name constants
            left: 'left',
            operator: 'operator',
            right: 'right'
        },
```

spoenemann:
Rethinking this, the property constants serve a very different purpose than the AstReflection service. The former are used in language-specific code to have stable access to AST types so you notice it with compiler errors when something has changed in the grammar spec. The latter is used for generic services that should work for all Langium languages.

I like @guiliannewiger's proposal in https://github.com/eclipse-langium/langium/issues/1184#issuecomment-1708519920 and updated this PR according to that.

I first thought it could be a problematic change to generate objects instead of string constants, but if you already have code like `context.container.$type === Person`, you'll get an error message:
```
This comparison appears to be unintentional because the types 'string' and '{ readonly $type: "Person" }' have no overlap.
```
The solution is to change the code to `Person.$type`. We should mark this very clearly in the CHANGELOG.

packages/langium/src/grammar/type-system/type-collector/types.ts
```diff
@@ -451,12 +451,15 @@ export function isMandatoryPropertyType(propertyType: PropertyType): boolean {
...
 }
 
-function addReflectionInfo(name: string): Generated {
+function addReflectionInfo(name: string, properties: Property[] = []): Generated {
     return expandToNode`
-        export const ${name} = '${name}';
+        export const ${name} = {
+            $type: '${name}'${properties.length > 0 ? ',' : ''}
+            ${joinToNode(properties.sort((a, b) => a.name.localeCompare(b.name)), prop => `${prop.name}: '${escapeQuotes(prop.name, "'")}'`, { separator: ',', appendNewLineIfNotEmpty: true })}
+        } as const;
```
msujew, line 459: Suggestion: Do we want to add a `create` function in here as a factory method? This would resolve https://github.com/eclipse-langium/langium/issues/1755. However, I'm fine with adding this later in a separate PR.

msujew (approved):
Looks great, thank you. I'll mention this in detail when writing the changelog for version 4.0 of Langium.

packages/langium/src/grammar/type-system/type-collector/types.ts
spoenemann: Wouldn't you just use normal object notation to create AST nodes programmatically?

Issue #841, opened 2022-12-15
Evaluate opportunities for indicating gray space in vs code


sailingKieler:
An initial draft of indicating smart template properties was contributed with https://github.com/langium/langium/commit/d4e0720563a9af9fbefdf070791f6d626f15ac5d
that I deactivated before Langium's 1.0 release.

Wish: Evaluate opportunities for taking this up again!

spoenemann:
See also #24.

PR #1943, opened 2025-06-10
Removed unused code from VS Code extension

spoenemann:
As long as we don't solve #841, we should remove the dead code.


sailingKieler (approved):
Fine for me 👍

PR #1951, opened 2025-06-12
Added grammar validation config, added strict mode for types

### LINKED ISSUE #385: The strict mode for types inference

spoenemann:
Closes #385.

This introduces a vscode setting for the Langium config file path and sets up configuration change watchers so all grammar files are revalidated when either the config file path or the config file content changes.

The actual strict mode is then a rather easy change, forbidding inferred types in your grammar definitions.

### DIFF
--- a/examples/statemachine/src/language-server/generated/ast.ts
+++ b/examples/statemachine/src/language-server/generated/ast.ts
@@ -70,7 +70,7 @@ export function isState(item: unknown): item is State {
     return reflection.isInstance(item, State);
 }
 
-/** A textual represntation of a state machine */
+/** A textual representation of a state machine */
 export interface Statemachine extends langium.AstNode {
     readonly $type: 'Statemachine';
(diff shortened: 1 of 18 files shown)


msujew (approved):
Looks good to me. The changes to the file watching interface also make a lot of sense 👍

PR #1954, opened 2025-06-13
Enable explicit types for infix rules

### LINKED ISSUE #1949: Allow to return declared types from infix rules
Currently the new infix rules cannot use `returns DeclaredTypeName` or `infers InferredTypeName` to change the generated type.

Additionally, if an adopter uses `returns`, the language server should validate it's usage against the expected field names `left`, `right` and `operator`.

spoenemann:
Closes #1949.

### DIFF
--- a/packages/langium-cli/src/generator/ast-generator.ts
+++ b/packages/langium-cli/src/generator/ast-generator.ts
@@ -48,8 +48,9 @@ function generateAstReflection(config: LangiumConfig, astTypes: AstTypes): Gener
             override readonly types = {
                 ${joinToNode(typeNames, typeName => {
                     const interfaceType = astTypes.interfaces.find(t => t.name === typeName);
-                    if (interfaceType) {
-                        const props = interfaceType.superProperties;
+                    const unionType = astTypes.unions.find(t => t.name === typeName);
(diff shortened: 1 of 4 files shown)


spoenemann:
I also included a fix for a bug introduced with #1942: we need to generate metadata for union types, too, so the AstReflection can include them in the subtype/supertype computation.

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -61,6 +61,17 @@ function generateAstReflection(config: LangiumConfig, astTypes: AstTypes): Gener
...
+                    if (unionType) {
+                        const superTypes = typeHierarchy.superTypes.get(typeName) || [];
+                        return expandToNode`
+                            ${typeName}: {
+                                name: ${typeName}.$type,
+                                properties: {},
+                                superTypes: [${superTypes.map(t => `'${t}'`).join(', ')}]
+                            }
+                        `;
+                    }
```
msujew: Suggestion: Why not merge the two branches? AFAIK the only difference is the `props`, which can be explicitly added for an `InterfaceType`. Also, we should use `${t}.$type` as well here.

PR #1965, opened 2025-06-23
Supplement of the generator API

sailingKieler:
This PR contributes 2 supplements to the generator API of 'langium/generate'.

1. `CompositeGeneratorNode.prepend(...)` & `CompositeGeneratorNode.prependIf(...)`:
    This enables prepending pre-composed generator node with a prefix, esp. depending on previously executed logic creates composite generator node conditionally. This new API facilitates expressions like
    ```js
   expandToNode`
     foo {
       ${generateBarIfPresent(foo.bar)?.prepend('bar: ')}${generateFlagIfPresent(foo.flag)?.prepend(' ')}
     }
    `
    ```
    that prepend some preceding text or whitespace to conditionally created content depending on its presence.

2. Multiple linebreak separator insertion by `joinToNode(list, toStringFunc, { appendNewLineIfNotEmpty: 3 })`:
   The current version `joinToNode(...., { appendNewLineIfNotEmpty: true })` permits single linebreaks only.

   Attempting to achieve the generation of elements of a list with the generated representations being separated by one or more empty lines — i.e. inserting more than one linebreak — quickly ends in a hassle, if those separating lines shall not be added to the end of the list or produced if the list is empty. For example:
   ```js
   expandToNode`
       ${joinToNode(model.declarations, generateDeclaration, { appendNewLineIfNotEmpty: 2, skipNewLineAfterLastItem: model.statements.isEmpty()})}
       ${joinToNode(model.statements,   generateStatement,   { appendNewLineIfNotEmpty: 2, skipNewLineAfterLastItem: true })}
   `
   ```
   This way all the `declarations` and `statements` are separated by 2 line breaks -> 1 empty line with:
      * no extra leading line breaks are added if `model.declarations` is empty
      * no extra trailing line breaks are added if `model.statements` is empty
      * no linebreaks are added at all if both lists are empty
    
   This contribution adds support for 1 to 5 empty separator lines requiring up to 6 linebreaks.
   Thus, `appendNewLineIfNotEmpty` now accepts values of the union `true | 1 | 2 | 3 | 4 | 5 | 6`.
   
   In order to implement this, I extended `NewLineNode` and the generator node processor correspondingly.

### DIFF
--- a/packages/langium/src/generate/generator-node.ts
+++ b/packages/langium/src/generate/generator-node.ts
@@ -1117,9 +1117,9 @@ export function traceToNodeIf<T extends AstNode>(condition: boolean, source: T |
  */
 export class IndentNode extends CompositeGeneratorNode {
 
-    indentation?: string;
-    indentImmediately = true;
-    indentEmptyLines = false;
+    readonly indentation?: string;
+    readonly indentImmediately;
+    readonly indentEmptyLines;
 
     constructor(indentation?: string | number, indentImmediately = true, indentEmptyLines = false) {
         super();
@@ -1133,19 +1133,22 @@ export class IndentNode extends CompositeGeneratorNode {
     }
 }
 
+export namespace NewLineNode {
+    /**
+     * Integer range type allowing to specify 1 to 6 consecutive line breaks, i.e. up to 5 empty lines with a single `NewLineNode`.
+     */
+    export type NoOfLineBreaks = 1 | 2 | 3 | 4 | 5 | 6;
+}
+
 /**
  * Implementation of @{link GeneratorNode} denoting linebreaks in the desired generated text.
  */
 export class NewLineNode {
-
-    lineDelimiter: string;
-
-    ifNotEmpty = false;
-
-    constructor(lineDelimiter?: string, ifNotEmpty = false) {
-        this.lineDelimiter = lineDelimiter ?? EOL;
-        this.ifNotEmpty = ifNotEmpty;
-    }
+    constructor(
+        public readonly lineDelimiter: string = EOL,
+        public readonly ifNotEmpty: boolean = false,
+        public readonly count: NewLineNode.NoOfLineBreaks = 1
+    ) {}
 }
 
 export const NL = new NewLineNode();
--- a/packages/langium/src/generate/generator-node.ts
+++ b/packages/langium/src/generate/generator-node.ts
@@ -1118,10 +1118,10 @@ export function traceToNodeIf<T extends AstNode>(condition: boolean, source: T |
 export class IndentNode extends CompositeGeneratorNode {
 
     readonly indentation?: string;
-    readonly indentImmediately;
-    readonly indentEmptyLines;
+    readonly indentImmediately: boolean;
+    readonly indentEmptyLines: boolean;
 
-    constructor(indentation?: string | number, indentImmediately = true, indentEmptyLines = false) {
+    constructor(indentation?: string | number, indentImmediately: boolean = true, indentEmptyLines: boolean = false) {
         super();
         if (typeof (indentation) === 'string') {
             this.indentation = indentation;
--- a/packages/langium/src/generate/generator-node.ts
+++ b/packages/langium/src/generate/generator-node.ts
@@ -198,6 +198,25 @@ export class CompositeGeneratorNode {
     }
 
     /**
+     * Prepends `strings` and instances of {@link GeneratorNode} to the content of `this` generator node.
+     *
+     * @param content a var arg mixture of `strings` or {@link GeneratorNode GeneratorNodes}.
+     *
+     * @returns `this` {@link CompositeGeneratorNode} for convenience.
+     *
+     * @example
+     *   generateSomeContent()?.prepend(
+     *      'Some preamble text:', NL
+     *   ).append(
+     *      'Some postamble text:', NL
+     *   );
+     */
+    prepend(...content: Generated[]): this {
+        this.contents.unshift(...content.filter(c => c !== undefined));
+        return this;
+    }
+
+    /**
      * Appends `strings` and instances of {@link GeneratorNode} to `this` generator node, if `condition` is equal to `true`.
      *
      * If `condition` is satisfied this method delegates to {@link append}, otherwise it returns just `this`.
@@ -218,6 +237,30 @@ export class CompositeGeneratorNode {
      */
     appendIf(condition: boolean, ...content: Array<Generated | ((node: CompositeGeneratorNode) => void)>): this {
         return condition ? this.append(...content) : this;
+    }
+
+    /**
+     * Prepends `strings` and instances of {@link GeneratorNode} to the content of `this` generator node, if `condition` is equal to `true`.
+     *
+     * If `condition` is satisfied this method delegates to {@link prepend}, otherwise it returns just `this`.
+     *
+     * @param condition a boolean value indicating whether to prepend the elements of `args` to the content of `this`.
+     *
+     * @param content a var arg mixture of `strings` or {@link GeneratorNode GeneratorNodes}.
+     *
+     * @returns `this` {@link CompositeGeneratorNode} for convenience.
+     *
+     * @example
+     *   generateSomeContent()?.prependIf(
+     *      generatePreamble === true,
+     *      'Some preamble', NL
+     *   ).appendIf(
+     *      generatePostamble === true,
+     *      'Some postamble', NL
+     *   );
+     */
+    prependIf(condition: boolean, ...content: Generated[]): this {
+        return condition ? this.prepend(...content) : this;
     }
 
     /**
--- a/packages/langium/src/generate/node-joiner.ts
+++ b/packages/langium/src/generate/node-joiner.ts
@@ -5,16 +5,15 @@
  ******************************************************************************/
 
 import type { AstNode, Properties } from '../syntax-tree.js';
-import { type Generated, isGeneratorNode } from './generator-node.js';
+import { CompositeGeneratorNode, type Generated, isGeneratorNode, NewLineNode, traceToNode } from './generator-node.js';
 import type { SourceRegion } from './generator-tracing.js';
-import { CompositeGeneratorNode, traceToNode } from './generator-node.js';
 
 export interface JoinOptions<T> {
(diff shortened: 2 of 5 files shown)


packages/langium/src/generate/generator-node.ts
```diff
@@ -1074,9 +1117,9 @@ export function traceToNodeIf<T extends AstNode>(condition: boolean, source: T |
  */
 export class IndentNode extends CompositeGeneratorNode {
 
-    indentation?: string;
-    indentImmediately = true;
-    indentEmptyLines = false;
+    readonly indentation?: string;
+    readonly indentImmediately;
```
JohannesMeierSE: To improve readability, I suggest to explicitly specify the types.

packages/langium/src/generate/generator-node.ts
```diff
@@ -1074,9 +1117,9 @@ export function traceToNodeIf<T extends AstNode>(condition: boolean, source: T |
...
+    readonly indentEmptyLines;
 
     constructor(indentation?: string | number, indentImmediately = true, indentEmptyLines = false) {
```
JohannesMeierSE: To improve readability, I suggest to explicitly specify the types for all parameters.

packages/langium/src/generate/generator-node.ts
```diff
@@ -1090,19 +1133,22 @@ export class IndentNode extends CompositeGeneratorNode {
     }
 }
 
+export namespace NewLineNode {
+    /**
+     * Integer range type allowing to specify 1 to 6 consecutive line breaks, i.e. up to 5 empty lines with a single `NewLineNode`.
+     */
+    export type NoOfLineBreaks = 1 | 2 | 3 | 4 | 5 | 6;
```
JohannesMeierSE, line 1140: Adding more than one line break is a nice improvement, but why 6? Why not 9 or 12? Is it a "random" number or is there an idea behind?

packages/langium/src/generate/node-joiner.ts
```diff
@@ -5,16 +5,15 @@
...
 import type { SourceRegion } from './generator-tracing.js';
-import { CompositeGeneratorNode, traceToNode } from './generator-node.js';
 
 export interface JoinOptions<T> {
     filter?: (element: T, index: number, isLast: boolean) => boolean;
     prefix?: Generated | ((element: T, index: number, isLast: boolean) => Generated | undefined);
     suffix?: Generated | ((element: T, index: number, isLast: boolean) => Generated | undefined);
     separator?: Generated;
-    appendNewLineIfNotEmpty?: true;
+    appendNewLineIfNotEmpty?: true | NewLineNode.NoOfLineBreaks;
```
JohannesMeierSE, line 36: Why not just `appendNewLineIfNotEmpty?: true | number`? Then the number of lines could even be programmatically calculated ...

packages/langium/test/generate/template-node.test.ts
```diff
@@ -1133,6 +1133,41 @@ describe('Joining lists', () => {
...
+              b
+            
+            
+            
+        `);
+    });
+    test('ForEach loop with 6 line breaks if not empty and skip line break after last line', () => {
+        const node = n`
+            Data:
+              ${joinToNode(stream(['a', '\t', 'b']), String, { appendNewLineIfNotEmpty: 6, skipNewLineAfterLastItem: true})}
```
JohannesMeierSE, line 1182: It would be nice to rename `skipNewLineAfterLastItem` to `skipNewLinesAfterLastItem` or `skipAllNewLinesAfterLastItem` to make this behaviour explicit. Maybe a breaking change would be OK for the upcoming v4.0?

packages/langium/test/generate/template-node.test.ts
```diff
@@ -1133,6 +1133,41 @@ describe('Joining lists', () => {
...
+              a
+            
+            
+              b
+            
+            
+            
+        `);
+    });
+    test('ForEach loop with 6 line breaks if not empty and skip line break after last line', () => {
```
JohannesMeierSE: suggestion: `'ForEach loop with 6 line breaks if not empty and skip all line breaks after last line with content'`

JohannesMeierSE (commented):
Thanks @sailingKieler for these nice improvements of the generator API! In particular, `prepend` is very helpful!

I found some details and have two general questions.

packages/langium/src/generate/generator-node.ts
sailingKieler: The idea is to not accept `-7` or `3.14`.
In my experience up to 3 line breaks (2 empty lines) are sufficient, and with a bit more room I ended up with 6 (5 consecutive empty lines).

packages/langium/src/generate/node-joiner.ts
sailingKieler: The idea is to not accept -7 or 3.14.

packages/langium/src/generate/generator-node.ts
JohannesMeierSE: In my experiments, this results in a TSC error:

```typescript
const calculated = 0 + 1; // some calculations based on the current AST
joinToNode(..., { appendNewLineIfNotEmpty: calculated });
```

Not sure, how often this use case appears in practice.

JohannesMeierSE: > The idea is to not accept `-7` or `3.14`.

Yes, that is a nice benefit of the approach!
And I agree, that 6 should be sufficient in practice.

However, I see also some drawbacks with it, in particular, calculating the value seems to be impossible.
I personally prefer the `number` approach, but the decision is up to you.

sailingKieler: Actually, I don't have this use case in mind.
> ```js
> const calculated = 0 + 1; // some calculations based on the current AST
> joinToNode(..., { appendNewLineIfNotEmpty: calculated });
> ```

In case you need to calculate the empty lines you can always fall back to
```js
joinToNode(.... { separator: new CompositeNode(calcNumberOfNLs(...)) } );
```

My intention here is support and simplify the common case, not the generality.

JohannesMeierSE: Could you sketch the fall-back solution in the comments? Then I am fine with your current approach.

packages/langium/test/generate/template-node.test.ts
sailingKieler: > suggestion: `'ForEach loop with 6 line breaks if not empty and skip all line breaks after last line with content'`

That's actually not correct. The `skip...` feature only affects the last item of the iteration.
Thus, the input of `['a', '\t', 'b', '']` would yield 6 trailing line-breaks added after `b`, before adding `''`.

I'll change it to `...  and skip line breaks after last line if not empty`, which is the accurate description.

packages/langium/test/generate/template-node.test.ts
sailingKieler: Actually I would like to keep it in sync with `appendNewLineIfNotEmpty`.
Yes, you could argue that both could be in plural, or that we both have them in singular and plural, but that's too noisy I think. I'll try to address it in the documentation.

JohannesMeierSE: BTW `appendNewLineIfNotEmpty()` could be extended to `appendNewLinesIfNotEmpty(lines: NewLineNode.NoOfLineBreaks = 1)`, then it is in sync again 🙂

packages/langium/test/generate/template-node.test.ts
JohannesMeierSE: or `... and skip all line breaks after the last item of the iteration, if this item is not empty`?

JohannesMeierSE: Some more ideas to rename the options:
- `appendNewLineIfNotEmpty` --> `appendNewLinesToNonEmptyItems`
- `skipNewLineAfterLastItem` --> `skipNewLinesAfterLastItem`

JohannesMeierSE: Question: In case of `{ appendNewLineIfNotEmpty: 2 }`, the input of ['b', ''] would result in 2 line breaks after `b`, but no line breaks after `''`. With other words, for this input, it does not matter, whether `skipNewLineAfterLastItem` is set to `true` or to `false`. Is that correct?

packages/langium/src/generate/generator-node.ts
sailingKieler: done.

packages/langium/src/generate/generator-node.ts
sailingKieler: done.

packages/langium/src/generate/generator-node.ts
sailingKieler: > Could you sketch the fall-back solution in the comments? Then I am fine with your current approach.

done. (in the definition/documentation of `appendNewLineIfNotEmpty`)

packages/langium/test/generate/template-node.test.ts
sailingKieler: > BTW `appendNewLineIfNotEmpty()` could be extended to `appendNewLinesIfNotEmpty(lines: NewLineNode.NoOfLineBreaks = 1)`, then it is in sync again 🙂

Actually, I thought about that, but I like the clarity of the pattern of having each call of `appendNewLine()` representing exactly one line break, and 2 line breaks requiring 2 calls of `.appendNewLine()`, like this
```
expandToNode`
    bla bla bla
`.appendNewLine()
 .appendNewLine()
```
However, that's my personal preference / monk...

packages/langium/test/generate/template-node.test.ts
sailingKieler: Correct.
`appendNewLineIfNotEmpty` activates the appending of `NewlineNodes` with their flags `ifNotEmpty` set to `true`.

packages/langium/src/generate/generator-node.ts
JohannesMeierSE: The comments for all join properties are very nice, thanks!

packages/langium/test/generate/template-node.test.ts
JohannesMeierSE: With this argumentation the solution is OK for me

PR #1966, opened 2025-06-24
Define Grammar AST types explicitly, drop topological sorting of interfaces, refine EBNF-based terminals to avoid synthetic capturing groups

sailingKieler:
This PR introduces `langium-types.langium` defining all the AST types of Langium's grammar language.
As an initial customization it contributes interface `TerminalElement extends AbstractElement`, and `AbstractElement#lookahead` is moved to `TerminalElement`.

The second commit refines the synthesis of RegExps representing the terminals s.t. synthetic pairs of parentheses that are added to the RegExp are marked as non-capturing groups `(?:...)`, while parentheses being present in the terminal definition are transferred as capturing groups `(...)`. To enable that an additional flag `TerminalElement#parenthesized` is introduced for capturing the information wether a sub terminal is enclosed in parenthesizes.
This gives adopters more control over the capturing groups within the RegExp, which is relevant when re-using the generated RegExps e.g. for value conversion, like
```js
const match = LangiumGrammarTerminals.RegexLiteral.exec(input);
const result = match && convert(match[1], match[2], match[3], match[4]);
```

### DIFF
--- a/packages/langium/src/grammar/langium-types.langium
+++ b/packages/langium/src/grammar/langium-types.langium
@@ -1,0 +1,248 @@
+// ******************************************************************************
+// Copyright 2025 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+// *****************************************************************************
+
+type AbstractRule = InfixRule | ParserRule | TerminalRule;
+
+type AbstractType = InferredType | InfixRule | Interface | ParserRule | Type;
+
+type Associativity = "left" | "right";
+
+type Condition = BooleanLiteral | Conjunction | Disjunction | Negation | ParameterReference;
+
+type FeatureName = "assoc" | "current" | "entry" | "extends" | "false" | "fragment" | "grammar" | "hidden" | "import" | "infer" | "infers" | "infix" | "interface" | "left" | "on" | "returns" | "right" | "terminal" | "true" | "type" | "with" | PrimitiveType | string;
+
+type PrimitiveType = "Date" | "bigint" | "boolean" | "number" | "string";
+
+type TypeDefinition = ArrayType | ReferenceType | SimpleType | UnionType;
+
+type ValueLiteral = ArrayLiteral | BooleanLiteral | NumberLiteral | StringLiteral;
+
+interface AbstractElement {
+    cardinality?: "*" | "+" | "?";
+}
+
+interface Action extends AbstractElement {
+    feature?: FeatureName;
+    inferredType?: InferredType;
+    operator?: "+=" | "=";
+    type?: @AbstractType;
+}
+
+interface Alternatives extends AbstractElement {
+    elements: AbstractElement[];
+}
+
+interface ArrayLiteral {
+    elements: ValueLiteral[];
+}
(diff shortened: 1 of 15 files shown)


packages/langium/src/grammar/langium-types.langium
```diff
@@ -0,0 +1,242 @@
+type AbstractRule = InfixRule | ParserRule | TerminalRule;
```
spoenemann, line 7: Please add the copyright header with _Copyright 2025_

packages/langium/src/grammar/langium-types.langium
```diff
@@ -0,0 +1,242 @@
...
+
+type PrimitiveType = "Date" | "bigint" | "boolean" | "number" | "string";
+
+type TypeDefinition = ArrayType | ReferenceType | SimpleType | UnionType;
+
+type ValueLiteral = ArrayLiteral | BooleanLiteral | NumberLiteral | StringLiteral;
+
+interface AbstractElement {
+    cardinality?: "*" | "+" | "?";
+    // parenthesized: boolean;
```
spoenemann: Let's remove it instead of leaving it commented out.

packages/langium/src/languages/generated/ast.ts
```diff
@@ -765,7 +726,63 @@ export function isKeyword(item: unknown): item is Keyword {
...
+    cardinality: 'cardinality',
+    lookahead: 'lookahead',
+    parenthesized: 'parenthesized'
+} as const;
+
+export function isTerminalElement(item: unknown): item is TerminalElement {
+    return reflection.isInstance(item, TerminalElement.$type);
+}
+
+export interface CharacterRange extends TerminalElement {
```
spoenemann: It looks like we're not sorting the types properly prior to generating them – the order shouldn't change after switching to declared types.

packages/langium/src/grammar/langium-types.langium
sailingKieler: done.

packages/langium/src/grammar/langium-types.langium
sailingKieler: done, was a leftover.

packages/langium/src/languages/generated/ast.ts
sailingKieler: That's indeed weird, especially since the interface definitions are in proper order in `langium-types.langium`. 🤔

sailingKieler: This was due to the topological sorting of interfaces in https://github.com/eclipse-langium/langium/pull/1966/files#diff-c032577d15ec02dc6ba352342e13d8c96cc37db00695d1238c7579f7d3b91346 that dated back to Langium v1.

@msujew confirmed that it's not required anymore so I dropped it entirely.

Now the types are alphabetically sorted because of https://github.com/eclipse-langium/langium/blob/1539b422e3f81f83f3ab4d7a0e045805724a4f03/packages/langium/src/grammar/type-system/ast-collector.ts#L61-L64

spoenemann (approved):
👍

Issue #1764, opened 2024-11-29
Option in the Yeoman Generator to generate a "blank" project (without "hello world")


ebousse:
## Problem

The Langium Yeoman generator is great to kickstart a working project around a small "hello world" DSL, and the provided examples of code to be written around the AST (eg. the example validator) are very useful to understand Langium.

However, a typical flow is:

1. generate a langium project with `yo langium`
2. modify the default `.langium` file,
3. run `npm run langium:generate`
4. run `npm run build`

If no other change is done, **step 4 will always fail** because the provided examples of generator, validator and CLI won't compile anymore. This is because these examples are made for the "hello world" default DSL, which is not there anymore because of step 2.

This means that two extra steps are in fact always required:

5. modify generator, validator and CLI  by commenting/changing/removing the parts that cannot compile
6. run `npm run build` once again, and hope it compiles

A problem I encountered when I was starting with Langium, and that I now currently encounter when I try to teach Langium to students, is that the additional **step 5 is quite difficult for people who are just starting with Langium**. Because at this stage one does not know how things works, it is hard to know how to make the changes, and it is easy to make mistakes (eg. commenting too many things).

## Idea

What I would find very useful would be to be offered the choice, in the Yeoman generator, to either have a project with the "hello world" example DSL (and accompanying generator, validator and CLI), or to have a "blank" project without any semantic model nor grammar rules. Such blank project could still provide files for the generator, validator and CLI, but where one would simply find `// TODO`  in each place where a change will be required for the component to work in the future.

This would allow kickstarting a project with steps 1 to 4, avoiding the hazardous steps 5-6 (especially for beginners).

msujew:
I agree, this makes a lot of sense. I've talked with a few people at conferences in the past who had this issue as well.

Mistourr:
Hello, as part of a software engineering course me and my classmates would really be willing to help solve this issue. However we're left wondering what would the Langium team prefers between : 
- Having two different templates ( possibly more maintenance )
- Use of a templating engine ( new dependency )

On top of that we still wonder if the minimal example should include comments for guidelines and if it should compile ( meaning it still needs a minimum grammar and some methods that will need changing )

spoenemann:
My view: 
- The CLI doesn't make sense without a concrete example, so I'd disable it if a blank project is desired
- The extension is not affected (contains only generic code)
- In the language package, the grammar file and validator are affected; for those two, I'd simply create different templates
- The tests don't make sense without an example, so I'd skip those

PR #1967, opened 2025-06-25
Add an option to generate a minimal project with Yeoman

JMazat:
This PR introduces in `generator-langium`, the ability to generate a minimal project instead of the current example "Hello World" project.

The aim of this PR is to **skip the tedious process of removing from the currently generated project, anything that is irrelevant to the target language**. This speeds up the setup process and avoids the need for error-prone refactoring.

This is the intended fix for ticket: #1764.

## What is new

We added questions to be able to choose between the minimal and example projects, **without changing the current default behaviour**:

- **"Generate example project?"**
Default: `Yes`
- **"Your grammar entry rule name:"**
Default: `Model`
This is only shown if the user answered `No` to the above question.
This lets the user choose their own `entry` name and skip the  refactoring that's currently needed.

The generated minimal project is a cleaned-up, de-contextualised version of the example project. It contains *TODOs* throughout the files to guide the user and provide useful resources, and it **should compile** as it is.

We can also find commented (non-compiling) code that can be used to quickly implement features and tests that are often needed.

If the user adds the CLI project, a minimal CLI project will also be generated.
 
## How it's done

We decided to keep it simple and use what had already been done as much as possible.

The templates `language` and `cli` have been renamed to `language-example` and `cli-example`. Separate folders, `language-minimal` and `cli-minimal`, contain the base files needed to generate the minimal projects. As **no files are shared between these projects**, they can be maintained independently and used with the current implementation.

Programmatically we simply change the source folders to be copied based on the user's answers. We have added a new tag `<%= EntryName %>` in `language-minimal` that is replaced by the chosen `entry` name.

### DIFF
--- a/packages/generator-langium/test/yeoman-generator.test.ts
+++ b/packages/generator-langium/test/yeoman-generator.test.ts
@@ -139,6 +139,7 @@ describe('Check yeoman generator works', () => {
             .withAnswers( <Answers>{
                 ...answersForCore,
                 extensionName,
+                includeExampleProject: true,
                 includeCLI: true,
                 includeTest: true
             }).then((result) => {


packages/generator-langium/templates/packages/cli-minimal/src/generator.ts
```diff
@@ -0,0 +1,20 @@
+import type { <%= EntryName %> } from '<%= language-id %>-language';
+import { expandToNode, toString } from 'langium/generate';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { extractDestinationAndName } from './util.js';
+
+export function generateJavaScript(model: <%= EntryName %>, filePath: string, destination: string | undefined): string {
```
spoenemann: In a minimal project, we shouldn't suggest that the target language must be JS. We could rename this to `generateOutput` and change the _destination_ option to refer to a file instead of a directory, so we don't have to guess the file extension.

JMazat:
I need some help please. I'm looking into this pipeline issue, but I don't understand how to resolve it.

#### On my branch merged with an up-to-date `main`:

 When I try to run either `npm run langium:generate` or `npm test`, I get errors relating to `AbstractElement`. As far as I understand, this is not something that the `generator-langium` project should have any influence over. 

#### On my fork's up-to-date `main`:

The same errors occur.

I also tried bumping the version and updating dependencies first, but the behaviour was the same.

What I am doing wrong here?  Thanks.

msujew:
@JMazat It seems like you merged `main` into your branch, which rewrote the history. This in turn makes GitHub think that you now want to merge all those changes into `main` again, leading to a bunch of merge conflicts (and likely the CI failure as well). I would recommend you to simply [rebase your change](https://git-scm.com/docs/git-rebase) on the latest main and force push your branch instead.

examples/arithmetics/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "langium-arithmetics-dsl",
   "displayName": "Arithmetics DSL",
-  "version": "3.5.0",
+  "version": "3.6.0",
```
msujew: Please don't update the package versions - we will do this ourselves when we perform the release.

msujew (commented):
Aside from https://github.com/eclipse-langium/langium/pull/1967#discussion_r2179291949 and my comment (see below), I think this looks quite good. We can always iterate on that after merge.

packages/generator-langium/src/index.ts
```diff
@@ -149,7 +167,17 @@ export class LangiumGenerator extends Generator {
...
+            } as PromptQuestion<Answers>,
+            {
+                type: 'input',
+                name: 'entryName',
+                prefix: description(
+                    'The name of the entry rule in your grammar file.'
+                ),
+                message: 'Your grammar entry rule name:',
+                default: 'Model',
+                when: (answers) => !answers.includeExampleProject
```
spoenemann: I think this makes sense with the Hello World example, too. Could you include this variable in the old template?

packages/generator-langium/src/index.ts
```diff
@@ -133,6 +141,16 @@ export class LangiumGenerator extends Generator {
                 message: 'Include VSCode extension?',
                 default: true
             } as PromptQuestion<Answers>,
+            {
+                type: 'confirm',
+                name: 'includeExampleProject',
+                prefix: description(
+                    'You can generate an example project to play around with Langium (with a "Hello world" grammar, generator and validator).',
```
spoenemann, line 157: The description of `extensionName` further above is outdated:

> This tool generates a VS Code extension with a "Hello World" language to get started quickly.

Could you change it to this?

> This tool generates one or more npm packages to create your own language based on Langium.

spoenemann (commented):
Great! Just two details below.

msujew (commented):
Please note that the tests are currently failing.

spoenemann (approved):
🚀

Issue #1974, opened 2025-07-01
Yarn workspace based projects can be broken by relative import in langium-cli.


RGHenderson:
Langium version: 3.5.0
Package name: langium-cli

## Steps To Reproduce

1. Have some yarn workspace based project, where one workspace requires langium only, and the other requires both langium and langium-cli
2. Run the langium generator, and note that the new EBNF generator fails with the unresolved relative import to the non-linked langium package (in some parent directory)
3. Downgrade to 3.4.0 and see that generatiion works as expected

## The current behavior

Langium generation fails

## The expected behavior

Langium generates successfully

RGHenderson:
@msujew thanks for releasing the fix. Does langium itself also need a 3.5.1? I see the "isInfixRule" from langiums generated ast is used in the cli, but not available.

msujew:
Ah, damn. I forgot about this. Let me quickly publish a new 3.5.2.

msujew:
It's published and is correctly based off the 3.5.0 version of Langium.

PR #1975, opened 2025-07-01
Remove relative import from BNF generator

RGHenderson:
Removes the relative import from the new BNF generator, which causes issues in yarn multi workspace projects. 

Closes: #1974

### DIFF
--- a/packages/langium-cli/src/generator/bnf-generator.ts
+++ b/packages/langium-cli/src/generator/bnf-generator.ts
@@ -4,10 +4,9 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { CstUtils, type Grammar } from 'langium';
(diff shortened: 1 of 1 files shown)


msujew (commented):
Hey @RGHenderson, thank you for the contribution!

You also need to remove the other `../../../langium/lib/languages/generated/ast.js` import for this to work (the current change doesn't do anything, since the `import type` statement is erased during compilation to JS.

Additionally, please sign the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/eca/), otherwise we cannot merge the changes.

RGHenderson:
Hi @msujew - totally right - homed in on exactly the wrong import. Should be fixed now.

msujew:
It still doesn't compile, there are some missing references to `ast`;

PR #1976, opened 2025-07-03
feat: pass ref info and context to createReferenceCompletionItem

cdietrich:


packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -476,9 +476,11 @@ export class DefaultCompletionProvider implements CompletionProvider {
      * To change the `documentation`, override the `DocumentationProvider` service instead.
      *
      * @param nodeDescription The description of a reference candidate
+     * @param _refInfo Information about the reference for which the candidate is proposed
+     * @param _context The completion context
      * @returns A partial completion item
      */
-    protected createReferenceCompletionItem(nodeDescription: AstNodeDescription): CompletionValueItem {
+    protected createReferenceCompletionItem(nodeDescription: AstNodeDescription, _refInfo: ReferenceInfo, _context: CompletionContext): CompletionValueItem {
```
cdietrich, line 483: alternatively i could make these two optional

cdietrich: alternatively switch the order to
context, refInfo, nodeDescription

cdietrich:
@msujew wdyt

msujew (approved):
Makes sense, thank you 👍

PR #1977, opened 2025-07-03
Added DocumentBuilder method to reset document state

### LINKED ISSUE #1913: Allow re-validation while specifying validation options.
Context: https://github.com/eclipse-langium/langium/discussions/1911

Make it possible to trigger revalidation while specifying validation options (e.g., `stopAfterLexingErrors: true`).

This is currently not possible (without overriding `DefaultDocumentBuilder`'s `build`) because the only way to [force re-running](https://github.com/eclipse-langium/langium/blob/main/packages/langium/src/workspace/document-builder.ts#L163) is to have `validation: true`, but `true` means that validation options cannot be specified.

Per @msujew's recommendation, creating this as a feature request.

> I believe the validation options in the document builder and the ones used in the document validator have drifted apart a bit. It might make sense to consolidate a bit for our v4.0. Do you mind creating a feature request for that?

spoenemann:
Closes #1913.

Now it's easier to reset the state of a document so you can repeat the phases you want (e.g. validation).


packages/langium/src/workspace/index-manager.ts
```diff
@@ -151,6 +161,17 @@ export class DefaultIndexManager implements IndexManager {
...
+    removeContent(uri: URI): void {
+        const uriString = uri.toString();
+        this.symbolIndex.delete(uriString);
+        this.symbolByTypeIndex.clear(uriString);
+    }
+
+    removeReferences(uri: URI): void {
+        const uriString = uri.toString();
+        this.referenceIndex.delete(uriString);
+    }
```
msujew: Suggestion: How about making `remove` simply call both `removeContent` and `removeReferences`?

msujew (approved):

packages/langium/src/workspace/index-manager.ts
spoenemann: 👍 I thought that would stringify the URI twice, but after a deeper look I saw that URI caches its string rep, and we call this only when deleting documents.

PR #1979, opened 2025-07-04
Improve generated `ast.ts` regarding multiple files and languages

JohannesMeierSE:
**Summary**

This PR reworks the content of the generated `ast.ts` in case of multiple `*.langium` files OR multiple languages inside the project (as defined in the `langium-config.json`).

**Motivation**

In a recent multi-language project, we wanted to write a formatter for a single language (not for all languages). In order to statically ensure by the TypeScript compiler, that we wrote definitions for all AST types in the language, we used the generated `XXXAstType = { Type1: Type1, Type2: Type2, ... }`. But this forced us to define formatting rules for all types of the project including types from other languages. Therefore this PR contributes (among others) a XXXAstType for each language.

**Contributions in more detail**

The `ast.ts` consists of:
1. terminals
2. keywords
3. type definitions (`export interface YYY extends langium.AstNode { ... }`)
4. `XXXAstType` (as described in the motivation)
5. reflection

While the whole content is merged for all files and languages, this PR splits the generated code for terminals (1), keywords (2) and XXXAstType (4), while type definitions (3) and reflection (5) remain unchanged.
In particular for the type definitions (3) it is important to have only one version of them:
- Otherwise imports in TS files during coding become non-unique
- Type definitions might be "incomplete", since it is possible to infer the same type (identified by its name) with different properties (which are merged together into the final type definition) by different parser rules, which might be part of different Langium files!

One instance of terminals, keywords and XXXAstType is created not only for each language, but also for each `*.langium` file in order to enable more fine-grained implementations.
Additionally, one instance of terminals, keywords and XXXAstType is created for the whole project.

**Discussion of (non) breaking changes**

- The `ast.ts` for a single-file project remains unchanged (try `npm run langium:generate` on the whole Langium repo on your own)!
- The terminals, keywords and XXXAstType created for the whole project have the same names as before this PR, therefore they can be used as before.
- Since the initial `grammar MyName` statement in Langium files is optional, it cannot be used as identifier for the instances for each file. Instead the file name is used with the following limitations (which are validated and reported by the CLI) and therefore can be seen as breaking changes:
    - Only the file name, but not the folder name is considered. Therefore `folder1/my.langium` and `folder2/my.langium` together are not supported.
    - The file name is converted to camel case, e.g. `my.langium` and `mY.langium` together are not supported.

**Additional contributions**

- Added test cases for the CLI generator for multi-file and multi-language projects (so far, only single-file projects are tested)
- Fixed bug in Langium in general: Even terminals of grammars with entry rules which are not used by these grammars are now part of the `ast.ts`, since they might be used by other grammars, since it is possible to import grammars with entry rules (considers also terminal fragments now). This can be seen as breaking changes.

### DIFF
--- a/examples/requirements/src/language-server/generated/ast.ts
+++ b/examples/requirements/src/language-server/generated/ast.ts
@@ -6,28 +6,85 @@
 /* eslint-disable */
 import * as langium from 'langium';
 
-export const RequirementsAndTestsTerminals = {
+export const RequirementsAndTestsFileCommonTerminals = {
     WS: /\s+/,
     ID: /[_a-zA-Z][\w_]*/,
+    INT: /[0-9]+/,
     STRING: /"(\\.|[^"\\])*"|'(\\.|[^'\\])*'/,
     ML_COMMENT: /\/\*[\s\S]*?\*\//,
     SL_COMMENT: /\/\/[^\n\r]*/,
 };
 
-export type RequirementsAndTestsTerminalNames = keyof typeof RequirementsAndTestsTerminals;
+export type RequirementsAndTestsFileCommonTerminalNames = keyof typeof RequirementsAndTestsFileCommonTerminals;
 
-export type RequirementsAndTestsKeywordNames =
+export type RequirementsAndTestsFileCommonKeywordNames =
+    | ":"
+    | "contact";
+
+export type RequirementsAndTestsFileCommonTokenNames = RequirementsAndTestsFileCommonTerminalNames | RequirementsAndTestsFileCommonKeywordNames;
+
+export const RequirementsAndTestsFileRequirementsTerminals = {
+};
+
+export type RequirementsAndTestsFileRequirementsTerminalNames = keyof typeof RequirementsAndTestsFileRequirementsTerminals;
+
+export type RequirementsAndTestsFileRequirementsKeywordNames =
     | ","
     | ":"
-    | "="
     | "applicable"
-    | "contact"
     | "environment"
     | "for"
-    | "req"
+    | "req";
+
+export type RequirementsAndTestsFileRequirementsTokenNames = RequirementsAndTestsFileRequirementsTerminalNames | RequirementsAndTestsFileRequirementsKeywordNames;
+
+export const RequirementsAndTestsFileTestsTerminals = {
+};
+
+export type RequirementsAndTestsFileTestsTerminalNames = keyof typeof RequirementsAndTestsFileTestsTerminals;
+
+export type RequirementsAndTestsFileTestsKeywordNames =
+    | ","
+    | "="
+    | "applicable"
+    | "for"
     | "testFile"
     | "tests"
     | "tst";
+
+export type RequirementsAndTestsFileTestsTokenNames = RequirementsAndTestsFileTestsTerminalNames | RequirementsAndTestsFileTestsKeywordNames;
+
+export const RequirementsAndTestsLanguageRequirementsTerminals = {
+    ...RequirementsAndTestsFileCommonTerminals,
+    ...RequirementsAndTestsFileRequirementsTerminals,
+};
+
+export type RequirementsAndTestsLanguageRequirementsTerminalNames = keyof typeof RequirementsAndTestsLanguageRequirementsTerminals;
+
+export type RequirementsAndTestsLanguageRequirementsKeywordNames = RequirementsAndTestsFileCommonKeywordNames | RequirementsAndTestsFileRequirementsKeywordNames;
+
+export type RequirementsAndTestsLanguageRequirementsTokenNames = RequirementsAndTestsLanguageRequirementsTerminalNames | RequirementsAndTestsLanguageRequirementsKeywordNames;
+
+export const RequirementsAndTestsLanguageTestsTerminals = {
+    ...RequirementsAndTestsFileCommonTerminals,
+    ...RequirementsAndTestsFileRequirementsTerminals,
+    ...RequirementsAndTestsFileTestsTerminals,
+};
+
+export type RequirementsAndTestsLanguageTestsTerminalNames = keyof typeof RequirementsAndTestsLanguageTestsTerminals;
+
+export type RequirementsAndTestsLanguageTestsKeywordNames = RequirementsAndTestsFileCommonKeywordNames | RequirementsAndTestsFileRequirementsKeywordNames | RequirementsAndTestsFileTestsKeywordNames;
+
+export type RequirementsAndTestsLanguageTestsTokenNames = RequirementsAndTestsLanguageTestsTerminalNames | RequirementsAndTestsLanguageTestsKeywordNames;
+
+export const RequirementsAndTestsTerminals = {
+    ...RequirementsAndTestsLanguageRequirementsTerminals,
+    ...RequirementsAndTestsLanguageTestsTerminals,
+};
+
+export type RequirementsAndTestsTerminalNames = keyof typeof RequirementsAndTestsTerminals;
+
+export type RequirementsAndTestsKeywordNames = RequirementsAndTestsLanguageRequirementsKeywordNames | RequirementsAndTestsLanguageTestsKeywordNames;
 
 export type RequirementsAndTestsTokenNames = RequirementsAndTestsTerminalNames | RequirementsAndTestsKeywordNames;
 
@@ -137,7 +194,18 @@ export function isTestModel(item: unknown): item is TestModel {
     return reflection.isInstance(item, TestModel.$type);
 }
 
-export type RequirementsAndTestsAstType = {
+export type RequirementsAndTestsFileCommonAstType = {
+    Contact: Contact
+}
+
+export type RequirementsAndTestsFileRequirementsAstType = {
+    Contact: Contact
+    Environment: Environment
+    Requirement: Requirement
+    RequirementModel: RequirementModel
+}
+
+export type RequirementsAndTestsFileTestsAstType = {
     Contact: Contact
     Environment: Environment
     Requirement: Requirement
@@ -145,6 +213,12 @@ export type RequirementsAndTestsAstType = {
     Test: Test
     TestModel: TestModel
 }
+
+export type RequirementsAndTestsLanguageRequirementsAstType = RequirementsAndTestsFileCommonAstType & RequirementsAndTestsFileRequirementsAstType
+
+export type RequirementsAndTestsLanguageTestsAstType = RequirementsAndTestsFileCommonAstType & RequirementsAndTestsFileRequirementsAstType & RequirementsAndTestsFileTestsAstType
+
+export type RequirementsAndTestsAstType = RequirementsAndTestsLanguageRequirementsAstType & RequirementsAndTestsLanguageTestsAstType
 
 export class RequirementsAndTestsAstReflection extends langium.AbstractAstReflection {
     override readonly types = {
--- a/examples/requirements/syntaxes/tests.tmLanguage.json
+++ b/examples/requirements/syntaxes/tests.tmLanguage.json
@@ -10,7 +10,7 @@
     },
     {
       "name": "keyword.control.tests-lang",
-      "match": "\\b(applicable|contact|for|testFile|tests|tst)\\b"
+      "match": "\\b(applicable|contact|environment|for|req|testFile|tests|tst)\\b"
     },
     {
       "name": "string.quoted.double.tests-lang",
--- a/packages/langium-cli/src/generate.ts
+++ b/packages/langium-cli/src/generate.ts
@@ -4,31 +4,29 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { AstNode, Grammar, LangiumDocument, Mutable } from 'langium';
-import type { LangiumConfig, LangiumLanguageConfig } from './package-types.js';
-import { URI } from 'langium';
-import { loadConfig } from './package.js';
-import { AstUtils, GrammarAST } from 'langium';
+import chalk from 'chalk';
+import fs from 'fs-extra';
+import { validate } from 'jsonschema';
+import { MultiMap, type AstNode, type Grammar, type LangiumDocument, type Mutable } from 'langium';
+import { AstUtils, GrammarAST, URI } from 'langium';
+import { createGrammarDiagramHtml, createGrammarDiagramSvg } from 'langium-railroad';
 import { createLangiumGrammarServices, resolveImport, resolveImportUri, resolveTransitiveImports } from 'langium/grammar';
 import { NodeFileSystem } from 'langium/node';
-import { generateAst } from './generator/ast-generator.js';
-import { serializeGrammar } from './generator/grammar-serializer.js';
-import { generateModule } from './generator/module-generator.js';
+import * as path from 'path';
+import { generateAstMultiFileProject, generateAstMultiLanguageProject, generateAstSingleFileProject } from './generator/ast-generator.js';
 import { generateBnf } from './generator/bnf-generator.js';
-import { generateTextMate } from './generator/highlighting/textmate-generator.js';
+import { serializeGrammar } from './generator/grammar-serializer.js';
 import { generateMonarch } from './generator/highlighting/monarch-generator.js';
 import { generatePrismHighlighting } from './generator/highlighting/prism-generator.js';
-import { getTime, log } from './generator/langium-util.js';
+import { generateTextMate } from './generator/highlighting/textmate-generator.js';
+import { getAstIdentifierForGrammarFile, getTime, log } from './generator/langium-util.js';
+import { generateModule } from './generator/module-generator.js';
 import { elapsedTime, getUserChoice, schema } from './generator/node-util.js';
-import { RelativePath } from './package-types.js';
-import { getFilePath } from './package.js';
-import { validateParser } from './parser-validation.js';
 import { generateTypesFile } from './generator/types-generator.js';
-import { createGrammarDiagramHtml, createGrammarDiagramSvg } from 'langium-railroad';
-import { validate } from 'jsonschema';
-import chalk from 'chalk';
-import * as path from 'path';
-import fs from 'fs-extra';
+import type { LangiumConfig, LangiumLanguageConfig } from './package-types.js';
+import { RelativePath } from './package-types.js';
+import { getFilePath, loadConfig } from './package.js';
+import { validateParser } from './parser-validation.js';
 
 export async function generate(options: GenerateOptions): Promise<boolean> {
     const config = await loadConfig(options);
@@ -260,7 +258,7 @@ export async function runGenerator(config: LangiumConfig, options: GenerateOptio
         grammarServices.validation.LangiumGrammarValidator.options = config.validation;
     }
 
-    const all = await buildAll(config);
+    const all = await buildAll(config); // all parsed *.langium documents with their "doc.uri.fsPath" as key
     const buildResult: (success: boolean) => GeneratorResult = (success: boolean) => ({
         success,
         files: Array.from(all.keys())
@@ -290,7 +288,8 @@ export async function runGenerator(config: LangiumConfig, options: GenerateOptio
         return buildResult(false);
     }
 
-    const grammars: Grammar[] = [];
+    // identify all relevant grammars
+    const topLevelGrammars: Grammar[] = [];
     const configMap: Map<Grammar, LangiumLanguageConfig> = new Map();
     const relPath = config[RelativePath];
     for (const languageConfig of config.languages) {
@@ -302,28 +301,20 @@ export async function runGenerator(config: LangiumConfig, options: GenerateOptio
                 log('error', options, chalk.red(`${absGrammarPath}: The entry grammar must start with the 'grammar' keyword.`));
                 return buildResult(false);
             }
-            grammars.push(grammar);
+            topLevelGrammars.push(grammar);
             configMap.set(grammar, languageConfig);
         }
     }
-
-    const grammarElements = mapGrammarElements(grammars);
-
-    const embeddedGrammars: Grammar[] = [];
-    for (const grammar of grammars) {
-        const embeddedGrammar = embedReferencedGrammar(grammar, grammarElements);
-        embeddedGrammars.push(embeddedGrammar);
-        configMap.set(embeddedGrammar, configMap.get(grammar)!);
-    }
-    // We need to rescope the grammars again
-    // They need to pick up on the embedded references
-    await relinkGrammars(embeddedGrammars);
-
-    for (const grammar of embeddedGrammars) {
-        // Create and validate the in-memory parser
-        const parserAnalysis = await validateParser(grammar, config, configMap, grammarServices);
-        if (parserAnalysis instanceof Error) {
-            log('error', options, chalk.red(parserAnalysis.toString()));
+    const importedGrammars = topLevelGrammars.flatMap(g => resolveTransitiveImports(sharedServices.workspace.LangiumDocuments, g));
+    const allGrammars = [ ...topLevelGrammars, ...importedGrammars ]
+        .filter((grammar, index, array) => array.indexOf(grammar) === index) // keep only the 1st occurance of each grammar => filter out duplicates in in-place way
+        .sort((g1, g2) => getAstIdentifierForGrammarFile(g1).localeCompare(getAstIdentifierForGrammarFile(g2))); // sort regarding their file names
+    // check that the identifiers of the grammars are unique
+    const mapCheckUnique: MultiMap<string, Grammar> = new MultiMap();
+    allGrammars.forEach(grammar => mapCheckUnique.add(getAstIdentifierForGrammarFile(grammar), grammar));
+    for (const [identifier, grammars] of mapCheckUnique.entriesGroupedByKey()) {
+        if (grammars.length >= 2) {
+            log('error', options, chalk.red(`The grammars ${grammars.map(g => AstUtils.getDocument(g).uri.toString()).join(', ')} result in the same identifier '${identifier}': Rename the file name(s) to make the grammar identifiers unique.`));
             return buildResult(false);
         }
     }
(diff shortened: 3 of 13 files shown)


sailingKieler:
First of all thank you very much for the detailed PR description, that's been a lot of effort! 🙏

examples/requirements/src/language-server/generated/ast.ts
```diff
@@ -6,29 +6,86 @@
...
+export type RequirementsAndTestsFileTestsTokenNames = RequirementsAndTestsFileTestsTerminalNames | RequirementsAndTestsFileTestsKeywordNames;
+
+export const RequirementsAndTestsLanguageRequirementsTerminals = {
+    ...RequirementsAndTestsFileCommonTerminals,
+    ...RequirementsAndTestsFileRequirementsTerminals,
+};
+
+export type RequirementsAndTestsLanguageRequirementsTerminalNames = keyof typeof RequirementsAndTestsLanguageRequirementsTerminals;
+
+export type RequirementsAndTestsLanguageRequirementsKeywordNames = RequirementsAndTestsFileCommonKeywordNames | RequirementsAndTestsFileRequirementsKeywordNames;
```
sailingKieler: Has it been like that previously in terms of keywords being contributed by non-called parser rules?

I assume that `File`-based types include all keywords that occur in the entire file, but I do not agree that the union of those types is equal to the union of keywords that are actually part of a language.

Instead, I would assume that non reachable keywords are not included here. What was the established approach in that?

examples/requirements/syntaxes/tests.tmLanguage.json
```diff
@@ -10,7 +10,7 @@
     },
     {
       "name": "keyword.control.tests-lang",
-      "match": "\\b(applicable|contact|for|testFile|tests|tst)\\b"
+      "match": "\\b(applicable|contact|environment|for|req|testFile|tests|tst)\\b"
```
sailingKieler: I think this change reflects my consideration from above.
Here it won't not hurt, I guess.
In general, however, keywords will be mixed with IDs in the highlighting and other features, if keywords are 'inherited' from grammar files, although they are not reachable via the entry rule.

packages/langium-cli/src/generate.ts
```diff
@@ -339,15 +330,57 @@ export async function runGenerator(config: LangiumConfig, options: GenerateOptio
         return buildResult(false);
     }
 
-    const genAst = generateAst(grammarServices, embeddedGrammars, config);
-    await writeWithFail(path.resolve(updateLangiumInternalAstPath(output, config), 'ast.ts'), genAst, options);
+    // ast.ts
+    const isSingleLanguage = config.languages.length === 1;
+    const isSingleFile = allGrammars.length === 1 && isSingleLanguage /* handles a special case: multiple languages use the same *.langium file */;
```
sailingKieler: I don't get the sense of the comment 🤔

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -4,47 +4,164 @@
...
 
     return expandToNode`
-        export type ${config.projectName}AstType = {
+        export type ${name}AstType = {
             ${joinToNode(typeNames, name => name + ': ' + name, { appendNewLineIfNotEmpty: true })}
         }
+    `.appendNewLine().appendNewLine();
+}
+
+function generateAstTypeComposed(identifiers: string[], name: string): CompositeGeneratorNode {
```
sailingKieler: something minor: return type `Generated` should be sufficient here if I don't overlook something

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -4,47 +4,164 @@
...
+        export type ${name}AstType = ${joinToNode(
+            identifiers.sort(),
+            identifier => `${identifier}AstType`,
+            { separator: ' & ' }
+        )}
+    `.appendNewLine().appendNewLine();
+}
 
-        export class ${config.projectName}AstReflection extends langium.AbstractAstReflection {
+function generateAstReflection(name: string, astTypes: AstTypes): CompositeGeneratorNode {
```
sailingKieler: something minor: return type `Generated` should be sufficient here if I don't overlook something

packages/langium-cli/test/generator/ast-generator.test.ts
```diff
@@ -695,11 +702,808 @@ async function testGenerated(grammar: string, expected: string, start: string, e
...
+    const typesFileContent = generateAstSingleFileProject(services.grammar, [result.value], config);
     let startIndex = typesFileContent.indexOf(start);
     for (let i = 0; i < startCount; i++) {
         startIndex = typesFileContent.indexOf(start, startIndex + start.length);
     }
     const relevantPart = typesFileContent.substring(startIndex, typesFileContent.indexOf(end)).trim();
     expect(relevantPart).toEqual(expectedPart);
 }
+
+describe('Ast generator (with multiple *.langium files)', () => {
```
sailingKieler, line 709: Considering that these additional new tests double the length of the file, I would prefer to create the test projects on disc rather in memory based on inlined code. That would also simplify the update of the expected parts.

Just reviewed the tests in VSCode and it is okay there, here on Github with side by side view it's basically unreadable.

packages/langium-cli/src/generate.ts
sailingKieler: Did you think about excluding grammar files that define interfaces & types only, like `langium-types.langium`, at least while considering keywords, token types and such parts?

I advocate that as otherwise many language projects are considered multi file projects, although they only contain a single file defining parser & terminal rules, like the Langium grammar 'project'.

JohannesMeierSE: If there are two languages with the same entry grammar, then it is a multi-language project with a single `*.langium` file.

packages/langium-cli/src/generator/ast-generator.ts
JohannesMeierSE: changed

packages/langium-cli/src/generator/ast-generator.ts
JohannesMeierSE: changed

packages/langium-cli/test/generator/ast-generator.test.ts
JohannesMeierSE: > That would also simplify the update of the expected parts.

I don't agree, since you need to change only one place with the current approach.

What do you think about splitting the `ast-generator.test.ts` file into two files, maybe `ast-generator-single.test.ts` and `ast-generator-multi.test.ts`?

packages/langium-cli/src/generate.ts
JohannesMeierSE: We need to ensure, that the complexity of the `ast.ts`-generator does not explode. Therefore I would not embed/inline grammar files containing only type definitions in general.

But we could interpret projects with one grammar file containing rules and additional grammar files containing _only_ type definitions as "single-file projects" as well. That would include the "Langium grammar project" itself.

sailingKieler: Yes, but the property is `true` for the `(1, 1)` case, so the context does not fit for me...

packages/langium-cli/test/generator/ast-generator.test.ts
sailingKieler: It is still okay for me to keep it, but it's quite on the edge.

The _files on disc_-approach has the advantage that you can run the `langium-cli` to update the files if necessary (probably with some kind of setup), which is much easier for others that didn't write the tests initially. In general, there's no obligation to do it manually, in contrast to the _in code_-approach.

JohannesMeierSE:
Thanks @sailingKieler for your review and discussing some critical points!

You are right regarding *reachability*, I focused on *completeness* in this PR.

> I assume that File-based types include all keywords that occur in the entire file

yes

> I do not agree that the union of those types is equal to the union of keywords that are actually part of a language.

You can argue in that way, yes.

This is a fundamental decision we need to decide.

I just checked, that the previous AST generator ignores unused rules (and therefore unused keywords), but it does consider unused types!

I see the following approaches (the project part would remain unchanged in all approaches):
1. Keep the current implementation of this PR, i.e. generate the *complete* set of elements, not the set of *reachable* elements.
    - Open question: Do we want the complete or the reachable elements?
2. Skip the file-based parts completely, i.e. generate code only for languages and the whole project.
    - Pro: reachable elements are easy
    - Contra: If the project uses `common.langium` files which are included by multiple languages, you don't have the elements of `common.langium` in order to define stuff only once and reuse it for multiple languages.
3. Keep the file-based parts as they are and generate the complete set of reachable elements for each language (without referring to the files).
    - Contra: unnecessary code clones
4. Generate the complete set of reachable elements for each language (without referring to the files), generate the complete set of elements for each file, and calculate the reachable elements for each file afterwards.
    - Pro: We have reachable elements for both files and languages
    - Contra: I am not sure, whether this approach is possible for terminals. The complexity grows.

I sketched the last approach for the keywords:

```typescript
// the three *.langium files with all contained keywords
export type KeywordsF1All = // f1.langium
    | 'F1used'
    | 'F1unused';
export type KeywordsF2All = // f2.langium
    | 'F2used'
    | 'F2unused';
export type KeywordsFCommonAll = // fcommon.langium
    | 'FCommonBoth'
    | 'FCommon1'
    | 'FCommon2'
    | 'FCommonNone';

// two languages with their reachable keywords
export type KeywordsL1Reachable = // language 1 uses f1.langium and fcommon.langium
    | 'F1used'
    | 'FCommonBoth'
    | 'FCommon1';
export type KeywordsL2Reachable = // language 2 uses f2.langium and fcommon.langium
    | 'F2used'
    | 'FCommonBoth'
    | 'FCommon2';

// the reachable keywords of the three *.langium files
export type KeywordsF1Reachable = KeywordsF1All & KeywordsL1Reachable;
export type KeywordsF2Reachable = KeywordsF2All & KeywordsL2Reachable;
export type KeywordsFCommonReachable = KeywordsFCommonAll & (KeywordsL1Reachable | KeywordsL2Reachable);
```

packages/langium-cli/src/generate.ts
JohannesMeierSE: We have two languages in the config => `isSingleLanguage` is `false` (line 334)

Both language configs point to the same `*.langium` file => `allGrammars.length === 1` is `true` => without the additional check `&& isSingleLanguage`, `isSingleFile` is `true` as well (line 335)

Since we check the single-file case first (line 337), this (strange) multi-language project is treated as a single-file project, i.e. the language-specific parts in the `ast.ts` are skipped, but it should be treated as a multi-language project. Therefore we need `&& isSingleLanguage`.

packages/langium-cli/test/generator/ast-generator.test.ts
JohannesMeierSE: I got your point. Do we have such "on disc"-test cases somewhere else in the Langium project?

Otherwise the complexity of the set-up grows unnecessarily. Since we usually don't change the AST generator, it is not worth the effort.

JohannesMeierSE: BTW: Is there an easy way to solve the indentation problem, i.e. to get rid of the strange `expandToStringIndented` solution?

spoenemann:
I agree that the file-based generation of additional types and constants is very noisy. I worry that it will cause confusion – the generated names are long and you have to look very closely to choose the correct one.

Idea: WDYT of wrapping all definitions generated for a particular file in a _namespace_? That might make the generated code more understandable.

Additionally, we should start adding comments to the generated definitions to make them easier to grasp.

JohannesMeierSE:
Decisions in the Langium dev meeting 2025-07-09 (@sailingKieler @spoenemann @msujew @JohannesMeierSE)

- We generate elements for each language, but not for each file in order to reduce complexity in the AST generator => the `ast.ts` for *all* single-language projects remain unchanged
- We generate only the used/reachable keywords, but all types.
- We try to structure the `ast.ts` for multi-language projects with namespaces and comments for a better readability

@JohannesMeierSE will update this PR accordingly

cdietrich:
looking forward what this does to our 12 language multi language project.

JohannesMeierSE:
@sailingKieler @spoenemann @msujew This PR is ready for another review

JohannesMeierSE:
There are some follow-up ideas for improvements, WDYT?

- Terminals which are called only by terminals are not listed in the terminals section: Why? I didn't got the reason for that ...
- For the type definitions, first all unions and second all interfaces are generated: We could think about mixing union and interface types and sort them by name.
- The terminals are not sorted by name so far.

cdietrich (changes requested):
can you please have a look the problem i reported

packages/langium-cli/src/generate.ts
```diff
@@ -159,20 +156,21 @@ function mapGrammarElements(grammars: Grammar[], visited: Set<string> = new Set(
                 grammar,
                 (grammar.rules as GrammarElement[])
                     .concat(grammar.types)
-                    .concat(grammar.interfaces)
+                    .concat(grammar.interfaces) // a new array is created here!
```
sailingKieler: What about this instead of the preceding 3 lines:
```suggestion
                [ ...grammar.rules, ...grammar.types, ...grammar.interfaces ]
```

packages/langium-cli/src/generate.ts
```diff
@@ -290,41 +288,31 @@ export async function runGenerator(config: LangiumConfig, options: GenerateOptio
...
-    await relinkGrammars(embeddedGrammars);
-
-    for (const grammar of embeddedGrammars) {
-        // Create and validate the in-memory parser
-        const parserAnalysis = await validateParser(grammar, config, configMap, grammarServices);
-        if (parserAnalysis instanceof Error) {
-            log('error', options, chalk.red(parserAnalysis.toString()));
-            return buildResult(false);
+            configMap.set(entryGrammar, languageConfig);
+            const embeddedGrammar = await embedGrammars([entryGrammar], configMap, config, options, grammarServices); // this changes the 'entryGrammar' and the 'configMap' in-place!
```
sailingKieler: In the former version all configs were visited once and `configMap.set(entryGrammar, languageConfig)` was performed before the first `configMap.get(...)` has been executed. That's not the the case in your version, see impl of `embedGrammars`. Does that yield the issue @cdietrich experienced?

BTW: In your test you do it 'the old way', i.e. populating `configMap` completely before calling `embedGrammars` for the first time. Maybe this entire function (`runGenerator(...)`) could be tested in one of your tests (or an additional one)?

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -4,47 +4,122 @@
...
+}
 
-        export class ${config.projectName}AstReflection extends langium.AbstractAstReflection {
+function generateAstTypeComposed(identifiers: string[], name: string): Generated {
+    identifiers.sort(); // in-place, for a stable order
+    return expandToNode`
+        export type ${name}AstType = ${joinToNode(
+            identifiers,
+            identifier => `${identifier}.AstType`,
+            { separator: ' & ' }
```
sailingKieler: I would prefer a one-liner version here, too, like in l.99, the the result will also be a one-liner.

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -4,47 +4,122 @@
...
         .concat(astTypes.unions.map(t => t.name))
         .sort();
-    const typeHierarchy = collectTypeHierarchy(mergeTypesAndInterfaces(astTypes));
 
     return expandToNode`
-        export type ${config.projectName}AstType = {
+        export type ${name}AstType = {
             ${joinToNode(typeNames, name => name + ': ' + name, { appendNewLineIfNotEmpty: true })}
         }
+    `.appendNewLine().appendNewLine();
```
sailingKieler: I suggest to drop the `appendNewLine()...` here and add empty lines in the calling template.
This will eliminate superfluous trailing newlines in `ast.ts` of the `requirements` project.

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -4,47 +4,122 @@
...
-        export class ${config.projectName}AstReflection extends langium.AbstractAstReflection {
+function generateAstTypeComposed(identifiers: string[], name: string): Generated {
+    identifiers.sort(); // in-place, for a stable order
+    return expandToNode`
+        export type ${name}AstType = ${joinToNode(
+            identifiers,
+            identifier => `${identifier}.AstType`,
+            { separator: ' & ' }
+        )}
+    `.appendNewLine().appendNewLine();
```
sailingKieler: I suggest to drop the appendNewLine()... here and add empty lines in the calling template.

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -112,28 +187,58 @@ function stringifyDefaultValue(value?: PropertyDefaultValue): string | undefined
...
 
-        export type ${config.projectName}TerminalNames = keyof typeof ${config.projectName}Terminals;
+        export type ${name}TerminalNames = keyof typeof ${name}Terminals;
 
-        export type ${config.projectName}KeywordNames =${keywordStrings.length > 0 ? keywordStrings.map(keyword => `${EOL}    | ${keyword}`).join('') : ' never'};
+        export type ${name}KeywordNames =${keywordStrings.length > 0 ? undefined : ' never;'}
+            ${joinToNode(
+                keywordStrings,
+                keyword => `| ${keyword}`,
+                { appendNewLineIfNotEmpty: true, skipNewLineAfterLastItem: true }
```
sailingKieler, line 241: Much nicer then before! 👌

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -112,28 +187,58 @@ function stringifyDefaultValue(value?: PropertyDefaultValue): string | undefined
...
+            )?.append(';')}
 
-        export type ${config.projectName}TokenNames = ${config.projectName}TerminalNames | ${config.projectName}KeywordNames;
-    `.appendNewLine();
+        export type ${name}TokenNames = ${name}TerminalNames | ${name}KeywordNames;
+    `.appendNewLine().appendNewLine();
```
sailingKieler: I suggest to drop the appendNewLine()... here and add empty lines in the calling template.

packages/langium/src/validation/validation-registry.ts
```diff
@@ -5,11 +5,11 @@
  ******************************************************************************/
 
 import type { CodeDescription, DiagnosticRelatedInformation, DiagnosticTag, integer, Range } from 'vscode-languageserver-types';
-import { assertUnreachable } from '../index.js';
 import type { LangiumCoreServices } from '../services.js';
 import type { AstNode, AstReflection, Properties } from '../syntax-tree.js';
 import type { CancellationToken } from '../utils/cancellation.js';
 import { MultiMap } from '../utils/collections.js';
+import { assertUnreachable } from '../utils/errors.js';
```
sailingKieler: I would prefer to just revert this change as it doesn't contribute anything relevant wrt. this PR's intention.

packages/langium-cli/test/generator/ast-generator.test.ts
```diff
@@ -424,6 +426,7 @@ describe('Ast generator', () => {
         };
     `));
 
+    // TODO warum sollen called Terminals nicht uncludiert werden?
```
sailingKieler: Please remove german comment.

packages/langium-cli/test/generator/ast-generator.test.ts
```diff
@@ -695,11 +698,675 @@ async function testGenerated(grammar: string, expected: string, start: string, e
...
+                    grammar OneGrammar
+                    import "two";
+                    entry Entry: A | B;
+                    A: 'a' a=ID;
+                    terminal ID: /[a-zA-Z0-9]+/;
+                    hidden terminal WS: /\s+/;
+                `,
+                expectedAstContentParts: [
+                    expandToStringIndented(2, expandToNode`
+                        A: {
```
sailingKieler: Regarding your question above:
Yes, there is. You can tell the evaluator to consider less indentation as template white space by inserting a substitution contributing _nothing_. That does also work with the substitution and `A` on the same line but looks a bit uglier. The proposed version expects a leading line break before `A` that is totally fine and not an issue here.
```suggestion
                ${''}
                        A: {
```

packages/langium-cli/test/generator/ast-generator.test.ts
```diff
@@ -695,11 +698,675 @@ async function testGenerated(grammar: string, expected: string, start: string, e
...
+    test.skip('Use this test case for debugging the generation of the Requirements example (multi-language project)', async () => {
+        await generate({ file: 'examples/requirements/langium-config.json' });
+    });
+});
+
+interface GrammarFile {
+    path: string;
+    languageID: string | undefined; // If not undefined, this grammar file is used as entry point for a language with the given value as 'id'
+    grammarContent: string;
+    expectedAstContentParts: string[];
```
sailingKieler: Some minor polishing suggestions:

```suggestion
    expectedAstContentParts: Generated[];
```

packages/langium-cli/test/generator/ast-generator.test.ts
```diff
@@ -695,11 +698,675 @@ async function testGenerated(grammar: string, expected: string, start: string, e
...
+}
+async function testMultiProject(grammarFiles: GrammarFile[], moreExpectedParts: string[]): Promise<void> {
```
sailingKieler: A varArg param allows to skip the array brackets above:

```suggestion
async function testMultiProject(grammarFiles: GrammarFile[], ...moreExpectedParts: Generated[]): Promise<void> {
```

packages/langium-cli/test/generator/ast-generator.test.ts
```diff
@@ -695,11 +698,675 @@ async function testGenerated(grammar: string, expected: string, start: string, e
...
+    // collect all information
+    const config: LangiumConfig = {
+        [RelativePath]: './',
+        projectName: 'test',
+        languages: [], // will be filled later
+    };
+    const documents: LangiumDocument[] = [];
+    const expectedParts: string[] = [];
```
sailingKieler, line 1710:
```suggestion
    const expectedParts: Generated[] = [];
```

packages/langium-cli/test/generator/ast-generator.test.ts
```diff
@@ -695,11 +698,675 @@ async function testGenerated(grammar: string, expected: string, start: string, e
...
+                languageConfig: e[1],
+                identifier: getLanguageIdentifier(config, entryGrammar), // the name defined after the 'grammar' keyword inside the *.langium file
+            });
+        }
+        languages.sort((e1, e2) => e1.identifier.localeCompare(e2.identifier));
+        generated = generateAstMultiLanguageProject(services.grammar, languages, config);
+    }
+
+    // check the generated content
+    expectedParts.forEach(part => expect(generated).toContain(part));
```
sailingKieler, line 1765:
```suggestion
    expectedParts.forEach(part => expect(generated).toContain(toString(part)));
```

packages/langium-cli/src/generate.ts
JohannesMeierSE: Yes, the use of `configMap` is not correct anymore. But during embedding grammars, there are even more issues, which I am currently investigating.

packages/langium-cli/src/generator/ast-generator.ts
JohannesMeierSE: Done, but line 99 results in multiple lines 🙂

packages/langium-cli/src/generate.ts
JohannesMeierSE: > Maybe this entire function (runGenerator(...)) could be tested in one of your tests (or an additional one)?

I thought about that, but `runGenerator` uses the file system and we have no test cases so far, which use temporary folders on disk ...

packages/langium/src/validation/validation-registry.ts
JohannesMeierSE: done

packages/langium-cli/test/generator/ast-generator.test.ts
JohannesMeierSE: done

packages/langium-cli/src/generator/ast-generator.ts
JohannesMeierSE: done

packages/langium-cli/src/generator/ast-generator.ts
JohannesMeierSE: done

packages/langium-cli/src/generator/ast-generator.ts
JohannesMeierSE: done

packages/langium-cli/test/generator/ast-generator.test.ts
JohannesMeierSE: nice improvement, thanks

packages/langium-cli/test/generator/ast-generator.test.ts
JohannesMeierSE: Since we are working with `string`s only, this suggestion makes the reading of test cases harder and not easier.

packages/langium-cli/test/generator/ast-generator.test.ts
JohannesMeierSE: OK, but here I need to add *more* indentation (not *less* indentation).

packages/langium-cli/src/generator/ast-generator.ts
```diff
@@ -4,47 +4,129 @@
...
+    const astTypes = collectAst(languages.map(l => l.embeddedGrammar), { services });
+
+    const fileNode = expandToNode`
+        ${generateAstHeader(config)}
+
+        ${joinToNode( // a namespace for each each language with its language-specific elements: reachable terminals & keywords, complete AstType list
+            languages,
+            language => expandToNode`
+                /** Contains the reachable terminals & keywords and all available types of the '${language.identifier}' language. */
+                export namespace ${language.identifier} {
```
spoenemann, line 54: What if the language name clashes with a type name? We should detect that case and try to resolve it e.g. by prefixing underscores `_`

Theoretically, this can happen with other things defined in ast.ts, too. I'm not sure whether we should handle them all.

JohannesMeierSE: > What if the language name clashes with a type name? We should detect that case and try to resolve it e.g. by prefixing underscores `_`

You are right, clashes are possible and not checked/validated so far.

I prefer `Project_Language` instead of `_Language`.

JohannesMeierSE: > Theoretically, this can happen with other things defined in ast.ts, too. I'm not sure whether we should handle them all.

True. We could validate that, but then we need access to the `langium-config.json` within the Langium language server, since it contains the `"projectName"` ...

cdietrich: funnily this is something we look into for our DSL atm too, also clashes with reserved keywords aka `export namespace default`

cdietrich:
ofc the file grows and is now 10k lines (from 6k)

packages/langium-cli/test/generator/ast-generator.test.ts
JohannesMeierSE: Fixed in a joint session

packages/langium-cli/src/generator/ast-generator.ts
spoenemann: I see three alternative approaches:
- Validation (make certain names forbidden)
- Automatically make generated names unique (e.g. add `_` until it's unique)
- Use characters that are not available in the grammar language (e.g. `$`)

But I don't think this has to be solved in the same PR. Let's keep it simple and extract this into a separate issue so we don't further delay the launch of v4.0.0.

JohannesMeierSE: @spoenemann We just decided in the Langium dev meeting, that I will add a validation, which checks that names of entry grammars of languages and names of types are different.

JohannesMeierSE:
According to our discussions in our Langium dev meeting yesterday, I pushed two more commits with these contributions (and updated the PR description accordingly). This PR is ready for another review.

* Since the name of the entry grammar is used as namespace name for the language parts in the `ast.ts`, there is a new validation, which ensures, that grammar names and rule/type names are different. This might result in new errors during the validation in existing projects, e.g. in our own statemachine example.
* The TypeScript definitions of unions and interfaces are not separated from each other, but mixed and sorted now. Therefore, type definitions in the `ast.ts` are not changed, but reordered in existing projects.
* There is a new validation, which ensures, that all grammars have a unique name.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -159,6 +160,14 @@ export class LangiumGrammarValidator {
                     data: diagnosticData(IssueCodes.GrammarNameUppercase)
                 });
             }
+            for (const otherGrammar of (this.documents.all.toArray()).map(doc => doc.parseResult.value).filter(ast.isGrammar)) {
```
msujew: Suggestion: While this works, this check could show false positives in the language server. If I have two grammars with the same name (which are not related) in my workspace, it will show this error, even though the generator should work just fine for both.

This should check on the imports to ensure that we don't accidentally import a grammar with the same name.

packages/langium/src/grammar/validation/validator.ts
```diff
@@ -814,11 +823,16 @@ export class LangiumGrammarValidator {
...
-            accept('error', `'${name}' is a reserved name of the JavaScript runtime.`, {
-                node,
-                property
-            });
+        if (typeof name === 'string') {
+            if (reservedNames.has(name)) {
+                accept('error', `'${name}' is a reserved name of the JavaScript runtime.`, { node, property });
+            }
+            // don't use names of grammars:
+            for (const grammar of this.documents.all.map(doc => doc.parseResult.value).filter(ast.isGrammar)) {
```
msujew: Suggestion: Similar issue to the one above. We should check:

1. The grammar that contains this node
2. The import that imports this rule

Otherwise we also get false positives.

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: I adjusted the validation and test cases.

packages/langium/src/grammar/validation/validator.ts
JohannesMeierSE: I reworked the implementation and added more test cases.
This PR is ready for another review.

JohannesMeierSE:
From the discussion today in the dev meeting:

```typescript
MetadataElementPropertyDefinition["name"][]
  export const EnumPropertyValues = [
    "description",
    "active",
  ] as const satisfies MetadataElementPropertyDefinition["name"][]

interface MetadataElementPropertyDefinition {
    name: 'extraData'
            | 'uuid'
            | 'description'
            | 'documentation'
            | 'label'
            | 'validation'
```

cdietrich:
created https://github.com/eclipse-langium/langium/issues/2045

PR #1981, opened 2025-07-08
make linker subclassing more easy

cdietrich:


packages/langium/src/references/linker.ts
```diff
@@ -78,9 +78,9 @@ export interface Linker {
 
 }
 
-const ref_resolving = Symbol('ref_resolving');
+export const ref_resolving = Symbol('ref_resolving');
```
msujew: Minor: Since we're exporting this now, we should give it a better name
```suggestion
export const RefResolving = Symbol('RefResolving');
```

PR #1987, opened 2025-07-22
chore: bump eslint

cdietrich:


package-lock.json
```diff
@@ -857,19 +857,32 @@
...
+      "integrity": "sha512-bkOp+iumZCCbt1K1CmWf0R9pM5yKpDv+ZXtvSyQpudrI9kuFLp+bM2WOPXImuD/ceQuaa8f5pj93Y7zyECIGNA==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@types/json-schema": "^7.0.15"
+      },
+      "engines": {
+        "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
+      }
+    },
```
msujew: Shouldn't we update our `eslint` as well? The current version relies on version `0.14.0` of `@eslint/core`. This change includes two different versions of `@eslint/core`.

cdietrich: good question ...

cdietrich: do we have some general practices how we update such things?

package.json
```diff
@@ -27,20 +27,20 @@
   },
   "devDependencies": {
     "@eslint/eslintrc": "~3.3.1",
-    "@eslint/js": "~9.27.0",
+    "@eslint/js": "~9.31.0",
     "@stylistic/eslint-plugin": "~4.2.0",
```
cdietrich, line 31: kept this one as the new one leads to new violations

package-lock.json
msujew: I don't think so.

PR #1990, opened 2025-07-29
AstUtils: added an optional `trace` map to `copyAstNode`

sailingKieler:
Just another convenience thing.


msujew (approved):

packages/langium/src/utils/ast-utils.ts
```diff
@@ -264,10 +264,16 @@ function copyDefaultValue(propertyType: PropertyType): PropertyType {
  * The resulting copy will only contain semantically relevant information, such as the `$type` property and AST properties.
  *
  * References are copied without resolved cross reference. The specified function is used to rebuild them.
+ *
+ * For the sake of tracking copied nodes and their originals a `trace` map can be provided.
```
msujew: Can you convert this into a `@param`?

sailingKieler: Yes.
I also just thought adding both directions to the map, i.e.
```js
    if (trace) {
        trace.set(node, copy);
        trace.set(copy, node);
    }
```

Any objections?

msujew: Fine with me.

PR #1991, opened 2025-07-30
Remove long-time deprecated properties

msujew:
Removes all fields/methods that were deprecated before the release of Langium v3. Adopters have had more than enough time to use the recommended alternatives.

### DIFF
--- a/packages/langium/src/lsp/semantic-token-provider.ts
+++ b/packages/langium/src/lsp/semantic-token-provider.ts
@@ -58,7 +58,7 @@ export const AllSemanticTokenModifiers: Record<string, number> = {
 };
 
 /**
- * @deprecated `SemanticTokenProvider` now supplies its own options.
+ * @deprecated Since 3.2.0. `SemanticTokenProvider` now supplies its own options.
  */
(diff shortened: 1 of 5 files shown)


JohannesMeierSE:
BTW Is this worth a note in the Changelog?

msujew:
@cdietrich What?

cdietrich:
grafik

packages/langium/src/lsp/semantic-token-provider.ts
```diff
@@ -58,7 +58,7 @@ export const AllSemanticTokenModifiers: Record<string, number> = {
 };
 
 /**
- * @deprecated `SemanticTokenProvider` now supplies its own options.
+ * @deprecated Since 3.2.0. `SemanticTokenProvider` now supplies its own options.
```
cdietrich, line 61: should this be since 4.0

msujew: The PR that actually changed this landed in 3.2.0 already. See https://github.com/eclipse-langium/langium/pull/1600. I just recently added the deprecation notice, since I didn't notice it was unused before that.

cdietrich (approved):

PR #2011, opened 2025-08-25
Fix empty infix parser error

msujew:
Follow up on https://github.com/eclipse-langium/langium/pull/1948, which improves on the error case, in case the array hasn't even been set yet.


packages/langium/src/parser/langium-parser.ts
```diff
@@ -466,21 +466,19 @@ export class LangiumParser extends AbstractLangiumParser {
     }
 
     private constructInfix(obj: InfixElement, precedence: Map<string, OperatorPrecedence>): any {
-        if (obj.parts.length === 1) {
+        const parts = obj.parts ?? [];
+        const operators = obj.operators ?? [];
+        if (parts.length < 2) {
```
sailingKieler: Wouldn't a check like this be sufficient here?
```suggestion
        if (!Array.isArray(obj.parts) || obj.parts.length <= 1) {
```
For the remainder of this method (parts >= 2) the assignment to a local variable doesn't make any difference, does it?

sailingKieler (commented):
I took some time to understand what's going on here. Pretty cool!
A dedicated unit test would make sense, wouldn't it? Or did I overlook it?

msujew:
@sailingKieler Added a test and updated the code. I've kept the local variable, because it just makes the code easier to read IMO.

packages/langium/src/parser/langium-parser.ts
sailingKieler: Update: It needs to be `!Array.isArray(obj.parts)` of course!

packages/langium/src/parser/langium-parser.ts
```diff
@@ -468,7 +468,7 @@ export class LangiumParser extends AbstractLangiumParser {
     private constructInfix(obj: InfixElement, precedence: Map<string, OperatorPrecedence>): any {
         const parts = obj.parts ?? [];
         const operators = obj.operators ?? [];
-        if (parts.length < 2) {
+        if (!Array.isArray(parts) || !Array.isArray(operators) || parts.length < 2) {
             // Captured just a single, non-binary expression
             // Simply return the expression as is.
             return parts[0];
```
sailingKieler, line 478: Does it work this way? Should be `parts?.[0]`, shouldn't it?

sailingKieler: And you can drop the `?? []` inits now 😉

msujew: Yeah, that's right 👍

sailingKieler: Makes sense (I had to look twice).

sailingKieler (approved):
🚀

PR #2020, opened 2025-09-01
langium: extended Multimap constructor to accept an Iterable of initial data.

sailingKieler:
added corresponding tests

### DIFF
--- a/packages/langium/src/utils/collections.ts
+++ b/packages/langium/src/utils/collections.ts
@@ -15,8 +15,8 @@ export class MultiMap<K, V> {
     private map = new Map<K, V[]>();
 
     constructor()
-    constructor(elements: Array<[K, V]>)
-    constructor(elements?: Array<[K, V]>) {
+    constructor(elements: Iterable<[K, V]>)
+    constructor(elements?: Iterable<[K, V]>) {
         if (elements) {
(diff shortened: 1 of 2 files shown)


packages/langium/src/utils/collections.ts
```diff
@@ -15,8 +15,8 @@ export class MultiMap<K, V> {
     private map = new Map<K, V[]>();
 
     constructor()
-    constructor(elements: Array<[K, V]>)
-    constructor(elements?: Array<[K, V]>) {
+    constructor(elements: Array<[K, V]> | Iterable<[K, V]>)
+    constructor(elements?: Array<[K, V]> | Iterable<[K, V]>) {
```
msujew: Question: Why make this a union? Isn't `Array` always an `Iterable`?

sailingKieler: I was in doubt about this. From what I can see `Iterable<[K, V]>` should be fine but I was not fully sure, and the union doesn't hurt. But can I drop the array part if you prefer.

msujew: It only adds confusion imo.

sailingKieler: Changed.
@msujew feel free to have another look.

PR #2022, opened 2025-09-03
langium: removed short cut guards in document builder's 'resetToState' method

sailingKieler:
in language implementations with transitively resolved cross-refs documents may contain resolved cross-refs although their state is still 'ComputedScopes' or even 'IndexContent' -> reset those docs properly, too;

follow-up on PR #1977


msujew (approved):
Makes sense to me, thanks 👍

PR #2023, opened 2025-09-03
Fix infix rule CST issues

msujew:
Follow up on https://github.com/eclipse-langium/langium/pull/2011 which fixes an issue where the infix rules object was stored in the CST `astNode` item by accident.


packages/langium/src/parser/cst-node-builder.ts
```diff
@@ -89,9 +89,9 @@ export class CstNodeBuilder {
 
     construct(item: { $type: string | symbol | undefined, $cstNode: CstNode }): void {
         const current: CstNode = this.current;
-        // The specified item could be a datatype ($type is symbol) or a fragment ($type is undefined)
+        // The specified item could be a datatype ($type is symbol), fragment ($type is undefined) or infix rule ($infix exists)
         // Only if the $type is a string, we actually assign the element
-        if (typeof item.$type === 'string') {
+        if (typeof item.$type === 'string' && !('$infix' in item)) {
```
sailingKieler: Wrt. performance: Would it be advantageous to add `$infix?: boolean` to the type def of `item` and just test `!item.$infix`?

msujew: I don't think it matters, but it makes it more readable 👍

cdietrich:
@msujew should sth like this still happen with 4.0.1

grafik

msujew:
@cdietrich Not really, but I'm not sure whether it's actually feasible to rewrite the whole CST during the binary operation AST rewrite. I'm quite unhappy with the feature myself - it works better in another project where we embed the CST in the AST (non-langium project).

PR #2024, opened 2025-09-05
refined `waitUntil` with document URI to fire as soon as document reaches required state

sailingKieler:
@msujew here's my proposal on a refinement of `DocumentBuilder.waitUntil(...)` as discussed.

Is the way of reporting wrong pre-conditions via rejecting with ResponseErrors the way to go?
I enforced them with VS Code but didn't see any outputs in the log or did notice any UI pop-ups...


msujew (approved):
Thanks, looks good to me 👍

packages/langium/src/workspace/document-builder.ts
```diff
@@ -500,26 +501,67 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+            return Promise.reject(
+                new ResponseError(
+                    LSPErrorCodes.RequestFailed,
+                    `Document state of ${uri.toString()} is ${DocumentState[document.state]}, requiring ${DocumentState[state]}, but workspace state is already ${DocumentState[this.currentState]}. Returning undefined.`
+                )
+            );
         }
+        return new Promise((resolve, reject) => {
+            const buildDisposable = this.onDocumentPhase(state, (doc) => {
+                if (doc.uri.toString() === uri.toString()) {
```
msujew: FYI, you could use `UriUtils.equals(doc.uri, uri)` here.

PR #2030, opened 2025-09-15
generator-langium: extension package: Remove lint call

kaisalmen:


msujew:
Note that the tests are still failing.

kaisalmen:
@msujew I just force-pushed a change. The `package.json` comparison failed. Should one of the test check the extension prepublish? WDYT?

PR #2032, opened 2025-09-16
Use readonly fields for type meta data

msujew:
Fixes an issue we noticed in https://github.com/eclipse-langium/langium/discussions/1997#discussioncomment-14006947:

Using `const` automatically makes all fields and arrays `readonly` in TypeScript. We use this in the `types` field of the generated AST reflection. We need to reflect this in the types expected by the `AstMetaData`, otherwise, we might run into compilation issues.


msujew:
Closing this as a false-positive. This was a result of an outdated version of TypeScript.

PR #2035, opened 2025-09-18
DocumentValidator: configure the validation of single nodes

JohannesMeierSE:
This PR extends the `DocumentValidator` to configure the validation of the current AstNode:
- whether the current node itself should be validated or not
- whether its children should be validated or not: Skipping sub-trees improves performance

### DIFF
--- a/packages/langium/test/workspace/document-validator.test.ts
+++ b/packages/langium/test/workspace/document-validator.test.ts
@@ -4,9 +4,9 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { AstNode, ValidationChecks } from 'langium';
+import type { AstNode, LangiumCoreServices, Module, PartialLangiumCoreServices, ValidateSingleNodeOptions, ValidationChecks, ValidationOptions } from 'langium';
 import { describe, expect, test } from 'vitest';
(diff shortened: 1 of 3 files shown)


packages/langium/test/workspace/document-validator.test.ts
```diff
@@ -93,6 +93,83 @@ describe('DefaultDocumentValidator', () => {
...
+    async function createServices() {
+        const grammar = `
+            grammar Test
+            entry Model:
+                foos+=Foo*;
+            Foo:
+                'foo' value=INT;
+            terminal INT returns number: /[0-9]+/;
+            hidden terminal WS: /\\s+/;
+        `;
```
msujew: Minor: Can we move this to the root scope? AFAIK this is a 1:1 copy of the other grammar in the code above.

msujew (commented):
Looks good to me, just one issue, see below.

packages/langium/test/workspace/document-validator.test.ts
JohannesMeierSE: done

JohannesMeierSE:
@msujew I just pushed updates, including to freeze the default config value.

PR #2047, opened 2025-09-26
DefaultWorkspaceManager: improve error handling in traverseFolder

### LINKED ISSUE #2044: Language server does not initialize properly if workspace-folder does not exist (anymore)
Langium version: 4.1.0
Package name: langium

## Steps To Reproduce

1. create 2 directories somewhere on your system (dir-a and dir-b)
2. Launch VS-Code (or if you have a langium-project in your current VS-Code instance launch an Extension Development Host using F5)
3. Use "File > Add Folder to Workspace" and add the folders you created above
4. Save the workspace using "Save workspace as"
5. Create a file with the langium language you are developing in "dir-a" and fill it with content
6. Shutdown your VS-Code instance (or Extension Development Host)
7. Remove "dir-b"
8. Restart your VS-Code on the workspace you saved or the Extension Development Host (it has remembered your workspace from above)
9. Open the file you created in dir-a
10. Notice that no outline is available, auto-complete does not work, ... - the only thing working is the syntax highlighting

I took me a bit but once you look in the Language Server Console Output of your language you'll notice something like the following

```
Error in WorkspaceManager initialization: Error: ENOENT: no such file or directory, scandir '/Users/tomschindl/bla/dir-b'
    at async readdir (node:internal/fs/promises:956:18)
    at async NodeFileSystemProvider.readDirectory (/Users/tomschindl/git/rsd-2.0/dsl/packages/extension/out/language/main.cjs:32507:21)
    at async DefaultWorkspaceManager.traverseFolder (/Users/tomschindl/git/rsd-2.0/dsl/packages/extension/out/language/main.cjs:30181:21)
    at async Promise.all (index 3)
    at async DefaultWorkspaceManager.performStartup (/Users/tomschindl/git/rsd-2.0/dsl/packages/extension/out/language/main.cjs:30151:5)
    at async DefaultWorkspaceManager.initializeWorkspace (/Users/tomschindl/git/rsd-2.0/dsl/packages/extension/out/language/main.cjs:30133:23)
    at async /Users/tomschindl/git/rsd-2.0/dsl/packages/extension/out/language/main.cjs:30087:24
    at async Promise.all (index 0)
    at async DefaultWorkspaceLock.performNextOperation (/Users/tomschindl/git/rsd-2.0/dsl/packages/extension/out/language/main.cjs:30085:5) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'scandir',
  path: '/Users/tomschindl/bla/foo'
}
```

The language server failed to initialize properly because it tried to read the an in existing directory.

## The current behavior

Language server does not initialize properly if workspace has a folder who does not exist anymore

## The expected behavior

Language server can deal with the fact that a workspace folder does not exist

## Proposed solution

I digged a bit into the code and I think the best option would be to check in `traverseFolder` the provided URI and if it does not exist simply return a resolved Promise.

tomsontom:
fixes #2044 

- check if directory exists eg in workspaces a directory could have been delete since the last start
- catch all errors while reading the directory content

### DIFF
--- a/packages/langium/src/workspace/workspace-manager.ts
+++ b/packages/langium/src/workspace/workspace-manager.ts
@@ -206,9 +206,8 @@ export class DefaultWorkspaceManager implements WorkspaceManager {
                     }
                 }
             }));
-        } catch(e) {
-            console.error('Failure while reading directory content', e);
-            return Promise.resolve();
+        } catch (e) {
+            console.error('Failure to read directory content of ' + folderPath.toString(true), e);
         }
     }
 


packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -195,16 +195,25 @@ export class DefaultWorkspaceManager implements WorkspaceManager {
...
-                if (entry.isDirectory) {
-                    await this.traverseFolder(entry.uri, uris);
-                } else if (entry.isFile) {
-                    uris.push(entry.uri);
-                }
+        try {
+            if( ! this.fileSystemProvider.existsSync(folderPath) ) {
+                console.info(`Directory ${folderPath} does not exist`);
+                return Promise.resolve();
             }
```
sailingKieler: What do you think about moving this diagnosis part into the `catch` block?, like
```
catch (e) {
    if(await this.fileSystemProvider.exists(folderPath) ) {
        console.error('Failure while reading directory content', e);
    } else {
        console.info(`Directory ${folderPath} does not exist`);
    }
}
```

msujew: On that note: Please don't use the `sync` file system methods in framework code - there is no guarantee that every file system is actually capable of supporting sync file system requests.

To be honest, I don't even see a reason to actually call `exists` at all, since the `e` error object will contain the information what kind of error occurred.

tomsontom: you are right I removed the check and let readDirectory fail

PR #2054, opened 2025-10-20
Validation for assigned calls of fragments checks also nested calls now

JohannesMeierSE:


msujew (approved):
Thanks, makes sense 👍

PR #2057, opened 2025-10-23
Extended checks for multiple assignments to the same feature

JohannesMeierSE:
This PR is a follow-up of #1412 and #1766 and improves and extends the checks of multiple assignments (in the grammar) to the same feature/property (in the AST):

- Validate the mixed use of `?=`, `=` and `+=` for the same feature: A warning (no error) is shown, if values are assigned with different operators to the same feature. This is prone to data loss, among others.
- Contributed a new code action to replace the assignment operator `=` with `+=`.
- Fixed bugs for the existing validation, when there are multiple assignments with `=` (or `?=`):
    - Create validation markers for critical assignments in other documents in the currently validated document instead (a called fragment might be located in another document).
    - Validate rewrite actions as starting point in order not to validate and mark them twice (rewrite actions might be located in fragments, which might be called twice).
    - Handle circular calls of fragments
- Wrote more test cases, in particular for the existing check for multiple assignments with `?=` as assignment operator to the same feature.
- Both validations (1st for mixing `?=`, `=`and `+=`, and 2nd for multiple assignments with `?=` or `=`) got an issue code, which is used in the 2nd case for a corresponding, new code action (for `=` only, see above) and which will be relevant for serializing ASTs back to text in the future.

### DIFF
--- a/packages/langium/src/grammar/validation/validator.ts
+++ b/packages/langium/src/grammar/validation/validator.ts
@@ -56,6 +56,7 @@ export function registerValidationChecks(services: LangiumGrammarServices): void
         ],
         ParserRule: [
             validator.checkParserRuleDataType,
+            validator.checkFragmentKeywords,
             validator.checkRuleParameters,
             validator.checkEmptyParserRule,
             validator.checkParserRuleReservedName,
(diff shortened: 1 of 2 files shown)


packages/langium/src/grammar/validation/validator.ts
```diff
@@ -1016,48 +1019,118 @@ export class LangiumGrammarValidator {
...
     }
 
-    private checkOperatorMultiplicitiesForMultiAssignmentsIndependent(startNodes: AstNode[], accept: ValidationAcceptor, map: Map<string, AssignmentUse> = new Map()): void {
+    checkActionAssignmentsToTheSameFeature(action: ast.Action, accept: ValidationAcceptor): void {
+        if (action.feature) {
+            // tree rewriting action
+            const mapForNewObject = new Map();
+            storeAssignmentUse(mapForNewObject, action.feature, 1, action); // remember the special rewriting feature
+            // all following nodes are put into the new object => check their assignments independently
+            const sibblings = ast.isGroup(action.$container) || ast.isUnorderedGroup(action.$container) ? action.$container.elements : [action];
```
emilkrebs: Minor typo: use `siblings` instead of `sibblings`

PR #2058, opened 2025-10-24
Validate that no return type is specified for fragments

JohannesMeierSE:


packages/langium/src/grammar/validation/validator.ts
```diff
@@ -971,11 +972,23 @@ export class LangiumGrammarValidator {
...
+        } else if (hasDatatypeReturnType && !dataTypeRule && !rule.fragment) { // fragments are validated in the next check below
             accept('error', 'Normal parser rules are not allowed to return a primitive value. Use a datatype rule for that.', { node: rule, property: rule.dataType ? 'dataType' : 'returnType' });
         }
     }
 
+    checkFragmentKeywords(rule: ast.ParserRule, accept: ValidationAcceptor): void {
+        if (rule.fragment) {
+            if (rule.dataType || rule.returnType || rule.inferredType) {
+                accept('error',
+                    "Fragments assign values to the object of the caller, but don't create a new object themselves. Therefore specifying the type of the returned object is not possible.",
```
msujew: Suggestion: The validation message is too long IMO.
```suggestion
                    "Fragment rules cannot specify a return type.",
```

emilkrebs: Makes sense, the longer description about why can be found in the documentation if the user requires more details: https://langium.org/docs/reference/grammar-language/#rule-fragments

JohannesMeierSE: I just pushed a new commit with the proposed, shorter error message.

PR #2064, opened 2025-11-06
Document BNF Generator

montymxb:
Was looking into the bnf generator and realized it wasn't documented externally yet. Went ahead & added that plus an associated log entry to match the other generators.


montymxb:
@spoenemann yep I can add those in today before pulling in the changes. No problem.

PR #2071, opened 2025-11-20
Unified the call of `resetToState` in the default document builder

JohannesMeierSE:
Motivation for this PR: In a Langium project, we annotated the `LangiumDocument` with additional properties to remember special situations in order to improve performance.
- Re-setting/un-setting these additional properties is done in a customization of `resetToState()` in the `DefaultDocumentBuilder`.
- During debugging, we found, that `resetToState()` is not called for documents which are explicitly given as changed documents to the document builder. `resetToState` is called only for documents which "should be relinked" due to the changed documents.

Therefore this PR aims to unify the call of `resetToState` in the default document builder.

Details and additional changes:
- Added call of `resetToState` for all explicitly given changed documents.
  - Exception to be discussed: `build(...)` with new validation categories
- As a consequence, `LangiumDocuments.invalidateDocument` became superflous, since all its statements are already part of `resetToState`.
  - Exception: `this.indexManager.removeContent(document.uri);` was missing. This didn't caused problems, since the content is _replaced later_ with `indexManager.updateContent(...)`. The same counts for `indexManager.removeReferences` and `indexManager.updateReferences`.
    - One could argue, that this is beneficial, since the information might be useful to identify other affected documents (`shouldRelink`).
    - But that strongly depends on the DSL. Additionally, it is inconsistent, since references of documents are reset at different points in time.
  - Additionally this improves separation of concerns a lot, since `LangiumDocuments` now focuses on storing documents (add, remove, get documents), while the document builder focuses on processing and updating document-related data. Disadvantage: This is a breaking API change!
  - Therefore the code is kept, but not used anymore with corresponding `@deprecated` comments.
- Fixed an important bug: If `resetToState` is used to mark documents to be updated with `document.state >= DocumentState.Linked` and `update` is called later without explicitly providing the URI of this document, this document is _not_ rebuild.
- Introduced `cleanUpDeleted` for deleted documents, this refactoring ensures a similar code structure.
- `getDocuments` was not exposed in the `LangiumDocuments` service.
- Fixed a gap: Update documents after updates of `updateBuildOptions`
- Might improve the performance and enable faster LSP responses: Mark documents as complete, when they are completed, not after completing the whole build
- Introduced `ValidationRegistry.getAllValidationCategories(LangiumDocument)` to easily integrate custom validations and provide different validation categories for different documents. By default, the `ValidationRegistry` just collects all validation categories, for which some checks are registered (plus the default categories).

### DIFF
--- a/packages/langium/src/workspace/document-builder.ts
+++ b/packages/langium/src/workspace/document-builder.ts
@@ -19,7 +19,7 @@ import { MultiMap } from '../utils/collections.js';
 import { OperationCancelled, interruptAndCheck, isOperationCancelled } from '../utils/promise-utils.js';
 import { stream } from '../utils/stream.js';
 import { UriUtils, type URI } from '../utils/uri-utils.js';
-import { ValidationCategory } from '../validation/validation-registry.js';
+import type { ValidationCategory } from '../validation/validation-registry.js';
 import { DocumentState } from './documents.js';
 import type { FileSystemProvider } from './file-system-provider.js';
 import type { WorkspaceManager } from './workspace-manager.js';
@@ -151,9 +151,10 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
     protected readonly langiumDocumentFactory: LangiumDocumentFactory;
     protected readonly textDocuments: TextDocumentProvider | undefined;
     protected readonly indexManager: IndexManager;
-    protected readonly serviceRegistry: ServiceRegistry;
     protected readonly fileSystemProvider: FileSystemProvider;
     protected readonly workspaceManager: () => WorkspaceManager;
+    protected readonly serviceRegistry: ServiceRegistry;
+
     protected readonly updateListeners: DocumentUpdateListener[] = [];
     protected readonly buildPhaseListeners = new MultiMap<DocumentState, DocumentBuildListener>();
     protected readonly documentPhaseListeners = new MultiMap<DocumentState, DocumentPhaseListener>();
@@ -268,15 +269,11 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
 
     protected findMissingValidationCategories(document: LangiumDocument, options: BuildOptions | undefined): ValidationCategory[] {
         const state = this.buildState.get(document.uri.toString());
-        const allCategories = this.getAllValidationCategories(document);
-        const executedCategories = new Set(state?.result?.validationChecks ?? (state?.completed ? allCategories : []));
+        const allCategories = this.serviceRegistry.getServices(document.uri).validation.ValidationRegistry.getAllValidationCategories(document);
+        const executedCategories = state?.result?.validationChecks ? new Set(state?.result?.validationChecks) : state?.completed ? allCategories : new Set();
         const requestedCategories = (options === undefined || options.validation === true) ? allCategories
             : typeof options.validation === 'object' ? (options.validation.categories ?? allCategories) : [];
-        return requestedCategories.filter(requested => !executedCategories.has(requested));
-    }
-
-    protected getAllValidationCategories(_document: LangiumDocument): readonly ValidationCategory[] {
-        return ValidationCategory.defaults;
+        return stream(requestedCategories).filter(requested => !executedCategories.has(requested)).toArray();
     }
 
     protected async findChangedUris(changed: URI): Promise<URI[]> {
--- a/packages/langium/test/validation/document-validator.test.ts
+++ b/packages/langium/test/validation/document-validator.test.ts
@@ -4,9 +4,9 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import { AstUtils, DefaultDocumentBuilder, type AstNode, type LangiumDocument, type Module, type ValidationAcceptor, type ValidationCategory, type ValidationChecks } from 'langium';
+import { AstUtils, type AstNode, type ValidationAcceptor, type ValidationChecks } from 'langium';
 import { createServicesForGrammar } from 'langium/grammar';
-import type { LangiumServices, LangiumSharedServices, PartialLangiumSharedServices } from 'langium/lsp';
+import type { LangiumServices } from 'langium/lsp';
 import type { ValidationHelperOptions, ValidationResult } from 'langium/test';
 import { validationHelper } from 'langium/test';
 import { beforeAll, describe, expect, test } from 'vitest';
@@ -272,20 +272,8 @@ describe('Register Before/AfterDocument logic for validations with state', () =>
         let validate: (input: string, options?: ValidationHelperOptions) => Promise<ValidationResult<AstNode>>;
 
         beforeAll(async () => {
-            class DocumentBuilderWithAdditionalValidationCategory extends DefaultDocumentBuilder {
-                // The `validationHelper` below uses "BuildOptions.validation = true" to execute all validation categories (see API documentation for the value `true`).
-                //  Therefore we need to register the custom validation category to be included in all these validation categories.
-                protected override getAllValidationCategories(document: LangiumDocument): readonly ValidationCategory[] {
-                    return [ ...super.getAllValidationCategories(document), 'user' ];
-                }
-            }
             const services = await createServicesForGrammar({
-                grammar,
-                sharedModule: <Module<LangiumSharedServices, PartialLangiumSharedServices>>{
-                    workspace: {
-                        DocumentBuilder: services => new DocumentBuilderWithAdditionalValidationCategory(services),
-                    },
-                },
+                grammar
             });
             const validationChecksUser: ValidationChecks<object> = {
                 AstNode: [
--- a/packages/langium/src/validation/validation-registry.ts
+++ b/packages/langium/src/validation/validation-registry.ts
@@ -14,7 +14,7 @@ import type { MaybePromise } from '../utils/promise-utils.js';
 import { isOperationCancelled } from '../utils/promise-utils.js';
 import type { Stream } from '../utils/stream.js';
 import { stream } from '../utils/stream.js';
-import type { DocumentSegment } from '../workspace/documents.js';
+import type { DocumentSegment, LangiumDocument } from '../workspace/documents.js';
 
 export type DiagnosticInfo<N extends AstNode, P extends string = Properties<N>> = {
     /** The AST node to which the diagnostic is attached. */
@@ -133,11 +133,13 @@ type ValidationCheckEntry = {
  * Manages a set of `ValidationCheck`s to be applied when documents are validated.
  */
 export class ValidationRegistry {
-    private readonly entries = new MultiMap<string, ValidationCheckEntry>();
-    private readonly reflection: AstReflection;
+    protected readonly entries = new MultiMap<string, ValidationCheckEntry>();
+    protected readonly knownCategories = new Set(ValidationCategory.defaults);
 
-    private entriesBefore: ValidationPreparation[] = [];
-    private entriesAfter: ValidationPreparation[] = [];
+    protected readonly reflection: AstReflection;
+
+    protected entriesBefore: ValidationPreparation[] = [];
+    protected entriesAfter: ValidationPreparation[] = [];
 
     constructor(services: LangiumCoreServices) {
         this.reflection = services.shared.AstReflection;
@@ -155,6 +157,7 @@ export class ValidationRegistry {
         if (category === 'built-in') {
             throw new Error("The 'built-in' category is reserved for lexer, parser, and linker errors.");
         }
+        this.knownCategories.add(category); // remember custom/user-defined categories
         for (const [type, ch] of Object.entries(checksRecord)) {
             const callbacks = ch as ValidationCheck | ValidationCheck[];
             if (Array.isArray(callbacks)) {
@@ -272,4 +275,7 @@ export class ValidationRegistry {
         return this.entriesAfter;
     }
 
+    getAllValidationCategories(_document: LangiumDocument): ReadonlySet<ValidationCategory> {
+        return this.knownCategories;
+    }
 }


packages/langium/src/workspace/documents.ts
```diff
@@ -465,19 +460,6 @@ export class DefaultLangiumDocuments implements LangiumDocuments {
...
-        const langiumDoc = this.documentTrie.find(uriString);
-        if (langiumDoc) {
-            const linker = this.serviceRegistry.getServices(uri).references.Linker;
-            linker.unlink(langiumDoc);
-            langiumDoc.state = DocumentState.Changed;
-            langiumDoc.localSymbols = undefined;
-            langiumDoc.diagnostics = undefined;
-        }
-        return langiumDoc;
-    }
```
sailingKieler, line 479: This is a breaking change that we cannot/should not merge into a minor release of Langium 4.
But we can delegate back to the document builder's `resetToState()` method if a document is present. ... and keep a ToDo of consolidating with the next major release — if that will ever emerge...

packages/langium/src/workspace/document-builder.ts
```diff
@@ -375,6 +372,11 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
         }
     }
 
+    resetToDeleted<T extends AstNode>(document: LangiumDocument<T>): void {
```
sailingKieler: I see your attempt to stay in line with `resetToState` but it sounds weird to me. What about
```suggestion
    cleanUpDeleted<T extends AstNode>(document: LangiumDocument<T>): void {
```

packages/langium/test/workspace/document-builder.test.ts
```diff
@@ -757,6 +757,187 @@ describe('DefaultDocumentBuilder', () => {
...
+
+            // update A
+            builder.checkBeforeBuild = () => {
+                expect(documentA.references).toHaveLength(0);
+                expect(documentA.localSymbols).toBe(undefined);
+                expect(documentA.diagnostics).toBe(undefined);
+                expect(documentB.localSymbols).not.toBe(undefined);
+                expect(documentB.diagnostics).toBe(undefined);
+            };
+            await builder.update([documentA.uri], []);
```
sailingKieler, line 973: The builder should behave the same when triggered this way, shouldn't it?
```suggestion
            await builder.update([], []);
```

packages/langium/src/workspace/document-builder.ts
```diff
@@ -220,23 +220,24 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
                 // This allows to cancel the parsing process later in case we need it.
-                const newDocument = this.langiumDocumentFactory.fromModel({ $type: 'INVALID' }, changedUri);
-                newDocument.state = DocumentState.Changed;
-                this.langiumDocuments.addDocument(newDocument);
+                changedDocument = this.langiumDocumentFactory.fromModel({ $type: 'INVALID' }, changedUri);
+                changedDocument.state = DocumentState.Changed; // required, since `langiumDocumentFactory.fromModel` marks the new document as `DocumentState.Parsed`
+                this.langiumDocuments.addDocument(changedDocument);
             }
+            this.resetToState(changedDocument, DocumentState.Changed);
             this.buildState.delete(changedUri.toString());
```
sailingKieler: I suggest to move this line into `resetToState` (at the beginning).
Then the `doc.state < DocumentState.Linked`/`doc.state < DocumentState.Validated` part below can be dropped completely I think.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -220,23 +220,24 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
             const deletedDocs = this.langiumDocuments.deleteDocuments(deletedUri);
             for (const doc of deletedDocs) {
                 deletedUris.push(doc.uri);
+                this.resetToDeleted(doc);
                 this.buildState.delete(doc.uri.toString());
```
sailingKieler: Consequently, this line can go into the method called before.

packages/langium/test/workspace/document-builder.test.ts
```diff
@@ -757,6 +757,187 @@ describe('DefaultDocumentBuilder', () => {
...
+            checkDocumentStateAfterBuild();
+
+            expect(builder.resetted).toHaveLength(1);
+            expect(builder.resetted[0]).toBe('/testA.txt: Changed');
+            expect(builder.actuallyBuilt).toHaveLength(2);
+            expect(builder.actuallyBuilt[0]).toBe('/testA.txt');
+            expect(builder.actuallyBuilt[1]).toBe('/testB.txt');
+        });
```
sailingKieler, line 1011: Very nice tests.
Could you add a sentence or two in comment on the practical use cases you had in mind for the latter two tests (if they exist)?

I would love to see one more test that cancels the build, e.g. after the scope computation phase or once one document has been linked, and resumes via another call of
```
await builder.update([], []);
```

packages/langium/src/workspace/documents.ts
JohannesMeierSE: Yes, the initial PR demonstrated the ideal case for discussion.

I now keep the initial API, but don't use them anymore.

packages/langium/src/workspace/document-builder.ts
JohannesMeierSE: done

packages/langium/test/workspace/document-builder.test.ts
JohannesMeierSE: Yes, there are some new test cases now, which use this approach.

packages/langium/src/workspace/document-builder.ts
JohannesMeierSE: I now finally checked, that the case of `resetToState` without clearing the `buildState` don't matter. Therefore I did this now.

I am still not sure, whether skipping `doc.state < DocumentState.Validated` is a nice idea, but yes, probably that is possible

packages/langium/src/workspace/document-builder.ts
JohannesMeierSE: done

packages/langium/test/workspace/document-builder.test.ts
JohannesMeierSE: Not sure, whether my updates fulfill your needs ...

packages/langium/src/workspace/documents.ts
```diff
@@ -404,14 +413,18 @@ export interface LangiumDocuments {
 
 export class DefaultLangiumDocuments implements LangiumDocuments {
 
+    protected readonly services: LangiumSharedCoreServices;
     protected readonly langiumDocumentFactory: LangiumDocumentFactory;
-    protected readonly serviceRegistry: ServiceRegistry;
+    protected _documentBuilder: DocumentBuilder;
+    protected get documentBuilder(): DocumentBuilder {
+        return this._documentBuilder ??= this.services.workspace.DocumentBuilder;
+    }
```
msujew: Minor: I don't think we actually need to cache this.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -254,15 +249,30 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+    protected resultsAreIncomplete(document: LangiumDocument, options: BuildOptions | undefined): boolean {
+        return this.findMissingValidationCategories(document, options).length >= 1;
+    }
+
+    protected findMissingValidationCategories(document: LangiumDocument, options: BuildOptions | undefined): ValidationCategory[] {
+        const state = this.buildState.get(document.uri.toString());
+        const executedCategories = new Set(state?.result?.validationChecks ?? (state?.completed ? ValidationCategory.all : []));
+        const requestedCategories = (options === undefined || options.validation === true) ? ValidationCategory.all
+            : typeof options.validation === 'object' ? (options.validation.categories ?? ValidationCategory.all) : [];
+        return stream(requestedCategories).filter(requested => !executedCategories.has(requested)).toArray();
```
msujew: Minor: Why create a stream and not just execute `.filter` on the `requestedCategories` array?

packages/langium/src/workspace/document-builder.ts
```diff
@@ -369,12 +375,26 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
 
+    // TODO open question, to be discussed during the review: Shall we expose this method in the Interface, as done for `resetToState`?
+    cleanUpDeleted<T extends AstNode>(document: LangiumDocument<T>): void {
+        this.buildState.delete(document.uri.toString());
+        this.indexManager.remove(document.uri);
+        // In the current implementation, this line is not necessary, since the state is already set before.
+        //  This line does not hurt and makes the code to be in sync with `resetToState`.
+        //  If `cleanUpDeleted` is made available outside (see TODO above), this line becomes necessary.
+        document.state = DocumentState.Changed;
+    }
```
msujew: Suggestion: I believe this is an implementation detail and should be `protected`. To phrase the question differently: Is there a reason for adopters to manually call this method, compared to `update` with the deleted URIs? I don't think so.

packages/langium/test/validation/document-validator.test.ts
```diff
@@ -328,11 +328,11 @@ describe('Register Before/AfterDocument logic for validations with state', () =>
...
         });
 
-        test('default categories', async () => {
+        test('default categories dont include user-defined categories', async () => {
             const validationResult = await validate('model test');
             const diagnostics = validationResult.diagnostics;
-            expect(diagnostics).toHaveLength(2);
-            expect(diagnostics.filter(d => d.severity === 1)).toHaveLength(1); // 1 error
+            expect(diagnostics).toHaveLength(1); // only 'fast' issues
+            expect(diagnostics.filter(d => d.severity === 1)).toHaveLength(0); // 0 error
```
msujew: Question: Is this a breaking change in the behavior?

packages/langium/src/workspace/documents.ts
spoenemann: 👍 instead of storing the `services` in a property, just define `documentBuilder` as a provider function
```typescript
    private documentBuilder: () => DocumentBuilder;
```
and set it in the constructor:
```typescript
this.documentBuilder = () => services.workspace.DocumentBuilder;
```
(we've already done it this way in other services to avoid cyclic references)

As the dependency is only here for backwards-compatibility of a deprecated method, I'd define the property as private so we can remove it later.

packages/langium/test/validation/document-validator.test.ts
spoenemann: Yes it is. The validation helper sets the validation options to `true` if not set. The current documentation (`document-builder.ts` l. 38) says:

> `true` enables all validation checks and forces revalidating the documents

The restriction to `'fast'` issues should only apply when documents are updated through LSP.

packages/langium/src/workspace/document-builder.ts
```diff
@@ -625,11 +656,12 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
      */
     protected async validate(document: LangiumDocument, cancelToken: CancellationToken): Promise<void> {
         const validator = this.serviceRegistry.getServices(document.uri).validation.DocumentValidator;
-        const validationSetting = this.getBuildOptions(document).validation;
-        const options = typeof validationSetting === 'object' ? validationSetting : undefined;
-        const diagnostics = await validator.validateDocument(document, options, cancelToken);
+        const options = this.getBuildOptions(document);
+        const validationOptions = typeof options.validation === 'object' ? { ...options.validation } : {};
+        validationOptions.categories = this.findMissingValidationCategories(document, options); // execute only not-yet-executed categories
```
spoenemann, line 663: I suspect the changed behavior observed in one of the tests is because of this additional call to `findMissingValidationCategories`. Why is it necessary?

The problem is that we don't know _all_ validation categories. We could fix it like this:
- Rename `ValidationCategory.all` to `ValidationCategory.defaults` (keep `all` and mark deprecated)
- Add a public property `allValidationCategories` to `DefaultDocumentBuilder` which is initialized with `ValidationCategory.defaults` – adopters (and our tests) can change this if they define additional categories
- Use the `allValidationCategories` property in `findMissingValidationCategories`

packages/langium/src/workspace/documents.ts
JohannesMeierSE: Very nice, thanks a lot!

packages/langium/src/workspace/document-builder.ts
JohannesMeierSE: done

packages/langium/src/workspace/document-builder.ts
JohannesMeierSE: I agree. Additionally, the state of the document is set to `Changed`, since we have no `Deleted`. Therefore the deletion is the document is not persisted and cannot be applied during a following call of `update(...)`.

packages/langium/test/validation/document-validator.test.ts
JohannesMeierSE: You are right. I made custom categories more explicit in the API documentation.

packages/langium/src/workspace/document-builder.ts
JohannesMeierSE: Yes, that is correct. I like your approach and realized it:
- I am not sure about the new name `defaults` (instead of `all`), but I found no better name
- Instead of a public property `allValidationCategories`, I introduced a method `protected getAllValidationCategories(LangiumDocument)` to enable different validation categories for different documents (that might be necessary in multi-language projects).

JohannesMeierSE:
@spoenemann @msujew Thanks for your reviews, I just pushed the corresponding changes. This PR is ready for another review.

Last small issue to discuss: I would like to call `resetToState` even in line 198 to be consistent, since we reset the state of the document. But that would loose the build state and the diagnostics of the already executed validation categories. I found no great solution for this (small) issue yet. Maybe you have an idea?

packages/langium/src/workspace/document-builder.ts
```diff
@@ -254,15 +250,35 @@ export class DefaultDocumentBuilder implements DocumentBuilder {
...
+    protected findMissingValidationCategories(document: LangiumDocument, options: BuildOptions | undefined): ValidationCategory[] {
+        const state = this.buildState.get(document.uri.toString());
+        const allCategories = this.getAllValidationCategories(document);
+        const executedCategories = new Set(state?.result?.validationChecks ?? (state?.completed ? allCategories : []));
+        const requestedCategories = (options === undefined || options.validation === true) ? allCategories
+            : typeof options.validation === 'object' ? (options.validation.categories ?? allCategories) : [];
+        return requestedCategories.filter(requested => !executedCategories.has(requested));
+    }
+
+    protected getAllValidationCategories(_document: LangiumDocument): readonly ValidationCategory[] {
```
spoenemann: Now that this is a service method, IMO it would make sense to move this to the `DocumentValidator` service interface.

JohannesMeierSE: @spoenemann  I like the idea of including `getAllValidationCategories` into the service interface, but IMO `ValidationRegistry` is slightly more appropriate than `DocumentValidator`: As a user of Langium, you work with the `ValidationRegistry` to register checks, including checks for custom validation category. Therefore it is easier to touch only one instead of two services.

Detail: I am not sure, whether the input parameter `document: LangiumDocument` makes sense here anymore, since the `ValidationRegistry` is already language-dependent.

Additionally, I found a nice way to collect custom categories in the `ValidationRegistry`, i.e. there is no need to customize a default implementation anymore in order to support new validation categories.

Issue #2074, opened 2025-12-04
"The assigned type is not compatible with the declared property" with enum-like type


ebousse:
Langium version: 4.1.0
Package name: ?

## Steps To Reproduce

1. Write the grammar below
2. Run `npm run langium:generate`

Code example:

```langium
grammar HelloWorld

interface Root {
    color: Color
}

type Color = 'red' | 'blue' | 'green';


entry RootRule returns Root:
    'color' color=ColorRule
;

ColorRule returns Color:
    'red' | 'blue' | 'green'
;

hidden terminal WS: /\s+/;
hidden terminal ML_COMMENT: /\/\*[\s\S]*?\*\//;
hidden terminal SL_COMMENT: /\/\/[^\n\r]*/;
```

## The current behavior

```
src/hello-world.langium:11:13 - The assigned type 'ColorRule' is not compatible with the declared property 'color' of type 'Color'.
```

Screenshot:
<img width="884" height="314" alt="image" src="https://github.com/user-attachments/assets/843e42ca-d191-418c-be63-cbc205a0fa61" />

## The expected behavior

The grammar should work without type error.

msujew:
Thanks for the report, see https://github.com/eclipse-langium/langium/pull/2076 for a fix.

PR #2076, opened 2025-12-04
Fix union type assignments

msujew:
Closes https://github.com/eclipse-langium/langium/issues/2074

I've noticed we're processing the `isValueType(to)` type to early, before completely "unwrapping" the `from` type. This eventually leads to some code which tries to assert that a union type can be assigned to a single value, which will quickly result in failure.

This change adjusts the order to move the unwrapping of the `to` type *after* the unwrapping of the `from` type.

### DIFF
--- a/packages/langium/test/grammar/type-system/type-validator.test.ts
+++ b/packages/langium/test/grammar/type-system/type-validator.test.ts
@@ -319,6 +319,22 @@ describe('Validate declared types', () => {
         `);
         expectNoIssues(validationResult);
     });
+
+    test('Can assign a union type value returned by a differently named rule', async () => {
+        const validationResult = await validate(`
+            interface ColorContainer {
+                color: Color
+            }
+            type Color = 'red' | 'blue' | 'green';
+            ColorContainer returns ColorContainer:
+                'color' color=ColorValue
+            ;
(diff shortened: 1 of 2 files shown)


packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -319,6 +319,22 @@ describe('Validate declared types', () => {
         `);
         expectNoIssues(validationResult);
     });
+
+    test('Can assign a union type to a differently named union type', async () => {
```
sailingKieler: Does this test title reflect the content of the test?
```suggestion
    test('Can assign a union type value returned by a differently named rule', async () => {
```

sailingKieler (approved):
Thank you very much @msujew for taking care!
I cannot assess the fix as I'm not too deep in the type system, the test looks pretty good to me, the title could be a bit more precise. 

One (two) more question(s) though:
in #2074 this error msg is mentioned that IMO has weaknesses, too:
```
src/hello-world.langium:11:13 - The assigned type 'ColorRule' is not compatible with the declared property 'color' of type 'Color'.
```
It seems to also treat the rule name as the type name.
Is that fixed with this change, too?
Could we have a corresponding validation test?

Issue #2075, opened 2025-12-04
Infix rules with explicit return type do not use that type in the AstNode


nk-coding:
Langium version: 4.1.2
Package name: langium

## Steps To Reproduce

1. Have a infix rule which with an explicit return type with a name different from the rule name.
2. The `$type` property of the AstNode is the name of the rule, not the name of the return type.

Grammar example:

```
grammar HelloWorld

entry Model:
    expression=OperatorExpression;

interface BaseExpression {}

interface IdExpression extends BaseExpression {
    id: string;
}

interface ComplexExpression extends BaseExpression {
    operator: string;
    left: BaseExpression;
    right: BaseExpression;
}

IdExpression returns IdExpression:
    id=ID;

infix OperatorExpression on IdExpression returns ComplexExpression:
    '%'
    > '^'
    > '*' | '/'
    > '+' | '-';

BOOLEAN returns boolean:
    'true' | 'false';

hidden terminal WS: /\s+/;
terminal ID: /[_a-zA-Z][\w_]*/;

hidden terminal ML_COMMENT: /\/\*[\s\S]*?\*\//;
hidden terminal SL_COMMENT: /\/\/[^\n\r]*/;
```

Code to parse:
```
a + b
```

## The current behavior
The AstNode for the `+` expression has `$type: "OperatorExpression"`

## The expected behavior
The AstNode for the `+` expression has `$type: "ComplexExpression"`

## Suspected cause

`computeRuleType` in [langium-parser.ts](https://github.com/eclipse-langium/langium/blob/a1f48f515f906a350c4fefe207043e9fe26f88f8/packages/langium/src/parser/langium-parser.ts#L262) directly returns `rule.name` for the infix case, ignoring the explicit return type.

### Possible solution

Use the same logic in the infix case as in the else case at the end.
Note: the type is still used for the [operatorPrecedence lookup](https://github.com/eclipse-langium/langium/blob/a1f48f515f906a350c4fefe207043e9fe26f88f8/packages/langium/src/parser/langium-parser.ts#L460), there, the rule name should still be used (which must thus be passed through in addition to the type).

PR #2077, opened 2025-12-04
Fix return type of infix rules

msujew:
Closes https://github.com/eclipse-langium/langium/issues/2075

Simply uses the `getTypeName` function to get the expected name of the AST node.

### DIFF
--- a/packages/langium/test/parser/langium-parser.test.ts
+++ b/packages/langium/test/parser/langium-parser.test.ts
@@ -204,7 +204,7 @@ describe('Infix operator parsing', async () => {
         expect((binExpr.right as PrimaryExpr).value).toBe('2');
     });
 
-    test('Should work with different infix rules that return the same type', async () => {
+    test('Should work with different infix rules that return the same type #1', async () => {
         const grammar = `
             grammar Test
             entry Model: expr=Expr;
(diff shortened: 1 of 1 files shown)


packages/langium/src/parser/langium-parser.ts
```diff
@@ -244,7 +244,7 @@ export class LangiumParser extends AbstractLangiumParser {
     }
 
     private registerPrecedenceMap(rule: InfixRule): void {
-        const name = rule.name;
+        const name = getTypeName(rule);
```
nk-coding: does this work? in theory, you could have different rules with different precedences which however return the same type

sth like

```
infix OperatorExpression on IdExpression returns ComplexExpression:
    '*' | '/'
    > '+' | '-';

infix OperatorExpression2 on IdExpression returns ComplexExpression:
    '%'
    > '^';
```

msujew: I guess that makes sense, thanks 👍

packages/langium/test/parser/langium-parser.test.ts
```diff
@@ -204,6 +204,46 @@ describe('Infix operator parsing', async () => {
...
+        const node = (document.parseResult.value as ExprModel).expr;
+        expect(node.$type).toBe('BinaryExpression');
+        const parser = services.parser.LangiumParser;
+        // Directly accessing the operatorPrecedence map for testing purposes
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const precedence = (parser as any).operatorPrecedence as Map<string, unknown>;
+        expect(precedence).toBeDefined();
+        expect(precedence.size).toBe(2);
+        expect(precedence.get('BinaryExpr2')).toBeDefined();
+        expect(precedence.get('BinaryExpr')).toBeDefined();
```
sailingKieler, line 244: Nice test!

Would it make sense to add one more taking up the pattern mentioned by @nk-coding like
```
  grammar Test
  entry Model: elements+=(A | B)*;
  A:
     'A' expr = BinaryExpr;
  B:
     'B' expr = BinaryExpr2

  infix BinaryExpr on PrimaryExpr returns BinaryExpression:
      '*' | '/';
  
  infix BinaryExpr2 on PrimaryExpr returns BinaryExpression:
      '+' | '-';
      
  ...
```
WDYT?
Is the completion parser able to properly deal with this?

msujew: Makes sense, I've added another test for this.

> Is the completion parser able to properly deal with this?

I think so, though we currently don't have tests for this.

PR #2083, opened 2025-12-16
move initial document loading into an overrideable method

cdietrich:
got an report that initialize workspace on a workspace with many files crashes on linux systems with ulimit set.
this change allows customization in such a case.

### DIFF
--- a/packages/langium/src/workspace/workspace-manager.ts
+++ b/packages/langium/src/workspace/workspace-manager.ts
@@ -163,12 +163,12 @@ export class DefaultWorkspaceManager implements WorkspaceManager {
             .distinct(uri => uri.toString())
             // Also ensure that the documents don't already exist
             .filter(uri => !this.langiumDocuments.hasDocument(uri));
-        await this.loadDocuments(uniqueUris, collector);
+        await this.loadWorkspaceDocuments(uniqueUris, collector);
         this._ready.resolve();
         return documents;
     }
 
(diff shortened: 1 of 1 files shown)


packages/langium/src/workspace/workspace-manager.ts
```diff
@@ -164,12 +163,16 @@ export class DefaultWorkspaceManager implements WorkspaceManager {
             .distinct(uri => uri.toString())
             // Also ensure that the documents don't already exist
             .filter(uri => !this.langiumDocuments.hasDocument(uri));
-        await Promise.all(uniqueUris.map(async uri => {
+        await this.loadDocuments(uniqueUris, collector);
+        this._ready.resolve();
+        return documents;
+    }
+
+    protected async loadDocuments(uris: Stream<URI>, collector: (document: LangiumDocument) => void): Promise<void> {
```
spoenemann: Suggestion: name this `loadWorkspaceDocuments` to differentiate better from the existing `loadAdditionalDocuments`

cdietrich: done

spoenemann (approved):
Thanks!

msujew (approved):
I'm fine with the changes as they are - but I remember we talked about a batching mechanism. Do we want to implement that as a default in here?

cdietrich:
@msujew need to do this at a lter point in time due to time restrictions

PR #2086, opened 2025-12-21
fix: follow element calculation

cdietrich:


packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -222,7 +222,9 @@ export class DefaultCompletionProvider implements CompletionProvider {
             }
         }
         const leftoverTokens = [...tokens].splice(parserResult.tokenIndex);
-        const features = findNextFeatures([parserResult.elementStack.map(feature => ({ feature }))], leftoverTokens);
+        // Only pass tokens that are after the cursor offset
+        const tokensAfterCursor = leftoverTokens.filter(token => token.startOffset >= offset);
```
cdietrich: @msujew do you remember why the current impl uses all leftover tokens?

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -303,7 +305,7 @@ export class DefaultCompletionProvider implements CompletionProvider {
                 ...partialContext,
                 tokenOffset: nextTokenStart,
                 tokenEndOffset: nextTokenEnd,
-                features: findFirstFeatures(parserRule.definition)
+                features: this.findFeaturesAt(textDocument, nextTokenStart)
```
cdietrich, line 310: cline told me this one is wrong too

cdietrich:
pushed another fix attempt

Issue #1955, opened 2025-06-14
Optional rule content breaks completion


AresEkb:
Langium version: 3.5.0
Package name: langium

## Steps To Reproduce

Use the following grammar:

```
grammar Test

entry Document:
    Annotation
    'document' name=ID;

fragment Annotation:
    ('@annotation' annotation=STRING)?;

terminal ID: /[_a-zA-Z][\w_]*/;
terminal STRING: /'(\\.|[^'])*'/; 

hidden terminal WS: /\s+/;
```

Run completion for an empty document

## The current behavior

It will show only one proposal: `@annotation`

## The expected behavior

But two proposals are expected: `@annotation` and `document`

## Additional information

If you replace fragment by the embedded rule, then completion works as expected:

```
grammar Test

entry Document:
    ('@annotation' annotation=STRING)?
    'document' name=ID;

terminal ID: /[_a-zA-Z][\w_]*/;
terminal STRING: /'(\\.|[^'])*'/; 

hidden terminal WS: /\s+/;
```

msujew:
It doesn't seem to be related to the fragment rule specifically, this grammar exhibits the same behavior:

```
grammar Test

entry Document:
    annotation=Annotation
    'document' name=ID;

Annotation:
    ('@annotation' annotation=STRING)?;
```

However, this works correctly:


```
grammar Test

entry Document:
    annotation=Annotation? // <-- moving the annotation here
    'document' name=ID;

Annotation:
    ('@annotation' annotation=STRING);
```

Therefore this also seems to work as a workaround for this issue (moving the optionality to the caller side of the fragment rule):

```
grammar Test

entry Document:
    Annotation?
    'document' name=ID;

fragment Annotation:
    '@annotation' annotation=STRING;

terminal ID: /[_a-zA-Z][\w_]*/;
terminal STRING: /'(\\.|[^'])*'/; 

hidden terminal WS: /\s+/;
```

AresEkb:
Thanks for the workaround! It works for me

Issue #1798, opened 2025-01-28
Completion invocation not always calling `completionFor`


nametable:
When completion is invoked, the relevant `completionFor` function is not always being called. Something appears to be wrong with the next features computation or the CompletionParser. It is also possible that I have a misunderstanding of how completions are supposed to work. See some screenshots below, relating to the reproduction steps:

Completions working
![Image](https://github.com/user-attachments/assets/263ef866-6778-45c0-833b-6228f545a3ab)

Completions not working
![Image](https://github.com/user-attachments/assets/74df5d7b-d1e6-4c11-867b-a56fae1604b7)

Langium version: `main`
Package name: `langium`

## Steps To Reproduce

1. Checkout `completion-issue-repro` from my fork of Langium - https://github.com/nametable/langium/tree/completion-issue-repro
2. Build the DomainModel example
3. Run the extension for DomainModel example
4. Open `blog.dmodel`
5. Trigger completion inside of string literals within `[Attribute2("")]` with Ctrl+Space
6. See that completion works for the top `Attribute2` but not the bottom `Attribute2`.

Link to code example:

https://github.com/nametable/langium/tree/completion-issue-repro

## The current behavior
The completion provider is not invoked for the lower `Attribute2` annotation, while it is for the upper.

## The expected behavior
Both `Attribute2` string literals should have working completion.

msujew:
Looks like something is going wrong in the token interpreter that we use to identify the current position. See [here](https://github.com/eclipse-langium/langium/blob/23d4a1441cb8b0f4b9cb0fc628ceac3c2c881b05/packages/langium/src/lsp/completion/follow-element-computation.ts#L223-L258). I'll take a look at this soon.

PR #2087, opened 2025-12-27
Refactor follow element computation

msujew:
Closes https://github.com/eclipse-langium/langium/pull/2086
Fixes https://github.com/eclipse-langium/langium/issues/1798
Fixes https://github.com/eclipse-langium/langium/issues/1955

It took me a bit to make sense of what I wrote >4 years ago - turns out, our original completion logic was pretty over-complicated and didn't make a lot of sense (explains the difficulty of understanding it in the first place).

This PR refactors the follow element computation pretty thouroughly, which should improve the completion quite massively in a lot of cases. It takes inspiration from the lookahead algorithm for LL(1) grammars. It essentially performs something that is analogous to the NFA->DFA transformation but for LL grammars.

The PR also adds a bunch of tests to ensure that we don't regress for these improvements.

### DIFF
--- a/packages/langium/src/lsp/completion/follow-element-computation.ts
+++ b/packages/langium/src/lsp/completion/follow-element-computation.ts
@@ -6,10 +6,8 @@
 
 import type { IToken } from 'chevrotain';
 import * as ast from '../../languages/generated/ast.js';
-import { isAstNode } from '../../syntax-tree.js';
-import { getContainerOfType } from '../../utils/ast-utils.js';
-import type { Cardinality } from '../../utils/grammar-utils.js';
 import { getCrossReferenceTerminal, getExplicitRuleType, getTypeName, isArrayCardinality, isOptionalCardinality, terminalRegex } from '../../utils/grammar-utils.js';
+import { getContainerOfType } from '../../utils/ast-utils.js';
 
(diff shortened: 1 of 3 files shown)


cdietrich:
arggg i am blind 🙈

packages/langium/src/lsp/completion/follow-element-computation.ts
```diff
@@ -46,187 +44,211 @@ export function findNextFeatures(featureStack: NextFeature[][], unparsedTokens:
...
+/**
+ * Determines recursively whether an element is optional. It is not sufficient to check only the cardinality of the element itself,
+ * because it could a group or alternative that contains only optional elements.
+ */
+function isElementOptional(feature: ast.AbstractElement, visited: Map<ast.AbstractElement, boolean>): boolean {
+    const visitedResult = visited.get(feature);
+    if (visitedResult !== undefined) {
+        return visitedResult;
+    }
+    visited.set(feature, false);
```
cdietrich, line 222: should we move this down to return for consistency

msujew: I don't think we can. We need to set this immediately, in case the recursion runs into the same element again (for example, via a rule call).

cdietrich: i see

PR #2089, opened 2026-01-05
Update Chevrotain to 11.1.0

msujew:
Updates Chevrotain to the latest version. The 11.1.0 contains updates to the lexer which reduces the need for some of the workarounds that we've added to our token builder.

Note that the new feature https://github.com/Chevrotain/chevrotain/pull/2138 isn't in there yet.


spoenemann (approved):
👍

PR #2091, opened 2026-01-05
langium/generate: fixed incorrect indentation replication for indented multiline substitutions

### LINKED ISSUE #2081: Indentation using Tabs in Template Expansion
Langium version: 4.1.0
Package name: CLI / Generator

## Steps To Reproduce

1. Run the following (indentation are tabs):
   ```
   console.log(toString(
   	expandToNode`
   		arst (
   			${joinToNode(["a", "b"], { separator: ",", appendNewLineIfNotEmpty: true })}
   		) wfp
   	`)
   )
   ```

## The current behavior

Output is (first indent is a tab, second a single space)
```
arst (
	a,
 b
) wfp
```

## The expected behavior

It should be 
```
arst (
	a,
	b
) wfp
```

##  Notes

Looking at [the code](https://github.com/eclipse-langium/langium/blob/a1f48f515f906a350c4fefe207043e9fe26f88f8/packages/langium/src/generate/template-node.ts#L529), I'm not sure why there's

```typescript
const indent: string = (i === 0 || isNewLineMarker(splitAndMerged[i - 1])) && typeof segment === 'string' && segment.length !== 0 ? ''.padStart(segment.length - segment.trimStart().length) : '';
```

instead of simply
```typescript
const indent: string = (i === 0 || isNewLineMarker(splitAndMerged[i - 1])) && typeof segment === 'string' && segment.length !== 0 ? segment) : '';
```

but of course I've only looked at this very locally and I'm not sure if `segment` could be anything other than whitespace in some cases - of course then it would not work.

sailingKieler:
Fixes #2081

### DIFF
--- a/packages/langium/src/generate/template-node.ts
+++ b/packages/langium/src/generate/template-node.ts
@@ -526,7 +526,7 @@ function composeFinalGeneratorNode(splitAndMerged: GeneratedOrMarker[]): Composi
                     //   }
                     // assuming that ${foo(bar)} yields a multiline result;
                     // the whitespace between 'return' and '${foo(bar)}' shall not add to the indentation of '${foo(bar)}'s result!
-                    const indent: string = (i === 0 || isNewLineMarker(splitAndMerged[i - 1])) && typeof segment === 'string' && segment.length !== 0 ? ''.padStart(segment.length - segment.trimStart().length) : '';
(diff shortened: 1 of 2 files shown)


msujew (approved):
LGTM, thanks 👍

Issue #2093, opened 2026-01-05
Publish package to NPM with provenance


RaederDev:
Some old versions of the package for example 3.0.0 have been published with provenance https://www.npmjs.com/package/langium/v/3.0.0#provenance

Tools like PNPM now provide settings to configure a trust policy to detect malicious package takeovers to prevent supply chain attacks. If this is set to "no-downgrade" which is a sensible thing to do this package cannot be installed since versions after 3.0.0 have been published without provenance. (Not a massive deal since you can whitelist packages but annoying).

I have no idea how much effort enabling this for future releases would be but it would be nice to have to prevent this from happening.

PR #2094, opened 2026-01-07
Automate package release process

msujew:
Related to https://github.com/eclipse-langium/langium/issues/2093.

Automatically publishes a package as soon as it is updated on the `main` or `maintenance/*` branches. This new flow can release packages individually, removing the need to publish them all at once.

This also automatically publishes the extensions to the vscode marketplace and open vsx.

You can test this by running the new `publish:updated` script with the `-- --dry` argument (`npm run publish:updated -- --dry`) after changing the `version` of the respective package.json file.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -47,5 +47,4 @@ jobs:
         GH_TOKEN: ${{ github.token }}
         RUN_ID: ${{ github.run_id }}
       run: |
-        npm install -g @vscode/vsce ovsx
         npm run publish:updated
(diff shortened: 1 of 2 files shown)


spoenemann (commented):
Looks good, but `vsce` and `ovsx` should be added to the devDependencies of `langium-vscode`. The script fails if those are not installed.

.github/workflows/publish.yml
```diff
@@ -32,27 +39,13 @@ jobs:
...
+        VSCE_TOKEN: ${{ secrets.VSCE_TOKEN }}
+        GH_TOKEN: ${{ github.token }}
+        RUN_ID: ${{ github.run_id }}
       run: |
         npm install -g @vscode/vsce ovsx
-        cd packages/langium-vscode
-        PACKAGE_VERSION=`npm pkg get version --workspaces=false | tr -d \"`
-        vsce package
-        ovsx publish langium-vscode-$PACKAGE_VERSION.vsix -p ${{ secrets.OVSX_TOKEN }}
+        npm run publish:updated
```
cdietrich, line 50: do we still need this? should we use the same versions?

cdietrich: ( i mean the install above)

msujew: Good point, thanks 👍

PR #2102, opened 2026-01-16
Fixes #2101: code completion should handle case insensitive languages

### LINKED ISSUE #2101: Case insensitivity not respected in code completion
When a language is case insensitive (I have set `"caseInsensitive": true` in langium-config.json), code completion does not work 

Langium version: v4.0.0
Package name: langium

## Steps To Reproduce

1. Go to https://langium.org/showcase/sql/
2. Type following
```
CREATE FUNCTION SUM(column REAL) AS REAL;

CREATE t
```
or just following
```
create t
```
3. You will see TABLE keyword being suggested.
4. Now write following
```
CREATE FUNCTION SUM(column REAL) AS REAL;

create t
```
5. You will see nothing being suggested.

Following works
Image
Image
 
this does not
Image

## The current behavior
No option being suggested

## The expected behavior
Since it is a case insensitive language TABLE keyword should be suggested

dhrubomoy:
Fixes https://github.com/eclipse-langium/langium/issues/2101: Code completion should handle case insensitive languages. When trying to match a keyword we can try to compare the content with the `tokenType.name` instead of token.image.

### DIFF
--- a/packages/langium/test/lsp/completion-provider.test.ts
+++ b/packages/langium/test/lsp/completion-provider.test.ts
@@ -483,7 +483,7 @@ describe('Common prefixes', async () => {
 
         const services = await createServicesForGrammar({ grammar, languageMetaData });
         const completion = expectCompletion(services);
-        const text = 'A b D <|>A <|>B <|>C';
+        const text = 'A b d <|>A <|>b <|>C';
         await completion({
             text,
             index: 0,


packages/langium/src/lsp/completion/follow-element-computation.ts
```diff
@@ -317,7 +317,7 @@ function getAbstractElementParent(element: ast.AbstractElement): ast.AbstractEle
 function featureMatches(feature: ast.AbstractElement, token: IToken): boolean {
     if (ast.isKeyword(feature)) {
         const content = feature.value;
-        return content === token.image;
+        return content === token.tokenType.name;
```
msujew, line 320: I don't think this actually fixes the case insensitive behavior, since this comparison will always be case sensitive. An actual fix would probably do something like taking the [`caseInsensitive`](https://github.com/eclipse-langium/langium/blob/37aab5513f1b473c4dc4293b2d90051b9ef816c7/packages/langium/src/languages/language-meta-data.ts#L14) flag into account.

packages/langium/test/lsp/completion-provider.test.ts
```diff
@@ -461,6 +461,46 @@ describe('Common prefixes', async () => {
...
+        const languageMetaData = {
+            caseInsensitive: true,
+            fileExtensions: ['.txt'],
+            languageId: 'UNKNOWN',
+            mode: 'development' as const
+        };
+
+        const services = await createServicesForGrammar({ grammar, languageMetaData });
+        const completion = expectCompletion(services);
+        const text = 'A b D <|>A <|>B <|>C';
```
msujew: To properly test this, we should probably use a mix of upper and lower case keywords here. This is only using upper case keywords.

dhrubomoy: Done. Used mix of upper and lower case

packages/langium/src/lsp/completion/follow-element-computation.ts
dhrubomoy: `tokenType.name` appears to handle case-insensitive scenarios since its lowercase. `content` string is also lowercased.

Another approach I was thinking initially, I noticed that for keywords, there's a `tokenType.PATTERN` property we could leverage. When `caseInsensitive` is true, the `PATTERN` regex appears to always be `/keyword/i`. We could implement something like:

```js
if (token.tokenType.PATTERN instanceof RegExp) {
    return token.tokenType.PATTERN.test(content);
}
return content === token.image;
```

Alternatively, we could pass the `caseInsensitive` flag down to this function (current call stack depth is 9). Which approach would work better here?

msujew: Yeah, I would prefer the `PATTERN` approach. Using `tokenType.name` is rather unintuitive and prone to breakage (we don't want to break code completion just because we change the capitalization of keyword token names).

msujew: Nevermind, thinking through this again, I think I'm fine with this. Thanks for the discussion.

dhrubomoy: Sounds good. Thanks. Also signed the ECA :)

msujew (approved):
Thanks, looks good to me 👍

PR #2120, opened 2026-02-23
Don't crash the language server by providing invalid URIs

JohannesMeierSE:
This PRs adds error handling in the LS for reading files `fromUri`:
- The existing implementation crashed the LS when providing an invalid URI or an URI for a non-existing file to read.
- Since the API always returns a `LangiumDocument`, a new document is created and the resulting error is put into it. Feel free to suggest a better strategy for error handling!

### DIFF
--- a/packages/langium/src/workspace/documents.ts
+++ b/packages/langium/src/workspace/documents.ts
@@ -155,6 +155,9 @@ export interface LangiumDocumentFactory {
 
     /**
      * Create an Langium document from a specified `URI`. The factory will use the `FileSystemAccess` service to read the file.
+     *
+     * If reading the file failed for some reasons (e.g. the URI is invalid or the file does not exist),
+     * a `LangiumDocument` is returned with a corresponding error message as content.
      */
     fromUri<T extends AstNode = AstNode>(uri: URI, cancellationToken?: CancellationToken): Promise<LangiumDocument<T>>;
 
@@ -181,7 +184,22 @@ export class DefaultLangiumDocumentFactory implements LangiumDocumentFactory {
     }
 
     async fromUri<T extends AstNode = AstNode>(uri: URI, cancellationToken = CancellationToken.None): Promise<LangiumDocument<T>> {
-        const content = await this.fileSystemProvider.readFile(uri);
+        const exists = await this.fileSystemProvider.exists(uri);
+        let content: string;
+        if (exists) {
+            const stats = await this.fileSystemProvider.stat(uri);
+            if (stats.isFile) {
+                try {
+                    content = await this.fileSystemProvider.readFile(uri);
+                } catch (err) {
+                    content = `// reading '${uri.toString()}' from file system failed: ${String(err)}`;
+                }
+            } else {
+                content = `// '${uri.toString()}' exists in the file system, but is no file`;
+            }
+        } else {
+            content = `// '${uri.toString()}' does not exist in the file system`;
+        }
         return this.createAsync<T>(uri, content, cancellationToken);
     }
 
--- a/packages/langium/src/workspace/documents.ts
+++ b/packages/langium/src/workspace/documents.ts
@@ -184,21 +184,21 @@ export class DefaultLangiumDocumentFactory implements LangiumDocumentFactory {
     }
 
     async fromUri<T extends AstNode = AstNode>(uri: URI, cancellationToken = CancellationToken.None): Promise<LangiumDocument<T>> {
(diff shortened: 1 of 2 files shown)


packages/langium/src/workspace/documents.ts
```diff
@@ -181,7 +184,22 @@ export class DefaultLangiumDocumentFactory implements LangiumDocumentFactory {
     }
 
     async fromUri<T extends AstNode = AstNode>(uri: URI, cancellationToken = CancellationToken.None): Promise<LangiumDocument<T>> {
-        const content = await this.fileSystemProvider.readFile(uri);
+        const exists = await this.fileSystemProvider.exists(uri);
+        let content: string;
+        if (exists) {
+            const stats = await this.fileSystemProvider.stat(uri);
+            if (stats.isFile) {
```
sailingKieler: I suggest to defer these fine diagnostics to the actual error cases, i.e., just wrap the original call of `readFile()` with some try-catch-block as below and move the calls of `exists()` and `stat()` into the catch block.
That way the (common & expected) good cases are not charged with additional efforts.

packages/langium/test/workspace/document-factory.test.ts
```diff
@@ -45,6 +46,40 @@ describe('DefaultLangiumDocumentFactory', () => {
...
+        const uri = URI.parse('file:///file/does/not/exist.langium');
+        const document = await documentFactory.fromUri(uri);
+        expect(document).not.toBe(undefined);
+        expect(document.textDocument.getText().trim()).toBe("// 'file:///file/does/not/exist.langium' does not exist in the file system");
+    });
+
+    test('Handle invalid URIs in the NodeFileSystem', async () => {
+        const services = createLangiumGrammarServices(NodeFileSystem).grammar;
+        const documentFactory = services.shared.workspace.LangiumDocumentFactory;
+        const uri = URI.parse('file:///file/does/not/exist.langium');
```
sailingKieler: I suggest to further obfuscate the dummy file name (maybe with unsupported characters?) as your path might exist unexpectedly, or other issues might pop up, like missing `r` or `x` permissions of the folder segments, and the like

packages/langium/test/workspace/document-factory.test.ts
```diff
@@ -45,6 +46,40 @@ describe('DefaultLangiumDocumentFactory', () => {
...
+            protected override traverseFolder(folderPath: URI, uris: URI[]): Promise<void> {
+                uris.push(URI.parse('file:///file/does/not/exist.langium'));
+                return super.traverseFolder(folderPath, uris);
+            }
+        }
+        const services = createLangiumGrammarServices(NodeFileSystem, { workspace: { WorkspaceManager: services => new TestWorkspaceManager(services) }}).grammar;
+        await services.shared.workspace.WorkspaceManager.initializeWorkspace([{ name: 'init', uri: 'start-somewhere' }]);
+        const documents = services.shared.workspace.LangiumDocuments.all.toArray();
+        expect(documents).toHaveLength(1);
+        expect(documents[0].textDocument.getText().trim()).toBe("// 'file:///file/does/not/exist.langium' does not exist in the file system");
```
sailingKieler: I suggest to swap these lines, the failure message is much more meaningful then.
You can make `documents[0].textDocument.getText().trim()` fail-safe by introducing a `?`: `documents[0]?.textDocument.getText().trim()`

sailingKieler (commented):
Thanks for caring @JohannesMeierSE.
Find some suggestions below.

packages/langium/src/workspace/documents.ts
JohannesMeierSE: Yes, that saves performances! I changed the implementation accordingly.

packages/langium/test/workspace/document-factory.test.ts
JohannesMeierSE: Added `*` in the file name, since it is not supported on Windows

packages/langium/test/workspace/document-factory.test.ts
JohannesMeierSE: I swapped the lines

msujew:
IMO, this change should ensure that call sites of `fromUri` can handle the case in which the file is missing/unreadable. Otherwise it becomes impossible to tell (from the outside) whether the document was successfully created.

I'd recommend: Continue throwing on invalid URIs/missing files and use try-catch correctly wherever `fromUri` is used.

msujew:
I would close this for now. I don't think we will go forward with this solution.

PR #2125, opened 2026-03-05
chore: another tar sec fix

cdietrich:


montymxb (approved):
Chore is good to go 🧹

PR #2128, opened 2026-03-10
chore: fix for simple-git + tar + minor bumps

cdietrich:


montymxb (approved):
Chore is good to go 🧹

PR #2129, opened 2026-03-12
chore: bump chevrotain to 11.2

cdietrich:
am not sure if anything needs to be changed in allstart with respect to LA_FAST @msujew 

in our measurements we see a 2-5% perf increase

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -443,42 +443,42 @@
       }
     },
     "node_modules/@chevrotain/cst-dts-gen": {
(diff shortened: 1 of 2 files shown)


msujew (approved):
We can merge this for now. I'll take a separate look at this in the `chevrotain-allstar` package wrt. `LA_FAST` - I believe we should be able to use it in there.

cdietrich:
thx @msujew

PR #2137, opened 2026-04-15
chore: bump action versions (to node 24 compatible ones)

cdietrich:


.github/workflows/generate-sbom.yml
```diff
@@ -35,13 +35,13 @@ jobs:
           echo "Product version: $VERSION"
 
       - name: Checkout repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
cdietrich: should we use pinned versions everywhere?

cdietrich:
see also https://github.com/eclipse-csi/workflows/issues/1

msujew (approved):
I don't think pinning the official actions makes a difference. It's mostly about third-party actions, where pinning can prevent malicious attacks.

PR #2154, opened 2026-05-11
Add missing langium package dependencies

### LINKED ISSUE #2153: Missing 'vscode-jsonrpc' in dependencies causes ESM resolution failure under strict layouts (pnpm GVS, etc.)
### Problem

`packages/langium/src/utils/event.ts` directly imports from `vscode-jsonrpc`:

```ts
// eslint-disable-next-line no-restricted-imports
export * from 'vscode-jsonrpc/lib/common/events.js';
```

…but `packages/langium/package.json` does not declare `vscode-jsonrpc` in `dependencies`. The package is only resolvable today because pnpm's default flat sibling layout makes it reachable through transitive deps (`vscode-languageserver` → `vscode-languageserver-protocol` → `vscode-jsonrpc`).

Under stricter ESM resolution layouts this breaks. I reproduced it on `pnpm@11` with `enableGlobalVirtualStore: true`, where `langium` lives at `~/Library/pnpm/store/v11/links/@/langium/.../node_modules/langium/` without a sibling `vscode-jsonrpc/` directory. Node itself fails to resolve the file:

```
$ node --input-type=module -e \
    "await import('.../langium/lib/utils/event.js')"

ERR_MODULE_NOT_FOUND: Cannot find package 'vscode-jsonrpc'
imported from .../langium/lib/utils/event.js
```

This also surfaces as a Turbopack production build failure ("The module has no exports at all") for any Next.js app that uses `mermaid` → `@mermaid-js/parser` → `langium` under pnpm GVS.

### Suggested fix

Add `vscode-jsonrpc` (matching the version pulled transitively via `vscode-languageserver@~9.0.1`) to `packages/langium`'s `dependencies`, mirroring PR #2121 which fixed the same kind of latent missing-dependency for `@chevrotain/regexp-to-ast`.

### Affects

- langium 4.2.3 on npm
- current `main` (4.2.0) — `vscode-jsonrpc` is still absent from `packages/langium/package.json` `dependencies`

msujew:
Closes https://github.com/eclipse-langium/langium/issues/2153

Also ensures that `langium-cli` is using the right import - but since it's a `type` import only, it's getting omitted anyway.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7044,9 +7044,9 @@
       "license": "MIT"
     },
     "node_modules/fast-uri": {
-      "version": "3.1.0",
-      "resolved": "https://registry.npmjs.org/fast-uri/-/fast-uri-3.1.0.tgz",
-      "integrity": "sha512-iPeeDKJSWf4IEOasVVrknXpaBV0IApz/gp7S2bb7Z4Lljbl2MGJRqInZiUrQwV16cpzw/D3S5j5Julj/gT52AA==",
+      "version": "3.1.2",
+      "resolved": "https://registry.npmjs.org/fast-uri/-/fast-uri-3.1.2.tgz",
+      "integrity": "sha512-rVjf7ArG3LTk+FS6Yw81V1DLuZl1bRbNrev6Tmd/9RaroeeRRJhAt7jg/6YFxbvAQXUCavSoZhPPj6oOx+5KjQ==",
       "dev": true,
       "funding": [
         {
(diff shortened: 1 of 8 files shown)


cdietrich:
npx -y knip --production --dependencies --workspace packages/langium
Unused dependencies (1)
vscode-languageserver  packages/langium/package.json:72:6

should we also fix generator-langium, langium-sprotty and packages/langium-vscode?

yamaaaaaa31:
Thanks @msujew — that's exactly the fix we landed on in the experiment. The langium-cli cleanup (`path` → `node:path`, dropping the direct `vscode-uri` import in favour of the `langium` re-export) is a nice bonus.

msujew:
> npx -y knip --production --dependencies --workspace packages/langium
Unused dependencies (1)
vscode-languageserver packages/langium/package.json:72:6

I get the same result - but it's wrong. We clearly use the `vscode-languageserver` dependency:

https://github.com/eclipse-langium/langium/blob/b766e919af928c61c164a1388829463b5f91809f/packages/langium/src/lsp/lsp-services.ts#L7

> langium-sprotty

Updated in latest commit 👍 

> generator-langium, langium-vscode

I don't think there's a need to touch those.

cdietrich (approved):

PR #2155, opened 2026-05-18
Add `isComplete` check to Reflection

montymxb:
Candidate feature for adding an `isComplete` checker to our reflection setup, so we can do built-in checks for whether nodes are missing non-optional properties. This is in response to a number of cases where nodes deviate from their declared type interfaces in the generated **ast.ts**, which can lead to some nasty surprises when doing validation checks, generation, etc. Really anything that expects the node to conform to it's type. As far as I'm aware, there's been a lot of manual sanity checking to avoid these kinds of cases _when_ they pop up, so the idea here is to try and make that easier to handle up front.

To support this I added in an `optional` property to the generated property metadata, so we can recover optionality at runtime without too much effort, but then this requires a regeneration of the ast.ts to get the feature working.

Definitely open for suggestions & feedback on this one, as this is the first approach I've come up with.

### DIFF
--- a/packages/langium/src/syntax-tree.ts
+++ b/packages/langium/src/syntax-tree.ts
@@ -266,7 +266,7 @@ export abstract class AbstractAstReflection implements AstReflection {
                 continue;
             }
 
-            const value = (node as unknown as Record<string, unknown>)[prop.name];
+            const value = (node as GenericAstNode)[prop.name];
             if (value === undefined || value === null) {
                 return false;
             }
--- a/packages/langium/src/syntax-tree.ts
+++ b/packages/langium/src/syntax-tree.ts
@@ -172,6 +172,7 @@ export interface AstReflection {
     isInstance(node: unknown, type: string): boolean
     isSubtype(subtype: string, supertype: string): boolean
     getAllSubTypes(type: string): string[]
+    isComplete(node: AstNode): boolean
 }
 
 /**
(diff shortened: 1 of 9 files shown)


packages/langium/src/syntax-tree.ts
```diff
@@ -251,6 +252,27 @@ export abstract class AbstractAstReflection implements AstReflection {
...
+     */
+    isComplete(node: AstNode): boolean {
+        const meta = this.getTypeMetaData(node.$type);
+        for (const key of Object.keys(meta.properties)) {
+            const prop = meta.properties[key];
+            if (prop.optional) {
+                continue;
+            }
+
+            const value = (node as unknown as Record<string, unknown>)[prop.name];
```
msujew: Suggestion: We have a type for this
```suggestion
            const value = (node as GenericAstNode)[prop.name];
```

msujew (commented):
Makes sense 👍

packages/langium/src/syntax-tree.ts
```diff
@@ -251,6 +252,27 @@ export abstract class AbstractAstReflection implements AstReflection {
...
+        for (const key of Object.keys(meta.properties)) {
+            const prop = meta.properties[key];
+            if (prop.optional) {
+                continue;
+            }
+
+            const value = (node as GenericAstNode)[prop.name];
+            if (value === undefined || value === null) {
+                return false;
+            }
```
msujew, line 272: Question: Now that I think about this - Arrays are always initialized as empty arrays, even if no explicit default value is set. But here we only check against `undefined` and `null`. This means that an array that is empty, but non-optional will still pass this check. Is this intended?

montymxb: In this case that's intended, but it's a good point. This is just meant to avoid accessing any prop that might otherwise crash. So the empty array case should be okay, so long as some value is present.

msujew: Alright, I see. Makes sense 👍

msujew (approved):
Looks good to me 👍

PR #2159, opened 2026-05-28
feat: use classes in linker for ref accessors

cdietrich:


packages/langium/src/references/linker.ts
```diff
@@ -91,6 +91,69 @@ export interface DefaultMultiReference extends MultiReference {
     _linkingError?: LinkingError;
 }
 
+function defineReferenceContext(reference: object, linker: DefaultLinker, node: AstNode, property: string): void {
+    Object.defineProperties(reference, {
```
cdietrich, line 95: am not sure if anyone will actually use this

packages/langium/src/references/linker.ts
```diff
@@ -214,100 +277,84 @@ export class DefaultLinker implements Linker {
...
-                            const linkedNode = linker.loadAstNode(description);
-                            if (linkedNode) {
-                                items.push({ ref: linkedNode, $nodeDescription: description });
-                            }
-                        }
+        return new DefaultMultiReferenceImpl(this, node, property, refNode, refText);
+    }
+
+    /** @internal */
+    resolveReference(reference: DefaultReference, node: AstNode, property: string): AstNode | undefined {
```
cdietrich, line 289: am not sure if we want to go this route or use
module-private symbol-keyed methods 
but then subclassing wont work

spoenemann: Maybe move this to protected methods of the new DefaultReferenceImpl classes?

packages/langium/src/references/linker.ts
spoenemann: Why not regular properties defined in the class itself?

cdietrich: the idea way to keek compatibility if someone uses Object.keys etc they would see more stuff if we have the props in the class.
otherwise we wont need it.

packages/langium/src/references/linker.ts
cdietrich: they use other stuff from the linker itself (getLinkedNode, loadAstNode, ...)

cdietrich: maybe we can introduce an interface to call back like

```
interface LinkerReferenceResolver {
  getLinkedNode(refInfo: ReferenceInfo): ...
  loadAstNode(description: AstNodeDescription): AstNode | undefined
  createLinkingError(...): LinkingError
  throwCyclicReferenceError(...): never
  getCandidates(refInfo: ReferenceInfo): ...
}
```
and create an instance like

```
private readonly referenceResolver = {
  getLinkedNode: (refInfo) => this.getLinkedNode(refInfo),
  loadAstNode: (descr) => this.loadAstNode(descr),
  createLinkingError: (...) => this.createLinkingError(...),
  throwCyclicReferenceError: (...) => this.throwCyclicReferenceError(...),
  getCandidates: (refInfo) => this.getCandidates(refInfo)
};
```

and then pass it but am not sure if this is any nicer

cdietrich: or we apply this pattern to 3 new methods

cdietrich: like
```
interface ReferenceResolver {
  resolveReference(reference: DefaultReference, node: AstNode, property: string): AstNode | undefined;
  resolveMultiReference(reference: DefaultMultiReference, node: AstNode, property: string): MultiReferenceItem[];
  getMultiReferenceError(reference: DefaultMultiReference, node: AstNode, property: string): LinkingError | undefined;
}
```

cdietrich: with

export class DefaultLinker implements Linker {
  ......

  protected resolveReference(...) { ... }
  protected resolveMultiReference(...) { ... }
  protected getMultiReferenceError(...) { ... }
}

cdietrich: sketch at https://github.com/eclipse-langium/langium/pull/2161

cdietrich:
closed in favour of #2161 2161

PR #2162, opened 2026-06-05
Fix: Preserve optionality when merging prop types

montymxb:
- [x] Failing test issue resolved

Fixes an small issue where the types we infer from parser rules can sometimes produce properties with incorrect optionality.

Ex. 
```langium
// marks `prop` as required
AB:
    'a' prop=ID | 'b' prop=ID?;

// marks `props` as _optional_
BA:
    'b' prop=ID? | 'a' prop=ID;
```

Effectively the resulting interface type in **ast.ts** would change based on the alternative order above. When we were merging the types of properties together, we were not merging whether that property was optional. So effectively we took the base optionality and that was it.

_See comment below for follow-up, the details below were from a draft state, kept for reference._

With that being said, this was a quick 5 min fix conveniently enough. _But_, there's a catch, with this fix we introduce another bug with a incorrect type for the following case and derivatives of it (i.e. Rules with actions that infer the same type):
```langium
// marks `ref` as _optional_
A: {infer A} ref=ID;
```

It's a side-effect of the fact we produce 2 types internally in the type-collector before they're merged together; one from rule `A` and another from the action inferring `A`. This is totally normal & expected, but the fix here kicks in _before_ we merge these types, creating an optional property where there shouldn't be one.

That one needs more time to dig into, given it's not so easy to change without negatively impacting the stability of the collector at large. I've already dug into it a bit, but I'm still working on it. So I figured I would at least bring the original patch up for the smaller fix first.

My thought here is that the aforementioned issue has likely caused problems due to erroneous types being generated, and should be prioritized. The side-effect of this fix also leads to invalid types, but only ones that will lead to redundant assertions & checks being required. I think, for now at least, this patched version is preferable to the existing one.

I've left an existing test case failing before I get some feedback on this. My initial thought was to patch it upfront w/ the 'unintended' behavior, but I felt like it's better to share the state first & see what everyone thinks.

We can also defer the PR until I patch the additional issue as well (which would be ideal), but I can't provide any guarantees on when I can provide that from a time perspective.

### DIFF
--- a/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
@@ -661,6 +661,8 @@ function flattenTypes(alternatives: TypePath[], part: TypePart): TypePath[] {
                 const existingProperty = properties.find(e => e.name === altProperty.name);
                 if (existingProperty) {
                     existingProperty.type = mergePropertyTypes(existingProperty.type, altProperty.type);
+                    // if either occurrence is optional, the merged property must be optional
+                    existingProperty.optional = existingProperty.optional || altProperty.optional;
                     altProperty.astNodes.forEach(e => existingProperty.astNodes.add(e));
                 } else {
                     properties.push({ ...altProperty });
--- a/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
@@ -433,6 +433,20 @@ function newTypePart(element?: AbstractParserRule | Action | string): TypePart {
 }
 
 /**
+ * Walks up the type path to attempt to find the name of the type we're currently building
+ */
+function nearestNamedParent(part: TypePart): string | undefined {
+    let current: TypePart | undefined = part;
+    while (current) {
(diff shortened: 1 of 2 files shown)


montymxb:
Following up here, finally came back around to correct the original issue noted above.

The fix walks back up the type graph to attempt to resolve the first ancestor with a name, and sanity-check that it isn't redundant. This works quite well for most cases. I also thought about carrying the type name we're currently resolving through the collection process, but that changes the existing signatures a bunch.

_An additional aside_

There are more complex grammar rules that can still be crafted with actions that can bypass this check (which just produces the same incorrect optionality results as originally reported). But this appears to be related to another separate issue. There's a loss of information occurring with some props being dropped entirely in the `ast.ts` for rules like so:

```langium
A: ('b' {infer B} cantDropThis=ID | 'a' propA=ID) {infer A} val=ID;
```

This is pretty contrived, but there's no error or diagnostics, and `cantDropThis` is lost in the as.ts entirely, resulting in `B` turning into a type rather than an interface. `val` is also inferred as optional for `A` in this case, but I'll keep digging to see whether these are related or not in the collection logic.

montymxb:
@msujew bump, when you have a moment to check this one

PR #2164, opened 2026-06-08
Replace publish ts script by github action

gfontorbe:
Attempt to replace the publish TS script by a GH action, moving the logic from the TS file to bash scripting in the yaml file.

The idea is similar: on push to main (or maintenance) branch, the versions of the published packages are checked against the public registry. In case of detected changes, the new version is automatically published to said registries.

This action should be easily applicable to other repositories with minimal changes.


Lotes:
I think tsconfig.json needs to be updated, because the actions folder is not covered by it yet, or?

gfontorbe:
@Lotes The action is gonna be moved to another repo eventually, this is here only for review purposes

spoenemann (commented):
@gfontorbe please add `environment: publish` to the publish job. This environment will be created after https://github.com/eclipse-langium/.eclipsefdn/pull/6 is merged.

.github/actions/publish/src/publish.ts
```diff
@@ -0,0 +1,191 @@
...
+    const ovsxVersion = await getOvsxVersion(fullName);
+    const shouldPublishVsce = compare(version, vsceVersion) === 1;
+    const shouldPublishOvsx = compare(version, ovsxVersion) === 1;
+    const fileName = `${name}-${version}.vsix`;
+
+    if (shouldPublishVsce || shouldPublishOvsx) {
+        console.log(`Extension ${fullName} has updates. Generating vsix...`);
+        if (!dryRun) {
+            // npx.cmd is needed on Windows; out of scope (publishes on ubuntu-latest)
+            execFileSync('npx', ['vsce', 'package', '-o', fileName], { cwd: packagePath });
```
spoenemann: *IMPORTANT:* We should allow pinning the version number of any command we run via `npx` (vsce, ovsx). The action should provide parameters to set a specific version. We can use `latest` as default.

```suggestion
            execFileSync('npx', [`vsce@${vsceVersion}`, 'package', '-o', fileName], { cwd: packagePath });
```

.github/workflows/publish.yml
```diff
@@ -11,41 +12,60 @@ on:
...
+          npm run langium:generate:production --workspace=langium
+          npm run build
+
+      - name: Test
+        if: success() || failure()
+        shell: bash
+        run: npm run test
+
+      - name: Publish
+        uses: TypeFox/gh-publish-npm@v0.2.0
```
msujew: Minor: I know this is our own package, but please pin the commit version.

.github/workflows/publish.yml
```diff
@@ -11,41 +12,60 @@ on:
...
+      - '**/package.json'
       - '.github/workflows/publish.yml'
-      - 'scripts/publish-updated-packages.ts'
 
 jobs:
   publish:
     name: Langium Publish
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    environment: publish
```
spoenemann: This depends on https://github.com/eclipse-langium/.eclipsefdn/pull/6

msujew (approved):
Thanks, output from the action looks reasonable. Let's see this in action once this is merged :)

spoenemann:
Shall we merge it without the publish environment and add it back after https://github.com/eclipse-langium/.eclipsefdn/pull/6 has been merged?

.github/workflows/publish.yml
gfontorbe: Done

spoenemann (commented):
Please remove scripts/publish-updated-packages.ts, remove `environment: publish`, and then let's merge.

msujew (approved):
Thanks, I think this is good enough for a test run. I'll let you know once I do the release whether there are any problems with the workflow/action 👍

PR #2161, opened 2026-06-01
feat: use classes in linker for ref accessors (alternative approach)

cdietrich:

PR #2165, opened 2026-06-08
feat: update to LSP 3.18

### LINKED ISSUE #2163: Update to LSP 3.18
now that LSP 3.18 is out and upstream libs are available we might consider supporting LSP 3.18 too
as the last update is so long ago i dont even remember: @msujew what is/was our update strategy in the past

cdietrich:
closes #2163

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -141,7 +141,7 @@ export default [{
             patterns: [ {
                 group: [ 'vscode-jsonrpc' ],
                 importNamePattern: '^(?!CancellationToken)',
-                message: 'Don\'t import types or symbols from "vscode-jsonrpc" (package index), as that brings a large overhead in bundle size. Import from "vscode-jsonrpc/lib/common/...js" and add a // eslint-disable..., if really necessary.'
+                message: 'Don\'t import types or symbols from "vscode-jsonrpc" (package index), as that brings a large overhead in bundle size. Add a // eslint-disable... comment if really necessary.'
             }]
         }],
         // use @typescript-eslint/no-unused-vars instead
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -128,22 +128,6 @@ export default [{
         'pluginHeader/header': [2, 'block', [
             { pattern: 'MIT License|DO NOT EDIT MANUALLY!' }
         ]],
-        'no-restricted-imports': ['error', {
-            paths: [{
-                name: 'vscode-jsonrpc',
-                importNames: [ 'CancellationToken' ],
-                message: 'Import "CancellationToken" via "Cancellation.CancellationToken" from "langium", or directly from "./utils/cancellation.ts" within Langium.'
-            }, {
-                name: 'vscode-jsonrpc/',
-                importNames: [ 'CancellationToken'],
-                message: 'Import "CancellationToken" via "Cancellation.CancellationToken" from "langium", or directly from "./utils/cancellation.ts" within Langium.'
-            }],
-            patterns: [ {
-                group: [ 'vscode-jsonrpc' ],
-                importNamePattern: '^(?!CancellationToken)',
-                message: 'Don\'t import types or symbols from "vscode-jsonrpc" (package index), as that brings a large overhead in bundle size. Add a // eslint-disable... comment if really necessary.'
-            }]
-        }],
         // use @typescript-eslint/no-unused-vars instead
         'no-unused-vars': 'off',
         // Disallow unnecessary escape characters
--- a/packages/langium-sprotty/src/diagnostic-marker-provider.ts
+++ b/packages/langium-sprotty/src/diagnostic-marker-provider.ts
@@ -6,7 +6,7 @@
 
 import type { AstNode } from 'langium';
 import type { SModelElement } from 'sprotty-protocol';
-import type { Diagnostic, Range } from 'vscode-languageserver';
+import { Diagnostic, type Range } from 'vscode-languageserver';
 import type { GeneratorContext } from './diagram-generator.js';
 import type { LangiumSprottyServices } from './sprotty-services.js';
 import { AstUtils } from 'langium';
@@ -40,7 +40,7 @@ export class DefaultDiagnosticMarkerProvider implements DiagnosticMarkerProvider
         }
         diagnostics.sort((a, b) => (a.severity ?? 0) - (b.severity ?? 0));
         const issues = diagnostics.map(diagnostic => ({
-            message: diagnostic.message,
+            message: Diagnostic.getMessageString(diagnostic),
             severity: convertSeverity(diagnostic)
         }) satisfies LangiumSprottyIssue);
         const marker: SDiagnosticMarker = {
--- a/packages/langium/src/lsp/language-server.ts
+++ b/packages/langium/src/lsp/language-server.ts
@@ -704,7 +704,7 @@ export function addTypeHierarchyHandler(connection: Connection, sharedServices: 
 }
 
 export function createHierarchyRequestHandler<P extends TypeHierarchySupertypesParams | TypeHierarchySubtypesParams | CallHierarchyIncomingCallsParams | CallHierarchyOutgoingCallsParams, R extends NonNullable<unknown>, _PR, E = void>(
-    serviceCall: (services: LangiumCoreAndPartialLSPServices, params: P, cancelToken: CancellationToken) => MaybePromise<R | null | undefined>,
+    serviceCall: (services: LangiumCoreAndPartialLSPServices, params: P, cancelToken: CancellationToken) => MaybePromise<R | null | undefined | ResponseError<E>>,
     sharedServices: LangiumSharedServices,
     requiredState?: ServiceRequirement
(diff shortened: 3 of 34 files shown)


packages/langium-sprotty/src/diagnostic-marker-provider.ts
```diff
@@ -40,7 +40,7 @@ export class DefaultDiagnosticMarkerProvider implements DiagnosticMarkerProvider
         }
         diagnostics.sort((a, b) => (a.severity ?? 0) - (b.severity ?? 0));
         const issues = diagnostics.map(diagnostic => ({
-            message: diagnostic.message,
+            message: Diagnostic.getMessageString(diagnostic),
```
cdietrich, line 43: ok with this behaviour?

packages/langium/src/lsp/language-server.ts
```diff
@@ -725,7 +724,7 @@ export function createHierarchyRequestHandler<P extends TypeHierarchySupertypesP
         } catch (err) {
             return responseError<E>(err);
         }
-    };
```
cdietrich, line 728: TODO check these

packages/langium/src/lsp/language-server.ts
```diff
@@ -54,6 +52,9 @@ export interface LanguageServer {
  */
 export type LangiumCoreAndPartialLSPServices = Omit<LangiumCoreServices & PartialLangiumLSPServices, 'shared'>
 
+type LangiumServerRequestHandler<P, R, E> = (params: P, token: CancellationToken) => HandlerResult<R | null, E, R | null>;
```
cdietrich, line 55: @msujew @spoenemann any idea how to make this "nice"

eslint.config.mjs
```diff
@@ -141,7 +141,7 @@ export default [{
             patterns: [ {
                 group: [ 'vscode-jsonrpc' ],
                 importNamePattern: '^(?!CancellationToken)',
-                message: 'Don\'t import types or symbols from "vscode-jsonrpc" (package index), as that brings a large overhead in bundle size. Import from "vscode-jsonrpc/lib/common/...js" and add a // eslint-disable..., if really necessary.'
+                message: 'Don\'t import types or symbols from "vscode-jsonrpc" (package index), as that brings a large overhead in bundle size. Add a // eslint-disable... comment if really necessary.'
```
cdietrich: it looks like vscode-jsonrpc has changed again so the previous workaround does no longer work

packages/langium/src/utils/cancellation.ts
```diff
@@ -5,4 +5,4 @@
  ******************************************************************************/
 
 // eslint-disable-next-line no-restricted-imports
-export * from 'vscode-jsonrpc/lib/common/cancellation.js';
+export { CancellationToken, type AbstractCancellationTokenSource, CancellationTokenSource } from 'vscode-jsonrpc';
```
cdietrich, line 7: sadly this old trick does no longer seem to work.
maybe there is a different workaround

packages/langium/src/utils/event.ts
```diff
@@ -5,4 +5,4 @@
  ******************************************************************************/
 
 // eslint-disable-next-line no-restricted-imports
-export * from 'vscode-jsonrpc/lib/common/events.js';
+export { type Event, Emitter } from 'vscode-jsonrpc';
```
cdietrich, line 7: maybe we also need a 

```
+export interface EmitterOptions {
+    onFirstListenerAdd?: (...args: unknown[]) => unknown;
+    onLastListenerRemove?: (...args: unknown[]) => unknown;
+}
```

packages/langium/src/lsp/normalized-text-documents.ts
```diff
@@ -10,13 +10,26 @@ import type {
...
+    LSPObject
 } from 'vscode-languageserver';
 import { TextDocumentSyncKind, Disposable, Emitter } from 'vscode-languageserver';
 import type { URI } from '../utils/uri-utils.js';
 import { UriUtils } from '../utils/uri-utils.js';
-// For some reason, this isn't exported by vscode-languageserver
-import type { NotebookDocumentChangeEvent } from 'vscode-languageserver/lib/common/notebook.js';
+
+// NotebookDocumentChangeEvent is not part of the public vscode-languageserver exports
+export type NotebookDocumentChangeEvent = {
```
cdietrich: am not sure if there is a better way or a way to avoid it

packages/langium-sprotty/src/diagnostic-marker-provider.ts
msujew: I think that makes sense. If the adopter uses markdown diagnostics, then they'll have to deal with this in their sprotty code correctly.

packages/langium/src/utils/cancellation.ts
msujew: No need for a workaround anymore. The issue with CJS is that bundlers don't perform tree-shaking on it (at least not well). Since this is now ESM, tree-shaking should work much better.

cdietrich: will remove it from the eslint config

cdietrich: hmmmm so i guess we would need to measure how bad it actually is

packages/langium/src/utils/event.ts
cdietrich: or we are fine with no longer exporting it

packages/langium/src/lsp/normalized-text-documents.ts
```diff
@@ -10,13 +10,17 @@ import type {
...
 } from 'vscode-languageserver';
 import { TextDocumentSyncKind, Disposable, Emitter } from 'vscode-languageserver';
 import type { URI } from '../utils/uri-utils.js';
 import { UriUtils } from '../utils/uri-utils.js';
-// For some reason, this isn't exported by vscode-languageserver
-import type { NotebookDocumentChangeEvent } from 'vscode-languageserver/lib/common/notebook.js';
+
+// `NotebookDocumentChangeEvent` is not re-exported from vscode-languageserver's public entry,
+// so we derive it from the public `NotebookDocuments#onDidChange` event instead of duplicating it.
+export type NotebookDocumentChangeEvent =
```
cdietrich, line 22: are we ok with this as a betetr solution istead of complety copy it?

packages/langium/src/utils/cancellation.ts
msujew: We should ;)

packages/langium/src/lsp/normalized-text-documents.ts
cdietrich: maybe just use NotebookDocumentChangeEvent from language-server-types?

cdietrich: (i assume there is some diffs)

packages/langium/src/lsp/normalized-text-documents.ts
cdietrich: @msujew are you actually using it in your codebases. if yes how.

msujew: Not using it - I think this is fine as it is 👍

packages/langium/src/utils/event.ts
msujew: I think that's ok 👍

packages/langium/src/lsp/normalized-text-documents.ts
cdietrich: good. i dont know why they have different types server and client side

cdietrich:
@msujew can we do build an alpha from a branch to ease testing

msujew:
I've published the changes across all packages using the `4.3.0-next.a52b426` version.

cdietrich:
first tests look good so far
(monaco and sprotty i could not completely test as they still depend on old version)

kaisalmen:
@cdietrich my plan is to update monaco-languageclient this week and finally create a new release.

msujew (approved):
Thanks, looks good to me 👍

Once this and #2161 are merged, we can create a 4.3.0 release.

packages/langium/src/utils/cancellation.ts
cdietrich: @msujew do you have testcases / test projects for that?

cdietrich: in a naive experiment with esbuild i ended up with complete vscode-jsonrpc

cdietrich: do we have an alternative idea?

msujew: Let me have a closer look at this 👍

msujew: Had a quick look. Even with the current version, we end up with a full `vscode-jsonrpc/common` directory in our build, because `vscode-languageserver-protocol` imports directly from `vscode-jsonrpc`. In fact, the new build is roughly ~8% smaller than previously. So this is in fact a win.

PR #2169, opened 2026-06-15
Update Langium packages to version 4.3.0

msujew:
We had a few improvements/changes that warrant a minor release. Especially #2165, since it contains breaking version changes to our runtime dependencies.

This removes the direct dependency to `vscode-jsonrpc`, since all symbols are transitively available from `vscode-languageserver-protocol`.


cdietrich:
@msujew looks like the EmitterOptions is there too, should we restore it?

update its not

cdietrich (approved):

PR #2175, opened 2026-06-25
fix: prevent prototype pollution in merge

cdietrich:


packages/langium/src/dependency-injection.ts
```diff
@@ -153,6 +153,9 @@ function _resolve<I, T>(obj: any, prop: string | symbol | number, module: Module
 function _merge(target: Module<any>, source?: Module<any>): Module<unknown> {
     if (source) {
         for (const [key, sourceValue] of Object.entries(source)) {
+            if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
```
cdietrich, line 156: alternative: skip instead of throw

msujew: I would just skip it actually. Maybe with a `console.warn` to ensure that it is somewhere logged with `Illegal object key %s found in dependency injection module`.

cdietrich: i will just siltently skip. am not sure if we log anywhere else

cdietrich: done

msujew (approved):
Thanks 👍

PR #2177, opened 2026-06-26
added types node to tsconfig in generator template

jpilgrim:
As VS Code (and newer typescript versions) do not automatically set node types as globally available types, one need to add this to the tsconfig.json in the yeoman templates as well. Otherwise the generated code contains compile errors.


msujew (approved):
Thanks! I'll create a follow up this week to pull up the default version of TypeScript to 6 - that way we have a regression test with our existing generator tests for this.

PR #2179, opened 2026-06-29
First iteration of AST-viewer view

gfontorbe:
Adds an AST viewer to the Langium VSCode extension.

Also contributes a package containing a protocol and utility functions to wire DSL-specific extensions to the viewer.

### DIFF
--- a/examples/arithmetics/package.json
+++ b/examples/arithmetics/package.json
@@ -36,6 +36,9 @@
   "activationEvents": [
     "onLanguage:arithmetics"
   ],
+  "extensionDependencies": [
+    "langium.langium-vscode"
+  ],
   "files": [
     "bin",
     "out",
@@ -69,6 +72,7 @@
     "chalk": "~5.6.2",
     "commander": "~15.0.0",
     "langium": "~4.3.0",
+    "langium-inspector": "~4.2.0",
     "vscode-languageclient": "~10.0.1",
     "vscode-languageserver": "~10.0.1"
   },
--- a/examples/arithmetics/package.json
+++ b/examples/arithmetics/package.json
@@ -72,7 +72,7 @@
     "chalk": "~5.6.2",
     "commander": "~15.0.0",
     "langium": "~4.3.0",
-    "langium-inspector": "~4.2.0",
+    "langium-inspector": "~4.3.0",
     "vscode-languageclient": "~10.0.1",
     "vscode-languageserver": "~10.0.1"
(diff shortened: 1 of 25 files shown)


examples/arithmetics/package.json
```diff
@@ -36,6 +36,9 @@
   "activationEvents": [
     "onLanguage:arithmetics"
   ],
+  "extensionDependencies": [
```
spoenemann, line 39: We should regard the langium-vscode extension as a pure _dev dependency_ and thus optional. I think the extensionDependencies field should be used for hard dependencies.

packages/langium-inspector/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "files": [
+        "lib",
+        "src"
+    ],
+    "exports": {
+        ".": {
+            "import": "./lib/index.js",
+            "types": "./lib/index.d.ts"
+        },
+        "./protocol": {
```
spoenemann, line 28: The index already exports everything from protocol, so no need for an additional export here.

packages/langium-vscode/src/ast-view.ts
```diff
@@ -0,0 +1,564 @@
...
+        const range = new vscode.Range(
+            new vscode.Position(message.range.start.line, message.range.start.character),
+            new vscode.Position(message.range.end.line, message.range.end.character),
+        );
+        const doc = await vscode.workspace.openTextDocument(uri);
+        await vscode.window.showTextDocument(doc, { selection: range, preserveFocus: true });
+    }
+
+    private buildWebviewHtml(): string {
+        return `<!DOCTYPE html>
```
spoenemann, line 118: This huge inline HTML is too difficult to maintain. Question: could the built-in tree view provided through VS Code API be good enough for this?

If that's not good enough, then we should extract the CSS and JS code from this inline string into separate files and load them from the HTML.

spoenemann (changes requested):
Great stuff, thank you!

I chose 'request changes' because of the inline HTML which should be refactored, either into a tree view controlled via extension API or into multiple files loaded by the webview HTML.

packages/langium-vscode/src/ast-view.ts
gfontorbe: I'll implement a second view using VSCode's treeview and we can compare the two

gfontorbe:
Replaced by #2185

PR #2181, opened 2026-07-01
Fix multi-ref json deserialization

### LINKED ISSUE #2180: JsonSerializer does not restore multi-references after deserialization
When an AST containing a multi-reference goes through a serialize/deserialize round trip, the reference is not restored as a MultiReference. The serialized JSON contains $refs, but the deserialized AST keeps it as a plain object, so the referenced items are no longer available.

Langium version: 4.3.0
Package name: langium

## Steps To Reproduce

1. Use a grammar with a multi-reference:

    ```langium
    grammar Repro

    entry Model: (persons+=Person | greetings+=Greeting)*;
    Person: 'person' name=ID;
    Greeting: 'hello' person=[+Person:ID];

    hidden terminal WS: /\s+/;
    terminal ID: /[_a-zA-Z][\w]*/;
    ```

2. Parse a document where the reference resolves to multiple targets:

    ```text
    person Alice
    person Bob
    person Alice
    hello Alice
    ```
3. Serialize and deserialize the AST:
    ```ts
    const serializer = services.serializer.JsonSerializer;
    const json = serializer.serialize(document.parseResult.value);
    const restored = serializer.deserialize<any>(json);
    ```

After this, `restored.greetings[0].person.items` is undefined.

## The current behavior
The serialized JSON contains $refs, but deserialization leaves it as a plain object instead of a MultiReference.

## The expected behavior
The deserialized AST should contain a MultiReference with items, like the original AST.

msujew:
Closes https://github.com/eclipse-langium/langium/issues/2180

Fixes an oversight in `isIntermediateReference` that prevented the deserialization of multi references to work as expected.

### DIFF
--- a/packages/langium/src/serializer/json-serializer.ts
+++ b/packages/langium/src/serializer/json-serializer.ts
@@ -105,7 +105,7 @@ interface IntermediateReference {
 }
 
 function isIntermediateReference(obj: unknown): obj is IntermediateReference {
-    return typeof obj === 'object' && !!obj && ('$ref' in obj || '$error' in obj);
+    return typeof obj === 'object' && !!obj && ('$ref' in obj || '$refs' in obj || '$error' in obj);
 }
 
 export class DefaultJsonSerializer implements JsonSerializer {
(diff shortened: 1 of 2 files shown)


spoenemann (approved):
👍

PR #2184, opened 2026-07-02
chore: bump vscode-languageclient vscode-languageserver vscode-languageserver-protocol

cdietrich:


msujew (approved):
AFAIK I don't think the update introduces anything interesting, right? So there's no real need to ship a new langium patch soon for this.

cdietrich:
yes, i also wonder if we should be less restrictive in our ranges to give consumers more freedom

msujew:
> yes, i also wonder if we should be less restrictive in our ranges to give consumers more freedom

I'd be open for this. We originally restricted this, because we had a dependency that introduced a breaking change in a minor version (I believe it was one of the vscode-languageserver libs). But ideally, it shouldn't be an issue. However, we should also add a note to the Langium readme that we cannot really follow semantic versioning across minor versions, since we sometimes break internal/protected APIs.

PR #2185, opened 2026-07-10
Add AST viewer to Langium VSCode extension

gfontorbe:
Adds a treeview contribution to the Langium VSCode extension to display the opened document's AST.


examples/arithmetics/src/extension.ts
```diff
@@ -52,5 +52,17 @@ async function startLanguageClient(context: vscode.ExtensionContext): Promise<La
...
     await client.start();
+    // Register with the Langium Inspector extension if present
+    context.subscriptions.push(await registerInspector(client, 'arithmetics'));
     return client;
 }
+
+async function registerInspector(client: LanguageClient, languageId: string): Promise<vscode.Disposable> {
+    const vscodeApi = await import('vscode');
+    try {
+        await vscodeApi.commands.executeCommand('langium-inspector.register', client, languageId);
```
spoenemann: Is this missing a `return`? Otherwise we always return a disposable that does nothing.

spoenemann (commented):
Looks good! Just a detail

examples/arithmetics/src/extension.ts
gfontorbe: It's missing the unregistration logic completely. Fixing that

PR #2189, opened 2026-07-21
 npm audit fixes for shell-quote, brace-expansion and others

cdietrich:


cdietrich:
https://github.com/eclipse-xtext/xtext/pull/3762
but there also seems to be infra problems with dependency repos

msujew (approved):

PR #2190, opened 2026-07-21
Implement TextDocumentContentProvider

msujew:
Adds support for the recent LSP 3.18 `workspace/textDocumentContent` request. Allows to circumvent the need for a dedicated file system provider for the given URI scheme in favor of implementing this request.

Can be (temporarily) tested by trying it out in the domainmodel example. Use the `Int` datatype somewhere and go to its definition to see how it works.

### DIFF
--- a/examples/domainmodel/src/language-server/domain-model-workspace-manager.ts
+++ b/examples/domainmodel/src/language-server/domain-model-workspace-manager.ts
@@ -1,0 +1,26 @@
+/******************************************************************************
+ * Copyright 2026 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { DefaultWorkspaceManager, URI, type LangiumDocument, type LangiumDocumentFactory, type WorkspaceFolder } from 'langium';
+import type { LangiumSharedServices } from 'langium/lsp';
+
+export class DomainModelWorkspaceManager extends DefaultWorkspaceManager {
+
+    private readonly documentFactory: LangiumDocumentFactory;
+
+    constructor(services: LangiumSharedServices) {
+        super(services);
+        this.documentFactory = services.workspace.LangiumDocumentFactory;
+    }
+
+    protected override async loadAdditionalDocuments(_folders: WorkspaceFolder[], collector: (document: LangiumDocument) => void): Promise<void> {
+        const document = this.documentFactory.fromString(
+            'datatype Int',
+            URI.parse('domain-model://domainmodel/Int.dmodel')
+        );
+        collector(document);
+    }
+}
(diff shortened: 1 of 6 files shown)


packages/langium/src/lsp/language-server.ts
```diff
@@ -594,6 +599,20 @@ export function addCodeLensHandler(connection: Connection, services: LangiumShar
     ));
 }
 
+export function addTextDocumentContentRequestHandler(connection: Connection, services: LangiumSharedServices): void {
```
cdietrich, line 602: should we await workspaceManager.ready

packages/langium/src/lsp/text-document-content-provider.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export interface TextDocumentContentProvider {
+    /**
+     * The URI schemes that this provider supports.
+     */
+    readonly schemes: string[];
+    /**
+     * Returns the text content of a file given its URI.
+     */
+    provideTextDocumentContent(params: TextDocumentContentParams, cancellationToken: CancellationToken): Promise<TextDocumentContentResult | undefined>;
```
cdietrich, line 28: do we have a common pattern how we do `| undefined` or `| null` also with respect to what the spec says?

packages/langium/src/lsp/language-server.ts
msujew: Good point, it's now awaited in the default implementation 👍

packages/langium/src/lsp/text-document-content-provider.ts
msujew: Not really. Generally, we discourage `null` in Langium's source code (eslint rule), while `undefined` should be used instead, if the result type needs some way to distinguish between empty and a none-result. For services that return an object, `| undefined` is required.

examples/domainmodel/src/language-server/domain-model-workspace-manager.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    constructor(services: LangiumSharedServices) {
+        super(services);
+        this.documentFactory = services.workspace.LangiumDocumentFactory;
+    }
+
+    protected override async loadAdditionalDocuments(_folders: WorkspaceFolder[], collector: (document: LangiumDocument) => void): Promise<void> {
+        const document = this.documentFactory.fromString(
+            'datatype Int',
+            URI.parse('domain-model://domainmodel/Int.dmodel')
```
JohannesMeierSE: Suggestion: Introduce `export const builtinScheme = 'domain-model';` and use it here and in the `domain-model-module.ts` to make the relationship between these "magical strings" more explicit.

JohannesMeierSE (approved):
Works fine for me, thanks @msujew!

I have only one small suggestion to make the example a bit more readable.

PR #2191, opened 2026-07-22
CST memory optimization

spoenemann:
This PR aims at improving Langium's memory usage, which was found to be relatively high [in this article](https://www.typefox.io/blog/fastbelt-architecture/). The Conrete Syntax Tree (CST) implementation is the main reason for this. These measures have been applied:

- `root` is no longer stored in all CST nodes, but derived by walking up the container hierarchy.
- `range` is no longer stored in leaf CST nodes, but created on demand from integer values.
- Optional properties (especially `_astNode` and `_rangeCache`) are initialized with `undefined` so the V8 engine includes them in the finalized object layout (this avoids the overhead of a separate `PropertyArray` for properties added later).
- `content` is now a `readonly` array and must be manipulated via the CstNodeBuilder so we avoid the CstNodeContainer overhead.

The first two changes save memory by doing a bit more computation for the `root`, `text`, and `range` properties, which mostly affects LSP-related services. The overhead of these computations is virtually negligible, and the benefits of the changes outweigh it by far as shown in the benchmarks below.

I created four benchmarks to measure the impact of these changes on the CST size, parsing, and document building:

- `cst-node-memory-benchmark.ts` measures the memory usage of a synthetically generated CST
    - **→ −49 % memory** compared to main 🎉
- `cst-parse-memory-benchmark.ts` measures the memory usage of a parsed CST
    - **→ −45 % memory** compared to main 🎉
- `parser-time-benchmark.ts` measures the execution time of the parser
    - **→ −33 % parse time** compared to main 🎉
- `workspace-build-benchmark.ts` measures the execution time and heap size of a full workspace build
    - **→ −14 % build time** and **−19 % heap size** compared to main 🎉


cdietrich:
in my measurements we see only 35% memory gain (on the total heap) but this is still a huge improvement. for pure cst it is almost 50%

CONTRIBUTING.md
```diff
@@ -49,8 +49,45 @@ Take a look at the examples readily available inside the repository to get start
...
+
+The test suite is based on [Vitest](https://vitest.dev). Run it from the repository root:
+
+```shell
+npm test               # run all tests once
+npm run test:watch     # re-run tests on file changes
+npm run coverage       # run tests with coverage report
+```
+
+You can also run the tests of a single package, e.g. `npm test --workspace=langium`.
```
cdietrich, line 65: is this correct?
https://github.com/eclipse-langium/langium/blob/0be7d7e2b0c85d2de6fa483a0fd4702fbeebc00f/packages/langium/package.json#L56

npm test --workspace=langium
npm error Lifecycle script `test` failed with error:
npm error workspace langium@4.3.0
npm error location /Users/dietrich/git/langium/packages/langium
npm error Missing script: "test"
npm error
npm error To see a list of scripts, run:
npm error   npm run --workspace=langium@4.3.0

cdietrich: maybe
```npm test -- packages/langium```

File: CONTRIBUTING.md [hunk 2]
```diff
@@ -28,7 +28,7 @@ In case you have a question, please look into the [documentation](https://langiu
 
 ## Prerequisites
 
-For developing Langium, you require at least Node.js version 16 and at least npm version 7.7.0 to be able to use npm workspaces.
+For developing Langium, you require at least Node.js version 20.10.0 (see the `engines` entry in [package.json](./package.json)). The npm version bundled with it supports npm workspaces, which this repository relies on.
```
cdietrich, line 31: in the nvmrc we have 24, package json requires 22

CONTRIBUTING.md
```diff
@@ -67,15 +104,15 @@ then you should follow some recommendations.
...
 
- If you execute `npm install` on the top-level of your Langium checkout all dependencies will be gathered locally inside `node_modules` and all four packages or sub-projects will be built. **Warning**: Do not run npm install directly inside the four Langium packages as it will mess up dependency resolution once you link packages globally.
+If you execute `npm install` on the top-level of your Langium checkout all dependencies will be gathered locally inside `node_modules` and all packages or sub-projects will be built. **Warning**: Do not run npm install directly inside the individual Langium packages as it will mess up dependency resolution once you link packages globally.
```
cdietrich, line 111: install wont build

examples/statemachine/test/workspace-build-benchmark.ts
msujew, line 1: Suggestion: `vitest` now has natively (experimental) integrated benchmarking tools. Would it make sense to use there here?

msujew (approved):
Thanks, looks quite good. One (optional) suggestion.

examples/statemachine/test/workspace-build-benchmark.ts
spoenemann: Thanks for the suggestion, I looked into `vitest bench`. I'd keep the standalone scripts for now:

- `vitest bench` (tinybench under the hood) measures wall-clock time only. There is no memory measurement, and three of the four benchmarks are about retained heap.
- It yields one number per `bench()` function and exposes no per-iteration hooks, so the parse/build split and the per-round cleanup in the workspace benchmark don't map onto it.
- Derived metrics (bytes per CST node, MB/s, diagnostics sanity check) would still need manual logging.
- The API is declared experimental and does not follow SemVer; Vitest still bundles tinybench 2.x while 3.x already changed the API.

If we want `vitest bench --compare` for regression tracking later, converting just `parser-time-benchmark.ts` (the only pure time benchmark) in a follow-up would be a good fit.

PR #2194, opened 2026-07-29
Sort keywords by length first (#2193)

### LINKED ISSUE #2193: Generated Highlightings "fail" if one keywords starts with another keyword as its prefix
Langium version: 4.2.0
Package name: langium-cli

## Steps To Reproduce

Suppose you have a language with the following keywords:
- local-date
- local-date-time

If you generate the the VS-Code extension the keywords are added to the generated `.tmLanguage.json` and `.monarch.ts` in lexical order. My understanding of the lexical highlighting is that the highlighting checks from left to right and so it does not highlight `local-date-time` but only the `local-date` part of it. See screenshot from my [project](https://github.com/BestSolution-at/rsd) in **The current behavior**.

I tested this by manually reordering the keyword entries in the files mentionned above (see screenshot of **The expected behavior**)

## The current behavior

Image

## The expected behavior

Image

## Suggested fix

There are different positions this could be fixed:
- change sorting in collectKeywords and first sort by length and then lexical
- sort in monarch/prism/textmate generators by length and then lexical

The only other place currently using the `collectKeywords` helper is the `ast-generator` so I think it is safe to change the sorting in `collectKeywords` although the fix is only needed for the highlighters, so maybe you prefer it there.

If you agree with me that reordering by length and then lexical I could prepare a patch if you tell me where you think this should be fixed.

tomsontom:
To make sure keywords who start with the same character sequence as another (shorter) keyword the keywords are sorted by length (longest first) and then lexical.

fixes #2193

### DIFF
--- a/examples/arithmetics/syntaxes/arithmetics.monarch.ts
+++ b/examples/arithmetics/syntaxes/arithmetics.monarch.ts
@@ -6,7 +6,7 @@ export default {
     operators: [
         '%','*','+',',','-','/',':',';','^'
     ],
-    symbols: /%|\(|\)|\*|\+|,|-|\/|:|;|\^/,
+    symbols: /-|,|;|:|\(|\)|\*|\/|%|\^|\+/,
 
     tokenizer: {
         initial: [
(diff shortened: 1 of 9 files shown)


msujew (approved):
Thanks, looks good to me 👍

PR #2202, opened 2026-08-08
Use declared datatype of datatype rule #2201

### LINKED ISSUE #2201: Data type rule with single terminal and different return types: Mismatch between inferred AST feature types and runtime behavior
Langium version: 4.3.0
Package name: langium

# Description

When a data rule declares a `returns string`, that declared return type is ignored for the generated TypeScript declarations. But the runtime parser respects it and produces a string in the AST. This can lead to a mismatch between the declared AST type and its actual value.

For example, when a data type happens to consist of only a single terminal and the declared return type of the terminal and the data type are different, there is a mismatch between the inferred AST type and the actual runtime type. 

```
// Inferred type for person.name in the generated Person interface will be number...
Person: 'person' name=PersonId;

// ...even though it is declared as string
PersonId returns string: INT;

// But at runtime, person.name is a string in the parsed AST

// Works as expected if there is more than one terminal
// PersonId returns string: INT '.' INT;
```

## Steps To Reproduce

Simplified example grammar that shows the use case (for section, name happens to be a single int, but for consistency, all should be string):

Grammar

```
grammar HelloWorld

entry Model:
    (sections+=Section | subsections+=Subsection | terms+=Term)*;

Section:
    'section' name=SectionId;

Subsection:
    'subsection' name=SubsectionId;

Term:
    'term' name=ID
    '[' 'rulebook' 'references' ']' (references+=RulebookReference)*;

RulebookReference: 
  ({infer RulebookSectionReference} 'section' section=[Section])
  ({infer RulebookSubsectionReference} 'subsection' subsection=[Subsection])
;

SectionId returns string: INT;
SubsectionId returns string: INT '.' INT;

hidden terminal WS: /\s+/;
terminal ID returns string: /[_a-zA-Z][\w_]*/;
terminal INT returns number: /[0-9]+/;
```

With this content:

```
section 100
section 101

subsection 101.1
subsection 101.2
subsection 101.3

term draw
[rulebook references]
  section 100
  subsection 101.1
```

## The current behavior

The inferred type of the feature in the AST is `number`:

```ts
export interface Section extends langium.AstNode {
    readonly $container: Model;
    readonly $type: 'Section';
    name: SectionId;
}

export type SectionId = number;
```

But during runtime, the actual type of `section.name` is string, which you can see either by logging it somewhere, or in the playground:

Image

## The expected behavior

The inferred type of the feature in the AST should be `string`.

(Or at least there should be no mismatch between the inferred AST feature types and the runtime behavior.)

Everything else seems to work, cross-references are recognized and resolved properly.

## Notes

* I have a converter that ensures `001 . 9` becomes the string `"001.9"` and does not get truncated to `"1.9"`, and also removes superfluous whitespaces.
* The issue can be worked around by declaring the interface for the data rule explicitly. I will be doing that once I'm done designing my grammar. (For now I'm still playing around with langium and trying out grammar ideas.)

## Observations

* Related to the `rule.dataType && rule.dataType !== 'string'` check in [getDataRuleType from inferred-types.ts](https://github.com/eclipse-langium/langium/blob/0be7d7e2b0c85d2de6fa483a0fd4702fbeebc00f/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts#L347)
* When I remove the `&& rule.dataType !== 'string'`, it works as I would expect 
* But then, the test `Should infer data type rules as unions` fails
  * `Strings returns string: 'a' | 'b' | 'c';` then gets inferred as `string`
  * Kinda weird though, why should it infer a union when the return type is set to string explicitly?
  * And if it should, shouldn't it also infer a union for `Numbers returns number: '1' | '2' | '3';` ?
* Test case for `packages/langium/test/grammar/type-system/inferred-types.test.ts` that fails and shows the issue:

```ts
    test('Should infer type of data type rule for single terminal', async () => {
        await expectTypes(`
            terminal INT returns number: /[0-9]+/;
            B returns string: INT;
            A: name=B a=B;
        `, expandToString`
            export interface A extends langium.AstNode {
                readonly $type: 'A';
                a: B;
                name: B;
            }
            export type B = string;
            
            export function isB(item: unknown): item is B {
                return typeof item === 'string';
            }
        `);
    });
```

blutorange:
Potential fix for #2201.

Reworked the implementation a bit, see [this comment below](https://github.com/eclipse-langium/langium/pull/2202#issuecomment-5301960698) for a description.

Closes #2201

### DIFF
--- a/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
@@ -402,7 +402,7 @@ function buildDataRuleType(element: AbstractElement, hasStringConstraint: boolea
                     regex = undefined;
                 }
                 return {
-                    primitive: hasStringConstraint ? 'string' :ref.type?.name ?? 'string',
+                    primitive: hasStringConstraint ? 'string' : ref.type?.name ?? 'string',
                     regex
                 };
             } else {
--- a/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
@@ -395,7 +395,7 @@ function buildDataRuleType(element: AbstractElement, hasStringConstraint: boolea
     if (visited.has(element)) {
         return visited.get(element);
     }
-    visited.set(element, { primitive: 'string' }); // default to string to avoid infinite recursion
+    visited.set(element, undefined);
 
     let type: PlainPropertyType | undefined;
     if (element.cardinality) {
--- a/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
@@ -371,7 +371,7 @@ function getDataRuleType(rule: ParserRule): PlainPropertyType {
     if (rule.returnType && isType(rule.returnType.ref)) {
         return { value: rule.returnType.ref.name };
     }
-    const type = buildDataRuleType(rule.definition, rule.dataType === 'string', new Map());
+    const type = buildDataRuleType(rule.definition, rule.dataType === 'string');
     return type ?? { primitive: 'string' };
 }
 
@@ -388,30 +388,23 @@ function getDataRuleType(rule: ParserRule): PlainPropertyType {
  *
  * @param element The AST element representing the definition of the data type rule.
  * @param hasStringConstraint Whether the data type rule is declared to return `string`.
- * @param visited A set of elements that have already been visited, to avoid infinite recursion.
  * @returns The inferred type of the data type rule, or undefined if the type cannot be inferred.
  */
-function buildDataRuleType(element: AbstractElement, hasStringConstraint: boolean, visited: Map<AbstractElement, PlainPropertyType | undefined>): PlainPropertyType | undefined {
-    if (visited.has(element)) {
-        return visited.get(element);
-    }
-    visited.set(element, undefined);
-
-    let type: PlainPropertyType | undefined;
+function buildDataRuleType(element: AbstractElement, hasStringConstraint: boolean): PlainPropertyType | undefined {
     if (element.cardinality) {
         // Multiplicity/optionality is not supported for types
-        type = undefined;
+        return undefined;
     }
     else if (isAlternatives(element)) {
         const types = element.elements
-            .map(e => buildDataRuleType(e, hasStringConstraint, visited))
+            .map(e => buildDataRuleType(e, hasStringConstraint))
             .filter(t => t !== undefined);
-        type = element.elements.length === types.length ? { types } : undefined;
+        return element.elements.length === types.length ? { types } : undefined;
     } else if (isGroup(element) || isUnorderedGroup(element)) {
         if (element.elements.length !== 1) {
-            type = undefined;
+            return undefined;
         } else {
-            type = buildDataRuleType(element.elements[0], hasStringConstraint, visited);
+            return buildDataRuleType(element.elements[0], hasStringConstraint);
         }
     } else if (isRuleCall(element)) {
         const ref = element.rule?.ref;
@@ -429,7 +422,7 @@ function buildDataRuleType(element: AbstractElement, hasStringConstraint: boolea
                 // converted to string, with a format that won't match the regex.
                 // E.g. 'Rule returns string: DATE' will use Date.toString() as a
                 // value in the AST.
-                type = {
+                return {
                     primitive: hasStringConstraint ? 'string' : terminalType,
                     regex: terminalType === 'string' ? regex : undefined,
                 };
@@ -438,37 +431,27 @@ function buildDataRuleType(element: AbstractElement, hasStringConstraint: boolea
                 // data type rules are not allowed to reference infix rules.
                 if (hasStringConstraint && isParserRule(ref)) {
                     if (isDataTypeRuleAssignableToString(ref)) {
-                        if (ref.fragment) {
-                            // Cannot reference a fragment type, since fragments
-                            // are not output as types. Instead, recurse into the
-                            // fragment and build the type from its definition.
-                            type = buildDataRuleType(ref.definition, hasStringConstraint, visited);
-                        } else {
-                            type = { value: ref.name };
-                        }
+                        return { value: ref.name };
                     } else {
-                        type = { primitive: 'string' };
+                        return { primitive: 'string' };
                     }
                 }
                 else {
-                    type = {
+                    return {
                         value: ref.name,
                     };
                 };
             }
         } else {
-            type = undefined;
+            return undefined;
         }
     } else if (isKeyword(element)) {
-        type = {
+        return {
             string: element.value
         };
     } else {
-        type = undefined;
+        return undefined;
     }
-
-    visited.set(element, type);
-    return type;
 }
 
 /**
--- a/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
@@ -344,10 +344,19 @@ function calculateInfixInterfaces(rules: InfixRule[]): PlainInterface[] {
 }
 
 function getDataRuleType(rule: ParserRule): PlainPropertyType {
-    if (rule.dataType && rule.dataType !== 'string') {
-        return {
-            primitive: rule.dataType
-        };
+    if (rule.dataType) {
+        // If a data type was declared, use that data type. Otherwise, infer the data type.
+        // Special cases to consider:
+        // "A return string: INT;" must result in a string type, not a number type.
+        // "A return string: (INT)+;" must result in a string type, not a number type.
+        // "A returns string: 'a' | 'b';" must result in a union type 'a' | 'b', not a plain string type.
+        // "A returns string: 'a';" must result in a string constant type 'a', not a plain string type.
+        const isAlternativeOrKeyword = isAlternatives(rule.definition) || isKeyword(rule.definition);
+        if (rule.dataType !== 'string' || !isAlternativeOrKeyword) {
+            return {
+                primitive: rule.dataType
+            };
+        }
     }
     let cancelled = false;
     const cancel = (): PlainPropertyType => {
--- a/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
@@ -272,7 +272,7 @@ export function collectInferredTypes(parserRules: ParserRule[], datatypeRules: P
     };
     for (const rule of parserRules) {
         const comment = commentProvider?.getComment(rule);
-        allTypes.push(...getRuleTypes(context, rule, services).map(typePath => ({...typePath, comment})));
+        allTypes.push(...getRuleTypes(context, rule, services).map(typePath => ({ ...typePath, comment })));
     }
     const infixInterfaces = calculateInfixInterfaces(infixRules);
     const interfaces = calculateInterfaces(allTypes, infixInterfaces);
@@ -438,9 +438,18 @@ function buildDataRuleType(element: AbstractElement, hasStringConstraint: boolea
                 // Only other remaining alternative is 'infix rule', and
                 // data type rules are not allowed to reference infix rules.
                 if (hasStringConstraint && isParserRule(ref)) {
-                    type = isDataTypeRuleAssignableToString(ref, visited)
-                        ? { value: ref.name }
-                        : { primitive: 'string' };
+                    if (isDataTypeRuleAssignableToString(ref, visited)) {
+                        if (ref.fragment) {
+                            // Cannot reference a fragment type, since fragments
+                            // are not output as types. Instead, recurse into the
+                            // fragment and build the type from its definition.
+                            type = buildDataRuleType(ref.definition, hasStringConstraint, visited);
+                        } else {
+                            type = { value: ref.name };
+                        }
+                    } else {
+                        type = { primitive: 'string' };
+                    }
                 }
                 else {
                     type = {
--- a/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
+++ b/packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
@@ -371,7 +371,7 @@ function getDataRuleType(rule: ParserRule): PlainPropertyType {
     if (rule.returnType && isType(rule.returnType.ref)) {
         return { value: rule.returnType.ref.name };
     }
-    const type = buildDataRuleType(rule.definition, rule.dataType === 'string');
+    const type = buildDataRuleType(rule.definition, rule.dataType === 'string', new Map());
     return type ?? { primitive: 'string' };
 }
 
@@ -391,7 +391,12 @@ function getDataRuleType(rule: ParserRule): PlainPropertyType {
  * @param visited A set of elements that have already been visited, to avoid infinite recursion.
  * @returns The inferred type of the data type rule, or undefined if the type cannot be inferred.
  */
-function buildDataRuleType(element: AbstractElement, hasStringConstraint: boolean): PlainPropertyType | undefined {
+function buildDataRuleType(element: AbstractElement, hasStringConstraint: boolean, visited: Map<AbstractElement, PlainPropertyType | undefined>): PlainPropertyType | undefined {
+    if (visited.has(element)) {
+        return visited.get(element);
+    }
+    visited.set(element, { primitive: 'string' }); // default to string to avoid infinite recursion
+
     let type: PlainPropertyType | undefined;
     if (element.cardinality) {
         // Multiplicity/optionality is not supported for types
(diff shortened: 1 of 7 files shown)


packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
```diff
@@ -344,10 +344,19 @@ function calculateInfixInterfaces(rules: InfixRule[]): PlainInterface[] {
...
+        // "A return string: INT;" must result in a string type, not a number type.
+        // "A return string: (INT)+;" must result in a string type, not a number type.
+        // "A returns string: 'a' | 'b';" must result in a union type 'a' | 'b', not a plain string type.
+        // "A returns string: 'a';" must result in a string constant type 'a', not a plain string type.
+        const isAlternativeOrKeyword = isAlternatives(rule.definition) || isKeyword(rule.definition);
+        if (rule.dataType !== 'string' || !isAlternativeOrKeyword) {
+            return {
+                primitive: rule.dataType
+            };
+        }
```
msujew: Suggestion: I don't think this is the correct fix for this, but it looks more like a workaround for me. Instead, the issue is that `buildDataRuleType` will simply return the type of the inner terminal rule call without checking that this is also the same primitive type as the `rule.dataType` value. The actual fix lives down at the `return type;` statement, where we should compare the `type.primitive` value with `rule.dataType`. If they are different, the `rule.dataType` should take precedence.

packages/langium/test/grammar/type-system/inferred-types.test.ts
```diff
@@ -87,6 +87,131 @@ describe('Inferred types', () => {
         `);
     });
 
+    test('Should infer number type for data type rule with no return type and single terminal of type number', async () => {
+        await expectTypes(`
+            terminal INT returns number: /[0-9]+/;
+            B: INT;
+            A: name=B b=B;
```
msujew: Minor: Is there a specific reason to have both `name` and `b` properties? They both test the same behavior. Please remove either one of those. The same applies to the other tests.

msujew (commented):
Thanks, for the contribution!

Please fix the remaining testing failures.

packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
blutorange: I think I might do that, since it also solves the test failure with the missing regex pattern.

But regarding the original code `rule.dataType && rule.dataType !== 'string'`. That felt weird to me. If the user declared a type, should we not just use that? `buildDataRuleType` only infers the type if none was declared, with the exception of the special `string` case. The only reason I found for this special case for `string` is so that

```langium
Strings returns string: 'a' | 'b';
```

Will result in a union. But this can already be done with an inferred type or an explicit type if one wants to be explicit:

```langium
Strings: 'a' | 'b';

type MyUnion = "a" | "b" | "c";
Strings returns MyUnion: 'a' | 'b';
// this is actually broken too right now as well,
// as buildDataRuleType ignores rule.returnType
```

(I feel like as long as this special case behavior exists, the code will look a bit workaround-ish.)

msujew: `returns string` is kind of a catch-all for data type rules. It's the simplest data type rule you can create, which simply takes the content of all its collected children and builds a new string from that. That's why it's written in such a way: We wanted to give accurate, but still useful output for people that use this catch-all mechanism.

I wouldn't want to break behavior now for such a minor fix. There are likely people that rely on this behavior.

packages/langium/test/grammar/type-system/inferred-types.test.ts
blutorange: I originally thought the issue was related to the special `name` feature for cross-references. But it turned out unrelated, so I guess it serves no purpose anymore. I'll remove it.

blutorange:
I reworked the handling of return type constraint for datatype rule type inference. Tests should now be passing.

(All examples below are a change in the current behavior that I would consider a bug fix)

**Respect return type**

When a data type rule declares a return type (`rule.returnType`) that references a type declaration, use that type. Previously the return type was completely ignored. Note that for datatype rules, the type declarations must be a primitive type.

_Why this fix?_ Ignoring the return type goes directly against what the user declared explicitly.

_Example_
```langium
// Results in
// export type OnlyAB = ABC;
// export type OnlyAC = ABC;
type ABC = "a" | "b" | "c";
OnlyAB returns ABC: 'a' | 'b';
OnlyAC returns ABC: 'a' | 'c';
```

**Respect string constraint during inference**

Datatype rules must always declare a return type. The current inference behavior simply ignores the return type constraint when it is `string`, and infers the type from the rules's content. This is problematic when the inferred type is not compatible with `string`, as the actual AST parser will produce a string.

With this commit, the `string` constraint is now respected when inferring the type. If the inferred type is not compatible with `string`, discards the inferred type and infers 'string' instead. For non-string constraints, preserves the current behavior where inference is skipped and the non-string type is used.

_Why this fix?_  The parsed AST will contain a string, so this brings the declared type into alignment with the runtime type. 

_Example (a)_
```langium
// Results in
// export type PlainText = string;
terminal INT returns number: /[0-9]+/;
terminal INT_ARABIC returns number: INT;
terminal INT_SINO returns number: /[◯一二三四五六七八九]+/;
PlainText returns string: INT_ARABIC | INT_SINO;
```

_Example (b)_
```langium
// Results in
// export type C = string;
A returns number: '0' | '1';
B returns number: '2' | '3';
C returns string: A | B;
```

_Example (c)_
```langium
// Results in
// export type ABInt = 'a' | 'b' | string;
terminal INT returns number: /[0-9]+/;
ABInt returns string: 'a' | 'b' | INT;
```

packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
blutorange: Yes, I did not mean to suggest a breaking change, compatibility has priority. Just that this special case makes the implementation harder that it would be otherwise.

Anyway, I reworked my solution a bit.  But `down at the "return type;" statement` is too late for the fix. The type could be e.g. a property union -- the fix needs to happen inside the `buildDataRuleType` method that does the inference. See [my comment below](https://github.com/eclipse-langium/langium/pull/2202#issuecomment-5301960698).

packages/langium/test/grammar/type-system/type-validator.test.ts
```diff
@@ -266,8 +266,9 @@ describe('Validate declared types', () => {
             RuleB returns B: propB=DTB;
             DTB returns Mytype: 'b';
         `);
-        expect(validationResult.diagnostics).toHaveLength(1);
+        expect(validationResult.diagnostics).toHaveLength(2);
         const grammar = validationResult.document.parseResult.value;
+        expectError(validationResult, /The assigned type 'DTB' is not compatible with the declared property 'propB' of type 'string'./, {});
```
sailingKieler: From my point of view, this expected error is wrong, as the expected type is a super set of the the actual type. This error shouldn't be raised.

packages/langium/test/grammar/type-system/inferred-types.test.ts
```diff
@@ -170,6 +288,78 @@ describe('Inferred types', () => {
...
+            }
+            export type B = number;
+
+            export function isB(item: unknown): item is B {
+                return typeof item === 'number';
+            }
+            export type C = string;
+        `);
+    });
+
```
sailingKieler: Would you please move theses tests up, they're put in-between some `action` related tests.

sailingKieler: I would also appreciate to see some few tests with chained datatype rules, like
```
        await expectTypes(`
            type ABC = "a" | "b" | "c";
            Zero: '0';
            ZeroAsNumber returns number: '0';
            OnlyAB returns ABC: 'a' | 'b';
            ZeroOrAB: Zero | OnlyAB;
            ZeroOrABasString returns string: Zero | OnlyAB;
            NumberOrAB: ZeroAsNumber | OnlyAB;
            NumberOrABAsString returns string: Zero | OnlyAB;
        `, expandToString`
              ...
        `)
```
to make sure the recursive cases make sense.

packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
```diff
@@ -393,13 +402,17 @@ function buildDataRuleType(element: AbstractElement, cancel: () => PlainProperty
...
-                    primitive: ref.type?.name ?? 'string',
+                    primitive: hasStringConstraint ? 'string' : ref.type?.name ?? 'string',
                     regex
                 };
             } else {
-                return {
-                    value: ref.name
-                };
+                if (hasStringConstraint && isParserRule(ref) && ref.dataType && ref.dataType !== 'string') {
+                    return { primitive: 'string' };
```
sailingKieler: I've doubts about this.
At least the case of
```
            type ABC = "a" | "b" | "c";
            Zero returns string: '0';
            OnlyAB returns ABC: 'a' | 'b';
            ZeroOrABasString returns string: Zero | OnlyAB;
```
will not produce
`export type ZeroOrABasString = Zero | OnlyAB;` but
`export type ZeroOrABasString = Zero | string;`, will it?

See also my comment regarding additional tests of recursive cases.

sailingKieler (commented):
Thank you @blutorange for taking care about this.
I also had a look and have some suggestions.
(you dug in a bit of a rabbit whole_ 😅)

packages/langium/test/grammar/type-system/inferred-types.test.ts
sailingKieler: I just noticed that I overlooked the constraint of
_Datatype rules must always declare a return type._,
my example is not fully correct/justified.
The recursive cases still deserve attention I think.

packages/langium/src/grammar/type-system/type-collector/inferred-types.ts
blutorange: That particular example is not a problem, it does result in `type ZeroOrABasString = OnlyAB | Zero;`. The data type rule `OnlyAB` does not have a `rule.dataType` (it has a `rule.returnType`), so it does not trigger the added if-check.

But you are right to have doubts. The following example is still wrong:

```
type OneOrTwo = number;
Rule1 returns OneOrTwo: '1' | '2';
Rule2 returns string: Rule1;
```

Since the check is not triggered, this results in 

```
type OneOrTwo = number;
type Rule1 = OneOrTwo;
type Rule2 = Rule1;
```

I.e. `type Rule2 = Rule1 = OneOrTwo = number`. Which contradicts the returned `returns string` constraint. 

So how can we solve this? If `rule.returnType` is set, we also need to check if that `returnType` is assignable to `string`. Not sure how yet, but let me see if I can figure out a way. The problem is that the type of the referenced data type rule also needs to inferred first, so there's a potential of a recursive loop that needs to be taken care of.

packages/langium/test/grammar/type-system/type-validator.test.ts
blutorange: I'm not sure, but this is at least only indirectly related to this PR. It happens due to the rule:
 
 ```
DTB returns Mytype: 'b'; 
```

Previously the return type got completely ignored, so it simply  inferred `type DTB = 'b'`. 

Now it actually respects the return type `type DTB = Mytype`. This in turn triggers the `types-validator.ts` to raise an error when it tries to compare the union type `Mytype | string` against `string` (previously it would only compare `b` with `string`).

One could argue the behavior is undefined since `type Mytype = Mytype | string` is recursive and not valid.

File: packages/langium/test/grammar/type-system/inferred-types.test.ts (hunk 4 above)
Comment (blutorange) [reply packages/langium/test/grammar/type-system/inferred-types.test.ts]: >  I just noticed that I overlooked the constraint of

Yeah, that also had me confused at first a bit. There are some existing tests without an explicit return type. But if you actually try it out e.g. in the langium playground, it gives you the error `Data type rules cannot infer a type.`

(Which is also kinda not true, since currently `returns string` means `infer the type`).

> you dug in a bit of a rabbit whole

I realized that as well when working on this : /

Can't really be changed for compatiblity, but: I'm still thinking it would be easier if data type rules would allow inferring a type. Then we could simply use the declared return type if one is specified, and infer it if not.

As a new user I also found it unexpected that `returns string` and `returns CustomType` gets ignored and triggers inference anyways.

File: packages/langium/src/grammar/type-system/type-collector/inferred-types.ts (hunk 5 above)
Comment (sailingKieler) [reply packages/langium/src/grammar/type-system/type-collector/inferred-types.ts]: Indeed, there're some challenges.

I think the established pattern of treating the rules in a (nested) loop assuming that in each iteration at least one rule's declared (string) datatype can be resolved (refined) might work here.
If the aforementioned condition is not met there's a cycle.

File: packages/langium/src/grammar/type-system/type-collector/inferred-types.ts [hunk 6]
```diff
@@ -344,19 +344,28 @@ function calculateInfixInterfaces(rules: InfixRule[]): PlainInterface[] {
...
+    // - a primitive type (string, number, boolean, ...)
+    // - a reference to a type declaration that is itself a primitive type (type X = 'foo' | 'bar')
     if (rule.dataType && rule.dataType !== 'string') {
         return {
             primitive: rule.dataType
         };
     }
+    if (rule.returnType && isType(rule.returnType.ref)) {
+        return {
+            value: rule.returnType.ref.name
```
blutorange: So the rabbit hole goes even deeper: I just noticed that this part is also problematic with the current behavior. Consider this toy grammar:

```
grammar Foo

entry Entry: 'value' value=Value;

type TNumber = number;
Value returns TNumber: '0' | '1';

hidden terminal WS: /\s+/;
```

The input `value 1` gets parsed into the AST `{ $type: "Entry",  value: "1"}`. That is, `value` gets parsed as a string, not as a number, as I would expect just looking at the grammar.

This is because the default value converter looks only at the immediately declared return type, finds `TNumber` and since `TNumber` is a reference, not a primitive, just returns string.

Although I guess one could argue this is fine because users need to add their own custom converter to `TNumber`?

packages/langium/test/grammar/type-system/inferred-types.test.ts
```diff
@@ -550,38 +454,573 @@ describe('Inferred types', () => {
...
+            export function isABC(item: unknown): item is ABC {
+                return isA(item) || isB(item) || item === 'c';
+            }
+            export type B = 'b';
+
+            export function isB(item: unknown): item is B {
+                return item === 'b';
+            }
+            export type Rule = A | ABC | B;
```
sailingKieler, line 958: Do we have an explanation why orderings got changed?
Are the declaration sorted lexicographically? Interesting that this even affects the alternatives in the union type here.

packages/langium/test/grammar/type-system/inferred-types.test.ts
```diff
@@ -550,38 +454,573 @@ describe('Inferred types', () => {
...
+    });
+
+    test('Should not error when a data type rule is part of a cycle with another rule', async () => {
+        await expectTypes(`
+            Rule1 returns string: Rule2;
+            Rule2 returns string: Rule1;
+        `, expandToString`
+            export type Rule1 = Rule2;
+
+            export type Rule2 = Rule1;
```
sailingKieler, line 991: This results in a Typescript compiler error!

IMO the grammar editor should error here, shouldn't it?

Image

packages/langium/test/grammar/grammar-validator.test.ts
```diff
@@ -2256,3 +2256,25 @@ describe('Strict type validation', () => {
...
+describe('Cyclic references', () => {
+    test('does not error on cyclic references in data type rules', async () => {
+        const validationResult = await validate(`
+        grammar Test
+        entry A: a=B;
+        B returns string: C;
+        C returns string: B;
+        `);
+        expectNoIssues(validationResult);
+    });
```
sailingKieler, line 2269: Duplicate of `'Should not error when a data type rule is part of a cycle with another rule'` that I commented, isn't it?

sailingKieler (commented):
Looks pretty good to me, thanks a lot @blutorange. 🙏
However, I have a doubt regarding the tolerance of cycles, which I prefer to forbid in order to avoid TS compiler errors, see below.

@msujew what do you think?

packages/langium/test/grammar/type-system/inferred-types.test.ts
sailingKieler: The same applies to the case above with `export type Rule = 'a' | 'b' | Rule;`

msujew:
I agree with @sailingKieler. We shouldn't encode bad data into code generator tests - since the bad data should be filtered out by validation in the first place.

packages/langium/test/grammar/type-system/inferred-types.test.ts
blutorange: The current behavior is that this input results in a `An error occurred during validation: Maximum call stack size exceeded`. I just fixed that issue, so it doesn't overflow the call  stack. 

But you mean, we should add additional validations so the grammar results in an error like `circular references are not allowed`?

packages/langium/test/grammar/type-system/inferred-types.test.ts
blutorange: I'm pretty sure that's an existing behavior. E.g. go to the main branch and run

```
npx vitest run  packages/langium/test/grammar/type-system/inferred-types.test.ts -t 'Should infer types for alternatives'
```

If you now open the test file and edit change the test input from e.g. `C: A | B;` to `C: B | A;`, the test still passes.

packages/langium/test/grammar/type-system/inferred-types.test.ts
sailingKieler: Hi @blutorange, my suggestion on this:

The test currently doesn't apply the validation and check the diagnostics, but it should do with your additional cases. Hence, replace this line
https://github.com/eclipse-langium/langium/blob/4094f25af763a715bf4a72438e011140e03fdef5/packages/langium/test/grammar/type-system/inferred-types.test.ts#L930 by
```ts
    const result = await helper(grammar, { validation: true });
    expect(printDiagnostics(result.diagnostics)).toBe('');
```
and update the test expectations and titles correspondingly, then the PR is good to go from my POV.

Thank you for your patience! 😉

sailingKieler: ... and check for duplicates in the tests once more.

PR #2205, opened 2026-08-11
chore: sec fixes for js-yaml and nanoid

cdietrich:


sailingKieler (approved):
Thanks @cdietrich, trusting you...

PR #2206, opened 2026-08-14
Fix VS Code ignore patterns during workspace indexing

shblue21:
`langium.build.ignorePatterns` was ignored during initial workspace indexing. Workspace folder URIs arrive as strings, while file entries use URI objects, so the containment check never matched them.

This change parses the workspace folder URI before checking it against each entry.

<details>
<summary>Before</summary>

langium-ignore-patterns-repro

</details>

<details>
<summary>After</summary>

langium-ignore-patterns-fixed

</details>

### DIFF
--- a/packages/langium-vscode/src/language-server/grammar-workspace-manager.ts
+++ b/packages/langium-vscode/src/language-server/grammar-workspace-manager.ts
@@ -53,7 +53,7 @@ export class LangiumGrammarWorkspaceManager extends DefaultWorkspaceManager {
     }
 
     override shouldIncludeEntry(entry: FileSystemNode): boolean {
-        const workspaceFolder = this.workspaceFolders?.find(folder => UriUtils.contains(folder.uri, entry.uri));
+        const workspaceFolder = this.workspaceFolders?.find(folder => UriUtils.contains(URI.parse(folder.uri), entry.uri));
         if (this.matcher && workspaceFolder) {
             // create path relative to workspace folder root: /user/foo/workspace/entry.txt -> entry.txt
             const relPath = path.relative(URI.parse(workspaceFolder.uri).path, entry.uri.path);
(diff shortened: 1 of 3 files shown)


packages/langium-vscode/src/language-server/grammar-workspace-manager.ts
```diff
@@ -53,7 +53,7 @@ export class LangiumGrammarWorkspaceManager extends DefaultWorkspaceManager {
     }
 
     override shouldIncludeEntry(entry: FileSystemNode): boolean {
-        const workspaceFolder = this.workspaceFolders?.find(folder => UriUtils.contains(folder.uri, entry.uri));
+        const workspaceFolder = this.workspaceFolders?.find(folder => UriUtils.contains(URI.parse(folder.uri), entry.uri));
```
sailingKieler: What is the actual issue here?
Consider that `UriUtils.contains()` does accept strings as first argument.
What are the actual strings being compared within `UriUtils.contains()`?
On which platform do you run the extension?
Is there a prefix like `file://` disturbing the comparison, or an upper case drive letter on windows?

sailingKieler (commented):
I'm a bit skeptical about this change, see below.

Can you investigate a bit further and share some details?

packages/langium-vscode/src/language-server/grammar-workspace-manager.ts
msujew: Looking into `contains`, I'm inclined to say that the function doesn't work as expected. If a given argument is a string, it's used as-is, whereas a `URI` only has its `path` being used for the comparison. A fix should actually:
1. Convert the input strings into URIs
2. Compare the URI schemes first, reject on mismatch
3. Then compare the paths (current logic)

shblue21: You were right about the file:// prefix. I checked the values on macOS

folder.uri: file:///private/tmp/...
entry.uri.path: /private/tmp/.../generated

So the comparison is between a full URI and a path. I’ll move the fix into `UriUtils.contains` as suggested and add a test for the mixed case.

msujew (approved):
Thanks, looks good to me 👍

shblue21:
Thanks for the review and merge!

PR #2207, opened 2026-08-15
Extracted methods to customize the file system access in the DefaultLangiumDocumentFactory

JohannesMeierSE:
Makes customizations easier for:
- dedicated error handling for issues during the file system access, when creating documents or updating documents
- take advanced URIs into account

### DIFF
--- a/packages/langium/src/workspace/documents.ts
+++ b/packages/langium/src/workspace/documents.ts
@@ -181,8 +181,12 @@ export class DefaultLangiumDocumentFactory implements LangiumDocumentFactory {
     }
 
     async fromUri<T extends AstNode = AstNode>(uri: URI, cancellationToken = CancellationToken.None): Promise<LangiumDocument<T>> {
-        const content = await this.fileSystemProvider.readFile(uri);
+        const content = await this.getInitialText(uri, cancellationToken);
         return this.createAsync<T>(uri, content, cancellationToken);
+    }
+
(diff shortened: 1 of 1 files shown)


packages/langium/src/workspace/documents.ts
```diff
@@ -181,10 +181,14 @@ export class DefaultLangiumDocumentFactory implements LangiumDocumentFactory {
     }
 
     async fromUri<T extends AstNode = AstNode>(uri: URI, cancellationToken = CancellationToken.None): Promise<LangiumDocument<T>> {
-        const content = await this.fileSystemProvider.readFile(uri);
+        const content = await this.getInitialText(uri, cancellationToken);
         return this.createAsync<T>(uri, content, cancellationToken);
     }
 
+    protected async getInitialText(uri: URI, _cancellationToken = CancellationToken.None): Promise<string> {
```
sailingKieler, line 188: I think similar to `getUpdatedText(...)` you can expect `_cancellationToken` to by given, like
```suggestion
    protected async getInitialText(uri: URI, _cancellationToken: CancellationToken): Promise<string> {
```

sailingKieler (approved):
Looks good to me, just one minor remark, see below.

packages/langium/src/workspace/documents.ts
JohannesMeierSE: Makes sense, changed

PR #2208, opened 2026-08-15
Extended existing assertions to accept error messages, which are calculated on-demand

JohannesMeierSE:
This enables to define demanding error messages, which are lazily calculated on-demand to save some performance.


sailingKieler (approved):
LGTM 👍

PR #2209, opened 2026-08-15
Use internal normalizeUri at one additional position in UriTrie (follow-up for #2069)

JohannesMeierSE:
Added missing case for #2069


sailingKieler (approved):
Ooops
🚀

PR #2210, opened 2026-08-15
Fixed edge case in MapScope: Don't use outer scope, if it is disabled

JohannesMeierSE:
The `MapScope` allows to disable a given outer scope under certain conditions by a dedicated option:
- Fixed an edge case, since the outer scope was applied, even when disabled by option.
- Targets only `getElements` for multi-refs
- Added test case for this scenario

### DIFF
--- a/packages/langium/src/references/scope.ts
+++ b/packages/langium/src/references/scope.ts
@@ -139,7 +139,7 @@ export class MapScope implements Scope {
         const localName = this.caseInsensitive ? name.toLowerCase() : name;
         const local = this.elements.get(localName);
         const arr = local ? [local] : [];
-        if ((this.concatOuterScope || arr.length > 0) && this.outerScope) {
+        if ((this.concatOuterScope || arr.length === 0) && this.outerScope) {
             return stream(arr).concat(this.outerScope.getElements(name));
         } else {
             return stream(arr);
--- a/packages/langium/test/references/scope.test.ts
+++ b/packages/langium/test/references/scope.test.ts
@@ -1,0 +1,49 @@
+/******************************************************************************
+ * Copyright 2026 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
+
+import { describe, expect, test } from 'vitest';
+import { MapScope } from '../../src/references/scope.js';
+import type { AstNodeDescription } from '../../src/syntax-tree.js';
+
+describe('Tests the default map scope implementation', () => {
+    const description: AstNodeDescription = {
+        name: 'MyName',
+        type: 'Type1',
+        documentUri: undefined!,
+        path: undefined!,
+    };
+
+    test('Do not use a given outer scope, if using the outer scope is disabled by option', async () => {
+        const outerScope = new MapScope([description]);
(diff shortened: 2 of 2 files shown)


sailingKieler (commented):
The change makes sense to me, the ` || arr.length > 0` condition looks odd.

However, I would appreciate some additional tests covering the remaining conditions.

msujew:
@sailingKieler @JohannesMeierSE The idea of the `arr.length > 0` check is to only refer to the outer scope, if the current scope doesn't return anything for the given input. After all, the option is called `concatOuterScope`. It doesn't actually concat, if the current scope is empty - it just returns the content of the outer scope. I agree, this could've been better documented.

To explain this in more detail: Let's say you have the following DSL code (with `X` being part of a multi-reference):

```
var X;
if (someVar) {
  var X;
  if (otherVar) {
    print(X);
  }
}
```

`print(X)` and the inner `var X` are in two different local scopes, so it will "fall through" to the inner `X`, but not merge it with the outer `X`, if the `concatOuterScope` option is set to false. With your changes, it won't pick up anything if `concatOuterScope` is set to false, which isn't the intended behavior.

sailingKieler:
> @sailingKieler @JohannesMeierSE The idea of the `arr.length > 0` check is to only refer to the outer scope, if the current scope doesn't return anything for the given input. After all, the option is called `concatOuterScope`. It doesn't actually concat, if the current scope is empty - it just returns the content of the outer scope. I agree, this could've been better documented.
> 
> To explain this in more detail: Let's say you have the following DSL code (with `X` being part of a multi-reference):
> 
> ```
> var X;
> if (someVar) {
>   var X;
>   if (otherVar) {
>     print(X);
>   }
> }
> ```
> 
> `print(X)` and the inner `var X` are in two different local scopes, so it will "fall through" to the inner `X`, but not merge it with the outer `X`, if the `concatOuterScope` option is set to false. With your changes, it won't pick up anything if `concatOuterScope` is set to false, which isn't the intended behavior.

So the condition implements _name shadowing_/literal _name scoping_ ... uhhh.
But shouldn't it then be like `(this.alwaysConcatOuterScope || arr.length === 0)...`,
i.e. if current scope's candidates list is empty then consider `outerScope`? 🧐

msujew:
Probably could have received a better name, yes. I'm also open to adjust the API to accept an enum or similar to have different behavior there.

sailingKieler:
My main doubt is about the binary operator that should be `=== 0` instead of `> 0`. Am I wrong?

msujew:
> My main doubt is about the binary operator that should be === 0 instead of > 0. Am I wrong?

Yeah, you're probably right. Now I'm confused what the original implementation was supposed to do...

msujew:
I've confused this with the `MultiMapScope` implementation:

https://github.com/eclipse-langium/langium/blob/c03fb6d4201dfc4f8bf4af9182823a48a661d043/packages/langium/src/references/scope.ts#L193-L197

Which correctly uses `local.length === 0`. We should align to this IMO.

JohannesMeierSE:
All three implementations (for `MapScope`, `MultiMapScope` and `StreamScope`) are in sync now. I added some more test cases.

msujew (approved):
Thanks, makes sense 👍

PR #2211, opened 2026-08-15
Completion provider supports both insert and replace mode now

### LINKED ISSUE #1924: Support `InsertReplaceEdit` for completion item results
Search for `InsertReplaceEdit` in the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) for more info.

JohannesMeierSE:
Fixes #1924 

The default completion provider supports the editor's insert/replace suggestion mode now.
In VS Code, this setting is `editor.suggest.insertMode`.

### DIFF
--- a/packages/langium/src/lsp/completion/completion-provider.ts
+++ b/packages/langium/src/lsp/completion/completion-provider.ts
@@ -4,7 +4,7 @@
  * terms of the MIT License, which is available in the project root.
  ******************************************************************************/
 
-import type { CompletionItem, CompletionParams, TextEdit } from 'vscode-languageserver-protocol';
+import type { CompletionClientCapabilities, CompletionItem, CompletionParams, InitializeParams, InsertReplaceEdit, TextEdit } from 'vscode-languageserver-protocol';
 import type { LangiumCompletionParser } from '../../parser/langium-parser.js';
 import type { NameProvider } from '../../references/name-provider.js';
 import type { ScopeProvider } from '../../references/scope-provider.js';
@@ -135,6 +135,7 @@ export class DefaultCompletionProvider implements CompletionProvider {
     protected readonly fuzzyMatcher: FuzzyMatcher;
     protected readonly grammarConfig: GrammarConfig;
     protected readonly astReflection: AstReflection;
+    protected clientCapabilities?: CompletionClientCapabilities;
     readonly completionOptions?: CompletionProviderOptions;
 
     constructor(services: LangiumServices) {
@@ -148,6 +149,14 @@ export class DefaultCompletionProvider implements CompletionProvider {
         this.grammarConfig = services.parser.GrammarConfig;
         this.astReflection = services.shared.AstReflection;
         this.documentationProvider = services.documentation.DocumentationProvider;
+
+        services.shared.lsp.LanguageServer.onInitialize(params => {
+            this.initialize(params);
+        });
+    }
+
+    protected initialize(params: InitializeParams): void {
+        this.clientCapabilities = params.capabilities.textDocument?.completion;
     }
 
     async getCompletion(document: LangiumDocument, params: CompletionParams, _cancelToken?: CancellationToken): Promise<CompletionList | undefined> {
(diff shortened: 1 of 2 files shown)


packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -595,18 +596,31 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
+     * A plain `TextEdit` forces the editor to use exactly that single range and disables the insert/replace distinction,
+     * so the suffix of an identifier after the cursor (e.g. `Pos|ition`) is never replaced.
+     * As a result "replace" mode appears broken.
+     *
+     * Returning an {@link InsertReplaceEdit} instead provides both ranges and lets the editor pick the
+     * right one based on the user's setting:
+     * - `insert`: from the token start up to the cursor (keeps the trailing text),
+     * - `replace`: from the token start up to the token end (replaces the whole identifier).
+     */
+    protected buildCompletionTextEdit(context: CompletionContext, label: string, newText: string): TextEdit | InsertReplaceEdit | undefined {
```
msujew, line 613: Suggestion: Whether or not to return a `TextEdit` vs a `InsertReplaceEdit` should depend on the client capabilities. The client indicates, whether it has the `insertReplaceSupport` [capability](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) at all.

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -595,18 +596,31 @@ export class DefaultCompletionProvider implements CompletionProvider {
...
+     *
+     * The range of a {@link TextEdit} ends at the cursor position (`context.offset`).
+     * A plain `TextEdit` forces the editor to use exactly that single range and disables the insert/replace distinction,
+     * so the suffix of an identifier after the cursor (e.g. `Pos|ition`) is never replaced.
+     * As a result "replace" mode appears broken.
+     *
+     * Returning an {@link InsertReplaceEdit} instead provides both ranges and lets the editor pick the
+     * right one based on the user's setting:
+     * - `insert`: from the token start up to the cursor (keeps the trailing text),
+     * - `replace`: from the token start up to the token end (replaces the whole identifier).
```
msujew, line 611: Suggestion: This reads like an artifact from an agent conversation. Please just document what the method does.

packages/langium/test/lsp/completion-provider.test.ts
```diff
@@ -26,6 +26,10 @@ describe('Langium completion provider', () => {
     const grammarServices = createLangiumGrammarServices(EmptyFileSystem).grammar;
     const completion = expectCompletion(grammarServices);
 
+    beforeEach(async () => grammarServices.shared.workspace.DocumentBuilder.update(
+        [], grammarServices.shared.workspace.LangiumDocuments.all.map((d) => d.uri).toArray()
+    ));
```
msujew, line 31: Question: Is this really required? I fail to understand why this would be necessary.

packages/langium/src/lsp/completion/completion-provider.ts
```diff
@@ -28,6 +28,7 @@ import { findDeclarationNodeAtOffset, findLeafNodeBeforeOffset, getDatatypeNode
 import { getEntryRule } from '../../utils/grammar-utils.js';
 import { stream, type Stream } from '../../utils/stream.js';
 import { findFirstFeatures, findNextFeatures } from './follow-element-computation.js';
+import type { InsertReplaceEdit } from 'vscode-languageserver-types';
```
msujew, line 31: Suggestion: There is already an import from `vscode-languageserver-protocol` in this file. Use that instead.

packages/langium/src/lsp/completion/completion-provider.ts
sailingKieler: I fully support Mark's remark.

What's IMO especially worth pointing out explicitly is the aspect that the token part in left of the cursor will overwritten in insert mode, too, i.e. the `Pos` part if completions are called at `Pos|ition`.
It's somehow identifiable in the `insert` `replace` clauses above, if the reader is aware of aspect/feature at all.

packages/langium/src/lsp/completion/completion-provider.ts
sailingKieler, line 24: At that, please replace
```
import { CompletionItemKind, CompletionList, Position } from 'vscode-languageserver';
```
by
```
import { CompletionItemKind, CompletionList, Position } from 'vscode-languageserver-types';
```

sailingKieler (commented):
Some additional two cents, see below.

packages/langium/src/lsp/completion/completion-provider.ts
JohannesMeierSE: Thanks for the hint, done now

packages/langium/src/lsp/completion/completion-provider.ts
JohannesMeierSE: reduced comments now

packages/langium/test/lsp/completion-provider.test.ts
JohannesMeierSE: `expectCompletion` internally adds new documents during each test. Deleting all documents afterwards ensures that each test case has only one document which can be easily found (otherwise you have to identify the newest document, which is more difficult).

I reworked the test cases, now this logic is not necessary anymore.

packages/langium/src/lsp/completion/completion-provider.ts
JohannesMeierSE: done

packages/langium/src/lsp/completion/completion-provider.ts
JohannesMeierSE: In these cases, the items are not used as `type`, but as `namespace`

msujew (approved):
Looks good, thanks 👍

PR #2214, opened 2026-08-25
Support priority read requests for LSP requests

msujew:
This is a new version of https://github.com/eclipse-langium/langium/pull/1640.

This is doing a bit less, but it's doing it a bit better. Should resolve intermittent errors that appear due to outdated data within LSP requests.

### DIFF
--- a/packages/langium/src/workspace/workspace-lock.ts
+++ b/packages/langium/src/workspace/workspace-lock.ts
@@ -22,14 +22,18 @@ export interface WorkspaceLock {
 
     /**
      * Performs a single action, like computing completion results or providing workspace symbols.
-     * Read actions will only be executed after all write actions have finished. They will be executed in parallel if possible.
+     * With {@link ReadPriority.Normal} priority (the default), read actions will only be executed after all write actions have finished.
+     * They will be executed in parallel if possible.
      *
      * If a write action is currently running, the read action will be queued up and executed afterwards.
      * If a new write action is queued up while a read action is waiting, the write action will receive priority and will be handled before the read action.
      *
+     * With {@link ReadPriority.Immediate} priority, the read action is executed right away, concurrently to any running write action.
+     * Use this only if the required workspace/document state has already been awaited by other means.
+     *
      * Note that read actions are not allowed to modify anything in the workspace. Please use {@link write} instead.
      */
-    read<T>(action: () => MaybePromise<T>): Promise<T>;
+    read<T>(action: () => MaybePromise<T>, priority?: ReadPriority): Promise<T>;
 
     /**
      * Cancels the last queued write action. All previous write actions already have been cancelled.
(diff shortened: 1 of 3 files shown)


packages/langium/src/workspace/workspace-lock.ts
```diff
@@ -22,14 +22,18 @@ export interface WorkspaceLock {
 
     /**
      * Performs a single action, like computing completion results or providing workspace symbols.
-     * Read actions will only be executed after all write actions have finished. They will be executed in parallel if possible.
+     * With {@link ReadPriority.Normal} priority (the default), read actions will only be executed after all write actions have finished.
+     * They will be executed in parallel if possible.
      *
      * If a write action is currently running, the read action will be queued up and executed afterwards.
      * If a new write action is queued up while a read action is waiting, the write action will receive priority and will be handled before the read action.
      *
```
JohannesMeierSE, line 30: Could you add a hint, that running read actions (independent of their priority) are not cancelled, when a write action is queued?

packages/langium/src/workspace/workspace-lock.ts
```diff
@@ -59,7 +75,19 @@ export class DefaultWorkspaceLock implements WorkspaceLock {
         return this.enqueue(this.writeQueue, action, tokenSource.token);
     }
 
-    read<T>(action: () => MaybePromise<T>): Promise<T> {
+    read<T>(action: () => MaybePromise<T>, priority: ReadPriority = ReadPriority.Normal): Promise<T> {
+        if (priority === ReadPriority.Immediate) {
```
JohannesMeierSE, line 80: Since the "immediate read" has a completely different implementation, what do you think about providing a separate method like `immediateRead<T>(...)` instead of adding a new `priority` parameter?

JohannesMeierSE (commented):
Looks fine for me, thanks @msujew!

I have only an idea for the new API.

packages/langium/src/workspace/workspace-lock.ts
msujew: While this makes the internal implementation cleaner, it makes external callers more complicated. See:

https://github.com/eclipse-langium/langium/blob/1cf1d00ed924e97b610afac380386c93fc74ddc0/packages/langium/src/lsp/language-server.ts#L757-L760

If we were to split this into multiple interface methods, we would need something like this:

```ts
const callback = async () => { 
  const document = await documents.getOrCreateDocument(uri); 
  return serviceCall(language, document, params, cancelToken); 
};
if (readPriority(requiredState) === ReadPriority.Immediate) {
  return await mutex.immediateRead(callback);
} else {
  return await mutex.read(callback);
}
```

Which is why I would favor the current API.

JohannesMeierSE: OK, both are fine for me

JohannesMeierSE (approved):

PR #2216, opened 2026-08-27
Add basic MCP support to language servers

msujew:
Adds experimental MCP server support for language servers. Right now, the default implementation only supports a single tool, which is a `get-diagnostics` call. I've confirmed that most models actually prefer this over any other approach to retrieve diagnostics from a given langium file.

Adopters can override it like this:

```ts
class CustomMcpService extends DefaultMcpService {
  override createServer() {
    const server = super.createServer();
    server.registerTool(...);
    return server;
  }
}
```

### DIFF
--- a/packages/langium-mcp/LICENSE
+++ b/packages/langium-mcp/LICENSE
@@ -1,0 +1,16 @@
+Copyright 2026 TypeFox GmbH
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+associated documentation files (the "Software"), to deal in the Software without restriction,
+including without limitation the rights to use, copy, modify, merge, publish, distribute,
+sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or
+substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
--- a/packages/langium-mcp/README.md
+++ b/packages/langium-mcp/README.md
@@ -1,0 +1,3 @@
+# Langium MCP Support
+
+This package provides support for integrating MCP servers into language servers built with Langium.
--- a/packages/langium-mcp/package.json
+++ b/packages/langium-mcp/package.json
@@ -1,0 +1,48 @@
+{
+  "name": "langium-mcp",
+  "version": "4.3.0",
+  "description": "Use Langium to build MCP Servers",
+  "homepage": "https://langium.org",
+  "keywords": [
+    "language",
+    "dsl",
+    "lsp",
+    "mcp"
+  ],
+  "license": "MIT",
+  "files": [
+    "lib",
+    "src"
+  ],
+  "sideEffects": false,
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./lib/index.d.ts",
+      "import": "./lib/index.js"
+    }
+  },
+  "scripts": {
+    "clean": "shx rm -rf lib coverage",
+    "build": "tsc",
+    "watch": "tsc --watch",
+    "publish:next": "npm --no-git-tag-version version \"$(semver $npm_package_version -i minor)-next.$(git rev-parse --short HEAD)\" && npm publish --tag next",
+    "publish:latest": "npm publish --tag latest --access public"
+  },
+  "dependencies": {
+    "@modelcontextprotocol/server": "~2.0.0",
+    "langium": "~4.3.0",
+    "vscode-languageserver-types": "~3.18.0",
+    "zod": "~4.4.3"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/eclipse-langium/langium.git",
+    "directory": "packages/langium-mcp"
+  },
+  "bugs": "https://github.com/eclipse-langium/langium/issues",
+  "author": {
+    "name": "TypeFox",
+    "url": "https://www.typefox.io"
+  }
+}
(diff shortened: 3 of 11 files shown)


packages/langium-mcp/LICENSE
```diff
@@ -0,0 +1,16 @@
+Copyright 2023 TypeFox GmbH
```
cdietrich, line 1: 2026

packages/langium-vscode/src/language-server/main.ts
```diff
@@ -25,3 +29,14 @@ const { shared, grammar } = createLangiumGrammarServices({ connection, ...NodeFi
...
+const mcpService = new DefaultMcpService(shared, {
+    name: 'Langium Grammar MCP Server',
+    version: '4.3.0'
+});
+
+const handler = createMcpHandler(() => mcpService.createServer());
+const nodeHandler = toNodeHandler(handler);
+createServer((req, res) => {
+    nodeHandler(req, res);
+}).listen(8999, '127.0.0.1');
```
cdietrich, line 42: should we make this dynamic

packages/langium-mcp/src/mcp-service.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    protected registerDiagnosticsTool(server: McpServer): void {
+        server.registerTool('get-diagnostics', {
+            title: 'Get Diagnostics',
+            description: 'Retrieves diagnostics from the language server for all files in the given directory, ordered by file.',
+            inputSchema: z.object({
+                directory: z.string().optional(),
+            })
+        }, async ({ directory }) => {
+            const severities = ['error', 'warning', 'information', 'hint'];
```
cdietrich, line 48: should we wait for workspace?

cdietrich, line 48: is this corrct. afaik error = 1

packages/langium-mcp/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "scripts": {
+    "clean": "shx rm -rf lib coverage",
+    "build": "tsc",
+    "watch": "tsc --watch",
+    "publish:next": "npm --no-git-tag-version version \"$(semver $npm_package_version -i minor)-next.$(git rev-parse --short HEAD)\" && npm publish --tag next",
+    "publish:latest": "npm publish --tag latest --access public"
+  },
+  "dependencies": {
+    "@modelcontextprotocol/server": "~2.0.0",
+    "langium": "~4.3.0",
```
montymxb, line 34: we can bump the Langium versions up to 4.4.0 throughout

packages/langium-mcp/README.md
```diff
@@ -0,0 +1,3 @@
+# Langium MCP Support
+
+This package provides support for integrating MCP servers into language servers built with Langium.
```
montymxb, line 3: Probably should include example/reference usage here, or just include examples outright.

packages/langium-vscode/src/language-server/main.ts
```diff
@@ -25,3 +29,14 @@ const { shared, grammar } = createLangiumGrammarServices({ connection, ...NodeFi
 registerLangiumConfigHandler(connection, shared, grammar);
 registerRailroadConnectionHandler(connection, grammar);
 startLanguageServer(shared);
+
+const mcpService = new DefaultMcpService(shared, {
+    name: 'Langium Grammar MCP Server',
+    version: '4.3.0'
```
montymxb, line 35: Is this the only spot we have a hardcoded version string? It would be great if we could inject a constant w/ esbuild or something similar, so we don't forget to bump it later.

packages/langium-vscode/src/language-server/main.ts
montymxb: I would second this, running another instance would be a problem here.

packages/langium-mcp/src/mcp-service.ts
```diff
@@ -0,0 +1,74 @@
...
+                directory: z.string().optional(),
+            })
+        }, async ({ directory }) => {
+            await this.services.workspace.WorkspaceManager.ready;
+            const severities = ['error', 'warning', 'information', 'hint'];
+            const lines: string[] = [];
+            // Wrap the diagnostics read in a lock, to ensure that the workspace build is already done
+            await this.services.workspace.WorkspaceLock.read(() => {
+                const documentManager = this.services.workspace.LangiumDocuments;
+                const documents = directory ? documentManager.getDocuments(URI.file(directory)) : documentManager.all.toArray();
```
montymxb, line 54: On a freshly opened workspace will this be empty? If so, I'm wondering if we want to manually trigger build w/ validation on the target docs, or expose an option to do so.

packages/langium-vscode/src/language-server/main.ts
```diff
@@ -5,13 +5,17 @@
...
 import { createLangiumGrammarServices } from 'langium/grammar';
 import { startLanguageServer, type LangiumSharedServices, type PartialLangiumSharedServices } from 'langium/lsp';
 import { NodeFileSystem } from 'langium/node';
 import { ProposedFeatures, createConnection } from 'vscode-languageserver/node';
 import { LangiumGrammarWorkspaceManager } from './grammar-workspace-manager.js';
 import { registerRailroadConnectionHandler } from './railroad-handler.js';
 import { registerLangiumConfigHandler } from './config-handler.js';
+import { DefaultMcpService } from 'langium-mcp';
+import { toNodeHandler } from '@modelcontextprotocol/node';
+import { createServer } from 'node:http';
```
montymxb, line 18: Are we interested in incorporating some of these within the langium-mcp package itself, so users don't need to set this up directly? Being able to start the mcp server with functions from `langium-mcp/node` could be helpful.

packages/langium-vscode/src/language-server/main.ts
```diff
@@ -25,3 +29,14 @@ const { shared, grammar } = createLangiumGrammarServices({ connection, ...NodeFi
...
 startLanguageServer(shared);
+
+const mcpService = new DefaultMcpService(shared, {
+    name: 'Langium Grammar MCP Server',
+    version: '4.3.0'
+});
+
+const handler = createMcpHandler(() => mcpService.createServer());
+const nodeHandler = toNodeHandler(handler);
+createServer((req, res) => {
```
montymxb, line 40: Would be good to error handle in case we can't bind, or for any other reason.

packages/langium-vscode/src/language-server/main.ts
```diff
@@ -25,3 +29,14 @@ const { shared, grammar } = createLangiumGrammarServices({ connection, ...NodeFi
...
+    nodeHandler(req, res);
```
montymxb, line 41: Looks like it's recommended to add some middleware in here from the mcp package for host & origin validation, specifically [localhostHostValidation](https://ts.sdk.modelcontextprotocol.io/v2/api/@modelcontextprotocol/express/middleware/hostHeaderValidation.html#localhosthostvalidation) and [localhostOriginValidation](https://ts.sdk.modelcontextprotocol.io/v2/api/@modelcontextprotocol/hono/middleware/originValidation.html#localhostoriginvalidation).

montymxb (commented):
Looks cool! Added some points & questions here and there.

PR #2231, opened 2026-09-17
fix(cli): keep doc comments of imported grammar elements when embedding

### LINKED ISSUE #2065: Comment Collection / Inlining
The comment collection / inlining feature
does not seem to work on imported grammars.

simonyang08:
Fixes #2065

## Root cause

`embedReferencedGrammar` (`packages/langium-cli/src/generate.ts`) copies imported grammar elements with `AstUtils.copyAstNode`, which drops the `` reference. Both comment consumers resolve comments through the CST:

- the grammar serializer (`comments: true` → `DefaultCommentProvider` → `findCommentNode(node.$cstNode, ...)`)
- the BNF generator (`findCommentNode(rule.$cstNode, ['ML_COMMENT'])`)

so rules and types from imported grammars lose their doc comments in the generated `grammar.ts` (``) and BNF output.

## Fix

Re-attach the original CST node to each copied element:

```ts
Object.defineProperty(copy, '', { value: element., writable: true });
```

Embedded grammars are only used for read-only generation, so sharing the CST node with the imported grammar's document is safe. (The `defineProperty` route is used because `` is typed readonly and `copyAstNode` deliberately skips `$`-properties.)

## Tests

New `packages/langium-cli/test/generator/embed-grammars.test.ts` drives the real `embedGrammars` path with a two-grammar setup (import + cross-reference) and locks down:

1. embedded elements keep their ``
2. the BNF output contains the imported rule's doc comment
3. the serialized grammar contains `` for imported rules and types

`vitest run packages/langium-cli/test`: 57 passed, 0 failed (54 existing + 3 new).

## Two adjacent observations from reproducing this

1. **Only multiline comments are collected.** `DefaultCommentProvider` follows `GrammarConfig.multilineCommentRules` (auto-derived from multiline comment terminal rules, so `SL_COMMENT`/`///` line docs are invisible by design), and the BNF generator hardcodes `['ML_COMMENT']` instead of going through `CommentProvider` — so a custom `multilineCommentRules` override wouldn't reach the BNF output. Worth a deliberate decision at some point; out of scope here.
2. **`ast.ts` never emits JSDoc.** We checked the current main and v3.5.0: the AST generator contains no comment emission at all, so the 'JSDoc appears on interfaces but not types' half of the original report doesn't reproduce on official code — the actual comment-loss symptoms are confined to `grammar.ts` and BNF output, both covered by this fix.

Reproduction project: two grammars (import + cross-reference), `bnf` target enabled — happy to share if useful.


msujew (approved):
Thanks, works well 👍

